PR #5, opened 2020-02-28
Hyperelastic solver object

jamiebramwell:
This adds a hyperelastic large deformation solver under the generic base solver class framework.


src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -0,0 +1,184 @@
...
+void NonlinearSolidQuasiStaticOperator::Mult(const mfem::Vector &k, mfem::Vector &y) const
+{
+  // Apply the nonlinear form
+  m_H_form->Mult(k, y);
+}
+
+// Compute the Jacobian from the nonlinear form
+mfem::Operator &NonlinearSolidQuasiStaticOperator::GetGradient(const mfem::Vector &x) const
+{
+  m_Jacobian = &m_H_form->GetGradient(x);
```
samuelpmishLLNL: confusing (to me) operator precedence-- is this any different than just `return m_H_form->GetGradient(x)`?

src/solvers/CMakeLists.txt
```diff
@@ -6,15 +6,15 @@
 
 set(solvers_sources
     base_solver.cpp
-    quasistatic_solver.cpp
+    nonlinear_solid_solver.cpp
```
samuelpmishLLNL: if I want to do dynamics with a nonlinear material, am I using nonlinear_solid_solver or dynamic_solver?

src/solvers/nonlinear_solid_solver.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+#ifndef NONLINSOLID_SOLVER
+#define NONLINSOLID_SOLVER
+
+#include "mfem.hpp"
+#include "base_solver.hpp"
+
+// Forward declaration
+class NonlinearSolidQuasiStaticOperator;
+class NonlinearSolidDynamicOperator;
```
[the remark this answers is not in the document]
jamiebramwell: Yes - This should include both the `DynamicSolver` and `QuasiStaticSolver` capabilities when it's done (similar to `ThermalSolver`). This is still very much a work in progress, however.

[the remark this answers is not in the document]
samuelpmishLLNL: Any objection to just making it "SolidMechanicsSolver" or similar? The other Solver objects don't distinguish between linear/nonlinear.

src/solvers/nonlinear_solid_solver.cpp
jamiebramwell: I don't think so. I save it in `m_Jacobian` because I can see a point where we might want to use it in other parts of the solver routines (for example returning sensitivities for the user). However, it may be better to just clean it up for the time being.

samuelpmishLLNL: Whoops! Good point, I thought it was a temporary at first glance.

src/solvers/nonlinear_solid_solver.hpp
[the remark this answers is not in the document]
jamiebramwell: That's a good point. I'm not sure whether we should combine nonlinear operators with linear ones as the underlying data structures are fairly different. However from a user perspective, I can see it being clearer with just a single interface and a flag to set whether it's linear or not. This is essentially what I'm doing with the static and dynamic options anyway.

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+void NonlinearSolidSolver::SetTractionBCs(mfem::Array<int> &trac_bdr, mfem::VectorCoefficient *trac_bdr_coef)
+{
+  SetNaturalBCs(trac_bdr, trac_bdr_coef);
+}
+
+void NonlinearSolidSolver::SetHyperelasticMaterialParameters(double mu, double K)
+{
+  delete m_model;
```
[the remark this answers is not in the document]
jamiebramwell: Yes! If you're handy with shared pointers, I would greatly appreciate at taking a stab at fixing `ThermalSolver`. It should be a bit more straight forward than this object, and we can convert this one after I get the old implementation operational.

jamiebramwell:
This is ready for review as all of the tests are passing.

src/solvers/nonlinear_solid_solver.hpp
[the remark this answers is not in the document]
jamiebramwell: I think I should keep it as it is for now since this PR is getting big enough as is. I'll add another issue to unify the nonlinear and linear implementations.

src/solvers/CMakeLists.txt
jamiebramwell: This should replace both `dynamic_solver` and `quasistatic_solver`.

.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
white238: you can't use comments in json... you can get around it by the following convention:

"_comment": "comment text goes here..."

jamiebramwell: Huh. This was autogenerated by vscode. I guess they strip this themselves.

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -0,0 +1,363 @@
...
+  // Form the gradient of the complete nonlinear operator
+  mfem::SparseMatrix *localJ = Add(1.0, m_M_form->SpMat(), m_dt, m_S_form->SpMat());
+  add(*m_v, m_dt, k, m_w);
+  add(*m_x, m_dt, m_w, m_z);
+  localJ->Add(m_dt*m_dt, m_H_form->GetLocalGradient(m_z));
+  m_jacobian = m_M_form->ParallelAssemble(localJ);
+  delete localJ;
+
+  // Eliminate the fixed boundary DOFs
+  mfem::HypreParMatrix *Je = m_jacobian->EliminateRowsCols(m_ess_tdof_list);
```
white238: Can you add a comment that you don't need the internally malloc returned matrix?

tests/serac_quasistatic_solver.cpp
```diff
@@ -38,64 +38,76 @@ TEST(quasistatic_solver, qs_solve)
...
+  // Pass the BC information to the solver object
+  solid_solver.SetDisplacementBCs(ess_bdr, &disp_coef);
+  solid_solver.SetTractionBCs(trac_bdr, &traction_coef);
+
+  // Set the material parameters
+  solid_solver.SetHyperelasticMaterialParameters(0.25, 10.0);
+
+  // Set the linear solver params
+  LinearSolverParameters params;
+  params.rel_tol = 1.0e-6;
```
white238: Should these magic numbers be in one place? I see a lot of them repeating and my inner CS person cringes.

static constexpr double whatever_it_should_be_called = 1.0e-6;

white238: This is more of a general question not to hold up the PR.

jamiebramwell: Unfortunately, these parameters will probably be different for every solve (I typically set them in all of my ALE3D decks). There should be reasonable defaults though. I think I put those in the struct definition?

jamiebramwell: I'll add an issue about this.

white238: I will take your word on that.

PR #18, opened 2020-03-11
Update Uberenv and allow testing Uberenv update in Serac

adrienbernede:
With this PR, I intend to update Uberenv to benefit from improvements and bug fixes.

I am also introducing multi-project integration at a small scale: The goal is to allow Uberenv to trigger a pipeline in Serac testing if updating Uberenv in Serac is safe. This would allow to test Uberenv in several projects instead of manually propagate.

Note:
Another work that went a bit off the rails.
This PR adds a specific workflow for master branch: builds with dependencies using uberenv.
This was necessary in this PR since otherwise uberenv was not used, and updating it wouldn’t make sense. So it’s kind of killing to birds with one stone.

I added some small improvements like the possibility to collapse log by steps, and better support for parallel compiling in build_and_test.sh (removing constraint on 4 parallel jobs only).


adrienbernede:
LGTM

adrienbernede:
(last comment is a sign off to trigger Gitlab CI)

jamiebramwell:
Is this PR obsolete with after your other one?

adrienbernede:
Sorry I missed the question. Which other PR?
This PR is not obsolete, but it’s not a priority neither.
I would first have to complete uberenv update.

jamiebramwell:
I'll reopen it since it's still in progress. Thanks for the update!

adrienbernede:
LGTM

.gitlab-ci.yml
```diff
@@ -14,6 +14,38 @@ stages:
...
+# Template
+.srun_build_script:
+  script:
+    - *run_update_uberenv
+    #Use pre-existing allocation if any
+    - JOBID=$(squeue -h --name=${PROJECT_ALLOC_NAME} --format=%A)
+    - ASSIGN_ID=$(if [[ -n "${JOBID}" ]]; then echo "--jobid=${JOBID}"; fi)
+    - EXEC_PREFIX="srun ${ASSIGN_ID}"
+    #DEPENDENCIES
+    - RESOURCES="-t 20 -N 1 -n 1 -c 4"
```
white238, line 31: The intel compiler took 22 minutes last time i ran it... might need to bump that up a bit.

adrienbernede: Yes, I just updated this branch, but haven’t merged it yet to my latest changes regarding allocation time. WIP.

adrienbernede:
LGTM

adrienbernede:
PR EDIT:
Another work that went a bit off the rails.
This PR adds a specific workflow for master branch: builds with dependencies using uberenv.
This was necessary in this PR since otherwise uberenv was not used, and updating it wouldn’t make sense. So it’s kind of killing to birds with one stone.

I added some small improvements like the possibility to collapse log by steps, and better support for parallel compiling in build_and_test.sh (removing constraint on 4 parallel jobs only).

.gitlab-ci.yml
```diff
@@ -7,12 +7,70 @@ variables:
...
 # There are no tests for now
 stages:
   - allocate_resources
   - build
   - release_resources
+  - build_with_deps
+
+# Whether and how to update uberenv
+.run_update_uberenv: &run_update_uberenv |
+  [[ -n "${UPDATE_UBERENV}" ]] && ./scripts/gitlab/update-uberenv.sh "${UPDATE_UBERENV}"
```
white238, line 21: Am I understanding this right that you are now updating the uberenv.py script every PR?  Because if that is so that is not okay.... PR's should be testing the code that was changed by the developer and not injecting possible failures itself.

adrienbernede: `run_update_uberenv` is called only in the master workflow (only for master branch) _and_ if the `UPDATE_UBERENV` is set. In short: it is called _only_ if the master pipeline is triggered by uberenv (never in the development process).

Also the update is local to the pipeline, not commited. The goal is to trace if updating is safe. Next step could be to add a manual job to effectively update uberenv if it has be proved safe.

adrienbernede: Again, this manual job would appear only is such pipeline. Providing an easy way to update `uberenv`.

This may seem of little use in serac. It has already prove useful to me for uberenv validation. But another benefit is to demonstrate this cross-project pipeline mechanism that can be very useful if other cases.

white238: Why not have that pipeline under uberenv? Since you are verifying uberenv not serac?

adrienbernede: There is a part in Uberenv. Here, I’m simply allowing the master pipeline to be triggered with and updated version of Uberenv. Placing this in Uberenv would only duplicate the existing pipeline.

Not all in this PR is dedicated to testing uberenv:
- This PR introduces a master pipeline that will also install the dependencies. So that this install get checked from time to time. (Could be made _scheduled_ only, like once a week).
- With an additional 10 lines of script, I can allow uberenv to use it to test itself in serac context.

white238: Okay, I understand better now.  Though, having a pipeline updating code that is not apart of the branch it's testing is confusing. Unless it is very specifically named to be that way, like "master_with_updated_uberenv".  "master" is also a very overloaded word in this context...

adrienbernede: You may think of it as in uberenv were a submodule. This is just testing if this submodule can be udated safely in master.

I see your point though, and it would be good idea to make this pipeline appear as a specific one more clearly. I can’t do that in this PR though, it’s a suggestion I will post on Gitlab :).

adrienbernede: Note: The pipeline does not run the master workflow at the moment because this branch is a PR... It certainly adds to the confusion. Sorry about that.

A demo of the master workflow running is found [here](https://lc.llnl.gov/gitlab/bernede1/serac/pipelines/11805).
Also, a demo of serac pipeline when triggered by uberenv is [here](https://lc.llnl.gov/gitlab/bernede1/serac/pipelines/11810).

adrienbernede: In the latter we see that the pipeline is triggered by an upstream project. So it’s less confusing.

jamiebramwell: So here's a dumb question: should uberenv just be a submodule?

jamiebramwell: So if I understand correctly, this is a test for an updated copy of uberenv. It sounds like the choice is either duplicate the Serac pipelines in the uberenv project or add a test uberenv pipeline to all of the dependent projects. Is that right @adrienbernede ? I don't really have an opinion yet. I'm just trying to understand the options.

.gitlab/ci/build_quartz.yml
```diff
@@ -4,62 +4,79 @@
...
+.pr_build_on_quartz:
   stage: build
-  script:
-    - JOBID=$(squeue -h --name=${PROJECT_ALLOC_NAME} --format=%A)
-    - JOBID=$(if [[ -n "${JOBID}" ]]; then echo "--jobid=${JOBID}"; fi)
-    - EXEC_PREFIX="srun ${JOBID} -t 20 -N 1 -n 1 -c 4"
-    - ${EXEC_PREFIX} scripts/gitlab/build_and_test.sh
+  extends: [.srun_build_script, .on_quartz, .pr_workflow]
+
+.master_build_with_deps_on_quartz:
```
jamiebramwell, line 42: This will definitely be nice to have (at least occasionally)!

adrienbernede: Currently, it’s setup to do this each time master is changed.
However, master is not mirrored... Most likely, someone with write access to master does not have 2 Factors Authentication (or is not in LLNL org?).

.gitlab-ci.yml
```diff
@@ -7,12 +7,70 @@ variables:
...
+####
+# Template
+.srun_build_script:
+  script:
+    #Use pre-existing allocation if any
+    - JOBID=$(squeue -h --name=${PROJECT_ALLOC_NAME} --format=%A)
+    - ASSIGN_ID=$(if [[ -n "${JOBID}" ]]; then echo "--jobid=${JOBID}"; fi)
+    - EXEC_PREFIX="srun ${ACCOUNT} ${ASSIGN_ID}"
+    #BUILD
+    - RESOURCES="-t 5 -N 1"
```
white238, line 40: How close are we to the 5 minute mark? That seems very close...

.gitlab/ci/build_quartz.yml
```diff
@@ -4,62 +4,79 @@
...
 # In pre-build phase, allocate a node for builds
 allocate_resources_build_quartz:
   variables:
     GIT_STRATEGY: none
-  extends: .on_quartz
+  extends: [.on_quartz, .pr_workflow]
   stage: allocate_resources
   script:
-    - salloc -N 1 -c 36 -p pdebug -t 30 --no-shell --job-name=${PROJECT_ALLOC_NAME}
+    - salloc ${ACCOUNT} -N 1 -c 36 -t 30 --no-shell --job-name=${PROJECT_ALLOC_NAME}
```
white238, line 22: This is a more general question I guess.  If this triggering pipelines from upstream projects becomes more common, would it be up to the downstream project to use their account? Is that even something we need to worry about? @jamiebramwell

adrienbernede: I never thought about it, but the way it is right now, the upstream project could in fact change the ACCOUNT value to use its own...
Here I am using radiuss account after approval of Rob that this is a valid use of it.

white238: I can see it both ways.. maybe we shouldn't make a fuss about whose account is used.

adrienbernede: Not yet... service users should become the norm, once it’s available. We shall re-discuss that then.

adrienbernede:
LGTM

Issue #23, opened 2020-03-13
Move away from mfem::Array types


jamiebramwell:
When possible, I think it would be best to go to `std::Vector` or `std::map` instead of `mfem::Array` types. This reduces the dependency on MFEM from a user perspective.

samuelpmishLLNL:
I just ran a small test where mfem::Array< FiniteElementState > (with std::shared_ptr) segfaults on deconstruction, but std::{map, vector}< FiniteElementState > containers don't. I don't know what the difference in implementation is, but it is a little bit concerning.

jamiebramwell:
I looks like it's using the fairly new memory manager classes [here](https://github.com/mfem/mfem/blob/v4.0/general/array.hpp), so I say we move to std::map/vector until we have a good reason not to.

samuelpmishLLNL:
A downside of the `std::map` for `m_state`: making an innocent typo in one of the field names will compile but result in potentially cryptic runtime errors. I wrote "displacment" instead of "displacement" initially, and it took me longer than I'd like to admit to understand what was going wrong.

white238:
Realize this is coming from a purely CS person. Do these have to be in the same data structure? It seems like an individual array that is named for each would be better. Or an array that holds a struct for each one.

jamiebramwell:
Yeah, I'm not sure what is the best. To have a common interface for each solver, we need to have some interface in the `BaseSolver` class which can lookup state information where each solver could have a different number of state variables. For example, the thermal solver only has a temperature field but the solid solver has displacement and velocity fields. Any thoughts?

jonwong12:
I think it depends on how you plan to arrange the solvers. One thing you could do is for solvers that are dependent on one another you could set the references to the gridfunctions from whatever code is organizing the coordinate flow of the solvers. Then you won't have to do any weird look-ups.

For the state, if you want strings, doesn't it make more sense to use std::unordered_map since there isn't really an order associated?

jamiebramwell:
I think there are two use casese we have to consider. The first is a simple proxy app type person who doesn't want to see any of the memory management guts. The second is a large integrated code that has their own data structures and memory managers. I'm not really sure what the best answer is, so we should probably have a WebEx to discuss this.

jamiebramwell:
I think this will be resolved by using a real data store as we discussed in the referenced webex.

PR #26, opened 2020-03-16
removing raw pointers and std::map for *Solver m_state

samuelpmishLLNL:
First stab at using some smart pointers in the solver classes. I didn't test the elasticity / solid dynamics examples much because that implementation is about to be significantly reworked by https://github.com/LLNL/serac/pull/5 in the near future.

I'm not really sure what the appropriate smart pointer type is for many of the members, but changing from `std::shared_ptr` to `std::unique_ptr` is pretty easy to do.

The new test `serac_dtor` used to fail for by `delete`ing things too many times, but passes now.


CMakeLists.txt
```diff
@@ -8,7 +8,7 @@
 # Basic CMake Setup
 #------------------------------------------------------------------------------
 
-cmake_minimum_required(VERSION 3.9)
+cmake_minimum_required(VERSION 3.8)
```
samuelpmishLLNL: Some of the host_configs for me were set to an older version of cmake, and this was just a quick workaround.

src/solvers/base_solver.cpp
```diff
@@ -11,25 +11,24 @@
...
-    m_gf_initialized(false)
-{
-  MFEM_ASSERT(state.Size() > 0, "State vector array of size 0 in BaseSolver constructor.");
-
-  MPI_Comm_rank(m_state[0].space->GetComm(), &m_rank);
-
-  m_ode_solver = new mfem::ForwardEulerSolver;
-
-}
+//BaseSolver::BaseSolver(std::map < std::string, mfem::Array<FiniteElementState> &state)
```
white238: Don't leave commented code.  It is either needed (and uncommented) or should be deleted.

src/solvers/base_solver.cpp
```diff
@@ -55,57 +54,56 @@ void BaseSolver::SetNaturalBCs(mfem::Array<int> &nat_bdr, mfem::Coefficient *nat
   m_nat_bdr_coef = nat_bdr_coef;
 }
 
-void BaseSolver::SetState(const mfem::Array<FiniteElementState> &state)
+void BaseSolver::SetState(const std::map<std::string, FiniteElementState> &state)
```
white238: Can you explain why you changed from an array to an ordered map?

samuelpmishLLNL: related to https://github.com/LLNL/serac/issues/23

Also in an early test, the mfem::Array was segfaulting on deconstruction when it contained std::shared_ptrs while std containers weren't.

src/solvers/base_solver.cpp
samuelpmishLLNL: Good point. I stubbed it out just to get things compiling. It wasn't used in any of the tests and is kind of a cumbersome constructor anyway, so I'm leaning toward just removing it.

src/common/serac_types.hpp
```diff
@@ -54,11 +56,11 @@ struct LinearSolverParameters {
 
 // Finite element information bundle
 struct FiniteElementState {
-  mfem::ParFiniteElementSpace* space = nullptr;
-  mfem::FiniteElementCollection* coll = nullptr;
-  mfem::ParGridFunction* gf = nullptr;
-  mfem::ParMesh* mesh = nullptr;
-  mfem::Vector* true_vec = nullptr;
+  std::shared_ptr < mfem::ParFiniteElementSpace > space;
```
jonwong12: I feel like space, coll, and mesh should be getters to gf. Otherwise there could be a danger that space, coll, mesh, gf could be inconsistent.

jonwong12:
It seems like m_state should have a Getter based on the "name" of the state. This ways you have an opportunity to check for "typos" and if the key exists. This would help with the spelling errors.

src/solvers/base_solver.cpp
jonwong12: As stated in #23, perhaps std::unordered_map would be another option as there isn't any explicit ordering with the state mapping.

samuelpmish:
I had already reverted some of the `std::map` stuff before our meeting. This PR will be just for the shared_ptrs.

src/solvers/base_solver.cpp
```diff
@@ -190,15 +174,3 @@ void BaseSolver::OutputState() const
     mfem::mfem_error("OutputType not recognized!");
   }
 }
-
-BaseSolver::~BaseSolver()
```
jamiebramwell: 👍

src/solvers/base_solver.hpp
```diff
@@ -7,6 +7,8 @@
 #ifndef BASE_SOLVER
 #define BASE_SOLVER
 
+#include <map>
```
jamiebramwell: Is this still needed?

src/solvers/thermal_solver.hpp
```diff
@@ -110,43 +113,43 @@ class DynamicConductionOperator : public mfem::TimeDependentOperator
...
   /// Pointer to the eliminated T matrix
-  mfem::HypreParMatrix *m_T_e_mat;
+  std::shared_ptr < mfem::HypreParMatrix > m_T_e_mat;
 
   /// Assembled RHS vector
-  mfem::Vector *m_rhs;
+  std::shared_ptr < mfem::Vector > m_rhs;
 
   /// RHS vector including essential boundary elimination
   mfem::Vector *m_bc_rhs;
```
jamiebramwell: I assume the raw pointer is needed here due to interactions with MFEM?

src/solvers/base_solver.hpp
samuelpmish: good call, I missed this one when reverting the std::map stuff

src/solvers/thermal_solver.hpp
samuelpmish: Nah, I was just on the fence about reworking the B.C. implementation. In hindsight though, it makes more sense to break those up into separate changes (e.g. move to smart pointers, then consider reorganizing boundary condition stuff). Hopefully I can get things building locally and finish inserting smart pointers into the rest of the code!

samuelpmish:
One other hiccup: my 2FA for my LLNL github account is tied to my laptop which is still on-site. So I'm stuck with my personal account for the next few weeks. This rules out my ability to merge pull requests or anything that requires write access to the repo.

jamiebramwell:
> 
> 
> One other hiccup: the 2FA for my LLNL github account is tied to my work laptop, which is still on-site. So I'm stuck with my personal account for the next few weeks. This rules out my ability to merge pull requests or anything that requires write access to the repo.

OK. Let me know when you think this is ready and I can merge it.

jamiebramwell:
@jonwong12 @white238 any other thoughts on this branch? I'd like to get this merged in since it fixes the ubuntu builds for local development.

src/common/serac_types.hpp
jonwong12: With MFEM I believe you always have to make sure the space, coll, mesh are in scope, both with raw and smart pointers otherwise the gridfunction doesn't really make sense. Maybe what you are getting at is how to best guarantee the existence of space, coll, mesh for a given rid function. I think `FiniteElementState` could take in a shared_ptrs and then construct a ParGridFunction for you. I forget what happens in the case of NCMeshes.

I still think this should be abstracted into a class. If it's not in this PR, we should put it in a separate one. There's a big benefit to using Getters to access Space() and so forth as we can change the underlying guts without messing with the rest of the code. This would allow us to change to enums or under the hood seamlessly.

src/solvers/base_solver.cpp
```diff
@@ -55,57 +44,56 @@ void BaseSolver::SetNaturalBCs(mfem::Array<int> &nat_bdr, mfem::Coefficient *nat
...
 }
 
-void BaseSolver::SetState(const mfem::Array<FiniteElementState> &state)
+void BaseSolver::SetState(const std::vector<FiniteElementState> &state)
 {
   MFEM_ASSERT(state.Size() > 0, "State vector array of size 0 in BaseSolver::SetState.");
   m_state = state;
 
-  MPI_Comm_rank(m_state[0].space->GetComm(), &m_rank);
+  MPI_Comm_rank(m_state.begin()->space->GetComm(), &m_rank);
```
white238: Please don't use the iterator unless you are iterating...

src/solvers/base_solver.cpp
```diff
@@ -127,25 +115,20 @@ int BaseSolver::GetCycle() const
   return m_cycle;
 }
 
-void BaseSolver::InitializeOutput(const OutputType output_type, std::string root_name,
-                                  const mfem::Array<std::string> names)
+void BaseSolver::InitializeOutput(const OutputType output_type, std::string root_name)
```
white238: Did names disappear as a conflict or should the assert be removed?

src/solvers/base_solver.cpp
```diff
@@ -175,13 +158,14 @@ void BaseSolver::OutputState() const
   }
 
   case OutputType::GLVis: {
-    for (int i=0; i<m_state.Size(); ++i) {
+
+    for (auto & state : m_state) {
```
white238: I find state vs m_state a bit confusing.  Should we rename m_state to m_states? I like to clarify when my containers that hold multiple things.....

src/solvers/elasticity_solver.cpp
```diff
@@ -71,15 +72,15 @@ void ElasticitySolver::CompleteSetup()
...
+  m_l_form = new mfem::ParLinearForm(displacement.space.get());
 
   // Add the traction integrator
   if (m_nat_bdr_vec_coef != nullptr) {
     m_l_form->AddBoundaryIntegrator(new mfem::VectorBoundaryLFIntegrator(*m_nat_bdr_vec_coef), m_nat_bdr);
     m_l_form->Assemble();
     m_rhs = m_l_form->ParallelAssemble();
   } else {
-    m_rhs = new mfem::HypreParVector(m_state[0].space);
+    m_rhs = new mfem::HypreParVector(displacement.space.get());
```
white238: Yay for names that mean something to dumb me.

src/solvers/thermal_solver.cpp
```diff
@@ -6,32 +6,32 @@
 
 #include "thermal_solver.hpp"
 
+const int num_fields = 1;
```
white238: Minor nitpick, please use "constexpr int" instead of "const int" since constexpr's are evaluated at compile time not run time when you do hard-coded "magic" numbers.

src/solvers/thermal_solver.hpp
```diff
@@ -22,38 +22,41 @@ class DynamicConductionOperator;
...
   /// Linear solver for the K operator
-  mfem::CGSolver *m_K_solver;
+  std::shared_ptr < mfem::CGSolver > m_K_solver;
 
   /// Preconditioner for the K operator
-  mfem::HypreSmoother *m_K_prec;
+  std::shared_ptr < mfem::HypreSmoother > m_K_prec;
 
   /// Conduction coefficient
   mfem::Coefficient *m_kappa;
```
white238: Is there a reason m_kappa and m_source didn't get turned into shared? Should the reason be in a comment?

samuelpmish:
So, it looks like the dynamic test is failing to set up boundary condition attributes correct, after merging the new implementation via nonlinear_solid_solver. I spent some time looking into this, but without much success.

jamiebramwell:
> 
> 
> So, it looks like the dynamic test is failing to set up boundary condition attributes correct, after merging the new implementation via nonlinear_solid_solver. I spent some time looking into this, but without much success.

This sounds vaguely familiar, so I'll take a look after this is merged.

src/solvers/base_solver.cpp
samuelpmish: oops, this was leftover from when I tried out std::map, so we had no m_states[0]! Fixed in upcoming code-style PR.

src/solvers/base_solver.cpp
samuelpmish: renamed to m_states

src/solvers/base_solver.cpp
samuelpmish: more leftover std::map stuff

src/solvers/thermal_solver.cpp
samuelpmish: [constexpr ALL the things!](https://www.youtube.com/watch?v=PJwd4JLYJJY). Addressed in next PR.

src/solvers/thermal_solver.hpp
samuelpmish: Good question, I was thinking of adopting a convention: std::shared_ptr / std::unique_ptr for pointers where we own the data, and raw pointers / references for views (kappa and source aren't given in the constructor, so we can't initialize reference variables there).

Comments to explain that are a good idea though!

PR #27, opened 2020-03-17
StdFunctionCoefficients, Attribute, and Essential BC infastructure

jonwong12:
This PR includes:
- A few coefficients that are useful in combining and making new temporary coefficients based off of functions.
- Helper functions for making lists of boundary conditions, attribute lists per element given a coefficient, and attribute lists for bdr element attributes
- AttributeModifierCoefficient which will assign temporary attributes on the fly and can effectively wrap another coefficient.

A set of unit tests are included to demonstrate some of these functions and coefficients.


src/coefficients/CMakeLists.txt
```diff
@@ -6,11 +6,13 @@
 
 set(coefficients_sources
     loading_functions.cpp
+    StdFunctionCoefficient.cpp
     )
 
 set(coefficients_headers
     loading_functions.hpp
     traction_coefficient.hpp
+    StdFunctionCoefficient.hpp
```
white238: Please follow the file name conventions of the repo. We can discuss what we should be using if anyone cares but please be consistent.

src/coefficients/StdFunctionCoefficient.hpp
```diff
@@ -0,0 +1,167 @@
...
+/**
+    @file StdFunctionCoefficient.hpp
+ */
+
+#ifndef STD_FUNCTION_COEFFICIENT_HPP
+#define STD_FUNCTION_COEFFICIENT_HPP
+#include <functional>
+#include "mfem.hpp"
+
+using namespace mfem;
```
white238: Please do not do any using outside of a function if at all.  It confuses where types are coming from.

src/coefficients/StdFunctionCoefficient.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+StdFunctionVectorCoefficient::StdFunctionVectorCoefficient(int dim,
+    std::function<void(mfem::Vector &, mfem::Vector &)> func) :
+  VectorCoefficient(dim),
+  func_(func)
+{ }
+
+void StdFunctionVectorCoefficient::Eval(Vector &V, ElementTransformation &T, const IntegrationPoint &ip)
+{
+  double x[T.GetSpaceDim()];
```
white238: Variable Length Arrays were dropped off the standard in C99.  They should be avoided due to undefined behavior and some compilers will not compile them.  "-pedantic" will add the warning.  We should probably address our warning/error flags.

Please use a shared_ptr or a unique_ptr.  There is a PR that @samuelpmishLLNL  is doing to look into moving towards those.  If you don't feel comfortable with those change to new/delete and we can update it after.

src/coefficients/StdFunctionCoefficient.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+   This is a place holder until the coefficient of the same name is merged into mfem.
+
+*/
+class StdFunctionCoefficient : public Coefficient
+{
+public:
+  /// Constructor that takes in an mfem Vector representing the coordinates and produces a double
+  StdFunctionCoefficient(std::function<double(mfem::Vector &)> func);
+  virtual ~StdFunctionCoefficient() {}
```
white238: Do you plan on someone inheriting this class? Otherwise the virtual is not necessary.

src/coefficients/StdFunctionCoefficient.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+  virtual double Eval (ElementTransformation &Tr, const IntegrationPoint &ip);
+private:
+  std::function<double(mfem::Vector &)> func_;
```
white238: Not sure i like having a trailing underscore to indicate its a function/callable object. If that is what you are doing.  Is this a convention you've seen before? Thoughts from the team?

src/coefficients/StdFunctionCoefficient.hpp
```diff
@@ -0,0 +1,167 @@
...
+   \param[in] m The mesh
+   \param[inout] attr_list Should be an array that will hold the attributes that correspond to each element
+   \param[in] c The coefficient provided that will be evaluated on the mesh
+   \param[in] digitize An optional function that can be called to assign attributes based on the value of c at a given projection point. By default, values of c at a given d.o.f that are > 0. are assigned attribute 2, otherwise attribute 1.
+
+*/
+void MakeAttributeList(Mesh &m, Array<int> &attr_list,
+                       Coefficient &c,
+                       std::function<int(double)> = [](double v)
+{
```
white238: I'm curious, was this formatting what astyle returned?

tests/StdFunctionCoefficient.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+using namespace mfem;
+using namespace std;
+
+
+int main(int argc, char **argv)
+{
+  MPI_Init(&argc, &argv);
+  int nranks;
+  MPI_Comm_size(MPI_COMM_WORLD, &nranks);
```
white238: This looks unnecessary.

src/coefficients/CMakeLists.txt
jonwong12: Alright. I will make the file names pot_hole_case. 

FYI. It seems at the moment the repo uses pot_hole_case for file names, camel_Pot_hole for member names, and UpperCamelCase for Classes and Method names. I would not mind making it further consistent.

src/coefficients/StdFunctionCoefficient.hpp
jonwong12: I don't. Do you think I should prohibit inheriting from this class?

src/coefficients/StdFunctionCoefficient.hpp
```diff
@@ -0,0 +1,167 @@
...
+// SPDX-License-Identifier: (BSD-3-Clause)
+// # Author: Jonathan Wong @ LLNL.
+
+/**
+    @file StdFunctionCoefficient.hpp
+ */
+
+#ifndef STD_FUNCTION_COEFFICIENT_HPP
+#define STD_FUNCTION_COEFFICIENT_HPP
+#include <functional>
```
white238: add new line after header guards please.

src/coefficients/StdFunctionCoefficient.hpp
white238: I say remove it for now and we can address inheritance when it comes up.  Or at least when we (and i really mean not me) sit down and figure a use case...

src/coefficients/StdFunctionCoefficient.hpp
jonwong12: Yes. I ran `make style`. I'm sure it ran because it said it changed a bunch of files.

white238: K.  At some point, I need to sit down and evaluate which is the best for lambda's.  Because that's not a helpful format.  Though depending on who you ask every tool is the worst and best at lambdas.

src/coefficients/StdFunctionCoefficient.hpp
jonwong12: Someone working on another project started the convention instead of "m_"-prefix. This is what happens when I re-use code from other projects.

src/coefficients/CMakeLists.txt
jamiebramwell: Yeah, I don't really care what we settle on as long as its consistent. I think I tried to come up with standards, but I probably missed some stuff. I don't really have opinions on one convention over another.

white238:
Do we want more than one class in a header?

jamiebramwell:
> 
> 
> Do we want more than one class in a header?

I have more than one in the solver classes for MFEM operators that are closely coupled with their related solvers. I think it's OK as long as they're clearly related, but I don't have a super strong opinion.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+*/
+void MakeAttributeList(mfem::Mesh &m, mfem::Array<int> &attr_list,
+                       mfem::Coefficient &c,
+                       std::function<int(double)> = [](double v)
+{
+  return v > 0. ? 2 : 1;
+});
+
+/**
```
jamiebramwell: I like this doxygen standard. I'll make myself an issue to get the other files in this format.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+protected:
+  const mfem::Array<int> & m_attr_list;
+  mfem::Coefficient *m_C;
+};
+
+/**
+   TransformedVectorCoefficient applies various operations to modify a VectorCoefficient
+*/
+class TransformedVectorCoefficient : public mfem::VectorCoefficient
```
jamiebramwell: Yeah, the more I'm looking at this, the more I think it should be separate files. That can be a follow-up change though.

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -0,0 +1,142 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+// # Author: Jonathan Wong @ LLNL.
```
jamiebramwell: So I realized I technically said to do this in the contributing file. However in full disclosure I just copied it from BLT. I think a statement about "if you would like to denote authorship, feel free to do so. Otherwise it is tracked by GitHub." would suffice.

PR #30, opened 2020-03-18
applying clang-format to each source file

samuelpmish:
In short: specify a format in the `.clang-format` file

The current formatting can be applied to each C++ file in the repo with the one-liner

`clang-format -i $(git ls-files *.cpp && git ls-files *.hpp)`

[this](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) is a useful link for describing the .clang-format file

[this](https://zed0.co.uk/clang-format-configurator/) is an interactive way to see how the different options affect the code


src/common/serac_types.hpp
```diff
@@ -11,10 +11,7 @@
 
 // Option bundling enums
 
-enum class OutputType {
-  GLVis,
-  VisIt
-};
+enum class OutputType { GLVis, VisIt };
```
white238: Dont like this.

samuelpmish: It seems to just be whether or not the definition can fit on a single line or not.

white238: Apparently to get around this you have to add a comment to every line. sigh...

enum Mode {
  NONE,         // Comments
  SPLIT_FILES,  // On
  SINGLE_FILE,  // Every
  INVALID       // Line
};

adrienbernede: You could locally deactivate clang format.

samuelpmish: there's a few simple workarounds:
- a single comment inside the enum (don't need all lines, but you can if you want)
- change the linewidth setting
- locally disable formatting with
```
// clang-format off
    void    unformatted_code  ;
// clang-format on
```

white238: Thats better.

white238:
I'm trying to hack a quick clang-format support into BLT... but that should hold up this PR.

src/common/serac_types.hpp
white238: I don't like the line width option for obvious reasons... just turning off clang-format is going to be helpful in a lot of places.

adrienbernede: Obvious but maybe still of personal taste?
Instead of using it in many places, maybe then changing the rule in clang-format config is the best option.

white238: I'll clarify.. Some people online suggest setting the line width to 0 so it doesn't do this.  Unless I read wrong, that line width option is not just for enums for the entire code.

On a similar note, stay with 120 length.

adrienbernede: Per [clang-format doc](https://clang.llvm.org/docs/ClangFormatStyleOptions.html):
bool AfterEnum Wrap enum definitions.
```
true:
enum X : int
{
  B
};

false:
enum X : int { B };
```

adrienbernede: Definitely agree the line width = 0 wouldn’t help!

white238:
https://github.com/LLNL/blt/pull/350

This PR should go through soon (probably tomorrow).  After it goes through we should update to the newest develop for BLT and utilize it.  It will enable the normal check and style build targets. To do so change this line: https://github.com/LLNL/serac/blob/5db79168d80c487ed7280b569fdcabd18fffb77e/cmake/SeracMacros.cmake#L50 to "CLANGFORMAT_CFG_FILE ${PROJECT_SOURCE_DIR}/.clang-format"

samuelpmish:
I'm not sure how closely we want to follow the Google C++ style guide. For example, our variable naming is in conflict with their convention, which would introduce quite a bit of churn. Some other things are easy enough to change though (i.e. [#include precedence / formatting](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes)).

.clang-format
```diff
@@ -0,0 +1,120 @@
...
+  - Regex: '^<.*\.h>'
+    Priority: 1
+  - Regex: '^<.*'
+    Priority: 2
+  - Regex: '.*'
+    Priority: 3
+IncludeIsMainRegex: '([-_](test|unittest))?$'
+IndentCaseLabels: true
+IndentPPDirectives: None
+IndentWidth: 3
```
white238: Sorry I know I gave this to you but go back to 2 indenting...

samuelpmish: same for 
```
ConstructorInitializerIndentWidth: 3
ContinuationIndentWidth: 3
```
as well?

white238: yea

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -44,18 +46,20 @@ void MakeEssList(mfem::Mesh &m, mfem::VectorCoefficient &c, mfem::Array<int> & e
   ess_vdof_list.SetSize(0);
 
   for (int v = 0; v < fes.GetNV(); v++)
```
white238: Is there an option to enforce { on all if's and for's. Because this is just asking for very hard to find mistakes.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -148,18 +151,20 @@ class TransformedVectorCoefficient : public mfem::VectorCoefficient
...
+    mfem::VectorCoefficient *v1,
+    mfem::VectorCoefficient *v2,
+    std::function<void(mfem::Vector &, mfem::Vector &, mfem::Vector &)> func);
+  virtual void Eval(mfem::Vector &V,
+                    mfem::ElementTransformation &T,
+                    const mfem::IntegrationPoint &ip);
 
-private:
+  private:
   mfem::VectorCoefficient *m_v1;
```
white238: private should not be at the same level as the member's.. private/public is usually not indented.

.clang-format
```diff
@@ -0,0 +1,120 @@
+# requires clang 9
+# https://releases.llvm.org/9.0.0/tools/clang/docs/ClangFormatStyleOptions.html
+---
+# BasedOnStyle:  Google
+Language: Cpp
+
+AccessModifierOffset: -3
```
white238: I think this got confused when we went from 3 to 2... try -2

src/drivers/serac.cpp
```diff
@@ -45,96 +45,117 @@ int main(int argc, char *argv[])
...
+  args.AddOption(&nonlin_params.print_level, "-npl", "--newton-print-level", "Newton print level.");
+  args.AddOption(&t_final, "-tf", "--t-final", "Final time; start time is 0.");
+  args.AddOption(&dt, "-dt", "--time-step", "Time step.");
 
   // Parse the arguments and check if they are good
   args.Parse();
-  if (!args.Good()) {
-    if (myid == 0) {
+  if (!args.Good())
+  {
```
white238: Can we go back to the old way on this { setting? I think its the brace wrapping settings.  Everyone's thoughts?

src/solvers/base_solver.cpp
```diff
@@ -4,193 +4,183 @@
...
-    m_ode_solver = std::make_shared< mfem::SDIRK23Solver >();
-    break;
-  case TimestepMethod::SDIRK34    :
-    m_ode_solver = std::make_shared< mfem::SDIRK34Solver >();
-    break;
-  default:
-    mfem::mfem_error("Timestep method not recognized!");
+  switch (m_timestepper)
+  {
+    case TimestepMethod::QuasiStatic: m_ode_solver = nullptr; break;
```
white238: its nice that the case is indented but the break got moved which is less than ideal.

white238:
This looks like a good step forward! A few tweaks and it should be good to merge.

src/drivers/serac.cpp
jamiebramwell: Slight preference for the old way, but I'm OK with either as long as it's consistent.

src/solvers/base_solver.cpp
jamiebramwell: I'm generally anti multiple statements on one line.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -74,12 +77,11 @@ void MakeEssList(mfem::Mesh &m, mfem::VectorCoefficient &c, mfem::Array<int> & e
...
-                       mfem::Coefficient &c,
-                       std::function<int(double)> = [](double v)
-{
-  return v > 0. ? 2 : 1;
-});
+void MakeAttributeList(
+  mfem::Mesh &m,
+  mfem::Array<int> &attr_list,
+  mfem::Coefficient &c,
+  std::function<int(double)> = [](double v) { return v > 0. ? 2 : 1; });
```
jamiebramwell: Same comment about single line definitions.

src/drivers/serac.cpp
```diff
@@ -168,10 +191,10 @@ int main(int argc, char *argv[])
   mfem::VectorCoefficient *defo_coef = new mfem::VectorFunctionCoefficient(dim, InitialDeformation);
 
   mfem::Vector velo(dim);
-  velo = 0.0;
+  velo                                       = 0.0;
```
jamiebramwell: This seems weird. Seems like it's keying off the the line below? Not sure if we can fix this. The equal spacing was really nice above...

.vscode/launch.json
```diff
@@ -1,20 +1,111 @@
...
+            "setupCommands": [
+                {
+                    "description": "Enable pretty-printing for gdb",
+                    "text": "-enable-pretty-printing",
+                    "ignoreFailures": true
+                }
+            ]
+        },
+        {
+            "name": "(gdb) mpi Dynamic Test",
```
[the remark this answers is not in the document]
samuelpmish: These need to get changed or removed, since they're specific to my paths and compiler choice. For some reason, with the mpi ones, the variable for the cmake build directory didn't expand correctly (missing slashes) for the prefix to the binary, so I had to put the entire path to get things working.

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -5,36 +5,38 @@
...
-StdFunctionVectorCoefficient::StdFunctionVectorCoefficient(int dim,
-    std::function<void(mfem::Vector &, mfem::Vector &)> func) :
-  mfem::VectorCoefficient(dim),
-  m_func(func)
-{ }
+StdFunctionVectorCoefficient::StdFunctionVectorCoefficient(
+  int dim,
+  std::function<void(mfem::Vector &, mfem::Vector &)> func)
+  : mfem::VectorCoefficient(dim)
+  , m_func(func)
```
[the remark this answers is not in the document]
samuelpmish: the clang-format file Chris gave me is based on Marble's so it's not really Google C++ style any more.

src/coefficients/stdfunction_coefficient.cpp
samuelpmish: I think clang-format is supposed to only do whitespace changes, so inserting characters to conform to the guidelines isn't an option. clang-tidy is capable of more complicated changes, but is more work to set up and maintain.

I prefer for loops with braces as well, so I'll go ahead and manually insert them in this case.

src/drivers/serac.cpp
samuelpmish: I'd prefer to not have separate rules about braces for loops, conditionals, structs, classes, etc. With that in my mind, my vote is braces-on-same-line for everything.

samuelpmish:
Should we consider reverting to the Google style? It seems like there are noticeably more objections to the code formatting since changing to Marble's clang-format file.

white238:
Sounds good @samuelpmishLLNL

samuelpmish:
Went back to Google style, with some alignment options enabled as per Jerome's comments. Previously, this PR had some warning fixes and suggestions from @white238 that I completely missed in the remove_raw_pointers PR. Now, this is purely formatting-related content, and the other fixes will be separated so they're not lost in the noise.

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -7,35 +7,32 @@
...
+StdFunctionVectorCoefficient::StdFunctionVectorCoefficient(
+    int dim, std::function<void(mfem::Vector &, mfem::Vector &)> func)
+    : mfem::VectorCoefficient(dim), m_func(func) {}
 
-void StdFunctionVectorCoefficient::Eval(mfem::Vector &V, mfem::ElementTransformation &T,
-                                        const mfem::IntegrationPoint &ip)
-{
+void StdFunctionVectorCoefficient::Eval(mfem::Vector &                V,
+                                        mfem::ElementTransformation & T,
+                                        const mfem::IntegrationPoint &ip) {
```
white238: Can you adjust the brace to the next line for functions?

white238:
Please go back to 120 ColumnLength

white238:
Thanks for the many iterations on this!

PR #36, opened 2020-03-25
Fixing serac CI on Gitlab: back to the basics [fix/ci]

adrienbernede:
### Intro
I have fixed Serac CI on LC Gitlab.

My mistake was to try to use the same pipeline template I use for other projects, while I could have kept things simple.

I simplified the pipeline, which means less optimization (produces more data, is longer to run), but better robustness.

### What
The result is a 3-stage pipeline, allocation/run/release. (I kept the "single allocation" optimization)

The run phase consists in three streps (in one job) aiming at reproducing the development workflow:
- install dependencies and generate the host-config file (uberenv)
- build with classic cmake commands (not using uberenv)
- test

This is done for 3 compilers (clang@4.0.0, gcc@8.1.0, intel@19.0.4) on quartz only and could (should?) be extended to lassen (just tell be what you want).

### Comments
The pipeline still fails, because some tests fails when built with gcc@8.1.0
Someone more into the project should give a look at this...

### TODO
- I found a way to control the collapse of logs by sections in GitLab, I will test it. It would make the all-in-one job log easier to read.
- I want to update Uberenv. (not necessarily in this PR)
- Add other targets (more recent clang? lassen?)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,9 +6,8 @@
 variables:
   GIT_SUBMODULE_STRATEGY: recursive
   SLURM_ALLOC_NAME: serac_$CI_COMMIT_SHORT_SHA
-  BUILD_ROOT: $CI_BUILDS_DIR/serac_$CI_COMMIT_REF_SLUG
-  LIBS_ROOT: $CI_BUILDS_DIR/serac_libs_$CI_COMMIT_REF_SLUG
-  INSTALL_ROOT: $CI_BUILDS_DIR/spack_uberenv
+  BUILD_ROOT: $CI_BUILDS_DIR/serac/${CI_COMMIT_REF_SLUG}_${CI_PIPELIN_ID}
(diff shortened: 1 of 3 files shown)


adrienbernede:
LGTM

adrienbernede:
(Last comment is a sign-off for Gitlab)

jamiebramwell:
This looks great!! Thank you for your work!

> This is done for 3 compilers (clang@4.0.0, gcc@8.1.0, intel@19.0.4) on quartz only and could (should?) be extended to lassen (just tell be what you want).

Eventually we should add lassen, but the core code isn't GPU-ized yet. It can wait for now.

> The pipeline still fails, because some tests fails when built with gcc@8.1.0

Yep, this is a known issue. [This recently merged PR](https://github.com/LLNL/serac/pull/33) should fix a bunch of it, and I have a new branch that should fix the rest.

adrienbernede:
Then I merge it, rather than adding non mandatory changes.

PR #38, opened 2020-03-26
Fixing the GCC tests

jamiebramwell:
This is work toward fixing the GCC tests.


jamiebramwell:
LGTM

adrienbernede:
@jamiebramwell Nice use of the sign-off mechanism!
I’m investigating why it failed...

jamiebramwell:
> 
> 
> @jamiebramwell Nice use of the sign-off mechanism!
> I’m investigating why it failed...

Ha! I was sneakily investigating this to show off your work in the group meeting today. I'm slowly learning... 🐢

adrienbernede:
Erf... I wish it worked then!

adrienbernede:
Found the cause, but lunch is calling!

adrienbernede:
Should be fixed by #39

jamiebramwell:
I'll check it out! This was probably it.

jamiebramwell:
Looks like it passes in serial, but not parallel. The hunt continues! Thanks for the help @samuelpmish !

jamiebramwell:
Getting close! Fixed the linear elastic test and now i'm going after the parallel dynamic one.

adrienbernede:
Good hunt!

jamiebramwell:
The tests now pass using intel, gcc, and clang. 🎉 I'm going to try the quartz builds using gitlab.

jamiebramwell:
LGTM

adrienbernede:
Can’t wait to see the results. Those three red dots and seeing the progress in your branch was my only thrill lately...

adrienbernede:
Yippee! 🎉

jamiebramwell:
Thanks @adrienbernede !!

src/drivers/serac.cpp
```diff
@@ -223,7 +219,7 @@ int main(int argc, char *argv[])
 
   bool last_step = false;
 
-  mfem::Array<std::string> names(2);
+  std::vector<std::string> names(2);
   names[0] = "Deformation";
   names[1] = "Velocity";
```
samuelpmish, line 224: slightly shorter option: 

`std::vector<std::string> names = {"Deformation", "Velocity"};`

src/solvers/elasticity_solver.cpp
```diff
@@ -25,7 +25,7 @@ ElasticitySolver::ElasticitySolver(int order, mfem::ParMesh *pmesh)
 {
   displacement.mesh     = pmesh;
   displacement.coll     = std::make_shared<mfem::H1_FECollection>(order, pmesh->Dimension(), mfem::Ordering::byVDIM);
-  displacement.space    = std::make_shared<mfem::ParFiniteElementSpace>(pmesh, displacement.coll.get());
+  displacement.space    = std::make_shared<mfem::ParFiniteElementSpace>(pmesh, displacement.coll.get(), pmesh->Dimension());
```
samuelpmish, line 28: Oops! Thanks for catching my mistake.

src/solvers/base_solver.hpp
```diff
@@ -22,10 +22,10 @@ class BaseSolver {
   mfem::BlockVector *m_block;
 
   /// Essential BC markers
-  mfem::Array<int> m_ess_bdr;
+  std::shared_ptr<mfem::Array<int> > m_ess_bdr;
```
samuelpmish, line 25: why move to `std::shared_ptr < mfem::Array >`?

.gitignore
```diff
@@ -6,3 +6,5 @@ spack-build*
 /src/docs/doxygen/html/
 /src/docs/doxygen/xml/
 /src/docs/sphinx/_build/
+*.cmake
+.vscode/
```
adrienbernede, line 10: Note: some projects (MFEM 😛) don’t like to use the local .gitignore to ignore IDE config files. The reason is to prevent .gitignore from containing all the possible IDEs config file.

I think it’s not a bad idea, especially since the alternative is quite convenient: 
You can ignore those file globally.

Add the vscode pattern in a file named, for example `~/.gitignore_global`
Run `git config --global core.excludesfile ~/.gitignore_global` (one-time command to register you global gitignore file)

src/solvers/base_solver.hpp
jamiebramwell: I was thinking that any user-facing wrappers should use vectors instead of MFEM arrays to reduce the dependence on MFEM. It was pretty easy to use the raw pointers to populate a data structure for use with the internal MFEM functions.

jamiebramwell: Ahh I realized I skipped a step in my response. I want the user to pass in a std::vector instead of an mfem::array. This means I need to construct an mfem::array internally on the fly instead of just passing by reference. I am trying to move away from my old way of just using raw pointers.

.gitignore
jamiebramwell: True. I generally like to have clean git status output and not pollute my global git config since I work on many projects. However, I don't feel that strongly about it so others (@white238 , @samuelpmish , @jonwong12 ) should definitely chime in.

src/solvers/base_solver.hpp
jamiebramwell: Moved to copying the std::vector into an mfem::array. The performance hit is acceptable since these will typically be small objects.

.gitignore
white238: It's fine the way it is.

PR #42, opened 2020-03-27
NonlinearformIntegrator wrappers

jonwong12:
This PR adds NonlinearFormIntegrator wrappers for BilinearFormIntegrators and LinearFormIntegrators such that existing linear problems can more easily be posited as a nonlinear problem.

- A test `wrapper_tests` is included to recast a linear problem as a nonlinear one using a NewtonSolver. Linear vs Nonlinear vs Nonlinear (MixedBilinearForm calls) is tested for a simple toy problem. Boundary conditions are designated by previous StdFunction calls.


src/integrators/wrapper_integrator.hpp
```diff
@@ -0,0 +1,86 @@
...
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#ifndef WRAPPER_INTEGRATOR_HPP
+#define WRAPPER_INTEGRATOR_HPP
+
+#include "mfem.hpp"
+
+/// A class to convert linearform integrators into a nonlinear residual-based one
+class LinearNonlinearFormIntegrator : public mfem::NonlinearFormIntegrator
```
jamiebramwell, line 13: I'm afraid this name might be a bit confusing, but I don't really have a better answer. Maybe `LinearToNonlinearFormIntegrator`?

jonwong12: Sure. I can change that. I'll do the same with the other ones.

jamiebramwell:
LGTM

src/integrators/wrapper_integrator.cpp
```diff
@@ -0,0 +1,80 @@
...
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+# include "wrapper_integrator.hpp"
+
+LinearToNonlinearFormIntegrator::LinearToNonlinearFormIntegrator( mfem::LinearFormIntegrator & f,
+    mfem::ParFiniteElementSpace * trial_fes)
+  :
+  m_f(f), m_trial_fes(trial_fes)
```
white238, line 12: Run make style on this PR please. It should work on any checked in host-config now

src/integrators/wrapper_integrator.hpp
```diff
@@ -0,0 +1,86 @@
...
+  /// Compute the residual vector => -F
+  virtual void AssembleElementVector( const mfem::FiniteElement &el, mfem::ElementTransformation &Tr,
+                                      const mfem::Vector &elfun, mfem::Vector & elvect);
+
+  /// Compute the tangent matrix = 0
+  virtual void AssembleElementGrad( const mfem::FiniteElement &el, mfem::ElementTransformation &Tr,
+                                    const mfem::Vector &elfun, mfem::DenseMatrix &elmat);
+
+private:
+  mfem::LinearFormIntegrator & m_f;
```
white238, line 32: Are you able to use std::shared_ptr here?

tests/wrapper_tests.cpp
```diff
@@ -0,0 +1,241 @@
...
+// # Author: Jonathan Wong @ LLNL.
+
+#include <gtest/gtest.h>
+
+#include "mfem.hpp"
+#include <memory>
+#include "coefficients/stdfunction_coefficient.hpp"
+#include "integrators/wrapper_integrator.hpp"
+
+using namespace mfem;
```
white238, line 15: Please don't do this not everyone knows where things are coming from... namely me.

src/integrators/wrapper_integrator.hpp
jonwong12: Good point. I do think std::shared_ptr is better here because it would further guarantee some other piece of code can't pull the rug out from this class.

tests/wrapper_tests.cpp
jonwong12: Should I change it in my previous test `StdFunctionCoefficient` also?

src/integrators/wrapper_integrator.cpp
jonwong12: What do you mean by checked in host-config? So just to clarify, the style changes have been made recently?

tests/wrapper_tests.cpp
jonwong12: I guess we've had this conversation partially in the past and perhaps it should be discussed further in person. I find it a bit tedious to refer to everything explicitly. I understand why it shouldn't be put into a header, but at some point it is a style preference and trade off.

Would you prefer I do this instead? (the Axom-way)
```
using Vector = mfem::Vector
```

samuelpmish: intellisense can see through `using` statements to clarify types (when hovering cursor over `u`):

intellisense

I agree `using namespace` shouldn't be in headers, but it's not a big deal otherwise.

samuelpmish: on an unrelated note: could you also change the variable with name `restrict` in the StdFunctionCoefficient tests to `restricted` or similar? It's a reserved word in C, and seems to confuse my syntax-highlighter!

src/integrators/wrapper_integrator.cpp
white238: In the repo there is a directory with host-configs that have shared libraries you can use on the network.  For example, host-configs/rzgenie-toss_3_x86_64_ib-gcc@8.1.0.cmake points at our public libraries with all the dev tools enabled.

./config-build.py -hc host-configs/rzgenie-toss_3_x86_64_ib-gcc@8.1.0.cmake
cd build-rzgenie-toss_3_x86_64_ib-gcc@8.1.0-debug
make style

tests/wrapper_tests.cpp
white238: I only like using's when they shorten namespaces (note remove completely) or are limited to a function in very limited quantity.  We can have a conversation about this at a later date in person.

white238: To clarify my comment, just leave it for now especially since its a test and not in a header.

jamiebramwell: I am also generally against `using` statements. However, I think tests are fine since they are self contained pieces of code. 

This brings up the fact that we should probably draft and discuss a 'best practices' guide at some point that goes beyond simple formatting. More work for the pile... 😭

jamiebramwell:
Going through @white238 comments, I think these are the outstanding items:
- [ ] `make style`
- [ ] use `std::shared_ptr` in `src/integrators/wrapper_integrator.cpp`

jamiebramwell:
@white238 @jonwong12 I think I address the comments. Let me know if this looks good to you. My next step will be using this to make a unified solid solver.

PR #50, opened 2020-04-14
Preventing Gitlab from cloning Serac for alloc job

adrienbernede:
I forgot this when creating the pipeline.

By default Gitlab will clone the project for every single job. However some jobs, like resource allocation, don’t need the source.


adrienbernede:
LGTM

PR #51, opened 2020-04-14
Improve regex partern, add verbose only messages

adrienbernede:
I reported here the changes I have made for another project using FindMFEM.cmake.
The goal is two-fold:
- Regex used to capture the matching sequence plus the rest of the file. Fixed to capture only one line. Not sure if the "strip" step is still necessary.
- Add VERBOSE and WARNING messages that were useful in another project.


adrienbernede:
LGTM

PR #53, opened 2020-04-20
Move the nonlinear solid solver to std::shared_ptrs

jamiebramwell:
This moves the nonlinear solid solver to a shared pointer interface.


src/solvers/nonlinear_solid_solver.hpp
```diff
@@ -73,7 +69,7 @@ class NonlinearSolidSolver : public BaseSolver {
   void SetTractionBCs(std::vector<int> &trac_bdr, mfem::VectorCoefficient *trac_bdr_coef);
 
   /// Set the viscosity coefficient
-  void SetViscosity(mfem::Coefficient *visc_coef);
+  void SetViscosity(std::shared_ptr<mfem::Coefficient> visc_coef);
```
jamiebramwell, line 72: @white238 @samuelpmish @jonwong12 Any thoughts on how to not require a user to use `std::shared_ptr`? I would like to have a pointer-type object of the base `mfem::Coefficient` class, but I can't figure out how to get `std::shared_ptr` to work with an argument of type `mfem::Coefficient &`.

jamiebramwell: To summarize our webex: we decided making the user pass a `shared_ptr` is the way to go.

jamiebramwell:
I think this is enough work for this PR. Let me know what you think!

jamiebramwell:
LGTM

src/solvers/base_solver.cpp
```diff
@@ -109,37 +100,36 @@ void BaseSolver::SetTimestepper(const TimestepMethod timestepper)
 
   switch (m_timestepper) {
     case TimestepMethod::QuasiStatic:
-      m_ode_solver = nullptr;
       break;
     case TimestepMethod::BackwardEuler:
-      m_ode_solver = std::make_shared<mfem::BackwardEulerSolver>();
+      m_ode_solver = std::make_unique<mfem::BackwardEulerSolver>();
```
white238, line 105: So nice... makes my CS heart sing.

src/solvers/base_solver.cpp
```diff
@@ -175,7 +163,7 @@ void BaseSolver::InitializeOutput(const OutputType output_type, std::string root
       mesh_name << m_root_name << "-mesh." << std::setfill('0') << std::setw(6) << m_rank - 1;
       std::ofstream omesh(mesh_name.str().c_str());
       omesh.precision(8);
-      m_state[0].mesh->Print(omesh);
+      m_state.begin()->mesh->Print(omesh);
```
white238, line 166: It is strange to use begin() when you aren't going to iterate over the vector. Since this returns an iterator not a reference to the first item.  Mind changing this to front()?

src/solvers/base_solver.cpp
```diff
@@ -152,20 +142,18 @@ double BaseSolver::GetTime() const { return m_time; }
...
   m_output_type = output_type;
 
   switch (m_output_type) {
     case OutputType::VisIt: {
-      m_visit_dc = new mfem::VisItDataCollection(m_root_name, m_state.begin()->mesh);
+      m_visit_dc = std::make_unique<mfem::VisItDataCollection>(m_root_name, m_state.begin()->mesh.get());
 
-      for (unsigned int i = 0; i < names.size(); i++) {
-        m_visit_dc->RegisterField(names[i], m_state[i].gf.get());
+      for (unsigned int i = 0; i < m_state.size(); i++) {
```
white238, line 155: You could be even fancier:

for(const auto& state: m_state) {
  m_visit_dc->RegisterField(state.name, state.gf.get());
}

Should work with the caveat I didn't actually try compiling it.

Also "gf" will always make me think of girlfriend and "state.gf.get()" makes me chuckle.

src/solvers/elasticity_solver.cpp
```diff
@@ -26,7 +26,7 @@ ElasticitySolver::ElasticitySolver(int order, mfem::ParMesh *pmesh)
   displacement.mesh = pmesh;
   displacement.coll = std::make_shared<mfem::H1_FECollection>(order, pmesh->Dimension(), mfem::Ordering::byVDIM);
   displacement.space =
-      std::make_shared<mfem::ParFiniteElementSpace>(pmesh, displacement.coll.get(), pmesh->Dimension());
+      std::make_shared<mfem::ParFiniteElementSpace>(pmesh.get(), displacement.coll.get(), pmesh->Dimension());
```
white238, line 29: Are we sure mfem isnt taking ownership of this?

white238: Nevermind.. tracked down the header and says explicitly its not owned.

jamiebramwell: Yep, this is a thing MFEM expects the users to handle.

src/solvers/thermal_operators.cpp
```diff
@@ -0,0 +1,134 @@
...
+  m_z.Add(1.0, *m_bc_rhs);
+  m_M_solver->Mult(m_z, du_dt);
+}
+
+void DynamicConductionOperator::ImplicitSolve(const double dt, const mfem::Vector &u, mfem::Vector &du_dt)
+{
+  MFEM_ASSERT(m_M_mat != nullptr, "Mass matrix not set in ConductionSolver::ImplicitSolve!");
+  MFEM_ASSERT(m_K_mat != nullptr, "Stiffness matrix not set in ConductionSolver::ImplicitSolve!");
+
+  m_y = u;
```
white238, line 99: That's weird naming to me.... u to y? having no idea what either of those are supposed to be.

jamiebramwell: Fine, I added a comment. 💬

white238: Don't get sassy with me...

jamiebramwell: 😎

jamiebramwell:
LGTM

src/drivers/serac.cpp
```diff
@@ -126,39 +125,36 @@ int main(int argc, char *argv[])
...
 
   mfem::Vector velo(dim);
   velo = 0.0;
 
-  mfem::VectorConstantCoefficient *velo_coef = new mfem::VectorConstantCoefficient(velo);
+  auto velo_coef = std::make_shared<mfem::VectorConstantCoefficient>(velo);
 
-  std::vector<mfem::VectorCoefficient *> coefs = {defo_coef, velo_coef};
+  auto coefs = {std::static_pointer_cast<mfem::VectorCoefficient>(defo_coef),
+                std::static_pointer_cast<mfem::VectorCoefficient>(velo_coef)};
```
samuelpmish, line 157: Hmm, this is a little confusing-- how is a user supposed to know that displacement goes first and velocity second? Maybe we make the ProjectState implementation specific to the solver, e.g. 
`NonlinearSolidSolver::ProjectDisplacement()` and `NonlinearSolidSolver::ProjectVelocity()`.

I'm also not entirely sure what "Project" means to MFEM. That word has multiple meanings depending on the context. If this is just setting the initial displacement/velocity fields, we might be better off calling them
`NonlinearSolidSolver::SetDisplacement()` and `NonlinearSolidSolver::SetVelocity()`.

tests/serac_dtor.cpp
```diff
@@ -32,59 +27,57 @@ TEST(serac_dtor, test1)
...
-  do_nothing(therm_solver);
-  do_nothing(therm_solver);
-  do_nothing(therm_solver);
+  therm_solver->CompleteSetup();
+
+  // Destruct the old thermal solver and build a new one
+  therm_solver.reset(new ThermalSolver(1, pmesh));
+
+  // Destruct the second thermal solver and leave the pointer empty
+  therm_solver.reset(nullptr);
```
samuelpmish, line 80: hmm, I wonder if it even makes sense to keep this test around if the solvers have trivial deconstructors.

jamiebramwell:
LGTM

src/drivers/serac.cpp
jamiebramwell: Good idea! I'll fix this.

tests/serac_dtor.cpp
jamiebramwell: I think it's a good idea to keep it around, especially once we get memory checking going. I can see people adding more complicated stuff, and this is a good check for it.

jamiebramwell:
LGTM

jamiebramwell:
LGTM

PR #56, opened 2020-04-29
don't use the symlink for cmake

white238:
I am moving us to use the system CMake but there is a small window that master will be broken during that time.  This will make that not happen.


white238:
LGTM

PR #57, opened 2020-04-30
Added test for non-homogenous component-wise boundary conditions

jamiebramwell:
As always, this PR got away from me. It adds a bunch of boundary-condition related functionality to enable integration with the Tribol patch test.

- Adds component-wise boundary condition enforcement to the base solver.
- Adds a component-wise Dirichlet condition test for the nonlinear solid solver. 
- Improves the output for the solid solver using the deformed mesh.
- Uses the isoparametric mesh representation in the solid solver.
- Adds new cuboid MFEM meshes to the `data` directory.
- Switches the WSL spack configs to use MPICH instead of OpenMPI.
- Adds a boundary condition information data storage bundle type.
- Adds output to the existing dynamic and quasi-static solid solver tests.


jamiebramwell:
I think this is approaching done. I want to do some more verification tests, but let me know if you see anything glaringly wrong.

src/solvers/thermal_solver.cpp
```diff
@@ -38,7 +38,9 @@ void ThermalSolver::SetTemperatureBCs(const std::vector<int> &ess_bdr, std::shar
   SetEssentialBCs(ess_bdr, ess_bdr_coef);
 
   // Get the essential dof indicies and project the coefficient onto them
```
samuelpmish, line 40: "indices"

src/solvers/thermal_solver.cpp
```diff
@@ -38,7 +38,9 @@ void ThermalSolver::SetTemperatureBCs(const std::vector<int> &ess_bdr, std::shar
...
-  temperature.space->GetEssentialTrueDofs(m_ess_bdr, m_ess_tdof_list);
+  for (auto &ess_bc_data : m_ess_bdr) {
+    temperature.space->GetEssentialTrueDofs(ess_bc_data->bc_markers, ess_bc_data->true_dofs);
+  }
```
samuelpmish, line 43: Does `true_dofs` refer to unconstrained degrees of freedom? This becomes a little bit more confusing with multiple entries in `m_ess_bdr`. e.g. a true_dof with respect to `m_ess_bdr[0]` may still be constrained after all.

I'm not sure what to do about it, just thinking aloud.

src/solvers/nonlinear_solid_operators.cpp
```diff
@@ -28,24 +28,24 @@ mfem::Operator &NonlinearSolidQuasiStaticOperator::GetGradient(const mfem::Vecto
...
+      m_ess_bdr(ess_bdr),
       m_lin_params(lin_params),
       m_z(height / 2)
 {
   // Assemble the mass matrix and eliminate the fixed DOFs
   m_M_mat.reset(m_M_form->ParallelAssemble());
-  auto Me = std::unique_ptr<mfem::HypreParMatrix>(m_M_mat->EliminateRowsCols(m_ess_tdof_list));
+  for (auto &ess_bc_data : m_ess_bdr) {
+    auto Me = std::unique_ptr<mfem::HypreParMatrix>(m_M_mat->EliminateRowsCols(ess_bc_data->true_dofs));
+  }
```
samuelpmish, line 48: is `Me` used for anything?

tests/serac_stdfunction_coefficient.cpp
```diff
@@ -194,3 +194,69 @@ TEST_F(StdFunctionCoefficientTest, EssentialBC)
...
   }
 }
+
+/**
+   Create the ess_vdof_list for all dofs that are x = 0
+ */
+TEST_F(StdFunctionCoefficientTest, EssentialBCCube)
+{
+  // Create an indicator function to set vertex at the origin
+  StdFunctionVectorCoefficient origin_bc(3, [](Vector &x, Vector &X) {
```
samuelpmish, line 204: The other ones use `pfes_v->GetVDim()` as the first argument. What is this test doing?

src/solvers/thermal_solver.cpp
jamiebramwell: Yeah, I'm considering it an analog to a "node list" that we associate with specific boundary conditions in low order codes. I'm welcome to any ideas to make it clearer! I'm also considering saving the total condensed essential degrees of freedom list saved in the solver object.

src/solvers/nonlinear_solid_operators.cpp
jamiebramwell: Not right now. It's the eliminated rows of the mass matrix, which is needed if you're enforcing changing displacement boundary conditions. I have this tagged as a feature to add, but haven't gotten around to it yet.

tests/serac_stdfunction_coefficient.cpp
jamiebramwell: @jonwong12 can answer since he's the one who wrote it, but I believe it's setting the appropriate dofs for a component-wise boundary condition.

jonwong12: The `3` is the vector dimension for the vector coefficient. I believe it is passed directly to the underlying base constructor for `VectorCoefficient`. I think `mesh->SpaceDim()` might actually be more appropriate.

src/solvers/thermal_solver.cpp
jonwong12: Disclaimer: This is how I understand MFEM `vdof` vs `tdof` terminology.

I think that `true_dofs` technically means that they are degrees of freedom that are "solved" in Ax=b. In mfem, I believe the `EliminiateRows*` calls just put a 1 on the diagonal, and update the RHS. So you are absolutely correct that any `true_dof` could be constrained.

`virtual_dofs` are those that are degrees of freedom that are "not solved" in Ax=b. They might have been condensed out or etc: ghost nodes, hanging nodes, etc.

jamiebramwell: This is my understanding as well. However, I've been burned before!

jamiebramwell:
LGTM

jamiebramwell:
LGTM

scripts/uberenv/spack_configs/linux-ubuntu18.04-broadwell/packages.yaml
```diff
@@ -2,7 +2,7 @@ packages:
   all:
     compiler: [gcc, clang]
     providers:
-      mpi: [openmpi]
+      mpi: [mpich]
```
white238, line 5: I was helping @samuelpmish about this file today.  A following PR is probably needed to see if you guys can come up with a common solution.  I have ideas.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -60,6 +60,19 @@ class StdFunctionVectorCoefficient : public mfem::VectorCoefficient {
   std::function<void(mfem::Vector &, mfem::Vector &)> m_func;
 };
 
+/**
+   \brief MakeTrueEssList takes in a FESpace, a vector coefficient, and produces a list
+   of essential boundary conditions
+
+   \param[in] pfes A finite element space for the constrained grid function
+   \param[in] c A VectorCoefficient that is porjected on to the mesh. All
```
white238, line 68: ... projected...

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -60,6 +60,19 @@ class StdFunctionVectorCoefficient : public mfem::VectorCoefficient {
...
+   d.o.f's are examined and those that are the condition (> 0.) are appended toe
```
white238, line 69: to.. though funny the other way

scripts/uberenv/spack_configs/linux-ubuntu18.04-broadwell/packages.yaml
adrienbernede: What was the issue?

white238: System installed packages being in different places, architectures being slightly different, versions not being updated in various ways..

src/coefficients/stdfunction_coefficient.hpp
jonwong12: I guess my spelling mistakes finally got caught the second time around. Typing is hard. I'll fix these.

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -30,6 +30,21 @@ void StdFunctionVectorCoefficient::Eval(mfem::Vector &V, mfem::ElementTransforma
   m_func(transip, V);
 }
 
+void MakeTrueEssList(mfem::ParFiniteElementSpace &pfes, mfem::VectorCoefficient &c, mfem::Array<int> &ess_tdof_list)
```
jonwong12, line 33: @jamiebramwell @white238 I actually think I like the loop here better than the one in `MakeEssList`. Do you think it would be better to update `MakeEssList` with the forloop here and then have `MakeTrueEssList` just remap the `GetLocalTDofNumber`? Just a thought.

jamiebramwell: I wasn't sure if you needed it the other way for some reason. I'll update this.

jonwong12: No. When I wrote the precursor code, I don't think these methods existed.

jamiebramwell:
LGTM

PR #58, opened 2020-05-08
Fixing the axom include

jamiebramwell:
This fixes the axom package finding function to use the new exported cmake target.


jamiebramwell:
LGTM

jamiebramwell:
LGTM

PR #61, opened 2020-05-15
Add AddressSanitizer and valgrind support

white238:
First shot at enabling memory checking.  I added two options: AddressSanitizer (seems to be the new hotness) and valgrind (tried and true, but cumbersome and slow).

I added documentation here:

https://serac.readthedocs.io/en/feature-white238-memcheck/memory_checking.html

The next step is to make it easier to use. Could be in another PR though.

Improvement ideas:

1) Outputting a script that has the options already defined in it
2) Script to consolidate the memory check reports after they run, there might be some plugins to gitlab or azure to help with this
3) Automate with CI for every PR so we don't slide.

Thoughts?


white238:
LGTM

jamiebramwell:
This looks great! I definitely would like to see ease-of-use scripts (#1) and integration with CI (#3) in a future PR. I think it's important to get it started now before we slide.

I also think keeping around memcheck and AddressSanitizer is a good idea for now until we can kick the tires a bit to see which is better. I know memcheck can be painfully slow on big problems, so I'm tempted to go with Asan.

src/docs/sphinx/memory_checking.rst
```diff
@@ -0,0 +1,87 @@
+.. ## Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
+.. ## other Serac Project Developers. See the top-level COPYRIGHT file for details.
+.. ##
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+
+===============
```
jamiebramwell, line 6: 👍

src/drivers/serac.cpp
```diff
@@ -34,7 +36,8 @@ int main(int argc, char *argv[])
   MPI_Comm_rank(MPI_COMM_WORLD, &myid);
 
   // mesh
-  const char *mesh_file = "../../data/beam-hex.mesh";
+  std::string base_mesh_file = std::string(SERAC_SRC_DIR) + "/data/beam-hex.mesh";
```
jamiebramwell, line 39: I assume `SERAC_SRC_DIR` is in the config header. Should we change the tests do it this way instead?

white238:
> This looks great! I definitely would like to see ease-of-use scripts (#1) and integration with CI (#3) in a future PR. I think it's important to get it started now before we slide.
> 
> I also think keeping around memcheck and AddressSanitizer is a good idea for now until we can kick the tires a bit to see which is better. I know memcheck can be painfully slow on big problems, so I'm tempted to go with Asan.

Also Asan is kinda the new hotness, there is usually value in going with what the community is converging on even if its a bit behind.

src/drivers/serac.cpp
white238: Would you like me to remove the command line path? I could push that to this PR if you want.  It is one less thing to remember / copy&paste.

jamiebramwell: I wouldn't worry about it now. I can add it to my next PR.

PR #64, opened 2020-05-21
add autodetection of WSL in uberenv

white238:
Couple people have complained about the WSL Spack build lately so I had thoughts.  Hard to test due to me not having WSL.  Since WSL2 is due any day I'll install it then.

I'll move the docs from gitlab to our sphinx docs if people think this is a good idea. As well as add some instructions about prereqs to install.   Probably a good idea to have users do an apt-get for cmake and mpi.

The people that use WSL what are your thoughts?


samuelpmish:
After our discussion the other day, I believe my issue with the build was less about it being on WSL (I think the same issue would happen on native Ubuntu 18.04) and more about the spack configs being very narrow in scope. It seems that if we don't provide a spack config for the user's exact combination of (OS, version, architecture), then uberenv will just ignore all our other spack_configs and select inappropriate versions of the dependencies.

It would be preferable if uberenv had either:
- a default/fallback spack_config for when the user's setup isn't covered by the configs we provide, or                           
- a way to select the spack_config that is "closest" to the user's machine (more desirable, but unclear what this means)

adrienbernede:
@samuelpmish,
Spack configs in uberenv are narrow by design: They are meant to lead to a reproducible installation on a given machine.
It is then expected that they would not work on another system, and uberenv will only use a spack config if it matches the system found by "detect_platform" which is very limited, or the system description you provided with `spack_config_dir` option (which answers you second point I think).
That is to say that there is nothing prepared for WSL or ubuntu, but a config can be crafted when needed.

There is another interesting point in your comment. What do you mean by "select inappropriate versions"?
If by default Serac is incompatible with certain dependencies version, then this should appear in its spack package. The default configuration for an unknown system should be the one Spack found (and it will basically try to get the newest version everywhere) based on what is describe in the package. The main drawback of this approach being that spack will then install mpi, cmake etc.

jamiebramwell:
Yeah, I think it's more ubuntu than WSL specific. I think a ubuntu spack config that works in concert with a sphinx doc on which packages to apt-get would be best. I would like to avoid building MPI, cmake, and other base libraries, but I unfortunately didn't have the time to get the correct combination working.

jonwong12:
I'll have to check again, but I have an older processor and I had to modify the WSL instructions a little bit. I've always been trying to document which packages I installed to try to make things faster.

Thanks for adding the WSL detection. I think it would be useful.

samuelpmish:
@adrienbernede Since there was no spack_config for my machine, uberenv downloaded MFEM 4.1 and other dependencies that are not supported, and it failed to build. It sounds like you're saying that we should try to address this not with spack_configs, but in the serac spack package definition itself. Sounds promising!

adrienbernede:
Yes that’s what I mean. I a new version of MFEM is out and we don’t currently support it, we can simply add something like:
`conflicts('mfem@4.1.0:')` to prevent spack from using it. The same is true for versions that are too old, of conflicting versions.
This way we limit the the possibilities for known limitations, but leave room for "improvisation" on unknown systems :)

white238:
> Yes that’s what I mean. I a new version of MFEM is out and we don’t currently support it, we can simply add something like:
> `conflicts('mfem@4.1.0:')` to prevent spack from using it. The same is true for versions that are too old, of conflicting versions.
> This way prevent concretization with known limitations, but leave room for "improvisation" on unknown systems :)

Is there a benefit that I am not seeing by conflicting in the serac package over listing the version that works?

adrienbernede:
Good point: as long as you use intervals I think it should be equivalent.
I just wanted to avoid a syntax like `depends_on('mfem@4.0.0')`.

white238:
> Good point: as long as you use intervals it’s equivalent.
> I just wanted to avoid a syntax like `depends_on('mfem@4.0.0')`.

But we only work with 4.0.0 =)

adrienbernede:
👍 OK, but what if MFEM releases 4.0.1 ? XD

white238:
> 👍 OK, but what if MFEM releases 4.0.1 ? XD

The real problem is changing library versions often (and usually) change floating point answers.  Which means they need to be revetted by a physicist/engineer/sciency-person.  All simulation codes I know of lock down not only library versions but also compilers and compiler flags as well.

jamiebramwell:
Simulation codes typically like to lock down specific versions of high impact libraries. This is because many of our tests are bit-for-bit restart checks and minor changes in libraries can cause frequent failures.

jamiebramwell:
@white238 As one of those engineering-y people. I don't like doing that vetting!

adrienbernede:
We talked about this with @white238 on a slightly different topic (variants and optimization). My feeling is that lock down makes sense in spack-configs.
And I do know vetting is painful. I used to do more of that. I just hope for a world where we can automate as much of this vetting as we can!

adrienbernede:
@white238, I asked Greg:
It is much better to specify a range of version in `depends_on` than exclude a range in `conflicts`, if only because conflicts are a post-concretization check...
I’ll edit my answer.

white238:
Let's have a short talk about this at the meeting at 4.

samuelpmish:
On my machine, this branch correctly identifies WSL (ubuntu 18.04) and uses the appropriate spack_config, but it still builds everything from scratch, as before. Now, my other non-WSL ubuntu 18.04 machine also builds everything from scratch, when it previously did not. It's possible I'm not using it properly-- I'm just cloning serac, checking out the branch and executing `python3 scripts/uberenv/uberenv.py`. Is this expected?

jonwong12:
Sorry. I'll check tonight. Need to wait for the temps to drop a little before I roast myself.

jamiebramwell:
I'm testing this out on a clean wsl install. I'll document the process and fix any issues that come up.

scripts/uberenv/packages/serac/package.py
```diff
@@ -87,8 +87,7 @@ class Serac(CMakePackage):
 
     # Basic dependencies
     depends_on("mpi")
-
-    depends_on("axom~openmp~fortran~raja~umpire")
+    depends_on("cmake@3.8:")
```
white238, line 90: Did you end up with an old version of CMake? The fact that this is a CMakePackage should add the dependency inheritantly

jamiebramwell:
I believe this works using the given instructions on WSL1 now.

jamiebramwell:
LGTM

jamiebramwell:
@white238 anything else we want to change before we merge this?

jamiebramwell:
> 
> 
> Has anyone tested it with WSL2 yet? I don't expect it to break, but I feel that should be our supported WSL platform moving forward.

I haven't, but I plan on testing it out today or tomorrow on my home desktop.

white238:
Is WSL2 out for general use as opposed to a beta sign up?

jonwong12:
@white238 It is out of beta, but MS is supposedly heavily throttling the update. So I have tried since last week to get it, but I guess my "not-even-broadwell" processor is not eligible yet.

jonwong12:
If someone is brave enough, you can go to "Windows Update" and supposedly it will bypass the built-in OS checker and install it.

jamiebramwell:
I am in the process of upgrading to WSL 2 on my desktop. Wish me luck!

jamiebramwell:
This branch is in the process of building on my fancy new WSL 2 installation. I did upgrade to Ubuntu 20.04 which changed a bunch of the package versions in the default distro. This points to the fact the spack configs should probably be labeled Ubuntu18.04 instead of (or maybe in addition to) wsl.

jamiebramwell:
It works in WSL 2!

jamiebramwell:
For now, I'm just going to put these configs in separate spack configs (ubuntu 18 and 20). I figure we can work out the autodetection stuff later.

scripts/uberenv/packages/serac/package.py
```diff
@@ -97,24 +96,31 @@ class Serac(CMakePackage):
...
-    depends_on("mfem~shared~zlib+hypre+metis+superlu-dist+lapack+mpi+debug", when="+debug")
-    depends_on("hypre~shared~superlu-dist+mpi")
-    depends_on("hypre~shared~superlu-dist+mpi+debug", when="+debug")
+    debug_deps = ["mfem@4.0.0~shared~zlib+hypre+metis+superlu-dist+lapack+mpi",
+                  "hypre@2.11.1~shared~superlu-dist+mpi"]
+    for dep in debug_deps:
+        depends_on("{0}".format(dep))
+        depends_on("{0}+debug".format(dep), when="+debug")
+
+    depends_on("hypre@2.11.1~shared~superlu-dist+mpi")
```
white238, line 105: Not sure how i mess this up but these two lines need to be removed because they are specified in the for loop above.

adrienbernede: Yup, also present in #66, but I. should have tracked it done to this PR: https://github.com/LLNL/serac/pull/66#discussion_r436901597

jamiebramwell:
I believe this is ready to go now. I have tested it on WSL1/Ubuntu 18.04 and WSL2/Ubuntu 20.04.

jamiebramwell:
LGTM

white238:
i say merge away after it passes gitlab tests.. thanks @jamiebramwell for finishing this up!

PR #65, opened 2020-06-03
Operator split simple thermal-structural solver

jamiebramwell:
This is a first cut at the structure of a coupled solver. It is a simple operator split thermal structural solver with a temperature dependent viscosity coefficient.


jamiebramwell:
Also, this contains the new wsl branch. Sorry! I was lazy and didn't want to rebuild my TPLs!

samuelpmish:
What is the motivation for changing all the FiniteElementState aliases into std::shared_ptrs?

jamiebramwell:
> 
> 
> What is the motivation for changing all the FiniteElementState aliases into std::shared_ptrs?

If we leave them as aliases, they need to be available in the constructor. I was having issues initializing the `m_state` object as I did not inherit from the single physics solvers. I think I could get back to references now that a first cut is done. However, since I am going with a composition model instead of aggregation model, I though having a separate pointer may be clearer.

samuelpmish:
does https://godbolt.org/z/fcpnLc do the `m_state` initialization you want?

jamiebramwell:
> 
> 
> does https://godbolt.org/z/fcpnLc do the `m_state` initialization you want?

Yep, that's what I was thinking we could go back to.

tests/serac_thermal_structural_solver.cpp
```diff
@@ -0,0 +1,168 @@
...
+void   InitialDeformation(const mfem::Vector &x, mfem::Vector &y);
+void   InitialVelocity(const mfem::Vector &x, mfem::Vector &v);
+double InitialTemperature(const mfem::Vector &x);
+
+TEST(dynamic_solver, dyn_solve)
+{
+  MPI_Barrier(MPI_COMM_WORLD);
+
+  // mesh
+  std::string base_mesh_file = std::string(SERAC_SRC_DIR) + "/data/beam-hex.mesh";
```
samuelpmish: I like this much better than the old way of finding the mesh files!

tests/serac_thermal_structural_solver.cpp
```diff
@@ -0,0 +1,168 @@
...
+  // define a boundary attribute array and initialize to 0
+  std::vector<int> ess_bdr(pmesh->bdr_attributes.Max(), 0);
+
+  // boundary attribute 1 (index 0) is fixed (Dirichlet)
+  ess_bdr[0] = 1;
+
+  auto deform = std::make_shared<mfem::VectorFunctionCoefficient>(dim, InitialDeformation);
+  auto velo   = std::make_shared<mfem::VectorFunctionCoefficient>(dim, InitialVelocity);
+  auto temp   = std::make_shared<mfem::FunctionCoefficient>(InitialTemperature);
+  auto kappa = std::make_shared<mfem::ConstantCoefficient>(0.5);
```
samuelpmish, line 56: Since the function bodies of Initial***() functions are small, I think it's more readable to use lambdas / stdfunctionCoefficient to include them in the body of the test

jonwong12:
You could also

> 
> 
> > does https://godbolt.org/z/fcpnLc do the `m_state` initialization you want?
> 
> Yep, that's what I was thinking we could go back to.

I guess it depends on whether you want one the state portion from the coupled solver to be the same as its counter parts or whether you want to manage the state between the different solvers. 

I guess the way it is written right now, you are (1) building an integrated coupled solver that owns two solvers internally vs (2) writing just the glue between two solvers that you may or may not own. Thus the reference is always valid for m_state during the lifetime of the coupled solver. If we want to allow people to use (2) then using the shared_ptrs is not a bad thing.

src/coefficients/linear_coefficient.hpp
```diff
@@ -0,0 +1,33 @@
...
+  {
+  }
+  double Eval(mfem::ElementTransformation &T, const mfem::IntegrationPoint &ip)
+  {
+    double val = m_coef->Eval(T, ip);
+    return m_scale * val + m_offset;
+  };
+};
+
+#endif
```
samuelpmish: Is this functionality already available in StdFunctionCoefficient?

tests/serac_thermal_structural_solver.cpp
jamiebramwell: Agreed! I will fix this.

tests/serac_thermal_structural_solver.cpp
jamiebramwell: @white238 is to thank for this one!

src/coefficients/linear_coefficient.hpp
jamiebramwell: Heh, yep. Well, I need to write a new class similar to `TransformedVectorCoefficient`, but for scalars. I'll do that instead.

tests/serac_thermal_structural_solver.cpp
```diff
@@ -0,0 +1,168 @@
...
+  return result;
+}
+
+void InitialDeformation(const mfem::Vector &x, mfem::Vector &y)
+{
+  y    = x;
+  y(1) = y(1) + x(0) * 0.01;
+}
+
+void InitialVelocity(__attribute__((unused)) const mfem::Vector &x, mfem::Vector &v) { v = 0.0; }
```
samuelpmish: there's a handful of `__attribute__((unused))` floating around in the repo. I think just not giving a name to the parameter silences the warning without the verbose attribute decorator

tests/serac_thermal_structural_solver.cpp
samuelpmish: Actually, SERAC_SRC_DIR is a little confusing, as there is a src/ directory in serac, but that's not what it's pointing to.. Maybe SERAC_DIR or SERAC_BASE_DIR or something would be better

src/coefficients/linear_coefficient.hpp
jonwong12: The reason why I didn't implement `TransformedCoefficient` in serac is because MFEM also has [one by the same name](https://mfem.github.io/doxygen/html/classmfem_1_1TransformedCoefficient.html). I found most of the time I could get away with using the mfem version.

Most of the time it works sans when you need to capture. I feel like TransformedVectorCoefficient and TransformedCoefficient (better implemented) should be upstreamed except I don't know how MFEM feels about smart pointers.

jamiebramwell: They are unfortunately generally against it. I will ask at the meeting today if that has changed though. One simple way out is to call it `TransformedScalarCoefficient`.

tests/serac_thermal_structural_solver.cpp
jamiebramwell: TIL! I'll try that out.

tests/serac_thermal_structural_solver.cpp
white238: SERAC_REPO_DIR?

jamiebramwell:
> 
> 
> You could also
> 
> > > does https://godbolt.org/z/fcpnLc do the `m_state` initialization you want?
> > 
> > 
> > Yep, that's what I was thinking we could go back to.
> 
> I guess it depends on whether you want the state portion from the coupled solver to be the same as its counterparts or whether you want to manage the state between the different solvers.
> 
> I guess the way it is written right now, you are (1) building an integrated coupled solver that owns two solvers internally vs (2) writing just the glue between two solvers that you may or may not own. Thus the reference is always valid for m_state during the lifetime of the coupled solver in (1). If we want to allow people to use (2) then using the shared_ptrs is not a bad thing.

After letting this percolate in my brain. I think (2) is a reasonable demand from users. This is especially true if we're wrapping legacy solvers instead of our own hand-rolled MFEM ones. For that reason, I think I'm going to keep them as `shared_ptr`s unless there are serious objections.

tests/serac_thermal_structural_solver.cpp
jamiebramwell: Done!

src/coefficients/linear_coefficient.hpp
jamiebramwell: Done! I think it's best to have a corresponding API for the scalars and vectors. Plus, then we get the capturing. I'll make a push toward upstreaming these to MFEM.

jamiebramwell:
LGTM

jamiebramwell:
I believe I've now addressed all of the things mentioned. The one notable exception is that I kept the `shared_ptr`s for the `FiniteElementState` members. I think this will allow greater flexibility when coupling solvers without much additional overhead. Any other thoughts @samuelpmish @jonwong12 @white238 ?

jamiebramwell:
LGTM

src/integrators/wrapper_integrator.cpp
```diff
@@ -12,19 +12,17 @@ LinearToNonlinearFormIntegrator::LinearToNonlinearFormIntegrator(std::shared_ptr
...
   elvect *= -1.;
 }
 
-void LinearToNonlinearFormIntegrator::AssembleElementGrad(const mfem::FiniteElement &                 el,
-                                                          mfem::ElementTransformation &               Tr,
-                                                          __attribute__((unused)) const mfem::Vector &elfun,
-                                                          mfem::DenseMatrix &                         elmat)
+void LinearToNonlinearFormIntegrator::AssembleElementGrad(const mfem::FiniteElement &  el,
+                                                          mfem::ElementTransformation &Tr, const mfem::Vector &,
+                                                          mfem::DenseMatrix &          elmat)
```
white238, line 25: Was this a compiler warning or just a cleaner way to do it?

src/solvers/base_solver.cpp
```diff
@@ -21,6 +21,11 @@ BaseSolver::BaseSolver(MPI_Comm comm) : m_comm(comm), m_output_type(OutputType::
 BaseSolver::BaseSolver(MPI_Comm comm, int n) : BaseSolver(comm)
 {
   m_state.resize(n);
+
+  for (auto &state : m_state) {
+    state = std::make_shared<FiniteElementState>();
```
white238, line 26: what is this doing?

src/integrators/wrapper_integrator.cpp
jamiebramwell: Just a cleaner way to do it. Also, in this particular case the variable actually was used, so it's also more correct.

src/common/serac_types.hpp
```diff
@@ -47,6 +47,13 @@ enum class Preconditioner
   BoomerAMG
 };
 
+enum class CouplingScheme
+{
+  OperatorSplit,
+  FixedPoint,
+  FullyCoupled
```
samuelpmishLLNL, line 54: I'm curious to see what you have in mind for these coupling schemes!

jamiebramwell: Ha! Me too! It's incredibly vague and notional in my head at this point...

PR #66, opened 2020-06-04
Update MFEM to v4.1.0

jamiebramwell:
This updates MFEM to v4.1.0. It builds and passes tests on the WSL but I need to test it on the LC systems.


scripts/uberenv/packages/axom/package.py
```diff
@@ -528,6 +528,8 @@ def cmake_args(self):
             options.append('-DENABLE_TESTS=OFF')
         else:
             options.append('-DENABLE_TESTS=ON')
+        src_dir = os.path.abspath(pjoin(self.stage.source_path, "src"))
```
white238, line 531: This is no longer needed if we go to the newest spack (like a week ago)

src/docs/sphinx/index.rst
```diff
@@ -33,6 +33,21 @@ Serac is hosted on GitHub `here <https://github.com/LLNL/serac>`_. Serac uses gi
...
+---------------------------------------------------------
+
+For faster installation of the Serac dependencies via Spack on Windows WSL/Ubuntu 18.04 systems, install cmake, MPICH, openblas, and OpenGL 
+using the following commands:
+
+.. code-block:: bash
+   $ sudo apt-get update
+   $ sudo apt-get upgrade
+   $ sudo apt-get install cmake libopenblas-dev libopenblas-base mpich mesa-common-dev libglu1-mesa-dev freeglut3-dev
+   $ sudo ln -s /usr/lib/x86_64-linux-gnu/* /usr/lib
```
white238, line 46: I wonder if this should be logged as a bug to spack...

scripts/uberenv/packages/serac/package.py
```diff
@@ -97,24 +96,31 @@ class Serac(CMakePackage):
     depends_on('py-sphinx', when="+devtools")
 
     # Libraries that support +debug
-    depends_on("mfem~shared~zlib+hypre+metis+superlu-dist+lapack+mpi")
-    depends_on("mfem~shared~zlib+hypre+metis+superlu-dist+lapack+mpi+debug", when="+debug")
-    depends_on("hypre~shared~superlu-dist+mpi")
-    depends_on("hypre~shared~superlu-dist+mpi+debug", when="+debug")
+    debug_deps = ["mfem@4.0.0~shared~zlib+hypre+metis+superlu-dist+lapack+mpi",
```
samuelpmish, line 99: should this be mfem@4.1.0?

src/docs/sphinx/index.rst
jamiebramwell: sigh... probably.

scripts/uberenv/packages/serac/package.py
jamiebramwell: Yep. Thanks for the catch!

jamiebramwell: Yeah, turns out I didn't push it. Whoops!

scripts/uberenv/packages/serac/package.py
```diff
@@ -97,24 +96,31 @@ class Serac(CMakePackage):
...
-    depends_on("hypre~shared~superlu-dist+mpi")
-    depends_on("hypre~shared~superlu-dist+mpi+debug", when="+debug")
+    debug_deps = ["mfem@4.1.0~shared+hypre+metis+superlu-dist+lapack+mpi",
+                  "hypre@2.11.1~shared~superlu-dist+mpi"]
+    for dep in debug_deps:
+        depends_on("{0}".format(dep))
+        depends_on("{0}+debug".format(dep), when="+debug")
+
+    depends_on("hypre@2.11.1~shared~superlu-dist+mpi")
+    depends_on("hypre@2.11.1~shared~superlu-dist+mpi+debug", when="+debug")
```
adrienbernede, line 106:
```suggestion
```

It seems to me that hypre is treated twice here.

adrienbernede, line 106:
```suggestion
```
It seems to me that hypre is treated twice here.

scripts/uberenv/packages/conduit/package.py
```diff
@@ -0,0 +1,520 @@
...
+        cfg.write("#######\n\n")
+        cfg.write("# c compiler used by spack\n")
+        cfg.write(cmake_cache_entry("CMAKE_C_COMPILER", c_compiler))
+        cfg.write("# cpp compiler used by spack\n")
+        cfg.write(cmake_cache_entry("CMAKE_CXX_COMPILER", cpp_compiler))
+
+        # SERAC PATCH BEGIN
+        # we need to specify the c++ std
+        cfg.write(cmake_cache_entry("BLT_CXX_STD", "c++11"))
+        # SERAC PATCH END
```
adrienbernede, line 342: @white238 , for my education:
- Why couldn’t this patch be applied in Conduit?
- What is BLT doing specifically with c++11?
- Why wouldn’t conduit add the exact standard feature requirement so that CMake can verify compiler compatibility and apply the necessary flags? (cmake compile features)

adrienbernede: I guess the answer to the second one is: RTFM...

white238: 1) It should, but first it needs to be verified that this solves the problem @jamiebramwell is running into
2) CMake has a couple variables it uses to control the C++ std (C++ and CUDA) so this unifies them but it is not strictly required
3) Libraries try to leave standards open as much as possible so they can be toggled by the host code but conduit should (and it is being added currently) default to C++11 if it is not explicitly stated.

jamiebramwell: Yeah, ideally this would go directly in conduit. I think @cyrush is hitting the same issue right now. I'm just working on it in my own space to see if I can help.

cyrush: @white238 after we discussed, I rechecked -- if the cxx standard isn't set -- conduit *does* set it to c++11, here is the logic:

https://github.com/LLNL/conduit/blob/f592a1067af1643e935488579ef5a2ae16da75c3/src/cmake/SetupBLT.cmake#L55

So not sure what is going on? 
If there is a bug in how it is done in the conduit build system, I would like to understand and fix it there.

white238: @jamiebramwell hardcoded BLT_CXX_STD as well on the platform/compiler she ran into the problem and it didn't make the build error go away as well.

cyrush: maybe an issue with the BLT I am using with that compiler? Maybe try with a newer external blt?

adrienbernede: The failure is in Axom build.
However Axom Spack package uses hostconfig files, and the toss3 Spack config does not have Clang9 compiler defined.
Maybe this is where ABI get mixed. Still investigating.

adrienbernede: Axom compiler config shouldn’t interfere though.

white238: There is a distinct possibility no one has tried the conduit->axom workflow with clang9.

adrienbernede: build commands in conduit have the `-std=cxx11`, but not `--gcc-toolchain=/usr/tce/packages/gcc/gcc-8.1.0`.

adrienbernede: Added cxx_flags definition in host-config file for conduit, seems to work!

white238: did we accidentally remove the writing of flags to the host-configs i put in recently? Though i don't think thats solving jamie's problem since her problem is showing up on WSL2 and doesn't have any global flags

white238: I confirmed. I think i gave bad advice and the conduit package has changes that are not in spack proper that need to be restored in our repo

white238: Actually now that I think about it, both changes from axom and conduit can now be pushed to spack proper now that a couple PR's in their repo have gone through.  I'll take care of that later...

white238: https://github.com/LLNL/serac/commit/cab14bef4579c36c4bdb1fb8b0808ea296b89053#diff-2cfc1f4f3c26f3bb4ba5de09656bde95
This file has a few more fixes in it.

jamiebramwell:
LGTM

Issue #67, opened 2020-06-05
Potential mishandling of boundary conditions


samuelpmish:
The most recent change to the boundary condition data structures now involves iterating over
each of the members of `m_ess_bdr` to do things like elimination of essential degrees of freedom:

```cpp
void ThermalSolver::CompleteSetup(){
  ...
  // Eliminate the essential DOFs from the stiffness matrix
  for (auto &ess_bc_data : m_ess_bdr) {
    m_K_e_mat.reset(m_K_mat->EliminateRowsCols(ess_bc_data->true_dofs));
  }
  ...
}

void ThermalSolver::QuasiStaticSolve(){
  ...
  // Apply the boundary conditions
  *m_bc_rhs = *m_rhs;
  for (auto &ess_bc_data : m_ess_bdr) {
    ess_bc_data->scalar_coef->SetTime(m_time);
    temperature.gf->ProjectBdrCoefficient(*ess_bc_data->scalar_coef, ess_bc_data->bc_markers);
    temperature.gf->GetTrueDofs(temperature.true_vec);
    mfem::EliminateBC(*m_K_mat, *m_K_e_mat, ess_bc_data->true_dofs, temperature.true_vec, *m_bc_rhs);
  }
  ...
}
```

It seems like `m_K_e_mat` is just being overwritten on each loop iteration, so it's not obvious to me that this is really handling the separate entries of `m_ess_bdr` correctly.

Changing the thermal test from
```cpp
TEST(thermal_solver, static_solve){
  ...
  // Initialize the temperature boundary condition
  std::vector<int> temp_bdr(pmesh->bdr_attributes.Max(), 1);
  auto u_0 = std::make_shared<StdFunctionCoefficient>([](const auto & x){
    return x.Norml2();
  });
  therm_solver.SetTemperatureBCs(temp_bdr, u_0);
  ...
}
```

to

```cpp
TEST(thermal_solver, static_solve){
  ...
  // Initialize the temperature boundary condition
  std::vector<int> temp_bdr(pmesh->bdr_attributes.Max(), 1);
  auto u_0 = std::make_shared<StdFunctionCoefficient>([](const auto & x){
    return 2 * x.Norml2();
  });
  therm_solver.SetTemperatureBCs(temp_bdr, u_0);

  // whoops, that first function was wrong--  
  // let me try to correct it by setting the right values!
  u_0 = std::make_shared<StdFunctionCoefficient>([](const auto & x){
    return x.Norml2();
  });
[...]

PR #68, opened 2020-06-11
Move eliminated matrix entries into BoundaryCondition objects

samuelpmishLLNL:
motivated by https://github.com/LLNL/serac/issues/67

The previous boundary condition implementation overwrites some data that we need to save. Added 2 new tests to the thermal solver:

`static_solve_multiple_bcs` describes the same physical problem as `static_solve`, but partitions the boundary into two pieces. Previously, having two boundary conditions would overwrite info from the first, leading to an incorrect result.

`static_solve_repeated_bcs` also describes the same problem as `static_solve`, but with a single set of boundary conditions. This test shows that the code 

```
  // Initialize the temperature boundary condition
  auto u_0 = std::make_shared<mfem::FunctionCoefficient>(BoundaryTemperature);
  auto u_1 = std::make_shared<mfem::FunctionCoefficient>(OtherBoundaryTemperature);

  std::vector<int> temp_bdr(pmesh->bdr_attributes.Max(), 1);

  // Set the temperature BC in the thermal solver
  therm_solver.SetTemperatureBCs(temp_bdr, u_1);
  therm_solver.SetTemperatureBCs(temp_bdr, u_0);
```
does not replace the first boundary condition with the second, but maintains both (conflicting) boundary conditions. This test currently fails (I'm not sure what we want to do here). Semantically,
the function name `SetTemperatureBCs` implies that if it is called multiple times with the same first argument, it should overwrite the old value, rather than append new constraints. `static_solve_repeated_bcs` is currently failing, indicating that we don't handle this case properly at the moment.

I'm hoping to get some feedback on the behavior we want to achieve with `SetTemperatureBCs` and other related functions.

There are also a couple small name changes that I think improves readability a bit: 
e.g. `BoundaryConditionData` -> `BoundaryCondition`


tests/serac_thermal_solver.cpp
```diff
@@ -91,6 +92,153 @@ TEST(thermal_solver, static_solve)
...
+
+  // Initialize the temperature boundary condition
+  auto u_0 = std::make_shared<mfem::FunctionCoefficient>(BoundaryTemperature);
+  auto u_1 = std::make_shared<mfem::FunctionCoefficient>(OtherBoundaryTemperature);
+
+  std::vector<int> temp_bdr(pmesh->bdr_attributes.Max(), 1);
+
+  // Set the temperature BC in the thermal solver
+  therm_solver.SetTemperatureBCs(temp_bdr, u_0);
+  therm_solver.SetTemperatureBCs(temp_bdr, u_1);
```
jamiebramwell, line 209: This now throws a warning.

tests/serac_thermal_solver.cpp
```diff
@@ -91,6 +92,153 @@ TEST(thermal_solver, static_solve)
...
+  // Set the time integration method
+  therm_solver.SetTimestepper(TimestepMethod::QuasiStatic);
+
+  // Initialize the temperature boundary condition
+  auto u_0 = std::make_shared<mfem::FunctionCoefficient>(BoundaryTemperature);
+  auto u_1 = std::make_shared<mfem::ConstantCoefficient>(0.0);
+
+  std::vector<int> marked_1(pmesh->bdr_attributes.Max(), 0);
+  std::vector<int> marked_2(pmesh->bdr_attributes.Max(), 0);
+  marked_1[0] = 1;
```
jamiebramwell, line 129: This now sets the temperature using a coefficient on two disjoint parts of the boundary. I've actually been considering having this data structure as a `std::set` of attributes instead of a `std::vector` of indicators and translate it to an `mfem::Array` in the solver routines.

jamiebramwell:
LGTM

jamiebramwell:
LGTM

src/solvers/base_solver.cpp
```diff
@@ -25,98 +25,124 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n) : BaseSolver(comm)
...
 }
 
 void BaseSolver::SetTrueDofs(const mfem::Array<int> &                 true_dofs,
                              std::shared_ptr<mfem::VectorCoefficient> ess_bdr_vec_coef)
 {
-  auto bc_data = std::make_shared<BoundaryConditionData>();
+  auto bc = std::make_shared<BoundaryCondition>();
 
-  bc_data->bc_markers.SetSize(0);
+  bc->markers.SetSize(0);
```
white238, line 63: Just curious... why is this calling SetSize(0)? Seems that is what the default constructor should be doing... This is quite the nitpick btw.

src/solvers/elasticity_solver.cpp
```diff
@@ -40,12 +40,7 @@ ElasticitySolver::ElasticitySolver(int order, std::shared_ptr<mfem::ParMesh> pme
 void ElasticitySolver::SetDisplacementBCs(std::vector<int> &                       disp_bdr,
                                           std::shared_ptr<mfem::VectorCoefficient> disp_bdr_coef, int component)
 {
-  SetEssentialBCs(disp_bdr, disp_bdr_coef, component);
-
-  // Get the list of essential DOFs
-  for (auto &ess_bc_data : m_ess_bdr) {
-    displacement.space->GetEssentialTrueDofs(ess_bc_data->bc_markers, ess_bc_data->true_dofs);
-  }
+  SetEssentialBCs(disp_bdr, disp_bdr_coef, *displacement.space, component);
```
white238, line 43: Just noticed displacement is a member variable that isnt prefixed with "m_" was confused where it was coming from because of this.

src/solvers/elasticity_solver.cpp
```diff
@@ -85,7 +80,7 @@ void ElasticitySolver::CompleteSetup()
   if (m_nat_bdr.size() > 0) {
     for (auto &nat_bc_data : m_nat_bdr) {
```
white238, line 81: should the "_data" be removed from this variable to be consistent with the rest of the changes?

src/solvers/base_solver.cpp
jamiebramwell: In theory yes, but I've been burned by the memory management of the `mfem::Array` class before.

src/solvers/elasticity_solver.cpp
jamiebramwell: Yep, I have this changed in my coupled solver branch.

src/solvers/elasticity_solver.cpp
jamiebramwell: So I've been leaving the `ElasticitySolver` alone for the most part because I expect it to be merged with the `NonlinearSolidSolver` once we figure out the material treatment issue.

jamiebramwell:
LGTM

Issue #12, opened 2020-03-04
Add entity set manager for boundary conditions


jamiebramwell:
Add a more sensible interface for boundary conditions based on lists of geometric entities (faceset, nodeset, edgeset, etc).

PR #75, opened 2020-07-09
Use std::set instead of std::vector to indicate boundary attributes for BCs

jamiebramwell:
This changes the user-facing interface for solvers to use `std::set`s instead of `std::vector`s of indicator values to denote the boundary attributes for applying boundary conditions

### DIFF
--- a/src/solvers/base_solver.cpp
+++ b/src/solvers/base_solver.cpp
@@ -29,26 +29,24 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n) : BaseSolver(comm)
   m_gf_initialized.assign(n, false);
 }
 
-void BaseSolver::SetEssentialBCs(const std::vector<int> &                 ess_bdr,
+void BaseSolver::SetEssentialBCs(const std::set<int> &                 ess_bdr,
                                  std::shared_ptr<mfem::VectorCoefficient> ess_bdr_vec_coef,
                                  mfem::ParFiniteElementSpace &fes, int component)
 {
   auto bc = std::make_shared<BoundaryCondition>();
 
-  bc->markers.SetSize(ess_bdr.size());
+  bc->markers.SetSize(m_state.front()->mesh->bdr_attributes.Max());
+  bc->markers = 0;
 
-  for (unsigned int i = 0; i < ess_bdr.size(); ++i) {
-    bc->markers[i] = ess_bdr[i];
-
-    if (bc->markers[i] == 1) {
-      for (auto &existing_bc : m_ess_bdr) {
-        if (existing_bc->markers[i] == 1) {
-          mfem::mfem_warning("Multiple definition of essential boundary! Using first definition given.");
-          bc->markers[i] = 0;
-          break;
-        }
+  for (int attr : ess_bdr) {
+    bc->markers[attr-1] = 1;
+    for (auto &existing_bc : m_ess_bdr) {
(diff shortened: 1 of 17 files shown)


jamiebramwell:
LGTM

tests/serac_component_bc.cpp
```diff
@@ -135,13 +132,9 @@ TEST(component_bc, qs_attribute_solve)
...
   NonlinearSolidSolver solid_solver(2, pmesh);
 
-  // define a boundary attribute array and initialize to 0
-  std::vector<int> ess_x_bdr(pmesh->bdr_attributes.Max(), 0);
-  std::vector<int> ess_y_bdr(pmesh->bdr_attributes.Max(), 0);
-
   // boundary attribute 1 (index 0) is fixed (Dirichlet) in the x direction
-  ess_x_bdr[0] = 1;
-  ess_y_bdr[1] = 1;
+  std::set<int> ess_x_bdr = {1};
```
white238, line 136: School me please, what does 1 vs 2 represent/mean?

white238 (approved):

tests/serac_component_bc.cpp
jamiebramwell: The input mesh has different markers (1,2,etc) for each part of the boundary. This is saying the boundary with the '1' markers are of the essential x boundary condition type. You could have multiple markers for a single boundary, so I thought a set was more appropriate. The mfem native way of dealing with this is a vector of length maximum number of attributes that is zero if an attribute isn't active and 1 if it is. I thought this was much more user friendly.

jamiebramwell:
LGTM

src/solvers/base_solver.cpp
```diff
@@ -29,26 +29,25 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n) : BaseSolver(comm)
...
-          bc->markers[i] = 0;
-          break;
-        }
+  bc->markers.SetSize(m_state.front()->mesh->bdr_attributes.Max());
+  bc->markers = 0;
+
+  for (int attr : ess_bdr) {
+    MFEM_ASSERT(attr <= bc->markers.Size(), "Attribute specified larger than what is found in the mesh.");
+    bc->markers[attr-1] = 1;
+    for (auto &existing_bc : m_ess_bdr) {
```
joshessman-llnl, line 44: I'm not sure how big the "real" meshes get, but depending on the number of boundary conditions/attributes, the solver could also store a set of attributes already assigned to a boundary condition and avoid the linear search here.  

It might also worth be comparing `std::set` and `std::unordered_set` for the boundary attribute indications, but for small sets the difference is negligible.

joshessman-llnl (approved):

src/solvers/base_solver.cpp
jamiebramwell: These will typically be extremely small. However, this is really a stop gap measure until we have a real "entity set" type of boundary marker. The problem here is that the internal MFEM mesh format can only have one attribute marker or "color" per boundary face. In complex multiphysics simulations, you can easily have overlapping areas for different boundary conditions. Production codes use face or nodesets to get around this. Development of this capability is logged in this [issue](https://github.com/LLNL/serac/issues/12).

jamiebramwell:
LGTM

jamiebramwell:
This changes the user-facing interface for solvers to use std::sets instead of std::vectors of indicator values to denote the boundary attributes for applying boundary conditions

PR #80, opened 2020-07-16
Renaming master to develop in CI scripts

jamiebramwell:
This updates the gitlab CI scripts with the new develop branch name.

PR #81, opened 2020-07-16
Modify CI scripts to reflect switch to gitflow and renaming of primary branch

### LINKED ISSUE #77: Be a real project and use the Gitflow workflow
https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow

We shouldn't just have one branch.  We should move to having most of our development be in a new "develop" branch and rename "master" to "main" (this seems to be the defacto new standard).

joshessman-llnl:
Picking back up on the discussion in #80 , does the build with dependencies need to happen for `develop`, `main`, or both?

Resolves #77

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -23,11 +23,11 @@ stages:
 
 .pr_workflow:
   rules:
-    - if: '$CI_COMMIT_BRANCH != "master" && $BUILD_DEPS != "ON"' #run only if ...
+    - if: '$CI_COMMIT_BRANCH != "main" && $CI_COMMIT_BRANCH != "develop" && $BUILD_DEPS != "ON"' #run only if ...
 
-.master_workflow:
+.main_workflow:
   rules:
-    - if: '$CI_COMMIT_BRANCH == "master" || $BUILD_DEPS == "ON"' #run only if ...
+    - if: '$CI_COMMIT_BRANCH == "main" || $CI_COMMIT_BRANCH == "develop" || $BUILD_DEPS == "ON"' #run only if ...
 
 ####
 # Template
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -23,11 +23,11 @@ stages:
 
 .pr_workflow:
   rules:
-    - if: '$CI_COMMIT_BRANCH != "master" && $BUILD_DEPS != "ON"' #run only if ...
+    - if: '$CI_COMMIT_BRANCH != "main" && $CI_COMMIT_BRANCH != "develop" && $BUILD_DEPS != "ON"' #run only if ...
 
-.master_workflow:
+.main_workflow:
```
white238, line 28: This might need to coincide with a change in uberenv. @adrienbernede  , who may not be available to answer atm?

white238 (approved):

.gitlab-ci.yml
adrienbernede: Processing...

adrienbernede:
The build with dependencies should happen for each merge into develop.
We don’t have a workflow defined for "main" yet, so we can leave the same as the one for "develop" for now.

.gitlab-ci.yml
adrienbernede: What I will do:
- Sign-off this branch (LGTM) so that it gets imported to Gitlab.
- Trigger a pipeline that mimics "main" or "develop" case on Gitlab to vet that it works fine. (we can manually trigger a pipeline, and setting BUILD_DEPS == "ON" will mimic the main workflow.)
- Push the corresponding change in Uberenv CI, so that it points to develop instead of master.

adrienbernede (approved):
If I wanted to be picky, I would have said "master" should be changed for "develop" in CI.
But the result is the same since we don’t have a "main" workflow yet. So let’s keep things as-is.
Thanks for diving into the CI for your first week, and welcome!

adrienbernede:
LGTM

adrienbernede:
LGTM

PR #82, opened 2020-07-16
Move everything in serac_types.hpp into namespace serac

### LINKED ISSUE #34: SeracTypes.h needs to be in a namespace
For clarity of where things are coming from....

joshessman-llnl:
Fixes #34


jamiebramwell (commented):
Looks great! My only thought is should we also add the solvers and other objects to the namespace? Any thoughts @white238 ?

samuelpmishLLNL:
Personally, I think it would be cleaner to have `using namespace serac` in our own cpp files, to make things a little less verbose. Any other thoughts?

jamiebramwell:
> 
> 
> Personally, I think it would be cleaner to have `using namespace serac` in our own cpp files, to make things a little less verbose. Any other thoughts?

I like this idea!

joshessman-llnl:
> Personally, I think it would be cleaner to have `using namespace serac` in our own cpp files, to make things a little less verbose. Any other thoughts?

There was some talk of Google's C++ style guide in #83 with regard to exceptions, so it might be worth mentioning that they're similarly strict when it comes to `using` directives:
> Do not use using-directives (e.g., `using namespace foo`).

As a sort of middle ground, we could individually pull the relevant types into global scope, e.g., `using serac::FiniteElementState;`, which the style guide does permit.

jamiebramwell:
@white238 , any thoughts on the google style conflict with the `using namespace` issue? I think this guideline is overly strict for a self-owned namespace, but I am admittedly not an expert in this area.

white238:
> @white238 , any thoughts on the google style conflict with the `using namespace` issue? I think this guideline is overly strict for a self-owned namespace, but I am admittedly not an expert in this area.

I am with @joshessman-llnl and google on this.  Never in a header (obviously) but you can have it in a source file. Also to restrict it as much as possible as long as you aren't bringing in multiple things. My main complaint with using is it makes it hard at a glance to not know where things are coming from especially when you are dealing with a lot of libraries but I know I'm in the minority on this.

jamiebramwell:
I opened a new issue to work out and document some of these style decisions: https://github.com/LLNL/serac/issues/86.

jamiebramwell (approved):
This is a good first step toward healthier namespace practices.

white238 (approved):

white238:
@joshessman-llnl can you swap the two std::cout calls to the proper SLIC macros? These were missed in the merging of the previous CLI11 PR.  Thanks!

joshessman-llnl:
LGTM

PR #78, opened 2020-07-15
Add basic logging functionality

white238:
* Adds SLIC logging functionality
* Add an exit function that cleans stuff up (serac::exit_gracefully(bool error=false)
* Starts using the new serac C++ namespace

I haven't gone through any of the code other than the driver yet but wanted to open this up for discussion.  This can and should be iterated over.  There is a lot of formatting of information in SLIC calls that can be done.

**Example error** (notice the automatic and human readable stacktrace, filename and line number, so fancy):

```
[ERROR (/usr/WS2/white238/serac/repo/src/drivers/serac.cpp:130)]

Can not open mesh file: fake
** StackTrace of 4 frames **
Frame 1: bin/serac() [0x95bf93]
Frame 2: bin/serac() [0x40bfd7]
Frame 3: __libc_start_main
Frame 4: bin/serac() [0x40b5a4]
=====
```

**Example info message:**

```
[INFO]: Opening mesh file: fake
```

If you run with more than one rank, a limited amount of ranks will show up at the beginning of the line like this:

```
[0,1,2][INFO]: Opening mesh file: fake
```

PR #83, opened 2020-07-17
Change from MFEM's command line arguments to CLI11

### LINKED ISSUE #76: Change from MFEM's command line arguments to CLI11
Axom has built in support for CLI11.  We should switch to that and away from MFEM's.  CLI11 is more industry standard and is C++ vs C (ie std::strings vs char*).

joshessman-llnl:
The only difference in functionality is that multi-character "short names" must now be preceded by `--` instead of `-`, e.g., `-rs` is now `--rs`.

For more info see my comment [here](https://github.com/LLNL/serac/issues/76#issuecomment-659614949).  Resolves #76.


white238:
This is a bit non-standard (shorter two letter and the -- for shorter options) but this may be a stop gap for now until we get the input deck functionality then we can review what command options we really need.

white238 (approved):

src/drivers/serac.cpp
```diff
@@ -75,46 +75,48 @@ int main(int argc, char *argv[])
...
+  app.add_option("--nabs, --newton-absolute-tolerance", nonlin_params.abs_tol, 
+                  "Absolute tolerance for the Newton solve.", true);
+  app.add_option("--nit, --newton-iterations", nonlin_params.max_iter,  "Maximum iterations for the Newton solve.", true);
+  app.add_option("--npl, --newton-print-level", nonlin_params.print_level,  "Newton print level.", true);
+  app.add_option("--dt, --time-step", dt,  "Time step.", true);
 
   // Parse the arguments and check if they are good
-  args.Parse();
-  if (!args.Good()) {
+  try {
```
jamiebramwell, line 112: Real exception throws! Yay!

jamiebramwell (approved):
Looks good!

jamiebramwell:
LGTM

src/drivers/serac.cpp
samuelpmishLLNL: We did discuss the possibility of adopting a Google C++ code style guide in the past, and their policy on exceptions is pretty severe:

`Do not use C++ exceptions.`

I have no objection to the use of exceptions here, but I think it might be worth restricting their use to the higher level parts of the code.

jamiebramwell: Agreed that exceptions should only be allowed on this "driver" level. We should also probably start drafting a code style guide.

jamiebramwell: Actually, this will probably get replaced by a SLIC call. See [this PR.](https://github.com/LLNL/serac/pull/78)

jonwong12: A dumb question, but this will be "thrown" if there is an undefined command line argument correct? On my other project I know we have waffled back and forth on this because it "insta-kills" batch jobs. 

I'm wondering if there is a way we can "trial-run-test" our command-line options without actually running anything?

joshessman-llnl: > it might be worth restricting their use to the higher level parts of the code.

We could also use a returncode-based wrapper function, possibly within the CLI11 built into axom, to avoid directly "infecting" user code, like drivers, with exceptions.

> but this will be "thrown" if there is an undefined command line argument correct?  I'm wondering if there is a way we can "trial-run-test" our command-line options without actually running anything?

Yes, CLI11 will throw exceptions for any invalid arguments.  Just to clarify, are you referring to something where you could run `serac --test-args -a -b --undefined-cmdline-arg` and it would only check/report the validity of arguments before exiting?

jamiebramwell:
LGTM

jamiebramwell:
LGTM

jamiebramwell:
@joshessman-llnl - that LGTM comment kicks off testing on our LC owned gitlab instance. You must be a member of the LLNL team and have 2FA enabled for this to work. We have a requirement that these tests pass before we can merge in addition to the cheaper azure driven tests.

Issue #29, opened 2020-03-18
Move to using a SidreDataCollection for state variable storage


jamiebramwell:
Look at [NamedFieldsMap](https://github.com/mfem/mfem/blob/b584a1a82dc17a0e46b5e3c760f8edcb3107ab8d/fem/datacollection.hpp#L29) for guidance on named lookups to the data collection.

joshessman-llnl:
## MFEMSidreDataCollection

The `MFEMSidreDataCollection` class is used to store the state of a simulation, namely, the mesh, associated fields, and simulation information like the current time and frame number.
The state is stored in a global Sidre tree (following the Conduit Blueprint structure) and can be saved in HDF5 (and JSON) format for visualization or restart purposes.

An abbreviated usage looks like the following:

```cpp
// Finite Element setup
mfem::Mesh mesh = loadMeshFromFile(mesh_filename);
mfem::H1_FEC fec(1,2);
mfem::FES fes(mesh, fec);
mfem::GridFunction soln(fes);

// Datastore Setup
MFEMSidreDataCollection dc;
dc.SetMesh(mesh);
dc.RegisterField("solution", soln);
dc.SetTime(0);

// for each iteration
dc.SetTime(t);
dc.Save();
```

## Restart

When initialized with a mesh, the `MFEMSidreDataCollection` extracts and retains information in the datastore sufficient to reconstruct a mesh.
It also contains data sufficient to reconstruct `mfem::GridFunctions`.  However, this is not implemented, so a newly constructed `MFEMSidreDataCollection` cannot simply
`Load()` from a file and have a state identical to the `MFEMSidreDataCollection` that was `Save()`d.  

For this reason, data is externally reconstructed in MARBL (pseudocode):

```cpp
// Finite Element setup
mfem::Mesh mesh = loadMeshFromFile(mesh_filename);
mfem::H1_FEC fec(1,2);
mfem::FES fes(mesh, fec);
mfem::GridFunction soln(fes);

// The data is "manually" pulled out of the datastore
// and used to repopulate the gridfunction 
DataStore ds = loadDSFromFile(datastore_filename);
FillFieldFromDataStore(ds, soln); 

MFEMSidreDataCollection dc(ds);

// mfem::Mesh& mesh = dc.GetMesh(); // Returns nullptr, mesh was not reconstructed
// Returns nullptr, field was not reconstructed even though its data is in the Datacollection
// mfem::GridFunction& soln = dc.GetField("solution");

[...]

PR #88, opened 2020-07-21
Add SidreDataCollection as an output option

### LINKED ISSUE #85: Add Sidre output option
In addition to VisIt and GLVis

joshessman-llnl:
Closes #85.


joshessman-llnl:
The Docker images used by the Azure pipelines may need to be updated, as MFEM support for Sidre is set at compile time.

white238:
This needs to be discussed before this gets pushed anywhere.  MFEM needs to not depend on Axom because Axom has to depend on MFEM.  I know it does currently but this is a circular dependency that because of time constraints has not been broken.  We need to do this right before this gets harder to clean up in the future.

joshessman-llnl:
I agree, this (and #29) present a good opportunity/motivation to resolve the circular dependency upstream.

samuelpmishLLNL (approved):

joshessman-llnl:
This is a first step toward integrating SidreDataCollection that incorporates it into the build and adds it as an output object as a sort of "smoke test".

src/physics/utilities/solver_config.hpp
```diff
@@ -25,7 +25,8 @@ enum class OutputType
 {
   GLVis,
   ParaView,
-  VisIt
+  VisIt,
```
white238, line 28: Would we ever want a VisIt output w/o going through Sidre? Seems like we should just hide our use of Sidre under the VisIt output type

joshessman-llnl: `VisitDataCollection` is probably more stable than `MFEMSidreDataCollection`, so it may be worth having it as an option as we tune SidreDC.

jamiebramwell: I agree with @joshessman-llnl . I would expect a fair amount of churn in the `MFEMSidreDataCollection`, so it's nice to have a stable option.

jamiebramwell:
@white238 Can you take another look at this?

src/physics/utilities/solver_config.hpp
white238: It's more of a philosophical concern.. `sidre` is not an output format.  As long as its temporary its fine but `SidreVisIt` is closer to the truth.

white238 (approved):

white238 (approved):

joshessman-llnl:
LGTM

Issue #99, opened 2020-07-23
Upgrade to C++17


jamiebramwell:
Upgrade the standard used to c++17. Compatibility with CUDA and BLT as well as compilers on ATS LC systems (like Sierra) needs to be investigated.

joshessman-llnl:
The clang 4.0.0 CI version appears to not be plug-and-play with C++17, per the [logs](https://dev.azure.com/llnl-serac/serac/_build/results?buildId=434&view=logs&j=ee5e116a-579b-5a02-f53f-47b9bf970bfe&t=768a03e3-8edf-5f95-65fc-a159c8edc275&l=308).  This may be in BLT's scope, as Clang 4 requires `-std=c++1z` as opposed to `-std=c++17` (which is what GCC 8 uses).

joshessman-llnl:
`nvcc` version 11 or higher is required per [the Nvidia website](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#cpp17-language-features).

white238:
clang@4.0.0 should be dropped.

jamiebramwell:
It looks like the [axom-maintained docker compiler containers](https://hub.docker.com/r/axom/compilers/tags) have newer versions now, so I fully support updating our clang version.

jamiebramwell:
"now" - looks like they have for a while and I was just lazy.

PR #105, opened 2020-07-28
Memory audit and general C++ correctness

### LINKED ISSUE #91: Audit the memory management strategy of the solver class
Now that a few solvers (including a composite solver) exist, a reexamination of the memory management strategies used for each object may be useful.

### LINKED ISSUE #95: Audit const correctness
I am sure in my rush to get initial functionality out, I missed some things which could (and should) be marked `const`.

joshessman-llnl:
There's more that can be done but in the interest of minimizing the size of PRs I think it makes sense to merge in what I have now, especially because this one is fairly large.

Main things addressed by this PR include:

- Using of the `const`, `explicit`, and `override` keywords to better communicate intent
- Limiting the use of raw pointers which have to be manually `new`'d and `delete`'d
- Limiting the use of `std::shared_ptr` to avoid shared mutable state (there is definitely more than can be done on this)
- Updating all the CI/docker scripts to reflect a switch from Clang 4 to Clang 6 (Clang 10 will be added pending a new axom Docker image)
- Conversion of out-params into return values to better communicate intent

There was also an interface change to `BaseSolver::setTrueDofs` due to an unintentionally zero-initialized variable that caused things to work correctly until some of the above memory changed were made.

An `m_` slipped through the cracks on the style PR (I think), so that was also fixed.

Partially addresses #91, #95, and #99.


CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ endif()
 
 # Tune BLT to our needs
 if (NOT BLT_CXX_STD)
-    set(BLT_CXX_STD "c++14" CACHE STRING "")
+    set(BLT_CXX_STD "c++17" CACHE STRING "")
```
jamiebramwell: Have we checked if BLT with CUDA supports this? I'd like to get started on GPU stuff and this might be a problem.

azure-pipelines.yml
```diff
@@ -19,13 +19,13 @@ strategy:
       COMPILER: 'g++'
       TEST_TARGET: 'linux_gcc8'
       HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@8.1.0'
-   linux_clang4:
+   linux_clang6:
```
jamiebramwell: 👍

host-configs/docker/docker-linux-ubuntu16.04-x86_64-clang@6.0.0.cmake
```diff
@@ -21,21 +21,25 @@ set(CMAKE_C_COMPILER "/usr/bin/clang" CACHE PATH "")
 
 set(CMAKE_CXX_COMPILER "/usr/bin/clang++" CACHE PATH "")
 
+set(CMAKE_CXX_FLAGS "-stdlib=libc++" CACHE PATH "")
```
jamiebramwell: @white238 I know BLT can have weirdness with stdlib definitions. Is it best practice to hard set this? I assume this is because we need new c++17 stuff.

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -34,68 +34,70 @@ void StdFunctionVectorCoefficient::Eval(mfem::Vector& V, mfem::ElementTransforma
   func_(transip, V);
 }
 
-void makeTrueEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoefficient& c, mfem::Array<int>& ess_tdof_list)
+mfem::Array<int> makeTrueEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoefficient& c)
```
jamiebramwell, line 37: I am unlearning the habits I have learned! Thanks for catching this!

src/common/serac_types.hpp
```diff
@@ -86,12 +87,12 @@ struct FiniteElementState {
 
 // Boundary condition information
 struct BoundaryCondition {
-  mfem::Array<int>                         markers;
-  mfem::Array<int>                         true_dofs;
-  int                                      component;
-  std::shared_ptr<mfem::Coefficient>       scalar_coef;
-  std::shared_ptr<mfem::VectorCoefficient> vec_coef;
-  std::unique_ptr<mfem::HypreParMatrix>    eliminated_matrix_entries;
+  using Coef = std::variant<std::shared_ptr<mfem::Coefficient>, std::shared_ptr<mfem::VectorCoefficient>>;
```
jamiebramwell: I really like this use of `std::variant` because a boundary condition should never have a scalar and vector coefficient at the same time.

src/integrators/inc_hyperelastic_integrator.hpp
```diff
@@ -33,7 +33,7 @@ class IncrementalHyperelasticIntegrator : public mfem::NonlinearFormIntegrator {
 
  public:
   /** @param[in] m  HyperelasticModel that will be integrated. */
-  IncrementalHyperelasticIntegrator(mfem::HyperelasticModel* m) : model_(m) {}
+  explicit IncrementalHyperelasticIntegrator(mfem::HyperelasticModel* m) : model_(m) {}
```
jamiebramwell, line 36: I assume it's best practice to use `explicit` whenever possible now?

src/solvers/base_solver.cpp
```diff
@@ -28,160 +29,95 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
   order_ = p;
   state_.resize(n);
 
-  for (auto& state : state_) {
-    state = std::make_shared<serac::FiniteElementState>();
-  }
+  std::generate(state_.begin(), state_.end(), std::make_shared<serac::FiniteElementState>);
```
jamiebramwell, line 32: 👍

src/solvers/base_solver.cpp
```diff
@@ -28,160 +29,95 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
-      }
+  serac::BoundaryCondition bc;
+
+  bc.markers.SetSize(state_.front()->mesh->bdr_attributes.Max());
+  bc.markers = 0;
+
+  for (const int attr : ess_bdr) {
+    SLIC_ASSERT_MSG(attr <= bc.markers.Size(), "Attribute specified larger than what is found in the mesh.");
+    bc.markers[attr - 1] = 1;
+    if (std::any_of(ess_bdr_.begin(), ess_bdr_.end(),
```
jamiebramwell: I've been meaning to learn the `std::algorithm` library better. Thank you for catching these!

CMakeLists.txt
jamiebramwell: https://github.com/LLNL/blt/issues/360 once this goes in and we move to cmake 3.18.0 we should be set!

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -31,7 +31,7 @@ class StdFunctionCoefficient : public mfem::Coefficient {
  public:
   /// Constructor that takes in an mfem Vector representing the coordinates and
   /// produces a double
-  StdFunctionCoefficient(std::function<double(mfem::Vector&)> func);
+  explicit StdFunctionCoefficient(std::function<double(mfem::Vector&)> func);
```
samuelpmishLLNL: I think there is some value in keeping this constructor not explicit.

src/common/serac_types.hpp
samuelpmishLLNL: it may be useful to move this typedef outside the BoundaryCondition scope, anyone else have thoughts on this?

src/solvers/base_solver.cpp
```diff
@@ -28,160 +29,95 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
-  for (int attr : ess_bdr) {
-    SLIC_ASSERT_MSG(attr <= bc->markers.Size(), "Attribute specified larger than what is found in the mesh.");
-    bc->markers[attr - 1] = 1;
-    for (auto& existing_bc : ess_bdr_) {
-      if (existing_bc->markers[attr - 1] == 1) {
-        SLIC_WARNING("Multiple definition of essential boundary! Using first definition given.");
-        bc->markers[attr - 1] = 0;
-        break;
-      }
+  serac::BoundaryCondition bc;
```
samuelpmishLLNL: hooray for value semantics!

src/solvers/base_solver.cpp
```diff
@@ -28,160 +29,95 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
-
-void BaseSolver::setState(const std::vector<std::shared_ptr<mfem::VectorCoefficient> >& state_vec_coef)
-{
-  SLIC_ASSERT_MSG(state_vec_coef.size() == state_.size(), "State and coefficient bundles not the same size.");
-
-  for (unsigned int i = 0; i < state_vec_coef.size(); ++i) {
-    state_[i]->gf->ProjectCoefficient(*state_vec_coef[i]);
+    // The generic lambda parameter, auto&&, allows the component type (mfem::Coef or mfem::VecCoef)
+    // to be deduced, and the appropriate version of ProjectCoefficient is dispatched.
+    std::visit([this, i](auto&& coef) { state_[i]->gf->ProjectCoefficient(*coef); }, state_coef[i]);
```
samuelpmishLLNL: @jamiebramwell this is the visitor pattern that goes with `std::variant`

src/solvers/elasticity_solver.cpp
```diff
@@ -70,7 +62,7 @@ void ElasticitySolver::completeSetup()
   SLIC_ASSERT_MSG(lambda_ != nullptr, "Lame lambda not set in ElasticitySolver!");
 
   // Define the parallel bilinear form
-  K_form_ = new mfem::ParBilinearForm(displacement_->space.get());
+  K_form_ = std::make_unique<mfem::ParBilinearForm>(displacement_->space.get());
```
samuelpmishLLNL, line 61: I believe we left some of these pointers raw because a handful of MFEM functions deallocate the pointers passed in. It may be necessary to call `std::unique_ptr::release` before those sink functions. Johnathan Wong started a list of some of the MFEM functions that deallocate, but it isn't complete.

adrienbernede: @jonwong12 ☝️

src/coefficients/stdfunction_coefficient.hpp
jamiebramwell: Do we want this to still work with a generic lambda rvalue? I'm not sure if this `explicit` prevents that.

src/solvers/base_solver.cpp
```diff
@@ -28,160 +29,95 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
-  }
+  bc.true_dofs = true_dofs;
 
-  bc->vec_coef  = nat_bdr_vec_coef;
-  bc->component = component;
+  bc.coef = ess_bdr_coef;
 
-  nat_bdr_.push_back(bc);
-}
+  SLIC_ERROR_IF(!std::holds_alternative<std::shared_ptr<mfem::VectorCoefficient>>(ess_bdr_coef) && (component == -1),
```
jamiebramwell: This is only true for vector valued basis functions. If a basis function is a single H1 (like the heat equation), you can have a scalar coefficient with a component of -1.

src/solvers/elasticity_solver.cpp
jonwong12: To my knowledge (not complete), only the `integrators` get deleted automatically by mfem. But @samuelpmishLLNL is correct that certain things like `integrators` or certain coefficients get automatically deleted :/.

src/solvers/base_solver.cpp
jamiebramwell: Things like this make me wonder if `std::variant` for the coefficients is worth it. Any thoughts @white238 ?

src/solvers/base_solver.cpp
```diff
@@ -28,160 +29,95 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
-  for (unsigned int i = 0; i < state_vec_coef.size(); ++i) {
-    state_[i]->gf->ProjectCoefficient(*state_vec_coef[i]);
+    // The generic lambda parameter, auto&&, allows the component type (mfem::Coef or mfem::VecCoef)
+    // to be deduced, and the appropriate version of ProjectCoefficient is dispatched.
+    std::visit([this, i](auto&& coef) { state_[i]->gf->ProjectCoefficient(*coef); }, state_coef[i]);
   }
 }
 
-void BaseSolver::setState(const std::vector<std::shared_ptr<serac::FiniteElementState> > state)
+void BaseSolver::setState(const std::vector<std::shared_ptr<serac::FiniteElementState>>& state)
```
jamiebramwell: are there any good guides to passing `shared_ptr` by reference or value? I tried to look into it when I did this the first time, and I went down an internet fight rabbit hole.

src/solvers/elasticity_solver.hpp
```diff
@@ -24,28 +24,28 @@ class ElasticitySolver : public BaseSolver {
   std::shared_ptr<serac::FiniteElementState> displacement_;
 
   /// Stiffness bilinear form
-  mfem::ParBilinearForm* K_form_;
+  std::unique_ptr<mfem::ParBilinearForm> K_form_;
```
jamiebramwell, line 27: Whoops! I haven't changed this one since the plan is to combine it with the `NonlinearSolidSolver`. Thanks for cleaning it up!

src/solvers/nonlinear_solid_operators.cpp
```diff
@@ -31,11 +31,12 @@ mfem::Operator& NonlinearSolidQuasiStaticOperator::GetGradient(const mfem::Vecto
 // destructor
 NonlinearSolidQuasiStaticOperator::~NonlinearSolidQuasiStaticOperator() {}
 
-NonlinearSolidDynamicOperator::NonlinearSolidDynamicOperator(
-    std::shared_ptr<mfem::ParNonlinearForm> H_form, std::shared_ptr<mfem::ParBilinearForm> S_form,
-    std::shared_ptr<mfem::ParBilinearForm>                         M_form,
-    const std::vector<std::shared_ptr<serac::BoundaryCondition> >& ess_bdr, mfem::NewtonSolver& newton_solver,
-    const serac::LinearSolverParameters& lin_params)
+NonlinearSolidDynamicOperator::NonlinearSolidDynamicOperator(std::shared_ptr<mfem::ParNonlinearForm>      H_form,
+                                                             std::shared_ptr<mfem::ParBilinearForm>       S_form,
```
jamiebramwell: What the heck caused the style to format it like it was before. That's gross!

src/solvers/elasticity_solver.hpp
```diff
@@ -81,10 +81,10 @@ class ElasticitySolver : public BaseSolver {
   void setLameParameters(mfem::Coefficient& lambda, mfem::Coefficient& mu);
 
   /// Set the vector-valued body force coefficient
-  void setBodyForce(mfem::VectorCoefficient& force);
+  void setBodyForce(const mfem::VectorCoefficient& force);
 
   /// Finish the setup and allocate the associate data structures
-  void completeSetup();
+  void completeSetup() override;
```
jamiebramwell, line 87: The `advanceTimestep` call is also an override.

jamiebramwell (commented):
I will continue to look at this later.

src/solvers/elasticity_solver.cpp
```diff
@@ -166,27 +158,20 @@ void ElasticitySolver::QuasiStaticSolve()
   // Apply the boundary conditions
   *bc_rhs_ = *rhs_;
   for (auto& bc : ess_bdr_) {
-    bc->vec_coef->SetTime(time_);
-    displacement_->gf->ProjectBdrCoefficient(*bc->vec_coef, bc->markers);
+    SLIC_ERROR_IF(!std::holds_alternative<std::shared_ptr<mfem::VectorCoefficient>>(bc.coef),
```
jamiebramwell: This is some more clunkiness with `std::variant`. I'm still on the fence if this is worth it.

src/solvers/nonlinear_solid_operators.hpp
```diff
@@ -115,21 +115,21 @@ class NonlinearSolidDynamicOperator : public mfem::TimeDependentOperator {
   mfem::NewtonSolver& newton_solver_;
 
   /// The fixed boudnary degrees of freedom
-  const std::vector<std::shared_ptr<serac::BoundaryCondition> >& ess_bdr_;
+  const std::vector<BoundaryCondition>& ess_bdr_;
```
jamiebramwell: While this is true now, I can see users changing the boundary conditions as the run progresses. That could be part of a larger boundary condition refactoring though.

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -111,7 +111,7 @@ void NonlinearSolidSolver::setSolverParameters(const serac::LinearSolverParamete
 void NonlinearSolidSolver::completeSetup()
 {
   // Define the nonlinear form
-  H_form_ = std::make_shared<mfem::ParNonlinearForm>(displacement_->space.get());
+  auto H_form_ = std::make_shared<mfem::ParNonlinearForm>(displacement_->space.get());
```
jamiebramwell: should this be `unique` and then we transfer it via `std::move` to the nonlinear operator?

src/common/serac_types.hpp
joshessman-llnl: My thinking when placing it in that scope was that `serac::BoundaryCondition::Coef` communicates that the type is used exclusively for boundary conditions, though I'm interested to hear what other people think

src/solvers/elasticity_solver.cpp
joshessman-llnl: I'm reassured by the lack of double-free errors from Valgrind, but I'll check the MFEM source to be 100% sure

host-configs/docker/docker-linux-ubuntu16.04-x86_64-clang@6.0.0.cmake
joshessman-llnl: The axom `compilers.yaml` (which propagated down to serac) uses this flag, so it needed to end up in the hostconfig for the TPLs to link properly (uberenv inserts the flag automatically).

jamiebramwell:
I think it might be better if you separate the `std::variant` stuff into a separate PR. I think it's hard to really evaluate this choice when it's co-mingled with all of the C++ corrections.

src/integrators/inc_hyperelastic_integrator.hpp
joshessman-llnl: Usually you want single-argument constructors to be `explicit`.  The [isocpp guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-explicit) have a section on it, but the main idea is that you want to avoid implicit conversions, e.g., if you have a function that takes an `mfem::Vector` (which has an integer size ctor), you could accidentally pass in the scalar value "4" and not catch the error until runtime since it would get promoted to a vector of size 4.

src/solvers/base_solver.cpp
joshessman-llnl: `cppcheck --enable=all` will sometimes (but not always) give you a hint about using a `std::algorithm` instead of a raw for loop.

src/coefficients/stdfunction_coefficient.hpp
joshessman-llnl: That makes sense, I'll go ahead and change it

src/solvers/base_solver.cpp
joshessman-llnl: There's some [stuff in the isocpp guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-smartptrparam) that gets discussed in further detail on [one of the C++ blogs I read](https://www.modernescpp.com/index.php/c-core-guidelines-passing-smart-pointer).

CMakeLists.txt
samuelpmish: I'm not sure how BLT implements that flag, but I found that the idiomatic CMake way (setting CMAKE_CUDA_STANDARD to 17) did not work for me on Windows. Although, that was with a 3.18rc-- the recently released 3.18 may have addressed that.

src/coefficients/stdfunction_coefficient.cpp
white238: Are you sure this is not doing a deep copy of the mfem::Array?

joshessman-llnl:
> I think it might be better if you separate the `std::variant` stuff into a separate PR. I think it's hard to really evaluate this choice when it's co-mingled with all of the C++ corrections.

That's fair, I've gone ahead and split it into a separate branch.  As a result, there's some boundary condition-related memory changes that are no longer in this PR (like the value semantics).  

All the C++17 changes (which may need to wait on BLT) are also in the other branch.

src/solvers/base_solver.hpp
```diff
@@ -13,133 +13,131 @@
 #include "common/serac_types.hpp"
 #include "mfem.hpp"
 
-namespace serac {
-
 /// This is the abstract base class for a generic forward solver
 class BaseSolver {
+ private:
```
jamiebramwell: It unfortunately looks like after your revert this missing the great "style-ing". Sorry!

src/solvers/base_solver.cpp
```diff
@@ -252,40 +190,80 @@ void BaseSolver::initializeOutput(const serac::OutputType output_type, std::stri
...
 
     default:
-      SLIC_ERROR_ROOT(mpi_rank_, "OutputType not recognized!");
-      serac::exitGracefully(true);
+      SLIC_ERROR_ROOT(m_rank, "OutputType not recognized!");
+      serac::ExitGracefully(true);
+  }
+}
+
+void BaseSolver::RegisterEssentialBC(std::shared_ptr<serac::BoundaryCondition> bc, const std::set<int> &ess_bdr,
```
jamiebramwell: Moving these to `private` makes a lot of sense.

src/solvers/base_solver.cpp
```diff
@@ -252,40 +190,80 @@ void BaseSolver::initializeOutput(const serac::OutputType output_type, std::stri
...
+          return existing_bc->markers[attr-1] == 1;
+        })) {
+      SLIC_WARNING("Multiple definition of essential boundary! Using first definition given.");
+      bc->markers[attr-1] = 0;
+    }
   }
+
+  bc->component = component;
+
+  // This function can and should be marked const in MFEM
```
jamiebramwell: 👍 They're pretty responsive on github and they have a good developer guide [here](https://github.com/mfem/mfem/blob/master/CONTRIBUTING.md) if you want to kick this up the chain.

src/solvers/base_solver.cpp
```diff
@@ -252,40 +190,80 @@ void BaseSolver::initializeOutput(const serac::OutputType output_type, std::stri
...
+  bc->markers = 0;
+
+  for (const int attr : nat_bdr) {  
+    SLIC_ASSERT_MSG(attr <= bc->markers.Size(), "Attribute specified larger than what is found in the mesh.");
+    bc->markers[attr-1] = 1;
+  }
+  bc->component   = component;
+
+  m_nat_bdr.push_back(bc);
+}
```
jamiebramwell: Newline missing

src/solvers/elasticity_solver.cpp
```diff
@@ -42,14 +42,14 @@ ElasticitySolver::ElasticitySolver(int order, std::shared_ptr<mfem::ParMesh> pme
...
+void ElasticitySolver::setDisplacementBCs(const std::set<int>&                     disp_bdr,
                                           std::shared_ptr<mfem::VectorCoefficient> disp_bdr_coef, int component)
 {
   setEssentialBCs(disp_bdr, disp_bdr_coef, *displacement_->space, component);
 }
 
-void ElasticitySolver::setTractionBCs(std::set<int>& trac_bdr, std::shared_ptr<mfem::VectorCoefficient> trac_bdr_coef,
-                                      int component)
+void ElasticitySolver::setTractionBCs(const std::set<int>&                     trac_bdr,
+                                      std::shared_ptr<mfem::VectorCoefficient> trac_bdr_coef, const int component)
```
jamiebramwell, line 40: If we're passing POD by value, is it still a good idea to decorate with `const`?

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -111,7 +111,7 @@ void NonlinearSolidSolver::setSolverParameters(const serac::LinearSolverParamete
 void NonlinearSolidSolver::completeSetup()
 {
   // Define the nonlinear form
-  H_form_ = std::make_shared<mfem::ParNonlinearForm>(displacement_->space.get());
+  auto H_form_ = std::make_unique<mfem::ParNonlinearForm>(displacement_->space.get());
```
jamiebramwell: Since `H_form_`, `M_form_`, and `S_form_` aren't members anymore, we should drop the trailing underscore.

tests/serac_component_bc.cpp
```diff
@@ -55,8 +55,7 @@ TEST(component_bc, qs_solve)
       }
   });
 
-  mfem::Array<int> ess_corner_bc_list;
-  makeTrueEssList(*solid_solver.getDisplacement()->space, zero_bc, ess_corner_bc_list);
+  mfem::Array<int> ess_corner_bc_list = makeTrueEssList(*solid_solver.getDisplacement()->space, zero_bc);
```
jamiebramwell, line 58: This is much nicer!

tests/serac_stdfunction_coefficient.cpp
```diff
@@ -213,8 +208,7 @@ TEST_F(StdFunctionCoefficientTest, EssentialBCCube)
       X[2] = 1.;
     }
   });
-  Array<int>                   ess_bottom_bc_list;
-  makeEssList(*pfes_v_, bottom_bc_z, ess_bottom_bc_list);
+  Array<int>                   ess_bottom_bc_list = serac::makeEssList(*pfes_v_, bottom_bc_z);
```
jamiebramwell: If you add an extra line, this weird spacing probably goes away.

jamiebramwell (commented):
This is seriously great. Thank you for taking the time to do this! I'm going to go over my old comments and make a list of todo items.

src/solvers/elasticity_solver.cpp
joshessman-llnl: I'll link this in the new PR.

The assertion itself doesn't need to be there, but if it ends up being `False`, the call to `std::get` on the next line will throw an exception.  I would argue that it's still preferable to the current approach with two variables, because if `vec_coef` remained uninitialized, you'd end up trying to dereference a nullptr.

src/solvers/nonlinear_solid_solver.cpp
joshessman-llnl: Good catch, I had left them that way to just test if they could work as locals.

jamiebramwell:
Here are my TODO items from the review:

- [x] fix the elasticity solver variable and function names to conform to the style guide
- [x] Add const to the component parameter in line 35 of elasticity_solver.cpp
- [x] Add newline in tests tests/serac_stdfunction_coefficient.cpp

src/solvers/elasticity_solver.cpp
```diff
@@ -42,14 +31,14 @@ ElasticitySolver::ElasticitySolver(int order, std::shared_ptr<mfem::ParMesh> pme
   displacement_->name = "displacement";
 }
 
-void ElasticitySolver::setDisplacementBCs(std::set<int>&                           disp_bdr,
+void ElasticitySolver::setDisplacementBCs(const std::set<int>&                     disp_bdr,
                                           std::shared_ptr<mfem::VectorCoefficient> disp_bdr_coef, int component)
```
jamiebramwell: This component should be `const`

src/solvers/elasticity_solver.cpp
```diff
@@ -79,68 +68,64 @@ void ElasticitySolver::completeSetup()
...
   }
 
   // Assemble the stiffness matrix
-  K_mat_ = K_form_->ParallelAssemble();
+  K_mat_ = std::unique_ptr<mfem::HypreParMatrix>(K_form_->ParallelAssemble());
 
   // Eliminate the essential DOFs
   for (auto& bc : ess_bdr_) {
-    K_e_mat_ = K_mat_->EliminateRowsCols(bc->true_dofs);
+    K_e_mat_.reset(K_mat_->EliminateRowsCols(bc->true_dofs));
```
white238, line 89: This continues to strike me as a very odd thing.  Looping over an object and destroying/assigning to the same object while keeping the last one to get there.  @jamiebramwell  assures me its fine but she may be messing with me.

white238: if its not needed why is it a member variable.

src/solvers/elasticity_solver.cpp
```diff
@@ -79,68 +68,64 @@ void ElasticitySolver::completeSetup()
...
-    K_minres->SetMaxIter(lin_params_.max_iter);
-    K_minres->SetPrintLevel(lin_params_.print_level);
-    K_minres->SetPreconditioner(*K_prec_);
-    K_solver_ = K_minres;
+    K_prec_ = std::move(K_hypreSmoother);
+
+    iter_solver = std::make_unique<mfem::MINRESSolver>(displacement_->space->GetComm());
   }
+
+  iter_solver->SetRelTol(lin_params_.rel_tol);
```
white238, line 122: 👍

src/solvers/elasticity_solver.cpp
jamiebramwell: These are the eliminated entries of the stiffness matrix. Since we don't currently support changing Dirichlet boundary conditions, it's OK to throw them away. However, we will need to keep these around (like in the `QuasiStaticSolve` function) if we want to handle changing BCs.

white238 (approved):
Approved pending @jamiebramwell 's list.

src/solvers/elasticity_solver.cpp
joshessman-llnl: The [isocpp section on this](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-const) indicates that it's not necessary (and pedantic!), but I like to do it because I've seen bugs before where a function was changing an argument when it was supposed to be changing a function-local variable.

jamiebramwell (approved):

joshessman-llnl:
LGTM

PR #108, opened 2020-07-29
Switch to C++17 and use std::variant for managing MFEM coefficients

### LINKED ISSUE #99: Upgrade to C++17

### LINKED ISSUE #121: Delete the intel compiler option
Intel has bugs with C++17 and I like gcc and clang better anyway.

joshessman-llnl:
This may need to wait on [BLT](https://github.com/LLNL/blt/issues/360).

This is an extension of #105.  There was some discussion there about the call to `std::visit` and whether its complexity is unacceptable.

Closes #99 
Closes #121


joshessman-llnl:
https://github.com/LLNL/serac/pull/105#discussion_r461951364

src/solvers/nonlinear_solid_operators.cpp
```diff
@@ -18,7 +18,7 @@ NonlinearSolidQuasiStaticOperator::NonlinearSolidQuasiStaticOperator(std::shared
 // compute: y = H(x,p)
 void NonlinearSolidQuasiStaticOperator::Mult(const mfem::Vector& k, mfem::Vector& y) const
 {
-  // Apply the nonlinear form
+  // Apply the nonlinear form H_form_->Mult(k, y);
```
white238, line 21: This is a bit redundant with the next line.

white238:
I was hesitant at first, but this cleans up a lot of duplicated code.  With comments on complicated sections (visitor and holds_alternative) and only using this feature when it is really needed, I feel we should move ahead with it.

src/common/serac_types.hpp
```diff
@@ -86,12 +87,12 @@ struct FiniteElementState {
 
 // Boundary condition information
 struct BoundaryCondition {
-  mfem::Array<int>                         markers;
-  mfem::Array<int>                         true_dofs;
-  int                                      component;
-  std::shared_ptr<mfem::Coefficient>       scalar_coef;
-  std::shared_ptr<mfem::VectorCoefficient> vec_coef;
-  std::unique_ptr<mfem::HypreParMatrix>    eliminated_matrix_entries;
+  using Coef = std::variant<std::shared_ptr<mfem::Coefficient>, std::shared_ptr<mfem::VectorCoefficient>>;
```
jamiebramwell, line 90: Naming the type `GeneralCoefficient` or something like that may be more descriptive. It does require more typing though...

src/solvers/base_solver.cpp
```diff
@@ -35,143 +35,76 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
   gf_initialized_.assign(n, false);
 }
 
-void BaseSolver::setEssentialBCs(const std::set<int>&                     ess_bdr,
-                                 std::shared_ptr<mfem::VectorCoefficient> ess_bdr_vec_coef,
+void BaseSolver::setEssentialBCs(const std::set<int>& ess_bdr, serac::BoundaryCondition::Coef ess_bdr_coef,
```
jamiebramwell, line 37: I really like removing all the duplication. This was a major source of errors when I was initially implementing this.

src/solvers/base_solver.cpp
```diff
@@ -35,143 +35,76 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
                                  mfem::ParFiniteElementSpace& fes, int component)
 {
-  auto bc = std::make_shared<serac::BoundaryCondition>();
+  serac::BoundaryCondition bc;
```
jamiebramwell, line 40: I will slowly start to learn to use value and move semantics!

src/solvers/base_solver.cpp
```diff
@@ -35,143 +35,76 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
       }
     }
   }
 
-  bc->scalar_coef = ess_bdr_coef;
-  bc->component   = component;
+  bc.component = component;
 
-  fes.GetEssentialTrueDofs(bc->markers, bc->true_dofs, component);
+  // This function can and should be marked const in MFEM
```
jamiebramwell, line 57: I think I said this in the other PR, but the MFEM team is fairly responsive to PR and has a good contributing guide in their repo.

src/solvers/base_solver.cpp
```diff
@@ -35,143 +35,76 @@ BaseSolver::BaseSolver(MPI_Comm comm, int n, int p) : BaseSolver(comm)
...
-
-void BaseSolver::setState(const std::vector<std::shared_ptr<mfem::VectorCoefficient> >& state_vec_coef)
-{
-  SLIC_ASSERT_MSG(state_vec_coef.size() == state_.size(), "State and coefficient bundles not the same size.");
-
-  for (unsigned int i = 0; i < state_vec_coef.size(); ++i) {
-    state_[i]->gf->ProjectCoefficient(*state_vec_coef[i]);
+    // The generic lambda parameter, auto&&, allows the component type (mfem::Coef or mfem::VecCoef)
+    // to be deduced, and the appropriate version of ProjectCoefficient is dispatched.
+    std::visit([this, i](auto&& coef) { state_[i]->gf->ProjectCoefficient(*coef); }, state_coef[i]);
```
jamiebramwell, line 105: I think now that I've had some time to let `std::visit` percolate in my brain, this is more readable.

src/solvers/nonlinear_solid_operators.cpp
```diff
@@ -31,11 +31,12 @@ mfem::Operator& NonlinearSolidQuasiStaticOperator::GetGradient(const mfem::Vecto
 // destructor
 NonlinearSolidQuasiStaticOperator::~NonlinearSolidQuasiStaticOperator() {}
 
-NonlinearSolidDynamicOperator::NonlinearSolidDynamicOperator(
-    std::shared_ptr<mfem::ParNonlinearForm> H_form, std::shared_ptr<mfem::ParBilinearForm> S_form,
-    std::shared_ptr<mfem::ParBilinearForm>                         M_form,
-    const std::vector<std::shared_ptr<serac::BoundaryCondition> >& ess_bdr, mfem::NewtonSolver& newton_solver,
-    const serac::LinearSolverParameters& lin_params)
+NonlinearSolidDynamicOperator::NonlinearSolidDynamicOperator(std::shared_ptr<mfem::ParNonlinearForm>      H_form,
```
jamiebramwell: I still don't know why the style was such garbage on this before...

src/solvers/nonlinear_solid_operators.cpp
```diff
@@ -159,14 +159,14 @@ mfem::Operator& NonlinearSolidReducedSystemOperator::GetGradient(const mfem::Vec
...
-    auto Je = std::unique_ptr<mfem::HypreParMatrix>(jacobian_->EliminateRowsCols(bc->true_dofs));
+    auto Je = std::unique_ptr<mfem::HypreParMatrix>(jacobian_->EliminateRowsCols(bc.true_dofs));
   }
   return *jacobian_;
 }
 
 NonlinearSolidReducedSystemOperator::~NonlinearSolidReducedSystemOperator() {}
 
-}  // namespace serac
+}  // namespace serac
```
jamiebramwell: What is the final verdict on EOF newlines? Github yells about it, but I'm not sure if it's really worth worrying about.

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -227,7 +242,7 @@ void NonlinearSolidSolver::completeSetup()
 }
 
 // Solve the Quasi-static Newton system
-void NonlinearSolidSolver::QuasiStaticSolve()
+void NonlinearSolidSolver::quasiStaticSolve()
```
jamiebramwell, line 248: I missed one! Thanks for catching this!

src/solvers/thermal_operators.cpp
```diff
@@ -51,8 +51,8 @@ DynamicConductionOperator::DynamicConductionOperator(
 void DynamicConductionOperator::setMatrices(std::shared_ptr<mfem::HypreParMatrix> M_mat,
                                             std::shared_ptr<mfem::HypreParMatrix> K_mat)
 {
-  M_mat_  = M_mat;
-  m_K_mat = K_mat;
+  M_mat_ = M_mat;
+  K_mat_ = K_mat;
```
jamiebramwell, line 55: I missed another one! shame!

src/solvers/thermal_operators.cpp
```diff
@@ -106,17 +106,17 @@ void DynamicConductionOperator::ImplicitSolve(const double dt, const mfem::Vecto
   x_       = 0.0;
 
   for (auto& bc : ess_bdr_) {
-    if (bc->scalar_coef != nullptr) {
-      bc->scalar_coef->SetTime(t);
+    if (std::holds_alternative<std::shared_ptr<mfem::Coefficient>>(bc.coef)) {
```
jamiebramwell: Should this be an `ASSERT` as well?

jamiebramwell (approved):
Now that my brain has groked `std::visit`, I fully support this. My one condition is to ensure that this makes sense with @jonwong12 's strategy for LiDO.

src/solvers/thermal_operators.cpp
joshessman-llnl: The "old" version looked like it was just ignoring `VectorCoefficient`s instead of crashing with an assertion failure, but if it's an invalid state (which it looks like it is because currently only `Coefficient`s  can be temperature boundary conditions), I can add an assertion.

src/solvers/nonlinear_solid_operators.cpp
joshessman-llnl: Linux doesn't consider something a line unless there's a newline at the end, so it's probably advisable.  We should enforce it with CI if `make style` can't fix it.

src/solvers/base_solver.cpp
joshessman-llnl: It's a fairly minor change, so I was planning on including it in my MFEM branch I'm currently using to move their sidre stuff to Axom.

src/solvers/thermal_operators.cpp
jamiebramwell: Yep, that was a bug. This should also be an assert.

src/common/serac_types.hpp
jonwong12: @jamiebramwell So in the past I think the idea was that LiDO would pass serac coefficients for the different primal and adjoint solves and serac would return a gridfunction. At the least having a `std::variant` as input to serac, could(?) possibly make things simpler. It seems I should be able to do something like `Coef c=ElasticiityTensor` and `Coef=LoadVector` and serac will know what I'm talking about. 

Do you know offhand if there's something that might not fit? I haven't thought too deeply about this.

jonwong12 (approved):
Thanks. Looks nice!

src/common/serac_types.hpp
jamiebramwell: I need to think about this a bit more, but I think it should work!

joshessman-llnl:
LGTM

jamiebramwell:
Looks like the intel compiler on LC didn't like it. 😢

joshessman-llnl:
There is a confirmed [compiler bug](https://community.intel.com/t5/Intel-C-Compiler/Converting-constructor-issue-with-std-variant/td-p/1182651?profile.language=en) in ICC that prevents the code from compiling.

joshessman-llnl:
Closes #121

joshessman-llnl:
LGTM

PR #114, opened 2020-08-03
Make a mesh builder helper function

### LINKED ISSUE #111: Make helper function for loading a mesh
I've noticed a lot of boilerplate code in the tests for opening a mesh from a file. This should be extracted and put in a utility helper function. See [this code](https://github.com/LLNL/serac/blob/9d34fbdbf8e237ddc2b3929a360e071aa15a20e6/tests/serac_thermal_structural_solver.cpp#L23) for an example.

jamiebramwell:
The consolidates a bunch of boilerplate code for reading a mesh from a file and performing uniform refinements. It addresses https://github.com/LLNL/serac/issues/111 .

### DIFF
--- a/src/common/mesh_utils.cpp
+++ b/src/common/mesh_utils.cpp
@@ -11,8 +11,8 @@
 
 namespace serac {
 
-std::shared_ptr<mfem::ParMesh> buildParallelMesh(const std::string& mesh_file, int refine_serial, int refine_parallel,
-                                                 MPI_Comm comm)
+std::shared_ptr<mfem::ParMesh> buildParallelMesh(const std::string& mesh_file, const int refine_serial,
+                                                 const int refine_parallel, const MPI_Comm comm)
 {
   // Get the MPI rank for logging purposes
   int rank = 0;
--- a/src/common/mesh_utils.cpp
+++ b/src/common/mesh_utils.cpp
@@ -1,0 +1,50 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "mesh_utils.hpp"
+
+#include "common/logger.hpp"
+#include "fmt/fmt.hpp"
+
+namespace serac {
+
+std::shared_ptr<mfem::ParMesh> buildParallelMesh(const std::string& mesh_file, int refine_serial, int refine_parallel,
+                                                 MPI_Comm comm)
+{
+  // Get the MPI rank for logging purposes
+  int rank = 0;
+  MPI_Comm_rank(comm, &rank);
+
+  // Open the mesh
+  std::string msg = fmt::format("Opening mesh file: {0}", mesh_file);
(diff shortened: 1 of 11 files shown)


src/common/mesh_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+#include "mesh_utils.hpp"
+
+#include "common/logger.hpp"
+#include "fmt/fmt.hpp"
+
+namespace serac {
+
+std::shared_ptr<mfem::ParMesh> buildParallelMesh(const std::string& mesh_file, const int refine_serial,
+                                                 const int refine_parallel, const MPI_Comm comm)
```
samuelpmishLLNL, line 15: maybe have defaults for the last 3 arguments

src/common/mesh_utils.cpp
```diff
@@ -0,0 +1,50 @@
...
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "mesh_utils.hpp"
+
+#include "common/logger.hpp"
+#include "fmt/fmt.hpp"
+
+namespace serac {
+
+std::shared_ptr<mfem::ParMesh> buildParallelMesh(const std::string& mesh_file, const int refine_serial,
```
joshessman-llnl, line 14: This belongs in a separate PR, but I think we can use a `unique_ptr<mfem::ParMesh>` if not just an `mfem::ParMesh` here and in the `FiniteElementState` struct.  Either way this function will make that refactoring a lot easier. :+1:

jamiebramwell:
LGTM

samuelpmishLLNL (approved):

src/drivers/serac.cpp
```diff
@@ -123,30 +124,7 @@ int main(int argc, char* argv[])
...
-  for (int lev = 0; lev < ser_ref_levels; lev++) {
-    mesh->UniformRefinement();
-  }
-
-  // create the parallel mesh
-  auto pmesh = std::make_shared<mfem::ParMesh>(MPI_COMM_WORLD, *mesh);
-  for (int lev = 0; lev < par_ref_levels; lev++) {
-    pmesh->UniformRefinement();
-  }
+  auto pmesh = serac::buildParallelMesh(mesh_file, ser_ref_levels, par_ref_levels);
```
joshessman-llnl, line 127: It's always fun to delete a bunch of code, since the only code without bugs is [no code](https://github.com/kelseyhightower/nocode).

joshessman-llnl (approved):

src/common/mesh_utils.cpp
jamiebramwell: I've thought a bit about this, and the problem is that each `FiniteElementState` needs access to the mesh. Certain physics algorithms actually change the mesh (via AMR or element erosion), so a `const Parmesh&` is not really valid. It becomes a question of who would "own" the mesh. I've considered putting it in the "vault-type" problem level object which owns a unique pointer, but that is definitely a separate PR.

jamiebramwell:
LGTM

jamiebramwell:
LGTM

white238 (approved):
Approved anyways!

PR #120, opened 2020-08-05
Guard CMake strings that could be empty

white238:
It's not safe to use strings in CMake without quoting them.  It causes syntax errors if they are empty because of the order they get expanded.

set(someemptystring)
string(TOUPPER ${someemptystring} outvar)

this turns into:

string(TOUPPER  outvar) //syntax error that string(TOUPPER takes two input variables


white238:
LGTM

jamiebramwell (approved):
This is one of those PRs where I'm afraid to ask how you found it out...

joshessman-llnl (approved):

white238:
> This is one of those PRs where I'm afraid to ask how you found it out...

Some Axom build work went wrong.  Copied this over there and one of Axom's spec failed due to one of them being empty.  I've hit this error a lot so i know whats wrong now.

PR #122, opened 2020-08-06
Add conflict with intel in spack package with explanation

adrienbernede:
Relates to #108 and #121 

Since Serac can’t be compiled with Intel, it’s worth mentioning it in the spack package.

I would have been further and kept the tests in CI, but simply allowing them to fail. That’s because we should be able to compile with intel again someday, and I like the idea of covering the major compilers in CI. But I’m probably being picky :).

### DIFF
--- a/.gitlab/ci/build_quartz.yml
+++ b/.gitlab/ci/build_quartz.yml
@@ -57,15 +57,16 @@ gcc_8_1_0 (PR build_on_quartz):
     HOST_CONFIG: "quartz-toss_3_x86_64_ib-gcc@8.1.0.cmake"
   extends: .pr_build_on_quartz
 
-# Intel C++ compiler has poor C++17 support and will no longer be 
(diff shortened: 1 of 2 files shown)


joshessman-llnl:
I think leaving them in is a good idea - not sure if it belongs in this PR, but we'd just have to uncomment the Intel jobs and add 
`allow_failure: true`

adrienbernede:
Exactly.
I was sounding to see if it would be OK :)

adrienbernede:
LGTM

joshessman-llnl (approved):

adrienbernede:
In the PR workflow the intel build still happens, because spack is not used there.
In the master workflow, the job is interrupted at spack level, which I quite like (https://lc.llnl.gov/gitlab/smith/serac/pipelines/14964).

jamiebramwell (approved):
Thanks for catching this!

PR #123, opened 2020-08-06
Deleting some leftover unused shared_ptrs

jamiebramwell:
I believe these `shared_ptr`s are leftovers from the refactoring work.

### DIFF
--- a/src/solvers/nonlinear_solid_solver.cpp
+++ b/src/solvers/nonlinear_solid_solver.cpp
@@ -173,12 +173,6 @@ void NonlinearSolidSolver::completeSetup()
   // The abstract viscosity bilinear form
   std::unique_ptr<mfem::ParBilinearForm> S_form;
 
-  // The abstract mass bilinear form
-  std::shared_ptr<mfem::ParBilinearForm> M_form_;
-
-  // The abstract viscosity bilinear form
-  std::shared_ptr<mfem::ParBilinearForm> S_form_;
-
   // If dynamic, create the mass and viscosity forms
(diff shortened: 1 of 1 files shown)


jamiebramwell:
LGTM

joshessman-llnl (approved):
Good catch, I wonder if any static analysis tools would have caught something like this

jamiebramwell:
> 
> 
> Good catch, I wonder if any static analysis tools would have caught something like this

I only noticed while resolving conflicts on my contact branch! It would be nice to have a static analysis thing look for stuff like this.

jamiebramwell:
LGTM

jamiebramwell:
@samuelpmishLLNL , could you take a look at this? Thanks!

samuelpmishLLNL (approved):

jamiebramwell:
LGTM

samuelpmishLLNL:
> 
> 
> Good catch, I wonder if any static analysis tools would have caught something like this

interesting that compilers will warn on unused basic types, but not on `std::shared_ptr`: https://godbolt.org/z/MW9xzn

jamiebramwell:
> 
> 
> > Good catch, I wonder if any static analysis tools would have caught something like this
> 
> interesting that compilers will warn on unused basic types, but not on `std::shared_ptr`: https://godbolt.org/z/MW9xzn

It might be due to the fact that uninitialized pointers (i.e. `nullptr`s) are a common design requirement.

samuelpmishLLNL:
Uninitialized is different than never used, though! I guess any type with a default constructor might actually be doing something in the background, so "unused" doesn't necessarily imply "does nothing". TIL

PR #124, opened 2020-08-07
Removing the unnecessary gets

jamiebramwell:
We really don't need the extra `get`s in the function calls that I originally added.


jamiebramwell:
LGTM

white238 (approved):

joshessman-llnl (approved):

PR #125, opened 2020-08-07
Add a signal handler

### LINKED ISSUE #100: Create a signal handler
Using [SCR](https://github.com/LLNL/scr) or [Sidre](https://github.com/LLNL/axom/tree/develop/src/axom/sidre) or somethine else, we should be able to handle system signals gracefully.

### LINKED ISSUE #133: Move non-logger functionality out of Logger.hpp
Done for expediency/ease of use, we overloaded the common/logger source and header files. We should move the non-logger functions to a more common place and/or move them to a better named file then create a common header file that includes all the specific headers (common/common.hpp?).

joshessman-llnl:
Registers handlers for common POSIX signals (segfaults, Ctrl+C, etc).

Once SCR is integrated we can also add a checkpoint or whatever else we need to the 'SignalHandler` function.

Resolves #100.


src/common/logger.cpp
```diff
@@ -6,11 +6,20 @@
 
 #include "common/logger.hpp"
 
+#include <csignal>
 #include <cstdlib>
 
 #include "axom/slic.hpp"
 #include "mpi.h"
 
+namespace {
```
samuelpmishLLNL: I'm unfamiliar with this use of an unnamed namespace. What is it doing?

samuelpmishLLNL (approved):

src/common/logger.cpp
joshessman-llnl: It makes sure that it isn't available for external linkage - I use it for little utility functions, mostly.  It's equivalent to declaring the function `static`.

The Google C++ style guide has a [little blurb on it](https://google.github.io/styleguide/cppguide.html#Unnamed_Namespaces_and_Static_Variables) as well.

joshessman-llnl: If the `static` keyword is more readable I'll switch it

samuelpmishLLNL: I have no objection, just hadn't seen that before. Thanks for the explanation!

src/common/logger.cpp
```diff
@@ -6,11 +6,20 @@
...
+void signalHandler(int signal)
+{
+  // Will clean up MPI via call to axom::Utilities::processAbort()
+  SLIC_ERROR("Received signal " << signal << ", exiting");
```
white238: I explicitly turned off SLIC's processAbort logic here:

https://github.com/LLNL/serac/blob/cb8c5295d72acb877029bc6c14a5342c33782be0/src/common/logger.cpp#L79

There has been a lot of complaints about how SLIC handles the aborting on parallel codes.  Did this exit?

src/common/logger.cpp
```diff
@@ -13,10 +13,42 @@
...
-  // Will clean up MPI via call to axom::Utilities::processAbort()
-  SLIC_ERROR("Received signal " << signal << ", exiting");
+  int mpi_finalized;
+  MPI_Finalized(&mpi_finalized);
+  if (!mpi_finalized) {
+    MPI_Finalize();
+  }
+  // This works, but is it UB to call MPI_Comm_rank here?
+  int rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
```
white238: I can't imagine calling MPI_Comm_rank after MPI_Finalized is safe or has expected behavior.

src/common/logger.cpp
joshessman-llnl: It didn't - I had misinterpreted the output.

src/common/logger.cpp
joshessman-llnl: Yeah, I just found the relevant part of the standard/manual.  I think the fact that MPI is dead by the time the signal handler is called means that having a signal handler in the first place isn't particularly useful.  SCR won't be of any help without MPI so I think all a "stranded" process could do is dump its local data and hope that it can be stitched together on restart.

white238:
Is there a reason you aren't just calling exitGracefully?

jamiebramwell:
Here are some comments from Rob Neely about this from the LLNL slack:

Rob Neely  4 hours ago
That’s actually a good idea for an Axom component. The problem with non-parallel-aware solutions like backtrace() is that it can cause your batch job to hang if one process calls it and the others move on to the next MPI_Barrier().  You really need a way to do the backtrace and then call MPI_Abort(), no?

Rob Neely  4 hours ago
I asked Rich if he knew - He thinks most codes have developed their own thing, but he’ll bring it up at the next Axom meeting. I offered to him that if there’s a good solution from an existing code that was lightweight and met all the requirements, we could fund some RADIUSS effort to refactor it into an Axom component.

joshessman-llnl:
> Is there a reason you aren't just calling exitGracefully?

The two things `exitGracefully` does before calling `exit` (clean up SLIC and clean up MPI) aren't necessary/have already happened by the time `mpirun` propagates the signal, so they'd have to be placed behind guards.  This is probably a good thing anyways, especially because MPI sighandling behavior is implementation-dependent.

src/common/logger.cpp
```diff
@@ -6,18 +6,55 @@
...
+ * Registers the signalHandler function to handle various fatal
+ * signals
+ * @note The behavior of MPI when a signal is sent to the mpirun
+ * process is implementation-defined.  OpenMPI will send a SIGTERM
+ * (which can be caught) and then a SIGKILL (which cannot) a few
+ * seconds later.  PlatformMPI will propagate the signal to the
+ * individual processes.  Before propagating or sending a signal,
+ * MPI may or may not call MPI_Finalize/MPI_Abort.
+ */
+void registerSignals()
```
white238: There has been a desire to move this registerSignals logic into Axom's core library and add windows support or at least guard against it.

This goes with the need to handle SLIC's aborting for a more complete error handling functionality.

joshessman-llnl: Ok, we can put this branch on hold until equivalent functionality is provided by Axom.  I can move what I've got now onto an Axom branch.

> This goes with the need to handle SLIC's aborting for a more complete error handling functionality.

Just to confirm, does this refer to addressing the issues with `processAbort` and MPI in addition to https://github.com/LLNL/axom/issues/27 and https://github.com/LLNL/axom/issues/321?

white238: More of a move forward and when we think we have a hardened/working function we can move the logic into Axom to share with others.

As for the other, yea there is talk on the Axom team about reworking/improving how SLIC handles the aborting of parallel programs.  Not specifically that you fix the entire world.

white238 (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

white238 (approved):

PR #126, opened 2020-08-07
Fix mfem cmake tpl search

jandrej:
This change fixes the case when a user builds mfem with CMake and uses the generated CMake configuration files to find mfem as a TPL.

### DIFF
--- a/cmake/thirdparty/FindMFEM.cmake
+++ b/cmake/thirdparty/FindMFEM.cmake
@@ -27,8 +27,8 @@ if(EXISTS ${_mfem_cmake_config})
     include(${_mfem_cmake_config})
 
     # TODO: This needs to be verified
-    set(MFEM_INCLUDE_DIRS  ${MFEM_INCLUDE_DIR} )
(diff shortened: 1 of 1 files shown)


cmake/thirdparty/FindMFEM.cmake
```diff
@@ -27,8 +27,8 @@ if(EXISTS ${_mfem_cmake_config})
     include(${_mfem_cmake_config})
 
     # TODO: This needs to be verified
```
white238: If you verified this works could you remove my long ignored TODO =)

jandrej: This was an oversight. I remove it.

white238 (approved):

joshessman-llnl (approved):

jamiebramwell:
LGTM

jamiebramwell:
@adrienbernede , It looks like the sign off isn't working for a PR from a forked repo. Any thoughts on how to fix it?

jamiebramwell:
It unfortunately looks like we can't test PRs from forks. Sorry for the bad news @jandrej . 😢

PR #127, opened 2020-08-10
Fix mfem cmake tpl search

jandrej:
This change fixes the case when a user builds mfem with CMake and uses the generated CMake configuration files to find mfem as a TPL.


jamiebramwell (approved):

jamiebramwell:
LGTM

jamiebramwell:
This is a copy of the previously approved PR (now in branch form).

joshessman-llnl (approved):

Issue #118, opened 2020-08-04
Consider refactoring to comply with ISOCPP core guidelines


joshessman-llnl:
There are couple maintainability improvements that have caught my eye and could be refactored - I've linked relevant sections of the ISOCPP guidelines:

1. The [two-phase initialization](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rnr-two-phase-init) for the solvers (constructor + `completeSetup`) - instead, a combination of reasonable defaults and constructor arguments (e.g. a class representing/encapsulating BCs) could be used to ensure that the solvers are always in a valid (solvable) state.
2. Switch to [`class`es and constructors](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-struct) for `BoundaryCondition` and `FiniteElementState` types - these have invariants, e.g. the FES should be constructed with the mesh, the GF should be constructed with the FES, etc
3. Separate [interface/behavior from state/implementation](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Ri-abstract) in `BaseSolver` - potentially there could be a separate helper class responsible for any common helper methods, in which case `BaseSolver` only determines the functions that a solver must implement (e.g. `advanceTimestep`)
4. There's still work to be done with regard to [function/constructor parameters and ownership](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-functions) as well.

Some of these improvements may be mutually inclusive with a switch to a boundary condition manager.

samuelpmishLLNL:
I agree with 1. We have no real way to enforce that the user has properly initialized a solver before calling `completeSetup()`. It would be much nicer to be able to tell a user at compile time that they have not properly initialized a solver object, but I'm not sure which approach to take to achieve this.

I'm not sure exactly what you mean with 4., but I think our plan for function parameters was to use raw pointers for non-owning refererences,  `std::shared_ptr` for references to objects the user might wish to change/refer to after calling `completeSetup()` and `std::unique_ptr` when passing ownership.

joshessman-llnl:
With regard to 4, I just meant that I think there are deviations from the ownership rules you mentioned.

joshessman-llnl:
>  It would be much nicer to be able to tell a user at compile time that they have not properly initialized a solver object

[...]

PR #128, opened 2020-08-10
Switch BoundaryCondition from struct to class to reflect invariants

joshessman-llnl:
This PR encapsulates the internal state of a boundary condition and adds various utility functions to "hide" access to this internal state behind functions for common uses of boundary condition objects.  Also, for Dirichlet BCs, the a reference to the corresponding field is stored for projection later.  Essential-BC-specific functionality is protected behind SLIC conditionals to ensure that a natural BC is not accidentally treated as an essential BC.

Now boundary conditions can be registered in the form 
`setEssentialBCs(ess_bdr, ess_bdr_coef, *temperature_);` indicating the coefficient `ess_bdr_coef` should be applied to nodes in the `ess_bdr` set on the `temperature_` field.

This allows for a solver class to iterate over the BCs and call `bc.projectBdr()` without needing to pass it any arguments, meaning that essential boundary conditions can constrain multiple variables.  Solvers using this approach cannot have boundary conditions whose true DOFs are provided explicitly, as these BCs are not associated with a field.  Potentially this `setTrueDofs` functionality needs to reconsidered and/or modified - e.g., individual solvers could wrap the BaseSolver version to specify a field.

There are two more classes coming of about equivalent size - I'm not sure if these classes (incl this one) need their own header files.  They are all related so my thinking would be that their protos stay in `serac_types.hpp` and any implementation goes in its own source file.

Partially addresses #118


src/common/serac_types.hpp
```diff
@@ -174,35 +178,178 @@ struct FiniteElementState {
...
   using Coef = std::variant<std::shared_ptr<mfem::Coefficient>, std::shared_ptr<mfem::VectorCoefficient>>;
 
   /**
-   * @brief The attribute marker array where this BC is active
+   * Constructor for setting up a boundary condition using a set of attributes
+   * @param[in] coef Either a mfem::Coefficient or mfem::VectorCoefficient representing the BC
+   * @param[in] component The zero-indexed vector component if the BC applies to just one component,
+   * should be -1 for all components
+   * @param[in] attrs The set of boundary condition attributes in the mesh that the BC applies to
+   * @param[in] num_attrs The total number of boundary attributes for the mesh
```
jamiebramwell: I hate having to pass `num_attrs` in, but I unfortunately can't think of a way around it.

src/common/serac_types.hpp
```diff
@@ -174,35 +178,178 @@ struct FiniteElementState {
 /**
  * @brief Boundary condition information bundle
  */
-struct BoundaryCondition {
+class BoundaryCondition {
```
jamiebramwell: This class is big enough it might be worth it to put it in its own header file.

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -124,9 +124,8 @@ void NonlinearSolidSolver::completeSetup()
   for (auto& nat_bc_data : nat_bdr_) {
     SLIC_ASSERT_MSG(std::holds_alternative<std::shared_ptr<mfem::VectorCoefficient>>(nat_bc_data.coef),
                     "Traction boundary condition had a non-vector coefficient.");
-    H_form->AddBdrFaceIntegrator(
-        new HyperelasticTractionIntegrator(*std::get<std::shared_ptr<mfem::VectorCoefficient>>(nat_bc_data.coef)),
-        nat_bc_data.markers);
+    H_form->AddBdrFaceIntegrator(nat_bc_data.newVecIntegrator<HyperelasticTractionIntegrator>().release(),
```
jamiebramwell, line 127: While this avoids a raw `new`, I'm not sure these factory methods really help much.

jamiebramwell: I could get used to it though. Anyone else have thoughts on this?

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -137,28 +136,11 @@ void NonlinearSolidSolver::completeSetup()
...
-                      "Displacement boundary condition contained all components but had a non-vector coefficient.");
-      displacement_->gf->ProjectCoefficient(*std::get<std::shared_ptr<mfem::VectorCoefficient>>(bc.coef), dof_list);
-    } else {
-      // If it is only a single component, project the scalar
-      SLIC_ASSERT_MSG(std::holds_alternative<std::shared_ptr<mfem::Coefficient>>(bc.coef),
-                      "Displacement boundary condition contained a single component but had a non-scalar coefficient.");
-      displacement_->gf->ProjectCoefficient(*std::get<std::shared_ptr<mfem::Coefficient>>(bc.coef), dof_list,
-                                            bc.component);
-    }
+    bc.project(*(displacement_->gf), *(displacement_->space));
```
jamiebramwell: I love seeing big chunks of code moved to a function!

src/solvers/base_solver.hpp
```diff
@@ -132,7 +132,7 @@ class BaseSolver {
    * @param[in] component The component to set (-1 implies all components are set)
    */
   virtual void setEssentialBCs(const std::set<int>& ess_bdr, serac::BoundaryCondition::Coef ess_bdr_coef,
-                               const mfem::ParFiniteElementSpace& fes, const int component = -1);
+                               FiniteElementState& state, const int component = -1);
```
jamiebramwell, line 54: I like this change. I think it makes more conceptual sense that you would send a `FiniteElementState` instead of a `FiniteElementSpace` when doing a boundary operation.

src/common/serac_types.hpp
```diff
@@ -174,35 +178,178 @@ struct FiniteElementState {
...
+   * Minimal constructor for setting the true DOFs directly
+   * @param[in] coef Either a mfem::Coefficient or mfem::VectorCoefficient representing the BC
+   * @param[in] component The zero-indexed vector component if the BC applies to just one component,
+   * should be -1 for all components
+   * @param[in] true_dofs The indices of the relevant DOFs
    */
-  mfem::Array<int> true_dofs;
+  BoundaryCondition(Coef coef, const int component, const mfem::Array<int>& true_dofs);
+
+  const mfem::Array<int>& getMarkers() const { return markers_; }
```
jamiebramwell: I have started to remove the `get` label from getter functions for brevity. I don't think it really adds much to label them.

src/common/serac_types.hpp
```diff
@@ -174,35 +178,178 @@ struct FiniteElementState {
...
+  void projectBdr(const double time, const bool should_be_scalar = true) const;
+
+  /**
+   * Allocates an integrator of type "Integrator" on the heap,
+   * constructing it with the boundary condition's vector coefficient,
+   * intended to be passed to mfem::*LinearForm::Add*Integrator
+   * @return An owning pointer to the new integrator
+   * @pre Requires Integrator::Integrator(mfem::VectorCoefficient&)
+   */
+  template <typename Integrator>
```
jamiebramwell: These are the only two methods that I do not think add much value. Anyone else have thoughts on this?

src/common/serac_types.hpp
```diff
@@ -174,35 +178,178 @@ struct FiniteElementState {
...
+
+  /**
+   * Eliminates the rows and columns corresponding to the BC's true DOFS
+   * from a stiffness matrix
+   * @param[inout] k_mat The stiffness matrix to eliminate from,
+   * will be modified.  These eliminated matrix entries can be
+   * used to eliminate an essential BC to an RHS vector with
+   * BoundaryCondition::eliminateToRHS
+   */
+  void eliminateFrom(mfem::HypreParMatrix& k_mat);
```
jamiebramwell: I would name this `eliminateFromMatrix` to be a bit more explicit.

src/common/boundary_condition.cpp
```diff
@@ -0,0 +1,117 @@
...
+  SLIC_ERROR_IF(!true_dofs_, "Only essential boundary conditions can be projected over all DOFs.");
+  // Value semantics for convenience
+  auto tdofs = *true_dofs_;
+  auto size  = tdofs.Size();
+  // FIXME: Why would the be a true dof array of size zero when running with MPI?
+  if (size) {
+    // Generate the scalar dof list from the vector dof list
+    SLIC_ASSERT_MSG(space, "Only BCs associated with a space can be projected.");
+    mfem::Array<int> dof_list(size);
+    std::transform(&tdofs[0], &tdofs[0] + size, &dof_list[0], [&space](int tdof) { return space.VDofToDof(tdof); });
```
jamiebramwell, line 49: I'm not sure we want to use a std algorithm with an MFEM type. I can see them changing it under the hood and this call causing issues.

jamiebramwell (commented):
Looks great! Two minor comments:

1) I'm not sure the integrator factory methods for weakly imposed conditions add much value.
2) I'm a bit leery of using std::algorithm with MFEM-native data structures.

However, I'm not 100% convinced either of these things are really an issue, so I look forward to seeing what others have to say.

src/common/serac_types.hpp
joshessman-llnl: Me too, at least it's hidden behind a function call and not exposed to the user directly.

src/common/serac_types.hpp
joshessman-llnl: Yep, I think all these new common classes I'm creating can go in separate headers and we can use a `common/common.hpp` to include all of them per https://github.com/LLNL/serac/issues/133

src/solvers/nonlinear_solid_solver.cpp
joshessman-llnl: Just to clarify, my thinking here was that wrapping up the construction of an object using part of internal class state (which I also did on my WIP FEState refactor) minimizes exposure of internal members (otherwise you'd need to write an get-function) without obfuscating what the code is doing (adding a new integrator based on that boundary condition).

Alternatively, the call could be structured as `bc.addBdrFaceVecIntegrator<HyperelasticTractionIntegrator>(H_form)` which in my opinion still converys the meaning "add a boundary face integrator of type `HyperelasticTractionIntegrator` to` H_form` based on boundary condition `bc`"

src/common/boundary_condition.cpp
joshessman-llnl: This usage just requires that `mfem::Array` is a contiguous data structure and not anything about internal class structure.  That being said, I'm fairly certain I've seen MFEM code that reinterprets an `mfem::Vector` as a `double*` which does make assumptions about class layout.

src/solvers/nonlinear_solid_solver.cpp
jamiebramwell: I think I like that call a bit better.

src/common/boundary_condition.cpp
jamiebramwell: Yep, I think it would be a fairly large breaking change if they did it, but we should be aware that it is a possibility.

samuelpmishLLNL (approved):

src/solvers/nonlinear_solid_solver.cpp
jamiebramwell: I've been thinking about this, and I think at the end of the day I've settled on the function that you originally developed. It is clear that you are building a new integrator and passing it off to the mfem bilinear form, so I think it wins out.

jamiebramwell (approved):

joshessman-llnl:
LGTM

PR #131, opened 2020-08-11
Wrap system solver into class

joshessman-llnl:
This allows for the duplicated code that applies linear/nonlinear solver params to system solvers to be removed.

### DIFF
--- a/src/common/serac_types.hpp
+++ b/src/common/serac_types.hpp
@@ -14,7 +14,6 @@
 #define SERAC_TYPES
 
 #include <memory>
-#include <optional>
 #include <variant>
 
 #include "mfem.hpp"
--- a/src/common/serac_types.hpp
+++ b/src/common/serac_types.hpp
@@ -4,6 +4,12 @@
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
 
+/**
+ * @file serac_types.hpp
+ *
+ * @brief This file contains common serac data structures
+ */
+
 #ifndef SERAC_TYPES
 #define SERAC_TYPES
 
@@ -14,14 +20,18 @@
 
 namespace serac {
 
-// Option bundling enums
-
+/**
+ * @brief Output file type associated with a solver
+ */
 enum class OutputType
 {
   GLVis,
   VisIt
 };
 
+/**
+ * @brief Timestep method of a solver
+ */
 enum class TimestepMethod
 {
   BackwardEuler,
@@ -37,6 +47,9 @@ enum class TimestepMethod
   QuasiStatic
 };
 
+/**
+ * @brief Linear solution method
+ */
 enum class LinearSolver
 {
   CG,
@@ -44,12 +57,18 @@ enum class LinearSolver
   MINRES
 };
 
+/**
+ * @brief Preconditioning method
+ */
 enum class Preconditioner
 {
   Jacobi,
   BoomerAMG
 };
 
+/**
+ * @brief Abstract multiphysics coupling scheme
+ */
 enum class CouplingScheme
 {
   OperatorSplit,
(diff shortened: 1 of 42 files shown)


src/common/serac_types.hpp
```diff
@@ -203,6 +204,68 @@ struct BoundaryCondition {
...
+
+  /**
+   * Sets a preconditioner for the underlying linear solver object
+   * @param[in] prec The preconditioner, of which the object takes ownership
+   * @note The preconditioner must be moved into the call
+   * @code(.cpp)
+   * solver.setPreconditioner(std::move(prec));
+   * @endcode
+   */
+  void setPreconditioner(std::unique_ptr<mfem::Solver>&& prec)
```
joshessman-llnl: Any thoughts on whether the MFEM naming convention should be used here since the class implements the `mfem::Solver` interface?

samuelpmishLLNL: my vote is to follow mfem's convention, as it is really an mfem thing living in serac.

src/common/serac_types.hpp
```diff
@@ -203,6 +204,68 @@ struct BoundaryCondition {
...
+   */
+  void SetOperator(const mfem::Operator& op) override { solver().SetOperator(op); }
+
+  /**
+   * Solves the system
+   * @param[in] x The system's RHS vector, "b" in Ax = b
+   * @param[out] y The system's solution vector, "x" in Ax = b
+   * @note Implements mfem::Operator::Mult
+   */
+  void Mult(const mfem::Vector& x, mfem::Vector& y) const override { solver().Mult(x, y); }
```
samuelpmishLLNL:
``` cpp
 /**
   * Solves the system A x = b
   * @param[in] b: rhs of the system of equations
   * @param[out] x: solution to the system of equations
   * @note Implements mfem::Operator::Mult
   */
  void Mult(const mfem::Vector& b, mfem::Vector& x) const override { solver().Mult(b, x); }
```

src/common/serac_types.hpp
jamiebramwell: I agree with @samuelpmishLLNL . However, I could go either way.

jamiebramwell:
I just added @jandrej because this would have implications for his flow solver.

jandrej (approved):
I don't see any issues with the concept for the flow solver

src/common/serac_types.hpp
```diff
@@ -203,6 +204,68 @@ struct BoundaryCondition {
   std::unique_ptr<mfem::HypreParMatrix> eliminated_matrix_entries;
 };
 
+/**
+ * Wraps a (currently iterative) system solver and handles the configuration of linear
+ * or nonlinear solvers
+ */
+class SystemSolver : public mfem::Solver {
```
jamiebramwell: `SystemSolver` is a bit overloaded. Maybe `LinearAlgegbraSolver` to denote that these are numerics operators?

jamiebramwell: Although the nonlinear one is here too. Hmmmm.

jamiebramwell: `AlgebraicSolver`?

src/common/serac_types.hpp
```diff
@@ -203,6 +204,68 @@ struct BoundaryCondition {
   std::unique_ptr<mfem::HypreParMatrix> eliminated_matrix_entries;
 };
 
+/**
```
jamiebramwell: This should probably go in it's own file.

src/common/serac_types.hpp
```diff
@@ -203,6 +204,68 @@ struct BoundaryCondition {
...
+   * @note Implements mfem::Operator::Mult
+   */
+  void Mult(const mfem::Vector& b, mfem::Vector& x) const override { solver().Mult(b, x); }
+
+  /**
+   * Returns the underlying solver object
+   * @return The underlying nonlinear solver, if one was configured
+   * when the object was constructed, otherwise, the underlying linear solver
+   */
+  mfem::IterativeSolver&       solver() { return (nonlin_solver_) ? **nonlin_solver_ : *iter_lin_solver_; }
```
jamiebramwell: I would explicitly label these as `linearSolver` and `nonlinearSolver` for clarity. I can see still wanting access to the underlying linear solver even when in a nonlinear setting.

src/common/serac_types.hpp
```diff
@@ -14,6 +14,7 @@
 #define SERAC_TYPES
 
 #include <memory>
+#include <optional>
```
jamiebramwell: Is this still needed?

src/solvers/algebraic_solver.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include <optional>
+#include <variant>
+
+#include "common/serac_types.hpp"
+#include "mfem.hpp"
+
+namespace serac {
+
+/**
+ * Wraps a (currently iterative) system solver and handles the configuration of linear
```
jamiebramwell, line 26: I would add a comment that this solves a generic global system of nonlinear algebraic equations.

src/solvers/algebraic_solver.hpp
```diff
@@ -0,0 +1,97 @@
...
+ * or nonlinear solvers
+ */
+class AlgebraicSolver : public mfem::Solver {
+public:
+  // TODO: Eliminate this once a dependency injection approach is used for the solvers
```
jamiebramwell, line 31: What do you mean by this TODO? What is the proposed dependency injection approach?

joshessman-llnl: That refers to the syntax I proposed [here](https://github.com/LLNL/serac/issues/118#issuecomment-668805621), in which a fully-constructed `AlgebraicSolver` is passed to a solver constructor, i.e.
```cpp
AlgebraicSolver as(MPI_COMM_WORLD, lin_params, nonlin_params);
NonlinearSolidSolver solver(as, /* other params */);
// ready to use
```

instead of 
```cpp
NonlinearSolidSolver solver(/* stuff */);
// invalid state
solver.set_params(lin_params, nonlin_params);
solver.completeSetup();
// now valid and ready to use
```

src/common/serac_types.hpp
joshessman-llnl: No, it's optional

jamiebramwell (approved):

joshessman-llnl:
LGTM

PR #132, opened 2020-08-11
Second memory audit

### LINKED ISSUE #91: Audit the memory management strategy of the solver class

joshessman-llnl:
This is intended to be another step in the general direction of hierarchical ownership, in which objects distribute non-owning handles to the sub-objects they create (e.g. a solver class giving a non-owning reference to an operator it creates).  Primarily this is via the removal of `shared_ptr`s which introduce [incidental data structures](https://sean-parent.stlab.cc/presentations/2016-08-08-data-structures/2016-08-08-data-structures.pdf).

The only remaining `shared_ptr`s in the solver classes are for boundary conditions, meshes, and FEStates, which are currently pending the merging of other branches.

This PR also addresses minor style issues like parameter names differing between function declarations and implementations.

### DIFF
--- a/src/docs/sphinx/style_guide.rst
+++ b/src/docs/sphinx/style_guide.rst
@@ -19,6 +19,16 @@ with the following amendments:
 The Google style guide is meant for style enforcement only. The design principles outlined in the 
 `C++ Core Guidelines <http://isocpp.github.io/CppCoreGuidelines/>`_ should be followed.
 
+Of particular importance are the guidelines proposed for managing object lifetime:
+
+    1. Use raw pointers and references [only] to denote non-ownership (R.3 - R.4)
+    #. Prefer ``unique_ptr`` to ``shared_ptr`` (R.21)
+
+For example, if an object ``A`` creates a subobject ``B`` whose constructor requires a reference
+to one of ``A``'s instance variables ``Foo f``, ``B`` should store a non-owning reference to ``f``,
+either ``Foo&`` or ``Foo*``.  This should be ``const`` if at all possible.  In this case, shared ownership
+is not required because the lifetime of ``B`` is entirely dependent on the lifetime of ``A``.
+
 Documentation
 -------------
 
--- a/src/solvers/nonlinear_solid_operators.hpp
+++ b/src/solvers/nonlinear_solid_operators.hpp
@@ -17,7 +17,7 @@
 
 #include "common/common.hpp"
 #include "mfem.hpp"
-#include "solvers/algebraic_solver.hpp"
+#include "solvers/equation_solver.hpp"
 
 namespace serac {
 
@@ -196,8 +196,8 @@ public:
    * This is the only requirement for high-order SDIRK implicit integration.
    *
    * @param[in] dt The timestep
-   * @param[in] vx The state vector
-   * @param[out] dvx_dt The implicit time derivative
+   * @param[in] vx The state block vector of velocities and displacements
+   * @param[out] dvx_dt The time rate of vx
    */
   virtual void ImplicitSolve(const double dt, const mfem::Vector& vx, mfem::Vector& dvx_dt);
 
@@ -230,7 +230,7 @@ protected:
   /**
    * @brief The CG solver for the mass matrix
    */
-  AlgebraicSolver M_solver_;
+  EquationSolver M_solver_;
 
   /**
    * @brief The reduced system operator for applying the bilinear and nonlinear forms
--- a/src/solvers/nonlinear_solid_operators.hpp
+++ b/src/solvers/nonlinear_solid_operators.hpp
@@ -251,10 +251,10 @@ public:
    * This is the only requirement for high-order SDIRK implicit integration.
    *
    * @param[in] dt The timestep
-   * @param[in] x The state vector
-   * @param[out] k The implicit time derivative
+   * @param[in] vx The state vector
+   * @param[out] dvx_dt The implicit time derivative
    */
-  virtual void ImplicitSolve(const double dt, const mfem::Vector& x, mfem::Vector& k);
+  virtual void ImplicitSolve(const double dt, const mfem::Vector& vx, mfem::Vector& dvx_dt);
 
   /**
    * @brief Destroy the Nonlinear Solid Dynamic Operator object
--- a/src/solvers/nonlinear_solid_operators.hpp
+++ b/src/solvers/nonlinear_solid_operators.hpp
@@ -28,7 +28,7 @@ protected:
   /**
    * @brief The nonlinear form
    */
-  std::shared_ptr<mfem::ParNonlinearForm> H_form_;
+  std::unique_ptr<mfem::ParNonlinearForm> H_form_;
 
   /**
    * @brief The linearized jacobian at the current state
@@ -41,7 +41,7 @@ public:
    *
    * @param[in] H_form The nonlinear form of the PDE
    */
-  explicit NonlinearSolidQuasiStaticOperator(std::shared_ptr<mfem::ParNonlinearForm> H_form);
+  explicit NonlinearSolidQuasiStaticOperator(std::unique_ptr<mfem::ParNonlinearForm> H_form);
 
   /**
    * @brief Get the Gradient of the nonlinear form
--- a/src/solvers/nonlinear_solid_solver.cpp
+++ b/src/solvers/nonlinear_solid_solver.cpp
@@ -172,7 +172,7 @@ void NonlinearSolidSolver::completeSetup()
     S_form->Finalize(0);
   }
 
-  solver_ = AlgebraicSolver(displacement_->space->GetComm(), lin_params_, nonlin_params_);
+  solver_ = EquationSolver(displacement_->space->GetComm(), lin_params_, nonlin_params_);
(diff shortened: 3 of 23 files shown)


src/solvers/thermal_operators.cpp
```diff
@@ -10,19 +10,19 @@
...
-DynamicConductionOperator::DynamicConductionOperator(std::shared_ptr<mfem::ParFiniteElementSpace> fespace,
-                                                     const serac::LinearSolverParameters&         params,
-                                                     std::vector<serac::BoundaryCondition>&       ess_bdr)
-    : mfem::TimeDependentOperator(fespace->GetTrueVSize(), 0.0),
-      fespace_(fespace),
+DynamicConductionOperator::DynamicConductionOperator(mfem::ParFiniteElementSpace&           fespace,
+                                                     const serac::LinearSolverParameters&   params,
+                                                     std::vector<serac::BoundaryCondition>& ess_bdr)
+    : mfem::TimeDependentOperator(fespace.GetTrueVSize(), 0.0),
+      // fespace_(fespace),
```
white238: Please remove commented code.

src/solvers/thermal_operators.hpp
```diff
@@ -76,14 +71,14 @@ class DynamicConductionOperator : public mfem::TimeDependentOperator {
   std::unique_ptr<mfem::HypreParMatrix> T_e_mat_;
 
   /**
-   * @brief Assembled RHS vector
+   * @brief Non-owning ptr to assembled RHS vector
    */
-  std::shared_ptr<mfem::Vector> rhs_;
+  mfem::Vector* rhs_ = nullptr;
```
white238: Changing this from a shared_ptr (which it is) to a unique_ptr+"non-owning ptr" is just abusing/stretching of the definition of unique.  Can you provide a good explanation of the benefit of this.  To me this just obfuscates the life-cycle of this pointer as well as opens us up to using a pointer that has been freed (even if it doesn't at the moment).

white238: I do commend your very explicit non-ownership description.

white238:
This looks good.  I like that it is more explicit on the ownership of each object.  As for my earlier comment, it is fine as long as it is used sparingly and is explicit in comments (which it is).  We should probably write up a small coding guideline on when/where to use each memory pattern for our code specifically.  I think that would help my general concern.

white238 (approved):

white238 (approved):

src/solvers/nonlinear_solid_operators.hpp
```diff
@@ -196,10 +196,10 @@ class NonlinearSolidDynamicOperator : public mfem::TimeDependentOperator {
    * This is the only requirement for high-order SDIRK implicit integration.
    *
    * @param[in] dt The timestep
-   * @param[in] x The state vector
-   * @param[out] k The implicit time derivative
+   * @param[in] vx The state vector
+   * @param[out] dvx_dt The implicit time derivative
```
samuelpmishLLNL:
```cpp
   * @param[in] vx The state block vector of velocities and displacements
   * @param[out] dvx_dt The time rate of vx
```

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -187,13 +190,13 @@ void NonlinearSolidSolver::completeSetup()
...
-    nonlinear_oper_                 = std::make_shared<NonlinearSolidQuasiStaticOperator>(std::move(H_form));
     solver_.solver().iterative_mode = true;
+    nonlinear_oper_                 = std::make_unique<NonlinearSolidQuasiStaticOperator>(std::move(H_form));
     solver_.SetOperator(*nonlinear_oper_);
   } else {
-    timedep_oper_ = std::make_shared<NonlinearSolidDynamicOperator>(
-        std::move(H_form), std::move(S_form), std::move(M_form), ess_bdr_, solver_.solver(), lin_params_);
     solver_.solver().iterative_mode = false;
+    timedep_oper_                   = std::make_unique<NonlinearSolidDynamicOperator>(
+        std::move(H_form), std::move(S_form), std::move(M_form), ess_bdr_, solver_.solver(), lin_params_);
```
samuelpmishLLNL, line 199: Following the recent discussion about satisfying finite-deformation constraints, if we're going to implement our own nonlinear solvers, will we still need these separate operators? @jamiebramwell

src/solvers/thermal_operators.cpp
```diff
@@ -10,46 +10,44 @@
...
-      y_(fespace->GetTrueVSize()),
-      x_(fespace->GetTrueVSize()),
+      z_(fespace.GetTrueVSize()),
+      y_(fespace.GetTrueVSize()),
+      x_(fespace.GetTrueVSize()),
       old_dt_(-1.0)
 {
   // Set the mass solver options (CG and Jacobi for now)
-  M_solver_ = AlgebraicSolver(fespace_->GetComm(), params);
+  M_solver_ = AlgebraicSolver(fespace.GetComm(), params);
```
samuelpmishLLNL: I guess I didn't get my review of the other PR in time, but `AlgebraicSolver` seems like a bad name to me. Algebraic equation has a specific meaning mathematically that I don't think applies here. For example, 
`exp(x) + x == 0` is not an algebraic equation, but is still something that we would expect to be able to solve with that class, I imagine.

src/solvers/thermal_operators.hpp
```diff
@@ -100,14 +94,14 @@ class DynamicConductionOperator : public mfem::TimeDependentOperator {
   AlgebraicSolver T_solver_;
 
   /**
-   * @brief Pointer to the assembled M matrix
+   * @brief Non-owning pointer to the assembled M matrix
```
samuelpmishLLNL, line 97: This is a good convention, so that intellisense can inform people about ownership on autocompletion without needing to dive into documentation.

src/solvers/nonlinear_solid_solver.cpp
jamiebramwell: Do you mean the time dependent operator vs the quasi-static one? I think it's still useful to have these algebraic constructs as separate objects since they are doing distinct algebraic solves (granted they are using the same building blocks to construct these solves). The time dependent operator framework is required to use MFEM timesteppers which is definitely useful. The built-in Newton solver, not so much. Maybe we could add the quasi-static functionality to the `TimeDependentOperator`? I welcome any ideas!

Anyway, I think this is an idea worth pursuing, but it probably goes beyond the scope of this PR.

src/solvers/thermal_operators.cpp
jamiebramwell: Yeah, the original idea was `SystemSolver`, but in my mind that comes with a notion of a full physics solver. The idea is this object is reduced to a simple mathematical form. `AlgebraicSolver` was the best I could come up with unfortunately but I definitely agree with your point. Any other naming thoughts?

jamiebramwell: `ExpressionSolver`? `EquationSolver`?

src/solvers/nonlinear_solid_solver.cpp
samuelpmishLLNL: You're right, it's not intended to really be a request for this PR! I was just wondering how mfem's time integrators interact with the nonlinear solves. It seems like a time integrator built on the implementations of `Mult()` and `ImplicitSolve()` will encounter the same issues with element inversion that Jerome mentioned.

src/solvers/thermal_operators.cpp
samuelpmishLLNL: I'm not sure what this object really does.

samuelpmishLLNL: Maybe EquationSolver, NonlinearSystemSolver, or RootFinder?

joshessman-llnl: It can be either linear or nonlinear (`LinearAlgebraSolver` was considered earlier), but maybe `EquationSolver` or `EquationSystemSolver` could work?

jamiebramwell: I like either `EquationSolver` or `EquationSystemSolver` with a slight preference for the former for brevity.

src/solvers/nonlinear_solid_solver.cpp
jamiebramwell: That's a fair point. I think it's worth investigating if we can put use those linearization fixes in with the abstract time integrator framework because it would be nice if we could use sundials. However, if we can't we can't.

src/docs/sphinx/style_guide.rst
```diff
@@ -19,6 +19,16 @@ with the following amendments:
 The Google style guide is meant for style enforcement only. The design principles outlined in the 
 `C++ Core Guidelines <http://isocpp.github.io/CppCoreGuidelines/>`_ should be followed.
 
+Of particular importance are the guidelines proposed for managing object lifetime:
```
jamiebramwell, line 22: 👍

jamiebramwell (approved):

jamiebramwell:
LGTM

joshessman-llnl:
LGTM

PR #136, opened 2020-08-13
Encapsulate FiniteElementState in class

joshessman-llnl:
This PR moves the FiniteElementState struct into a class.  Direct member accessors are still used in several places but some functionality has been abstracted into methods to minimize direct member exposure.

The `variant` coefficient has been moved out of the `BoundaryCondition` namespace and has been renamed `GeneralCoefficient`.  

Because `BoundaryCondition` depends on `FiniteElementState`, some changes were made to this class, in particular a helper method to apply a boundary condition to a linear system (called `apply`).

Partially addresses #118


src/solvers/elasticity_solver.cpp
```diff
@@ -13,21 +13,12 @@ namespace serac {
 constexpr int NUM_FIELDS = 1;
 
 ElasticitySolver::ElasticitySolver(int order, std::shared_ptr<mfem::ParMesh> pmesh)
-    : BaseSolver(pmesh->GetComm(), NUM_FIELDS, order), displacement_(state_[0])
+    : BaseSolver(pmesh->GetComm(), NUM_FIELDS, order),
+      displacement_(std::make_shared<FiniteElementState>(
+          order, pmesh, FESOptions{.ordering = mfem::Ordering::byVDIM, .name = "displacement"}))
```
samuelpmishLLNL: designated initializers aren't officially supported until C++20, right?

samuelpmishLLNL: well, it looks like clang and gcc support it with C++17, but MSVC doesn't: https://godbolt.org/z/zTMGfv

samuelpmishLLNL (commented):
it's pretty tricky to review these PRs that change types between values/pointers, the diffs become long enough that it's easy to miss stuff.

src/solvers/elasticity_solver.cpp
joshessman-llnl: Yes, it's not official until C++20, but most compilers (even Intel) have supported it for almost a decade.  Designated initializers were part of C99 so the porting effort for the language implementers was minimal.  Does anyone feel strongly about adding a dependency on technically-not-standardized language features?

samuelpmishLLNL: can't we just write a 2 argument constructor for `FESOptions` instead? By the same token, we still don't use the widely (but unofficially) supported `#pragma once` either, presumably for the same reason.

white238: I vote this is a pretty useful and innocuous compiler feature, that if our compiler set support it we can move ahead with it.  We can take these compiler features on a case by case basis.

joshessman-llnl: The issue is that there are (currently) four options (all with defaults) and different solvers need to specify different options, so you'd need a constructor for each item in the powerset of options (2^n = 16 ctors).  Designated initializers let you specify only what you care about and thus avoid the problem of having to worry about things like having default arguments in the right order.

src/common/serac_types.hpp
```diff
@@ -136,40 +136,7 @@ struct NonlinearSolverParameters {
...
-   * @brief The parallel mesh
-   */
-  std::shared_ptr<mfem::ParMesh> mesh;
-
-  /**
-   * @brief Name of the state variable
-   */
-  std::string name = "";
-};
+using GeneralCoefficient = std::variant<std::shared_ptr<mfem::Coefficient>, std::shared_ptr<mfem::VectorCoefficient>>;
```
jamiebramwell, line 139: Is there a reason this is a variant of two flavors of `shared_ptr` vs. the two direct coefficient types?

src/common/finite_element_state.hpp
```diff
@@ -0,0 +1,159 @@
...
+#include <type_traits>
+
+#include "common/serac_types.hpp"
+
+namespace serac {
+
+/**
+ * Structure for optionally configuring a FiniteElementState
+ */
+// The optionals are explicitly default-constructed to allow the user to partially aggregrate-initialized
```
jamiebramwell, line 27: 👍

src/common/finite_element_state.hpp
```diff
@@ -0,0 +1,159 @@
...
+  std::optional<int> space_dim = std::optional<int>();
+  /**
+   * The FECollection to use - defaults to an H1_FECollection
+   */
+  std::optional<std::unique_ptr<mfem::FiniteElementCollection>> coll =
+      std::optional<std::unique_ptr<mfem::FiniteElementCollection>>();
+  /**
+   * The DOF ordering that should be used interally by MFEM
+   */
+  mfem::Ordering::Type ordering = mfem::Ordering::byNODES;
```
jamiebramwell: The default ordering should probably be `Ordering::byVDIM` since that is the one that works with BoomerAMG.

src/common/finite_element_state.hpp
```diff
@@ -0,0 +1,159 @@
...
+class FiniteElementState {
+public:
+  /**
+   * Main constructor for building a new state object
+   * @param[in] order The order of the problem
+   * @param[in] pmesh The problem mesh
+   * @param[in] options The options specified, namely those relating to the dimension of the FESpace, the type of
+   * FEColl, the DOF ordering that should be used, and the name of the field
+   */
+  FiniteElementState(const int order, std::shared_ptr<mfem::ParMesh> pmesh, FESOptions&& options = FESOptions());
```
jamiebramwell: The order should probably go with the `FESOptions` since you construct the collection using this order.

src/solvers/base_solver.cpp
```diff
@@ -53,28 +50,25 @@ void BaseSolver::setEssentialBCs(const std::set<int>& ess_bdr, serac::BoundaryCo
...
-void BaseSolver::setState(const std::vector<serac::BoundaryCondition::Coef>& state_coef)
+void BaseSolver::setState(const std::vector<serac::GeneralCoefficient>& state_coef)
 {
   SLIC_ASSERT_MSG(state_coef.size() == state_.size(), "State and coefficient bundles not the same size.");
 
   for (unsigned int i = 0; i < state_coef.size(); ++i) {
-    // The generic lambda parameter, auto&&, allows the component type (mfem::Coef or mfem::VecCoef)
-    // to be deduced, and the appropriate version of ProjectCoefficient is dispatched.
-    std::visit([this, i](auto&& coef) { state_[i]->gf->ProjectCoefficient(*coef); }, state_coef[i]);
+    state_[i]->project(state_coef[i]);
```
jamiebramwell, line 71: 👍

src/solvers/elasticity_solver.cpp
jamiebramwell: Yeah, this makes it way cleaner. As long as nvcc or whatever GPU compiler du jour can do this I strongly support it.

src/solvers/elasticity_solver.cpp
```diff
@@ -142,14 +133,12 @@ void ElasticitySolver::QuasiStaticSolve()
   *bc_rhs_ = *rhs_;
   for (auto& bc : ess_bdr_) {
     bool should_be_scalar = false;
-    bc.projectBdr(*displacement_->gf, time_, should_be_scalar);
-    displacement_->gf->GetTrueDofs(*displacement_->true_vec);
-    mfem::EliminateBC(*K_mat_, *K_e_mat_, bc.getTrueDofs(), *displacement_->true_vec, *bc_rhs_);
+    bc.apply(*K_mat_, *bc_rhs_, *displacement_, time_, should_be_scalar);
```
jamiebramwell, line 135: This brings me joy!

src/common/finite_element_state.hpp
```diff
@@ -0,0 +1,159 @@
...
+#include "common/serac_types.hpp"
+
+namespace serac {
+
+/**
+ * Structure for optionally configuring a FiniteElementState
+ */
+// The optionals are explicitly default-constructed to allow the user to partially aggregrate-initialized
+// with only the options they care about
+struct FESOptions {
```
jamiebramwell: `FEStateOptions` may be preferable since `FESpace` is also a thing in MFEM.

src/common/finite_element_state.hpp
```diff
@@ -0,0 +1,159 @@
...
+  void distributeSharedDofs() { gf_->SetFromTrueDofs(true_vec_); }
+
+  /**
+   * Utility function for creating a tensor, e.g. mfem::HypreParVector,
+   * mfem::ParBilinearForm, etc on the FESpace encapsulated by an FEState object
+   * @return An owning pointer to a heap-allocated tensor
+   * @pre Tensor must have the constructor Tensor::Tensor(ParFiniteElementSpace*)
+   */
+  template <typename Tensor>
+  std::unique_ptr<Tensor> createTensorOnSpace()
```
jamiebramwell: While I like the idea of this function, I'm not sure `Tensor` is appropriate since it can be used to construct explicitly nonlinear things. Maybe `createObjectOnSpace` or just `createOnSpace`? Any other ideas?

jamiebramwell:
> 
> 
> it's pretty tricky to review these PRs that change types between values/pointers, the diffs become long enough that it's easy to miss stuff.

Yeah, I agree. My hope is these large sweeping refactoring/style PRs should slow down soon. I am a big fan of small PRs!

jamiebramwell:
This has some great code readability and general quality of life improvements! I have a few comments we should discuss, but overall this looks great!

src/solvers/base_solver.cpp
```diff
@@ -143,9 +137,9 @@ void BaseSolver::initializeOutput(const serac::OutputType output_type, const std
 
   switch (output_type_) {
     case serac::OutputType::VisIt: {
-      visit_dc_ = std::make_unique<mfem::VisItDataCollection>(root_name_, state_.front()->mesh.get());
+      visit_dc_ = std::make_unique<mfem::VisItDataCollection>(root_name_, &state_.front()->mesh());
       for (const auto& state : state_) {
-        visit_dc_->RegisterField(state->name, state->gf.get());
```
white238, line 148: It's a sad day. gf will always live on in our hearts.

src/solvers/elasticity_solver.cpp
white238: It's also easy to remove if it causes a problem in nvcc.

src/solvers/thermal_operators.cpp
```diff
@@ -10,35 +10,35 @@
...
 
-DynamicConductionOperator::DynamicConductionOperator(std::shared_ptr<mfem::ParFiniteElementSpace> fespace,
-                                                     const serac::LinearSolverParameters&         params,
-                                                     std::vector<serac::BoundaryCondition>&       ess_bdr)
-    : mfem::TimeDependentOperator(fespace->GetTrueVSize(), 0.0),
+DynamicConductionOperator::DynamicConductionOperator(mfem::ParFiniteElementSpace&           fespace,
+                                                     const serac::LinearSolverParameters&   params,
+                                                     std::vector<serac::BoundaryCondition>& ess_bdr)
+    : mfem::TimeDependentOperator(fespace.GetTrueVSize(), 0.0),
       fespace_(fespace),
```
white238: Should this be fe_space_?

src/common/boundary_condition.cpp
```diff
@@ -46,26 +47,27 @@ void BoundaryCondition::project(mfem::ParGridFunction& gf, const mfem::ParFinite
     // Generate the scalar dof list from the vector dof list
     SLIC_ASSERT_MSG(space, "Only BCs associated with a space can be projected.");
     mfem::Array<int> dof_list(size);
-    std::transform(&tdofs[0], &tdofs[0] + size, &dof_list[0], [&space](int tdof) { return space.VDofToDof(tdof); });
+    std::transform(&tdofs[0], &tdofs[0] + size, &dof_list[0],
```
white238: Can you add a unit test to ensure that mfem::Array continues to work with std::transform?  This could be a very bad time if this changes from under us.

white238: Especially since we use different C++ standards than they do.

src/solvers/elasticity_solver.cpp
jamiebramwell: Once we get a "dumb" GPU build and CI up and running, I will feel much more comfortable with these unofficial language features. The `#pragma once` would also be really nice...

joshessman-llnl:
> My hope is these large sweeping refactoring/style PRs should slow down soon. I am a big fan of small PRs!

I agree, but to some extent it's a catch-22: eliminating or reducing tight couplings between components means that future modifications won't need to modify as much code (as long as the right abstractions are chosen), but can require a fair amount of upfront modification.

src/common/serac_types.hpp
joshessman-llnl: They are abstract base classes, so they have to be wrapped up by a pointer or reference.  As work continues on #118 there will probably need to be a rethinking of which classes should be responsible for what things, so this could possibly change to `unique_ptr`.

src/solvers/elasticity_solver.cpp
joshessman-llnl: I don't like having a function with so many parameters but I still think it's an improvement

src/common/finite_element_state.hpp
joshessman-llnl: Yes, this is kind of the same problem as `SystemSolver` (TBD?).  I was hoping for something more descriptive but since the function really only enforces the existence of a matching constructor maybe `createOnSpace` is appropriate.

src/solvers/thermal_operators.cpp
jamiebramwell: probably

jamiebramwell:
> 
> 
> > My hope is these large sweeping refactoring/style PRs should slow down soon. I am a big fan of small PRs!
> 
> I agree, but to some extent it's a catch-22: eliminating or reducing tight couplings between components means that future modifications won't need to modify as much code (as long as the right abstractions are chosen), but can require a fair amount of upfront modification.

Yep, that's why I'm OK with this for now. It's kind of a necessary evil.

jamiebramwell (approved):

white238 (approved):

white238:
LGTM

PR #139, opened 2020-08-14
Update to newest version of Uberenv, fixing bugs

adrienbernede:
This PR only updated Uberenv script.

The main fix I wanted is that Uberenv used to ignore a change in the Spack commit section of the `project.json`.
Updating Spack was a pain, or required removing uberenv_libs directory.

### DIFF
--- a/scripts/uberenv/uberenv.py
+++ b/scripts/uberenv/uberenv.py
@@ -1,7 +1,7 @@
 #!/bin/sh
 "exec" "python" "-u" "-B" "$0" "$@"
 ###############################################################################
-# Copyright (c) 2014-2019, Lawrence Livermore National Security, LLC.
+# Copyright (c) 2014-2020, Lawrence Livermore National Security, LLC.
 #
 # Produced at the Lawrence Livermore National Laboratory
 #
@@ -60,6 +60,7 @@ import platform
 import json
 import datetime
 import glob
+import re
 
 from optparse import OptionParser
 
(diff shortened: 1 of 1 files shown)


adrienbernede:
LGTM

adrienbernede:
Tested both on PR and Develop workflow. (That’s why it points to the latter in Gitlab CI link).

adrienbernede:
LGTM

scripts/uberenv/uberenv.py
```diff
@@ -224,12 +232,6 @@ def is_darwin():
 def is_windows():
     return "windows" in platform.system().lower()
 
-def is_wsl():
```
white238, line 227: @jamiebramwell is this still needed for WSL/WSL2 support?

jamiebramwell: I don't think so.

scripts/uberenv/uberenv.py
```diff
@@ -272,8 +271,6 @@ def detect_platform(self):
         res = None
         if is_darwin():
             res = "darwin"
-        elif is_wsl():
```
jamiebramwell, line 275: This is the only case where it is used. We could have wsl users manually specify the spack config (this is what I do for WSL2 anyway), but we should update the docs to reflect this.

white238 (approved):

scripts/uberenv/uberenv.py
white238: yea we should just have them specify the directory in the docs.

jamiebramwell: 👍 Another PR is incoming to fix the docs.

jamiebramwell (approved):

PR #141, opened 2020-08-18
Add initialize function for use in drivers

### LINKED ISSUE #134: Create a driver initialize function
The serac driver (and subsequent drivers) will have a lot of shared logic for initializing basic functionality.  We should create one function that does this.

Such as:
* call register signal handler (registerSignals() )
* init mpi
* init slic

joshessman-llnl:
This adds a function that encapsulates the three "boilerplate" components of a serac driver, namely:
- MPI init
- logger init
- signal handler init

Before merging, some possible additions:

1. Right now this is probably overkill, but we could look at RAII to handle init/cleanup:
```cpp
int main(argc, argv)
{
  // Calls init functions
  serac::Driver driver(argc, argv);
  // ... all the driver code ...
} // <- "driver" goes out of scope here, destructor would call MPI_Finalize, logger finalize, etc
```

2. Depending on how subsequent drivers may work, the CLI parse + help logic could also be handled by this `initialize` function, but this is probably less of a concern if Inlet will be replacing most of the CLI arguments.

Resolves #134

### DIFF
--- a/src/common/CMakeLists.txt
+++ b/src/common/CMakeLists.txt
@@ -7,6 +7,7 @@
 set(common_headers
     boundary_condition.hpp
     common.hpp
+    initialize.hpp
     serac_types.hpp
     logger.hpp
     mesh_utils.hpp
@@ -15,6 +16,7 @@ set(common_headers
 
 set(common_sources
     boundary_condition.cpp
+    initialize.cpp
     logger.cpp
     mesh_utils.cpp
     terminator.cpp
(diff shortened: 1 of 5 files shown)


white238 (approved):

white238:
We might actually be able to move the CLI logic in here after using Inlet because most command line options will be very generic at that point.  But that will obviously have to be found out when we move to inlet.

white238:
LGTM

jamiebramwell (approved):

joshessman-llnl:
LGTM

PR #142, opened 2020-08-19
Create a mesh shared_ptr accessor

jamiebramwell:
In the tribol work, I create a new `FiniteElementState` based on another one. This means I need access to the underlying `shared_ptr` to the mesh instead of just a reference. This is one simple solution, but there may be better ones. Any thoughts @joshessman-llnl @white238 ?

### DIFF
--- a/src/common/finite_element_state.hpp
+++ b/src/common/finite_element_state.hpp
@@ -83,7 +83,7 @@ public:
   mfem::ParMesh& mesh() { return *mesh_; }
 
   /**
-   * Returns a owning shared_ptr to the internal mesh object
+   * Returns an owning shared_ptr to the internal mesh object
    */ 
   std::shared_ptr<mfem::ParMesh> meshPointer() { return mesh_; }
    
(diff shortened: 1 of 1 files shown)


white238:
Does tribol use shared_ptrs?

jamiebramwell:
> 
> 
> Does tribol use shared_ptrs?

No - this is for our own internal bookkeeping. I have to make new state objects specifically for contact. I've been chatting with @joshessman-llnl , and we think the best answer might be to move the mesh object to the solver.

white238:
> Does tribol use shared_ptrs?

Oh i see what you are saying, its another one of our FiniteElementState class holding onto the same ParMesh.  I don't see why that would be a problem.  Though maybe this begs a different question, should the individual FiniteElementStates hold the mesh class or a single higher level class?

jamiebramwell:
After talking with @joshessman-llnl , I think we settled on moving the mesh to the solver instead of the state. I am going to close this PR in favor of that solution.

PR #144, opened 2020-08-19
Have the solver (engine?) objects own the primary mesh

### LINKED ISSUE #143: Move the mesh to the solver
Right now, the `FiniteElementState` logically owns the mesh instead of the `BaseSolver` class. This becomes an issue when we want to make new `FiniteElementState` objects outside of the constructor. After talking with @joshessman-llnl and @white238 , we think the mesh shared pointer should be moved to the `BaseSolver` class. For reference, this is currently needed for the Tribol contact library integration.

joshessman-llnl:
A future PR may eliminate the need for sharing the mesh at all (i.e. use `unique_ptr`) once the exact interface for things like AMR is defined.

This PR intends that solvers that require more than one mesh should initialize the BaseSolver object with the primary mesh and store remaining meshes within the relevant derived solver object.  Further work on #118 may separate interface, data, and functionality that is currently all handled by the BaseSolver class, in which case this intention may change.

Resolves #143

### DIFF
--- a/src/common/finite_element_state.hpp
+++ b/src/common/finite_element_state.hpp
@@ -60,12 +60,12 @@ class FiniteElementState {
 public:
   /**
    * Main constructor for building a new state object
-   * @param[in] pmesh The problem mesh
+   * @param[in] mesh The problem mesh (object does not take ownership)
    * @param[in] options The options specified, namely those relating to the order of the problem,
    * the dimension of the FESpace, the type of FEColl, the DOF ordering that should be used,
    * and the name of the field
    */
-  FiniteElementState(mfem::ParMesh& pmesh, FEStateOptions&& options = FEStateOptions());
+  FiniteElementState(mfem::ParMesh& mesh, FEStateOptions&& options = FEStateOptions());
 
   /**
    * Returns the MPI communicator for the state
--- a/src/common/finite_element_state.hpp
+++ b/src/common/finite_element_state.hpp
@@ -65,7 +65,7 @@ public:
    * the dimension of the FESpace, the type of FEColl, the DOF ordering that should be used,
    * and the name of the field
    */
-  FiniteElementState(std::shared_ptr<mfem::ParMesh> pmesh, FEStateOptions&& options = FEStateOptions());
+  FiniteElementState(mfem::ParMesh& pmesh, FEStateOptions&& options = FEStateOptions());
 
   /**
    * Returns the MPI communicator for the state
@@ -80,7 +80,7 @@ public:
   /**
    * Returns a non-owning reference to the internal mesh object
    */
-  mfem::ParMesh& mesh() { return *mesh_; }
+  mfem::ParMesh& mesh() { return mesh_; }
 
   /**
    * Returns a non-owning reference to the internal FESpace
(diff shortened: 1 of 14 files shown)


src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -201,8 +202,7 @@ void NonlinearSolidSolver::advanceTimestep(double& dt)
   displacement_->initializeTrueVec();
 
   // Set the mesh nodes to the reference configuration
-  displacement_->mesh().NewNodes(*reference_nodes_);
-  velocity_->mesh().NewNodes(*reference_nodes_);
+  mesh_->NewNodes(*reference_nodes_);
```
jamiebramwell, line 204: Yeah, this also cleans this stuff up a lot as well

jamiebramwell (approved):
Thanks for looking into this!

src/common/finite_element_state.hpp
```diff
@@ -65,7 +65,7 @@ class FiniteElementState {
    * the dimension of the FESpace, the type of FEColl, the DOF ordering that should be used,
    * and the name of the field
    */
-  FiniteElementState(std::shared_ptr<mfem::ParMesh> pmesh, FEStateOptions&& options = FEStateOptions());
+  FiniteElementState(mfem::ParMesh& pmesh, FEStateOptions&& options = FEStateOptions());
```
white238: Should we be labeling in parameters as non-owning?

src/solvers/base_solver.hpp
```diff
@@ -30,18 +30,18 @@ class BaseSolver {
...
+   * @param[in] pmesh The primary mesh
    */
-  BaseSolver(MPI_Comm comm);
+  BaseSolver(std::shared_ptr<mfem::ParMesh> pmesh);
 
   /**
    * @brief Constructor that creates n entries in state_ of order p
    *
-   * @param[in] comm MPI communicator
+   * @param[in] pmesh The primary mesh
```
white238: p for primary or parallel is a tiny bit confusing also their is an llnl code called pmesh if it even matters.

jamiebramwell: This is an unfortunately MFEM idiom I picked up. Since we don't use the serial flavor of meshes, we should just call it `mesh`.

src/common/finite_element_state.hpp
jamiebramwell: Are references ever owning? I always assume references are owned by the caller.

white238: You could explicitly delete it but that would be bad. I was talking more for about consistency and clarity sake. We do return it out of this class below and label it as non-owning... how do you feel about this @joshessman-llnl ?

joshessman-llnl: I don't mind adding it in for clarity, but Jamie is right that references are almost never owning  - you'd have to do something really weird to get an "owning reference":
```cpp
int& a = *(new int);
// ...
delete &a;
```

jamiebramwell: I think a note saying not to do this in the style guide would suffice.

samuelpmishLLNL (approved):

white238 (approved):

jamiebramwell:
LGTM

PR #148, opened 2020-08-20
Add basic support for profiling with Caliper

### LINKED ISSUE #119: Add option to use Caliper performance analysis library
We should have a build option to use [Caliper](https://github.com/LLNL/Caliper) for performance analysis and timing metrics.

joshessman-llnl:
This introduces the build system changes required to install and link against Caliper (with the `+caliper`) variant, documentation for source code annotation, and a simple example usage in a fairly hot loop within one of the integrators.

The example has no overhead when profiling is disabled, but it can be placed elsewhere if that would be more helpful.

Resolves #119


src/common/profiling.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <string>
+
+#include "serac_config.hpp"
+
+#ifdef SERAC_USE_CALIPER
+
+#include "caliper/cali-manager.h"
+#include "caliper/cali.h"
+
+#define MARK_FUNCTION CALI_CXX_MARK_FUNCTION
```
white238: Paranoid Chris: Please prefix these with SERAC_ so we avoid name conflicts

white238 (approved):

white238:
LGTM

src/drivers/serac.cpp
```diff
@@ -29,6 +30,9 @@ int main(int argc, char* argv[])
 {
   auto [num_procs, rank] = serac::initialize(argc, argv);
 
+  // Initialize profiling if enabled
```
jamiebramwell: Should this be in initialize?

joshessman-llnl: Oops, I added it to initialize but forgot to remove it from the driver

jamiebramwell (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #158, opened 2020-08-26
Add coefficient accessors to the BC class

jamiebramwell:
This adds accessors for the underlying boundary condition coefficients. It also uses the vanilla mfem constructors instead of special factory methods for the boundary integrators.

### DIFF
--- a/src/common/boundary_condition.cpp
+++ b/src/common/boundary_condition.cpp
@@ -48,7 +48,7 @@ void BoundaryCondition::project(FiniteElementState& state) const
     SLIC_ASSERT_MSG(space, "Only BCs associated with a space can be projected.");
     mfem::Array<int> dof_list(size);
     std::transform(tdofs.begin(), tdofs.end(), dof_list.begin(),
-                   [& space = std::as_const(state.space())](int tdof) { return space.VDofToDof(tdof); });
+                   [&space = std::as_const(state.space())](int tdof) { return space.VDofToDof(tdof); });
 
     if (component_ == -1) {
       // If it contains all components, project the vector
--- a/src/common/boundary_condition.cpp
+++ b/src/common/boundary_condition.cpp
@@ -48,7 +48,7 @@ void BoundaryCondition::project(FiniteElementState& state) const
     SLIC_ASSERT_MSG(space, "Only BCs associated with a space can be projected.");
     mfem::Array<int> dof_list(size);
     std::transform(tdofs.begin(), tdofs.end(), dof_list.begin(),
-                   [&space = std::as_const(state.space())](int tdof) { return space.VDofToDof(tdof); });
+                   [& space = std::as_const(state.space())](int tdof) { return space.VDofToDof(tdof); });
 
     if (component_ == -1) {
       // If it contains all components, project the vector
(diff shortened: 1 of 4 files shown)


jamiebramwell:
Let's discuss how we want to handle complex boundary integrators. This is more in line with how we handle the domain integrators. Any thoughts?

jamiebramwell:
LGTM

src/common/boundary_condition.cpp
```diff
@@ -123,4 +123,32 @@ void BoundaryCondition::apply(mfem::HypreParMatrix& k_mat_post_elim, mfem::Vecto
   eliminateToRHS(k_mat_post_elim, state.trueVec(), rhs);
 }
 
+const mfem::Coefficient& BoundaryCondition::scalarCoefficient() const
+{
+  SLIC_ERROR_IF(!std::holds_alternative<std::shared_ptr<mfem::Coefficient>>(coef_),
+                "Asing for a scalar coefficient on a BoundaryCondition that contains a vector coefficient.");
```
jandrej: s/Asing/Asking

jamiebramwell: Now that all of our communication is virtual, I am realizing how many typos I make...

src/common/boundary_condition.cpp
```diff
@@ -123,4 +123,32 @@ void BoundaryCondition::apply(mfem::HypreParMatrix& k_mat_post_elim, mfem::Vecto
...
+  return *std::get<std::shared_ptr<mfem::Coefficient>>(coef_);
+}
+
+mfem::Coefficient& BoundaryCondition::scalarCoefficient()
+{
+  SLIC_ERROR_IF(!std::holds_alternative<std::shared_ptr<mfem::Coefficient>>(coef_),
```
jamiebramwell, line 135: Should this be an assert? It should only come up in solver development.

samuelpmishLLNL (commented):
It feels like these changes are trying to return to the non-variant storage of coefficients. I'm not against this-- I think std::variant has its uses but maybe it's not the right fit for this use case.

src/common/boundary_condition.cpp
joshessman-llnl: If a user-facing solver ever exposes `GeneralCoef` (the variant) then this might still be useful as an error condition.

src/common/boundary_condition.hpp
```diff
@@ -52,6 +52,46 @@ class BoundaryCondition {
...
+   */
+  const mfem::Coefficient& scalarCoefficient() const;
+
+  /**
+   * @brief Accessor for the underlying scalar coefficient
+   *
+   * This method performs an internal check to verify the underlying GeneralCoefficient
+   * is in fact a scalar.
+   *
+   * @return A non-owning reference to the underlying scalar coefficient
```
joshessman-llnl, line 91: :+1:

joshessman-llnl (approved):

joshessman-llnl:
> It feels like these changes are trying to return to the non-variant storage of coefficients. I'm not against this-- I think std::variant has its uses but maybe it's not the right fit for this use case.

From my point of view this allows some of the physics-specific stuff to "assert" information about a particular BC, but allows for uniform storage and in some cases application.  I'd really like to see if we can figure out a more exhaustive list of the ways that boundary conditions might be used so we can determine what a general BC interface would look like.

jamiebramwell:
> 
> 
> It feels like these changes are trying to return to the non-variant storage of coefficients. I'm not against this-- I think std::variant has its uses but maybe it's not the right fit for this use case.

Yeah, I still think the use of `std::variant` cleans up a lot of the internal duplication in the `BaseSolver` and `BoundaryCondition` classes, so I'm on the side of keeping it for now. However, I would like to keep solver developers away from the `std::variant` syntax. At that level, one should know whether to interpret the boundary condition coefficient as a scalar or vector (or hopefully tensor in the future).

jamiebramwell:
> 
> 
> > It feels like these changes are trying to return to the non-variant storage of coefficients. I'm not against this-- I think std::variant has its uses but maybe it's not the right fit for this use case.
> 
> From my point of view this allows some of the physics-specific stuff to "assert" information about a particular BC, but allows for uniform storage and in some cases application. I'd really like to see if we can figure out a more exhaustive list of the ways that boundary conditions might be used so we can determine what a general BC interface would look like.

Sure enough, this is a much clearer way to say what I was trying to describe!

samuelpmishLLNL (approved):

jamiebramwell:
LGTM

PR #162, opened 2020-08-28
Add +cuda Spack variant and a smoketest for CUDA

### LINKED ISSUE #137: Add dummy GPU build
Add a GPU-enabled build to the gitlab CI to test the advanced architecture compilers for our use of new language features.

joshessman-llnl:
BLT's built-in smoketest is too trivial to determine whether a CUDA C++17 environment is correctly set up (doesn't require any new stdlib features), so a new smoketest was added based on NVIDIA's "hello world".  

On Lassen, CUDA will not function properly unless the correct CUDA driver (CUDA 11) is loaded via the `ml`/`module` command.  I am not sure whether this should be documented in the sphinx docs or somewhere else.

Resolves #137


host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@upstream_gfortran-cuda.cmake
```diff
@@ -0,0 +1,82 @@
...
+
+#------------------------------------------------------------------------------
+# Cuda
+#------------------------------------------------------------------------------
+
+set(ENABLE_CUDA ON CACHE BOOL "")
+
+set(CUDA_TOOLKIT_ROOT_DIR "/usr/tce/packages/cuda/cuda-11.0.182" CACHE PATH "")
+
+set(CMAKE_CUDA_COMPILER "${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc" CACHE PATH "")
```
samuelpmishLLNL, line 38: I would expect these to be populated by `find_package(CUDA REQUIRED)`, but I guess it doesn't hurt to be explicit

samuelpmishLLNL (approved):
Looks good-- I don't understand uberenv enough to make heads or tails of that part, though.

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@upstream_gfortran-cuda.cmake
white238: BLT's CUDA supports started before what is currently in CMake and also supports separable compiliation which the current CMake functionality does not support yet.  Because of this, we have a mix between FindCUDA (deprecated but still necessary) and the CMake first class citizen language support.

It is possible its not needed though...

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@upstream_gfortran-cuda.cmake
```diff
@@ -0,0 +1,82 @@
...
+
+set(SUPERLUDIST_DIR "${TPL_ROOT}/superlu-dist-5.4.0" CACHE PATH "")
+
+set(MFEM_DIR "${TPL_ROOT}/mfem-4.1.0p1" CACHE PATH "")
+
+#------------------------------------------------------------------------------
+# Devtools
+#------------------------------------------------------------------------------
+
+set(ENABLE_DOCS OFF CACHE BOOL "")
```
white238: Did something go wrong with the devtools detection so none of them show up or are they just not built yet on this platform?

scripts/uberenv/packages/serac/package.py
```diff
@@ -241,6 +248,51 @@ def hostconfig(self, spec, prefix, py_site_pkgs_dir=None):
...
+
+            if not spec.satisfies('cuda_arch=none'):
+                cuda_arch = spec.variants['cuda_arch'].value
+                flag = '-arch sm_{0}'.format(cuda_arch[0])
+                # CXX flags will be propagated to the host compiler
+                cuda_flags = ' '.join([flag, cxxflags])
+                cfg.write(cmake_cache_string("CMAKE_CUDA_FLAGS", cuda_flags))
+                cfg.write(cmake_cache_string("CMAKE_CUDA_ARCHITECTURES", ' '.join(cuda_arch)))
+
+            if '+deviceconst' in spec:
```
white238: This shouldnt be needed. I assume you copied this from umpire.

CMakeLists.txt
```diff
@@ -44,6 +44,12 @@ set(ENABLE_ASTYLE      OFF CACHE BOOL "")
 set(ENABLE_CLANGQUERY  OFF CACHE BOOL "")
 set(ENABLE_UNCRUSTIFY  OFF CACHE BOOL "")
 
+if (ENABLE_CUDA AND ON_BLUEOS)
+    # Very specific fix for working around CMake adding implicit link directories returned by the BlueOS
+    # compilers to link CUDA executables 
+    set(BLT_CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES_EXCLUDE "/usr/tce/packages/gcc/gcc-4.9.3/lib64/gcc/powerpc64le-unknown-linux-gnu/4.9.3;/usr/tce/packages/gcc/gcc-4.9.3/lib64" CACHE STRING "")
```
white238: I'd like to keep all this machine specific stuff in one place.  Can you move this into the spack package then have it generated into host-config?  I'd also be concerned this wouldn't get loaded early enough for when the checks are called (though i assume you tested this).

joshessman-llnl: That variable just needs to be set before BLT is included (as that's when all the setup stuff happens).

I had initially had this set in the serac `package.py` but changed it to avoid (sort of) introducing an explicit BLT dependency in the Spack config stuff.  If that's not a concern I'll move it back.

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@upstream_gfortran-cuda.cmake
samuelpmishLLNL: CUDA_SEPARABLE_COMPILATION has been in CMake since 3.6, it seems

samuelpmishLLNL: whether or not it works is another question!

white238: It was in FindCUDA CMake since then, but it is still not supported in the new CUDA CMake support, see: https://cmake.org/cmake/help/v3.18/release/3.18.html#languages

CMakeLists.txt
white238: definitely not a concern.

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@upstream_gfortran-cuda.cmake
joshessman-llnl: They hadn't been built yet when this hostconfig was generated, will update accordingly

white238 (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@upstream_gfortran-cuda.cmake
samuelpmishLLNL: I think that only applies to compiling CUDA with clang directly (CUDA language support in clang lags behind nvcc considerably). Separable compilation works fine with CMake + nvcc in my experience.

white238: find_package(CUDA) is NOT the new cmake CUDA as a first case language capability.

There are several codes at the lab that use BLT's CUDA support that relies on a mixture of both that utilize the CUDA separable compilation but that specifically relies on the deprecated find_package(CUDA)

PR #165, opened 2020-08-31
Feature/essman/expr templates

joshessman-llnl:
This adds a set of operator overloads for making MFEM code more readable.  The overloads themselves do not have Doxygen annotations as they are just syntactic sugar over the expression classes, but I'm happy to add it in.  Ideally more operations would be added in future PRs.

The tricky lifetime issues regarding returning expressions from lambdas are addressed in a new Sphinx page, and are tested (though not exhaustively) in a new set of unit-ish tests.

The new templates are spread across three headers: one for the base class (such that the primary interface is accessible), one for the internal derived template classes, and one for the operator overloads (which is just a bunch of copy-and-paste).  

The operator overload boilerplate could be partially reduced with macros, but that could be equally messy.  

The "UnaryExpr" and "BinaryExpr" abstractions don't eliminate a ton of code duplication right now given the small set of currently implemented operations but as more are added the hope is that they will be useful.


src/solvers/thermal_operators.cpp
```diff
@@ -59,16 +60,13 @@ void DynamicConductionOperator::Mult(const mfem::Vector& u, mfem::Vector& du_dt)
...
   }
 
   // Compute:
   //    du_dt = M^{-1}*-K(u)
   // for du_dt
-  K_mat_->Mult(y_, z_);
-  z_.Neg();  // z = -zw z_.Add(1.0, *bc_rhs_);
-  z_.Add(1.0, *bc_rhs_);
-  M_solver_.Mult(z_, du_dt);
+  du_dt = M_solver_ * (-(*K_mat_ * y_) + *bc_rhs_);
```
samuelpmishLLNL: I think we might rename some of our EquationSolvers and matrices to be in more agreement with these overloaded operators,

e.g. substituting {`M_solver` -> `invM` and `K_mat` -> `K`} would make the code read

`du_dt = invM_ * (f_ - K_ * y_);`

src/common/CMakeLists.txt
```diff
@@ -7,12 +7,15 @@
 set(common_headers
```
jamiebramwell, line 7: `common` is getting fairly large. Is it worth splitting this up a bit? I'm not sure how we would go about doing that though.

samuelpmishLLNL (dismissed):
I think this all looks good! You mentioned in a private discussion that you measured the performance of the previous implementation against the expression template one. It might be good to include timers in the unit test (or should we have a separate set of performance tests @jamiebramwell ?) to verify both correctness but also similarity in performance.

jamiebramwell (commented):
> 
> 
> I think this all looks good! You mentioned in a private discussion that you measured the performance of the previous implementation against the expression template one. It might be good to include timers in the unit test (or should we have a separate set of performance tests @jamiebramwell ?) to verify both correctness but also similarity in performance.

I think having a performance test is a good idea, but I don't have an opinion where it should go. Any thoughts @white238 ?

white238:
> > I think this all looks good! You mentioned in a private discussion that you measured the performance of the previous implementation against the expression template one. It might be good to include timers in the unit test (or should we have a separate set of performance tests @jamiebramwell ?) to verify both correctness but also similarity in performance.
> 
> I think having a performance test is a good idea, but I don't have an opinion where it should go. Any thoughts @white238 ?

I would add it to the test and just print it out to the screen for now.  Do whatever is easiest until we implement a performance suite along with our integrated testing.

joshessman-llnl:
I've added a couple simple performance tests, though the results that run in CI should be taken with a grain of salt as the following was observed with the (default) debug builds:
```
Expression templates took 14.6052 times as long as the raw MFEM calls
...
Expression templates took 19.2782 times as long as the raw MFEM calls             
```
versus the release builds:
```
Expression templates took 4.0012 times as long as the raw MFEM calls
...
Expression templates took 0.498103 times as long as the raw MFEM calls
```

The larger the expression, the better the expression templates perform (as one would expect).

samuelpmishLLNL:
It is worth noting that this implementation of expression templates cannot eliminate a memory allocation when assigning the results to an mfem::Vector. With smaller vectors and fewer operations in the expression, that memory allocation overhead becomes more significant.

However, if we were able to implement an mfem::Vector constructor that takes a vector expression as an argument, this memory allocation could be eliminated.

white238:
With the large increase in times, I would worry about this creeping into non-prototyping code because it is so readable (credit to Josh), I may not even realize it's there.

Thought on how to stop this: ifdef'ing the template expressions against non-release builds in hopes when we test the release build in the future it would be very apparently if this crept into the mainline? Maybe spit out a warning if it's used in a release build?

joshessman-llnl:
> Thought on how to stop this: ifdef'ing the template expressions against non-release builds in hopes when we test the release build in the future it would be very apparently if this crept into the mainline? Maybe spit out a warning if it's used in a release build?

I went ahead and added a compiler warning for release builds, but the impact varies significantly and per above isn't always negative.  It may be useful to further investigate the inflection point at which the expression templates are faster in a release build.

samuelpmishLLNL:
oof, wrong PR, ignore the last comment!

samuelpmishLLNL (approved):

src/common/CMakeLists.txt
joshessman-llnl: It seems like there may be a reasonable split between "utility" stuff - loggers, initialization, termination, and FEM-related stuff - boundary conditions, mesh stuff, etc.

src/docs/sphinx/expr_templates.rst
```diff
@@ -0,0 +1,78 @@
+.. ## Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
+.. ## other Serac Project Developers. See the top-level COPYRIGHT file for details.
+.. ##
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+
+====================
+Expression Templates
+====================
+
+Expression templates are a C++ technique that leverages static polymorphism (with the `Curiously
```
white238, line 10: Add a note here to say to only use this for prototyping

samuelpmishLLNL: I don't believe it is only for prototyping.

jamiebramwell: I think with the current [performance hit](https://github.com/LLNL/serac/pull/165#issuecomment-685968525), it's best to keep it for prototyping purposes (at least in hot loops) for now. Once we can beat down these slowdowns, we can put them in production kennels.

jamiebramwell: I view this current functionality similar to using auto differentiation for calculating residual gradients. It's good for generating baselines for performant versions.

samuelpmishLLNL: When I spoke to Josh earlier, he said the 20x slowdown was from comparing times from debug expression templates w/ release mfem, which isn't really an honest comparison.

On my home machine, comparing release serac w/ release mfem:
```
[==========] Running 12 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 12 tests from expr_templates
[ RUN      ] expr_templates.basic_add
[       OK ] expr_templates.basic_add (0 ms)
[ RUN      ] expr_templates.basic_add_lambda
[       OK ] expr_templates.basic_add_lambda (0 ms)
[ RUN      ] expr_templates.move_from_temp_lambda
[       OK ] expr_templates.move_from_temp_lambda (0 ms)
[ RUN      ] expr_templates.move_from_temp_vec_lambda
[       OK ] expr_templates.move_from_temp_vec_lambda (0 ms)
[ RUN      ] expr_templates.small_matvec
[       OK ] expr_templates.small_matvec (0 ms)
[ RUN      ] expr_templates.small_mixed_expr
Expression templates took 1.58392 times as long as the raw MFEM calls
[       OK ] expr_templates.small_mixed_expr (9 ms)
[ RUN      ] expr_templates.large_mixed_expr
Expression templates took 1.02729 times as long as the raw MFEM calls
[       OK ] expr_templates.large_mixed_expr (1980 ms)
[ RUN      ] expr_templates.complex_expr_lambda
[       OK ] expr_templates.complex_expr_lambda (0 ms)
[ RUN      ] expr_templates.small_vec_single_add_time_check
Expression templates took 0.916667 times as long as the raw MFEM calls
[       OK ] expr_templates.small_vec_single_add_time_check (0 ms)
[ RUN      ] expr_templates.large_vec_single_add_time_check
Expression templates took 0.84289 times as long as the raw MFEM calls
[       OK ] expr_templates.large_vec_single_add_time_check (23 ms)
[ RUN      ] expr_templates.small_vec_multi_add_time_check
Expression templates took 0.83571 times as long as the raw MFEM calls
[       OK ] expr_templates.small_vec_multi_add_time_check (2 ms)
[ RUN      ] expr_templates.large_vec_multi_add_time_check
Expression templates took 0.309412 times as long as the raw MFEM calls
[       OK ] expr_templates.large_vec_multi_add_time_check (79 ms)
[----------] 12 tests from expr_templates (2094 ms total)
[----------] Global test environment tear-down
[==========] 12 tests from 1 test suite ran. (2094 ms total)
[  PASSED  ] 12 tests.
```

These tests show significant speedups, despite the fact that our expression template implementation is inherently flawed in the sense that we can't modify mfem::Vector to eliminate some unnecessary memory allocations. The exceptional 1.5x slowdown case is from a test with tiny vectors, where the extra overhead from dynamically allocating memory in the loop hurts.

jamiebramwell: Ahhhh that's much better!! I think with these numbers using this in production is not only allowed, it should be encouraged.

jamiebramwell: (that's what I get for trying to understand a technical discussion while chasing my kid!)

jamiebramwell: Also, if we can show how nice this is I can see making the argument to upstream the fix as well as this implementation to mfem.

jamiebramwell: Also, if we can show how nice this is I can see making the argument to upstream the fix as well as this implementation to mfem.

samuelpmishLLNL: For the sake of completeness, comparing debug serac against debug mfem:

```
[==========] Running 12 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 12 tests from expr_templates
[ RUN      ] expr_templates.basic_add
[       OK ] expr_templates.basic_add (0 ms)
[ RUN      ] expr_templates.basic_add_lambda
[       OK ] expr_templates.basic_add_lambda (0 ms)
[ RUN      ] expr_templates.move_from_temp_lambda
[       OK ] expr_templates.move_from_temp_lambda (0 ms)
[ RUN      ] expr_templates.move_from_temp_vec_lambda
[       OK ] expr_templates.move_from_temp_vec_lambda (0 ms)
[ RUN      ] expr_templates.small_matvec
[       OK ] expr_templates.small_matvec (0 ms)
[ RUN      ] expr_templates.small_mixed_expr
Expression templates took 1.17761 times as long as the raw MFEM calls
[       OK ] expr_templates.small_mixed_expr (59 ms)
[ RUN      ] expr_templates.large_mixed_expr
Expression templates took 0.995274 times as long as the raw MFEM calls
[       OK ] expr_templates.large_mixed_expr (6551 ms)
[ RUN      ] expr_templates.complex_expr_lambda
[       OK ] expr_templates.complex_expr_lambda (0 ms)
[ RUN      ] expr_templates.small_vec_single_add_time_check
Expression templates took 2.75278 times as long as the raw MFEM calls
[       OK ] expr_templates.small_vec_single_add_time_check (6 ms)
[ RUN      ] expr_templates.large_vec_single_add_time_check
Expression templates took 6.58961 times as long as the raw MFEM calls
[       OK ] expr_templates.large_vec_single_add_time_check (321 ms)
[ RUN      ] expr_templates.small_vec_multi_add_time_check
Expression templates took 1.61061 times as long as the raw MFEM calls
[       OK ] expr_templates.small_vec_multi_add_time_check (23 ms)
[ RUN      ] expr_templates.large_vec_multi_add_time_check
Expression templates took 2.97274 times as long as the raw MFEM calls
[       OK ] expr_templates.large_vec_multi_add_time_check (1425 ms)
[----------] 12 tests from expr_templates (8386 ms total)
[----------] Global test environment tear-down
[==========] 12 tests from 1 test suite ran. (8386 ms total)
[  PASSED  ] 12 tests.
```

It is true that expression templates tend to not perform as well in debug builds, as they rely pretty heavily on inlining (which is typically disabled in debug builds).

white238: Any idea how much of the runtime this part of the code is?  Those numbers are a lot more promising for sure. Especially if they would improve with some changes on the MFEM side.

joshessman-llnl:
image

The total time spent in the expression templates is 0.3% of total runtime so it looks like the overall impact is negligible regardless of which method is used.

src/common/CMakeLists.txt
jamiebramwell: I'll add an issue on this

src/docs/sphinx/expr_templates.rst
jamiebramwell: It depends on the physics operator, but these calls would probably be small compared to the global linear solves.

src/docs/sphinx/expr_templates.rst
```diff
@@ -0,0 +1,78 @@
...
+to another expression.  This allows for chaining, e.g., ``-a + b + 0.3 * c``.
+
+Because these expression objects are imperfect closures (as with lambdas in C++), care should be taken to
+ensure that objects are not used after they go out of scope.  
+
+Consider a case where an intermediate expression is assigned to a variable (perhaps for readability):
+
+.. code-block:: cpp
+
+    auto lambda = [](const auto& a, const auto& b) {
```
jamiebramwell, line 40: Thank you for these examples!

src/solvers/thermal_operators.cpp
```diff
@@ -21,75 +22,73 @@ DynamicConductionOperator::DynamicConductionOperator(mfem::ParFiniteElementSpace
...
   }
 
   // Compute:
   //    du_dt = M^{-1}*-K(u)
   // for du_dt
-  K_mat_->Mult(y_, z_);
-  z_.Neg();  // z = -zw z_.Add(1.0, *bc_rhs_);
-  z_.Add(1.0, *bc_rhs_);
-  M_solver_.Mult(z_, du_dt);
+  du_dt = M_inv_ * (-(*K_ * u) + *bc_rhs_);
```
jamiebramwell, line 69: This is so much nicer!

src/common/vector_expression.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+namespace serac {
+/**
+ * @brief A base class representing a vector expression
+ * @tparam T The base vector type, e.g., mfem::Vector, or another VectorExpr
+ * @note This class should never be used directly
+ */
+template <typename T>
+class VectorExpr {
+#ifdef NDEBUG
```
jamiebramwell: Do we still need this?

jamiebramwell (commented):
This looks great and I can't wait to use it. I'm holding off on approval because I want to make sure @white238 has a say in this.

src/common/vector_expression.hpp
white238: I spoke to @joshessman-llnl and after futher investigation this (and the note i mentioned that didn't get implemented) can be removed.  Small performance hits should be taken for such a large readability/productivity boosts.  And its not even clear this has a performance hit overall.  Also this is a very small section of the total runtime.

white238 (approved):

jamiebramwell (approved):

joshessman-llnl:
LGTM

white238:
Azure doesn't seem to have reported and/or github doesn't seem to be receiving the last commit's status.  I am attempting to rerun the specific branch/sha through azure directly but honestly never hit this before so it may not work.  If this doesnt work make a small change to a random file to retrigger.

white238:
I win!

PR #166, opened 2020-09-01
fix: 10 more minutes for full dependency build

joshessman-llnl:
The dependency stage of the build (unrelated to the Lassen/CUDA stuff or the devtools) is now failing on Quartz as it appears to be taking ~30 seconds longer than usual.  It finishes the build but is [killed](https://lc.llnl.gov/gitlab/smith/serac/-/jobs/92528#L242) when setting permissions.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -54,7 +54,7 @@ stages:
     - *run_update_uberenv
     - EXEC_PREFIX="srun ${SLURM_ACCOUNT}"
     #DEPENDENCIES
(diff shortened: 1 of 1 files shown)


joshessman-llnl:
LGTM

jamiebramwell (approved):

white238 (approved):
Funnily enough I moved over work I did in Axom (like a month ago) to skip the permission/group fixing on non-shared area builds in my current PR to add netcdf.  But its good to do 30 anyways here.

PR #167, opened 2020-09-01
Add netcdf and fix static linking

### LINKED ISSUE #164: Add netcdf and petsc build options
Some users want the netcdf and petsc capabilities in mfem. We should add this as an optional build configuration.

white238:
Had enough fun with netcdf.  I'll do Petsc in a follow up PR.


scripts/uberenv/packages/netcdf-c/package.py
```diff
@@ -0,0 +1,237 @@
...
+        if self.spec.satisfies('@4.3.2:'):
+            config_args += self.enable_or_disable('jna')
+
+        # Starting version 4.1.3, --with-hdf5= and other such configure options
+        # are removed. Variables CPPFLAGS, LDFLAGS, and LD_LIBRARY_PATH must be
+        # used instead.
+        hdf5_hl = self.spec['hdf5:hl']
+        cppflags.append(hdf5_hl.headers.cpp_flags)
+        ldflags.append(hdf5_hl.libs.search_flags)
+        if hdf5_hl.satisfies('~shared'):
```
white238, line 184: I added these two lines (184,185).. I will be creating a spack PR soon. Everything else is mainline spack.

Side note: the comment above this makes me angry.

scripts/uberenv/packages/axom/package.py
```diff
@@ -51,7 +51,7 @@ class Axom(CMakePackage, CudaPackage):
     homepage = "https://github.com/LLNL/axom"
     git      = "https://github.com/LLNL/axom.git"
 
-    version('master', branch='master', submodules=True)
+    version('main', branch='main', submodules=True)
```
white238, line 54: These changes are the from the new Axom package...

scripts/uberenv/packages/mfem/package.py
```diff
@@ -369,21 +365,27 @@ def find_optional_library(name, prefix):
         cxxflags = spec.compiler_flags['cxxflags']
 
         if cxxflags:
+            # Add opt/debug flags if they are not present in global cxx flags
```
white238, line 368: All changes except the added hdf5 lines are just updating the package from the spack repo.

scripts/uberenv/packages/mfem/package.py
```diff
@@ -497,10 +499,16 @@ def find_optional_library(name, prefix):
                 ld_flags_from_dirs([spec['gslib'].prefix.lib], ['gs'])]
 
         if '+netcdf' in spec:
+            netcdf_libdirs = [spec['netcdf-c'].prefix.lib]
```
white238: These were required if linking a static netcdf into mfem, it required the transitive hdf5 brought along for the ride. Spack PR coming soonish.

scripts/uberenv/packages/serac/package.py
```diff
@@ -99,11 +100,12 @@ class Serac(CMakePackage):
     depends_on('py-sphinx', when="+devtools")
 
     # Libraries that support +debug
-    debug_deps = ["mfem@4.1.0p1~shared+hypre+metis+superlu-dist+lapack+mpi",
+    debug_deps = ["mfem@4.1.0p1~shared+metis+superlu-dist+lapack+mpi",
```
white238: hypre variant was removed because it did nothing.

scripts/uberenv/spack_configs/toss_3_x86_64_ib/packages.yaml
```diff
@@ -10,123 +10,140 @@ packages:
       mpi: [mvapich2]
       gl: [opengl]
       glu: [openglu]
-
   opengl:
-    paths:
-      opengl@1.7.0: /usr
-    buildable: False
+    buildable: false
```
white238, line 14: This is just an update to the syntax of this file. @becker33 gave me a script to do this so I'll run the rest through it.  Thanks Greg!

scripts/llnl/common_build_functions.py
```diff
@@ -295,17 +295,27 @@ def set_group_and_perms(directory):
...
-    print "[changing group to smithdev]"
-    sexe("chgrp -f -R smithdev %s" % (directory),echo=True,error_prefix="WARNING:")
-    # change group perms to rwX
-    print "[changing perms for smithdev members to rwX]"
-    sexe("chmod -f -R g+rwX %s" % (directory),echo=True,error_prefix="WARNING:")
-    # change perms for all to rX
-    print "[changing perms for all users to rX]"
-    sexe("chmod -f -R a+rX %s" % (directory),echo=True,error_prefix="WARNING:")
-    print "[done setting perms for: %s]" % directory
+    skip = True
```
white238, line 298: Tested this skipping long enough in Axom so I am bringing it over.

scripts/uberenv/packages/mfem/package.py
white238: I'd like Greg to look at this section before i throw it up on a Spack PR.  Since there are like 3 ways to build this list in the mfem package.

becker33: @white238 I think that mostly looks fine, I'd prefer if we could avoid the hardcoding of the library names. What about this?
```
netcdf_libraries = spec['netcdf-c'].libs
if spec['hdf5'].satisfies('~shared'):
    netcdf_libraries += spec['hdf5'].libs
...
    ld_flags_from_library_list(netcdf_libraries)
```

If what I suggested doesn't work (modulo obvious typos) then I'm fine with what you have here.

white238: implemented a modified version of greg's..

white238:
LGTM

scripts/uberenv/packages/axom/package.py
```diff
@@ -77,16 +79,16 @@ class Axom(CMakePackage, CudaPackage):
 
     variant("mfem",     default=False, description="Build with mfem")
     variant("hdf5",     default=True, description="Build with hdf5")
-    variant("lua",      default=False, description="Build with Lua")
+    variant("lua",      default=True, description="Build with Lua")
```
jamiebramwell, line 82: 👍

scripts/uberenv/packages/netcdf-c/package.py
```diff
@@ -0,0 +1,237 @@
...
+        if self.spec.satisfies('@4.4:'):
+            if '+mpi' in self.spec:
+                config_args.append('--enable-parallel4')
+            else:
+                config_args.append('--disable-parallel4')
+
+        if self.spec.satisfies('@4.3.2:'):
+            config_args += self.enable_or_disable('jna')
+
+        # Starting version 4.1.3, --with-hdf5= and other such configure options
```
jamiebramwell, line 178: 👎

scripts/uberenv/packages/serac/package.py
```diff
@@ -89,8 +86,12 @@ class Serac(CMakePackage, CudaPackage):
     variant("devtools",  default=False,
             description="Build development tools (such as Sphinx, AStyle, etc...)")
 
+    variant('caliper', default=False, 
+            description='Build with hooks for Caliper performance analysis')
     variant('glvis', default=False,
             description='Build the glvis visualization executable')
+    variant('netcdf', default=False,
```
jamiebramwell: How long does this take to build? I'm thinking this would be nice to have on by default.

jamiebramwell (approved):

scripts/uberenv/packages/serac/package.py
white238: on toss3... clang took 14min, gcc took 16min.  There are a whole set of libraries that crept in.  I can look to see if i can utilize some system libraries to keep them down.  I'll look into that after I have the blueos build working again.

joshessman-llnl: What broke on blueos?

scripts/uberenv/packages/netcdf-c/package.py
joshessman-llnl: "Replace use of --with-hdf5= and other such configure options that violate conventions and causes build problems. Set environment variables CPPFLAGS, LDFLAGS, and LD_LIBRARY_PATH instead, before running configure script. "

...why

joshessman-llnl (approved):

scripts/uberenv/packages/axom/package.py
```diff
@@ -96,15 +98,13 @@ class Axom(CMakePackage, CudaPackage):
...
-    depends_on("conduit~shared~python+hdf5", when="+hdf5~python")
-    depends_on("conduit~shared~python~hdf5", when="~hdf5~python")
+    depends_on("conduit+python", when="+python")
+    depends_on("conduit~python", when="~python")
+    depends_on("conduit+hdf5", when="+hdf5")
+    depends_on("conduit~hdf5", when="~hdf5")
 
     # HDF5 needs to be the same as Conduit's
-    depends_on("hdf5@1.8.19:1.8.999~mpi~cxx~shared~fortran", when="+hdf5")
+    depends_on("hdf5@1.8.19:1.8.999~cxx~shared~fortran", when="+hdf5")
```
jandrej, line 107: I don't understand the fixed version here. Does spack not prevent you from using axom with conduit when you use out of sync hdf5? e.g.
```
spack install axom ^hdf5@1.12~cxx~shared~fortran
```
should install conduit with hdf5 1.12 (if conduit allows it) or fail (if conduit restricts the version).

jandrej (approved):

scripts/uberenv/packages/serac/package.py
white238: * mvapich compiler wrapper couldnt compile anything
* the python path change caused the devtools variant to go crazy because i was comparing the root path of python and doxygen.

scripts/uberenv/packages/axom/package.py
white238: It has to do with how Spack locks in the concretization on a first come first serve nature.  Axom will get their first and default to the highest version, then conduit restricts that to the lower version and it fails.  If you don't all agree then you can't install both at the same time with one command like "spack install axom@develop"

scripts/uberenv/packages/axom/package.py
adrienbernede: I was corrected in a Spack PR when changing "master" for "main" in the first argument: "master" is a specific value for spack. In order to preserve the original behavior (master is always newer that a release), then the correct syntax is:
```suggestion
    version('master', branch='main', submodules=True)
```

adrienbernede: That’s definitely messy, and I hope they will soon put "main" at the same level as "master".

scripts/uberenv/packages/axom/package.py
adrienbernede: @jandrej for the record:
https://github.com/LLNL/serac/pull/48#discussion_r409999658

However, there might be a change of strategy coming... ;)

scripts/uberenv/packages/axom/package.py
```diff
@@ -524,12 +536,17 @@ def cmake_args(self):
...
         else:
             options.append('-DENABLE_TESTS=ON')
-        src_dir = os.path.abspath(pjoin(self.stage.source_path, "src"))
-        options.append(src_dir)
+
+        if "+shared" in spec:
+            options.append('-DBUILD_SHARED_LIBS=ON')
+        else:
+            options.append('-DBUILD_SHARED_LIBS=OFF')
+
```
adrienbernede, line 549: Should we consider putting all the CMake arguments in the host-config file?

FYI, I am asking because this is what I had to do on Umpire/RAJA/CHAI, so I made the host-config file name unique using the spack spec hash. 
Pros:
- Reusing the host-config file then leads to the exact same build.
- All the CMake parameters are handle at the same level in the Spack package.

Off-topic though...

scripts/uberenv/packages/serac/package.py
jamiebramwell: If it takes that long I think we should leave it as optional.

scripts/uberenv/packages/axom/package.py
adrienbernede: Oops, that’s Axom...

scripts/uberenv/packages/serac/package.py
white238: That was total time.  I havent timed netcdf by itself. I also haven't tested what it was before the change since I ran into a rather bad Spack bug.

scripts/uberenv/spack_configs/blueos_3_ppc64le_ib_p9/devtools/packages.yaml
```diff
@@ -22,86 +21,99 @@ packages:
...
+    buildable: false
   fujitsu-mpi:
-    buildable: False
+    buildable: false
   mpilander:
-    buildable: False
+    buildable: false
   mpt:
-    buildable: False
+    buildable: false
```
adrienbernede:
```
mpi:
  buildable: false
```
?

scripts/uberenv/spack_configs/toss_3_x86_64_ib/packages.yaml
adrienbernede: I want this script!

adrienbernede (approved):
Only non-critical comments.

scripts/uberenv/packages/serac/package.py
jamiebramwell: Ahhh OK. Then I think it's a good idea to have it required.

scripts/uberenv/spack_configs/blueos_3_ppc64le_ib_p9/devtools/packages.yaml
white238: Apparently mpt is another mpi implementation.  Are these even needed at this point? half my spack configs dont list them.  It could just be a relic of a bygone era

adrienbernede: Sorry, I meant Spack now supports "buildable: false" at the provider level.
So you could replace all this by the snippet above.

adrienbernede: It’s still useful to prevent building any mpi provider, since that’s the only way to force spack to use the one specified as external or fail, _in any circumstances_.

white238: woo! cleaning up spack configs even more! they are almost readable at this point.

scripts/uberenv/packages/axom/package.py
white238: I actually got a fix from greg in one of my PR's to add main as a known branch in Spack.So it should be good now.

scripts/uberenv/packages/axom/package.py
white238: So i generally put things in the host-config.. but I felt this one leaned more towards the individual build option then needing to be in there for reproducibility.

scripts/uberenv/packages/netcdf-c/package.py
white238: I cannot comprehend the logic of whoever did that in netcdf... it's actively antagonistic to their users and goes against all general autoconf practices.

scripts/uberenv/packages/serac/package.py
```diff
@@ -266,7 +270,9 @@ def hostconfig(self, spec, prefix, py_site_pkgs_dir=None):
             cfg.write(cmake_cache_entry("CMAKE_CUDA_COMPILER",
                                         cudacompiler))
 
-            if not spec.satisfies('cuda_arch=none'):
+            if spec.satisfies('cuda_arch=none'):
```
white238, line 273: Added a comment because I forget things...

scripts/uberenv/spack_configs/toss_3_x86_64_ib/packages.yaml
white238: I emailed this earlier to a few people, @adrienbernede included..

scripts/uberenv/spack_configs/toss_3_x86_64_ib/packages.yaml
```diff
@@ -10,123 +10,134 @@ packages:
...
+      prefix: /usr/tce/packages/python/python-3.8.2
   py-sphinx:
-    version: [3.0.0]
-    paths:
-      py-sphinx: /usr/WS2/smithdev/devtools/toss_3_x86_64_ib/latest/py-sphinx-3.0.0
-    buildable: False
+    version: [2.2.0]
+    buildable: false
+    externals:
+    - spec: py-sphinx@2.2.0
```
white238, line 142: I am a bit concerned on the multiple versions sphinx we have now... though i do all my docs building through readthedocs so maybe it doesn't actually matter.

tests/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ if (ENABLE_GTEST)
         serac_linelastic_solver.cpp
         serac_quasistatic_solver.cpp
         serac_thermal_solver.cpp
-	serac_thermal_structural_solver.cpp
+        serac_thermal_structural_solver.cpp
```
white238, line 20: There was a tab here!! A TAB!!!! (screams in horror)

tests/serac_cuda_smoketest.cpp
```diff
@@ -1,9 +1,33 @@
+#include <cuda_runtime.h>
 #include <gtest/gtest.h>
 
 #include <algorithm>
 
 void vector_add(float* out, float* a, float* b, int n);
 
+TEST(cuda_smoketest, cuda_version)
```
white238, line 8: Another test to tell me when i forgot to load the right CUDA module..... that definitely hasn't happend to me twice today...

white238: I think I am going to push this up to BLT because its amazingly helpful

white238:
LGTM

white238:
Well this PR got away from me.  But it is done (i hope)

scripts/uberenv/spack_configs/blueos_3_ppc64le_ib_p9/packages.yaml
```diff
@@ -46,110 +44,103 @@ packages:
...
-      - spec: mvapich2%clang
-        prefix: /usr/tce/packages/mvapich2/mvapich2-2020.01.09-clang-ibm-2019.10.03/
-      - spec: mvapich2%xl@coral
-        prefix: /usr/tce/packages/mvapich2/mvapich2-2020.01.09-xl-2020.06.25
-    buildable: False
+    buildable: false
   netlib-lapack:
     externals:
-      # This is the one that has the trailing underscores, needed for MFEM compat
+      # This is the one that has the trailing underscores in name mangling, needed for MFEM compat
```
joshessman-llnl, line 112: 👍

joshessman-llnl (approved):

jamiebramwell:
LGTM

PR #168, opened 2020-09-01
Add a DSuperLU solver option

jamiebramwell:
This is a work-in-progress toward enabling the direct solver DSuperLU for linear solves. Comments and suggestions welcome!

### DIFF
--- a/src/solvers/equation_solver.cpp
+++ b/src/solvers/equation_solver.cpp
@@ -15,34 +15,74 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_
 {
   // Preconditioner configuration is too varied, maybe a PrecondParams is needed?
   // Maybe a redesign to better support custom preconditioners as well
-  switch (lin_params.lin_solver) {
-    case LinearSolver::CG:
-      iter_lin_solver_ = std::make_unique<mfem::CGSolver>(comm);
-      break;
-    case LinearSolver::GMRES:
-      iter_lin_solver_ = std::make_unique<mfem::GMRESSolver>(comm);
-      break;
-    case LinearSolver::MINRES:
-      iter_lin_solver_ = std::make_unique<mfem::MINRESSolver>(comm);
-      break;
-    default:
-      SLIC_ERROR("Linear solver type not recognized.");
-      exitGracefully(true);
-  }
-  iter_lin_solver_->SetRelTol(lin_params.rel_tol);
-  iter_lin_solver_->SetAbsTol(lin_params.abs_tol);
-  iter_lin_solver_->SetMaxIter(lin_params.max_iter);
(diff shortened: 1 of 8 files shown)


src/solvers/equation_solver.cpp
```diff
@@ -15,28 +15,38 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_
...
-      exitGracefully(true);
-  }
-  iter_lin_solver_->SetRelTol(lin_params.rel_tol);
-  iter_lin_solver_->SetAbsTol(lin_params.abs_tol);
-  iter_lin_solver_->SetMaxIter(lin_params.max_iter);
-  iter_lin_solver_->SetPrintLevel(lin_params.print_level);
+  
+  if (lin_params.lin_solver == LinearSolver::SuperLU) {
+    lin_solver_ = std::make_unique<mfem::SuperLUSolver>(comm);
+  } else {
```
joshessman-llnl, line 21: I think since the LinearSolverParams don't get used here the existing design should be reconsidered.  One option would be to have constructors corresponding to each configuration (just iterative, just SuperLU, iterative + newton, superLU + newton), but that doesn't scale too well.  Another option might be to just have the physics solvers store a reference to an mfem::Solver and define Solver wrappers for each component (e.g. an iterative solver class that takes a LinearSolverParams by ctor arg, and a nonlinear solver class that takes a NonlinearParams by ctor arg).

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -173,13 +173,13 @@ void NonlinearSolidSolver::completeSetup()
...
-    solver_.solver().iterative_mode = true;
+    solver_.nonlinearSolver()->iterative_mode = true;
     nonlinear_oper_                 = std::make_unique<NonlinearSolidQuasiStaticOperator>(std::move(H_form));
     solver_.SetOperator(*nonlinear_oper_);
   } else {
-    solver_.solver().iterative_mode = false;
+    solver_.nonlinearSolver()->iterative_mode = false;
     timedep_oper_                   = std::make_unique<NonlinearSolidDynamicOperator>(
-        std::move(H_form), std::move(S_form), std::move(M_form), ess_bdr_, solver_.solver(), lin_params_);
+        std::move(H_form), std::move(S_form), std::move(M_form), ess_bdr_, *solver_.nonlinearSolver(), lin_params_);
```
joshessman-llnl, line 182: The reason we expose the solvers directly is mostly to set the iterative mode.  We could use the EquationSolver interface in the NonlinearSolidDynamicOperator (instead of the underlying solver) if we expose the "getConverged" method.  Would it be reasonable to associate the timestep method with the EquationSolver object instead of the physics solver?  It could still be exposed via an accessor and would keep all equation solver configuration within the same object.

PR #169, opened 2020-09-02
Encapsulate all boundary conditions in a separate class

joshessman-llnl:
In the interest of minimizing PR size this is just a first iteration demonstrating some useful functionality that becomes possible when boundary conditions are stored in their own class. 

This PR also removes the nonlinear-specific logic in the solid solver that sets the full list of DOFs in the NLForm and replaces it with some elementary boundary condition operations that were already in use - zeroing out the entries of a vector corresponding to the DOF indices, eliminating from a matrix, etc.  The use of a single class to manage DOFs also simplifies these operations by maintaining a single list of DOFs for all Dirichlet BCs, removing the need to iterate over them for each operation.

Accessors to the underlying lists of BCs are still used liberally but these will be eventually removed.

This PR is another step toward the compositional approach described in #118.


src/common/boundary_condition.hpp
```diff
@@ -245,6 +241,116 @@ class BoundaryCondition {
   mutable std::unique_ptr<mfem::HypreParMatrix> eliminated_matrix_entries_;
 };
 
+class BoundaryConditionManager {
```
white238: Can you move this to its own file?

src/common/boundary_condition_manager.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * @brief Set the natural boundary conditions from a list of boundary markers and a coefficient
+   *
+   * @param[in] nat_bdr The set of mesh attributes denoting a natural boundary
+   * @param[in] nat_bdr_coef The coefficient defining the natural boundary function
+   * @param[in] state The finite element state to which the BC should be applied
+   * @param[in] component The component to set (-1 implies all components are set)
+   */
+  void addNatural(const std::set<int>& nat_bdr, serac::GeneralCoefficient nat_bdr_coef, FiniteElementState& state,
```
jamiebramwell: It would be nice to append each boundary condition by a name (e.g. "pressure", "traction", "symmetry" ) set by either the user or the physics module. We could have a third `addGenericBoundary` call to catch the ones that don't fall exactly under Dirichlet or Neumann categories.

jamiebramwell: (This extra term would probably live in the `BoundaryCondition` class)

src/solvers/base_solver.cpp
```diff
@@ -34,34 +34,18 @@ BaseSolver::BaseSolver(std::shared_ptr<mfem::ParMesh> mesh, int n, int p) : Base
...
-                    [attr](auto&& existing_bc) { return existing_bc.markers()[attr - 1] == 1; })) {
-      SLIC_WARNING("Multiple definition of essential boundary! Using first definition given.");
-      bc.removeAttr(attr);
-    }
-  }
-
-  bc.setTrueDofs(state);
-
-  ess_bdr_.emplace_back(std::move(bc));
+  bcs_.addEssential(ess_bdr, ess_bdr_coef, state, component);
```
jamiebramwell: I think we can delete this functionality from the base class and just put `bcs_.addEssential(...)` etc. in the individual solvers.

jamiebramwell: This is a step toward better following of [this guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i25-prefer-abstract-classes-as-interfaces-to-class-hierarchies).

src/common/boundary_condition_manager.hpp
```diff
@@ -0,0 +1,137 @@
...
+
+  /**
+   * @brief The vector of essential boundary conditions
+   */
+  std::vector<BoundaryCondition> ess_bdr_;
+
+  /**
+   * @brief The vector of natural boundary conditions
+   */
+  std::vector<BoundaryCondition> nat_bdr_;
```
jamiebramwell, line 272: I think a third container of "other_bdr_" containing the non-essential and natural ones is appropriate here. Ideally, all boundary condition data for all solvers would live in this class.

src/solvers/nonlinear_solid_operators.hpp
```diff
@@ -64,6 +65,11 @@ class NonlinearSolidQuasiStaticOperator : public mfem::Operator {
    * @brief The linearized jacobian at the current state
    */
   mutable std::unique_ptr<mfem::Operator> Jacobian_;
+
+  /**
+   * @brief The fixed boudnary degrees of freedom
```
jamiebramwell: This is all of the BC data - not just the essential part.

tests/serac_boundary_cond.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+  MPI_Barrier(MPI_COMM_WORLD);
+  constexpr int      N    = 15;
+  constexpr int      ATTR = 1;
+  mfem::Mesh         mesh(N, N, mfem::Element::TRIANGLE);
+  mfem::ParMesh      par_mesh(MPI_COMM_WORLD, mesh);
+  FiniteElementState state(par_mesh);
+
+  for (int i = 0; i < par_mesh.GetNE(); i++) {
+    par_mesh.SetAttribute(i, ATTR);
```
jamiebramwell: I believe this changes the attribute on the volumetric element - not the boundary. I think you want something like this
```
for (int i = 0; i < num_boundary_elements; i++) {
  boundary[i]->SetAttribute(ATTR);
}
````
(only using appropriate accessor functions)

**Review by jamiebramwell — COMMENTED** [review]
Looks good! My major comments are more cleaning: specifically adding "extra" BCs to the manager and moving the boundary condition management stuff out of the `BaseSolver` class.

File: src/solvers/base_solver.cpp (hunk 3 above)
Comment (joshessman-llnl) [reply src/solvers/base_solver.cpp]: I was thinking that eventually this would be initialized separately (as in https://github.com/LLNL/serac/issues/118#issuecomment-668805621) but I agree that it can be removed from the base in this PR

File: src/common/boundary_condition.hpp [hunk 7]
```diff
@@ -48,6 +48,24 @@ class BoundaryCondition {
...
+  /**
+   * @brief Sets the tag for the BC
+   * @tparam Tag The template type for the tag (label)
+   * @param[in] The new tag
+   * @pre Template type "Tag" must be an enumeration
+   */
+  template <typename Tag>
+  void setTag(const Tag tag)
+  {
+    static_assert(std::is_enum_v<Tag>, "Only enumerations can be used to tag a boundary condition.");
```
jamiebramwell, line 78: Ha! Modern C++ in fact had what we needed!

src/common/boundary_condition_manager.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+void BoundaryConditionManager::addTrueDofs(const mfem::Array<int>& true_dofs, serac::GeneralCoefficient ess_bdr_coef,
+                                           int component)
+{
+  ess_bdr_.emplace_back(ess_bdr_coef, component, true_dofs);
+  all_dofs_valid_ = false;
+}
+
+void BoundaryConditionManager::updateAllDofs() const
```
jamiebramwell: I think `updateAllEssentialDofs` would be a more descriptive name.

src/common/boundary_condition_manager.hpp
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  /**
+   * @brief Eliminates all essential BCs from a matrix
+   * @param[inout] matrix The matrix to eliminate from, will be modified
+   * @return The eliminated matrix entries
+   * @note The sum of the eliminated matrix and the modified parameter is
+   * equal to the initial state of the parameter
+   */
+  std::unique_ptr<mfem::HypreParMatrix> eliminateAllFromMatrix(mfem::HypreParMatrix& matrix) const
```
jamiebramwell: I would rename this `eliminateAllEssentialDofsFromMatrix`. I know it's verbose, but these calls are tricky enough I think it's worth it.

src/common/boundary_condition_manager.hpp
```diff
@@ -0,0 +1,299 @@
...
+   * @brief View over all "other"/generic boundary conditions with a specific tag
+   * @tparam Tag The template type for the tag
+   * @param tag The tag to filter with
+   * @pre Tag must be an enumeration type
+   */
+  template <typename Tag>
+  const auto genericsWithTag(const Tag tag)
+  {
+    static_assert(std::is_enum_v<Tag>, "Only enumerations can be used to tag a boundary condition.");
+    return FilterView(other_bdr_.begin(), other_bdr_.end(), [tag](const auto& bc) { return bc.tag() == tag; });
```
jamiebramwell: Nice!

src/common/boundary_condition_manager.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+void BoundaryConditionManager::addNatural(const std::set<int>& nat_bdr, serac::GeneralCoefficient nat_bdr_coef,
+                                          const int component)
+{
+  nat_bdr_.emplace_back(nat_bdr_coef, component, nat_bdr, num_attrs_);
+  all_dofs_valid_ = false;
+}
+
+void BoundaryConditionManager::addTrueDofs(const mfem::Array<int>& true_dofs, serac::GeneralCoefficient ess_bdr_coef,
```
jamiebramwell: `addEssentialTrueDofs`?

samuelpmishLLNL (approved):
Sorry for taking so long to look at this, I would like to immediately use a number of these features in the time-dependent boundary condition branch!

tests/serac_boundary_cond.cpp
```diff
@@ -0,0 +1,101 @@
...
+  for (int i = 0; i < N; i++) {
+    bcs.addGeneric({}, coef, TestTag::Tag1, 1);
+    bcs.addGeneric({}, coef, TestTag::Tag2, 1);
+  }
+
+  int bcs_with_tag1 = 0;
+  for (const auto& bc : bcs.genericsWithTag(TestTag::Tag1)) {
+    EXPECT_EQ(TestTag::Tag1, bc.tag());
+    bcs_with_tag1++;
+  }
```
samuelpmishLLNL: won't this also match an unrelated enum with the same underlying values?

jamiebramwell: Oh good point. Is there any way to check if the registered BC makes sense with the correct associated solver?

joshessman-llnl: I've added a check to make sure that a tag is compared with the correct type (will warn and return false when a different enum is used)

samuelpmishLLNL:
We discussed a similar idea before Josh showed up, and my basic idea was something a little bit different (promoting different types of boundary conditions to different types). Two skeleton implementations of the container class are shown here (one with std::variant, one with std::tuple)
https://godbolt.org/z/P368e6

Having different types allows the boundary condition implementation to limit the scope of its application, and each combination of physics can clearly describe the list of supported boundary conditions.

jamiebramwell:
> 
> 
> We discussed a similar idea before Josh showed up, and my basic idea was something a little bit different (promoting different types of boundary conditions to different types). Two skeleton implementations of the container class are shown here (one with std::variant, one with std::tuple)
> https://godbolt.org/z/P368e6
> 
> Having different types allows the boundary condition implementation to limit the scope of its application, and each combination of physics can clearly describe the list of supported boundary conditions.

Ahhh yes. I think this goes along with the physics-specific parameters and material models. I think a physics-agnostic part of the "manager" is nice for stuff like dirichlet operations, general bookeeping, and a unified lookup API, but it would be nice to include strong scopes.

I think this is worth merging as-is since it is still a significant improvement over what's there. Maybe the next task should be coming up with ideas for these physics-specific data types.

jamiebramwell (approved):

samuelpmishLLNL:
> Ahhh yes. I think this goes along with the physics-specific parameters and material models. I think a physics-agnostic part of the "manager" is nice for stuff like dirichlet operations, general bookeeping, and a unified lookup API, but it would be nice to include strong scopes.

That was what I was trying to hint at in the godbolt link: all specializations of boundaryconditionmanagers support the same API for dirichlet/neumann, since they are common to every PDE. However, by explicitly naming the other supported bc types, we can prevent a user (at compile time) from trying to add a radiation boundary condition to a displacement field, for example.

> I think this is worth merging as-is since it is still a significant improvement over what's there. 

I agree! I'm a little bit roadblocked on another branch trying to figure out how to do some of the things in this PR, so the sooner the better in my book.

joshessman-llnl:
LGTM

joshessman-llnl:
> Maybe the next task should be coming up with ideas for these physics-specific data types.

As we gain more customers/solvers I think we'll be able to do this more intelligently - in particular I'm wondering if we'll see any common operations, but even the existing stuff mentioned above plus the ability to iterate over conditions of a specific type is definitely an improvement.

white238 (approved):

joshessman-llnl:
LGTM

PR #173, opened 2020-09-03
Add an option for direct solver with SuperLU

joshessman-llnl:
Extension of #168 

This avoids `dynamic_cast` with a more type-safe `variant`, at the cost of possibly-less-clear usage of `std::visit`.


tests/serac_thermal_solver.cpp
```diff
@@ -57,11 +57,7 @@ TEST(thermal_solver, static_solve)
 
   // Define the linear solver params
   serac::LinearSolverParameters params;
-  params.rel_tol     = 1.0e-6;
-  params.abs_tol     = 1.0e-12;
-  params.print_level = 0;
-  params.max_iter    = 100;
-  params.lin_solver  = LinearSolver::CG;
+  params.lin_solver = LinearSolver::SuperLU;
```
jamiebramwell, line 60: 👍

src/solvers/nonlinear_solid_solver.cpp
```diff
@@ -173,13 +173,13 @@ void NonlinearSolidSolver::completeSetup()
 
   // Set the MFEM abstract operators for use with the internal MFEM solvers
   if (timestepper_ == serac::TimestepMethod::QuasiStatic) {
-    solver_.solver().iterative_mode = true;
-    nonlinear_oper_                 = std::make_unique<NonlinearSolidQuasiStaticOperator>(std::move(H_form));
+    solver_.nonlinearSolver().iterative_mode = true;
+    nonlinear_oper_                          = std::make_unique<NonlinearSolidQuasiStaticOperator>(std::move(H_form));
```
jamiebramwell, line 177: These two line giant spaces are making me twitch. I'll see if there's a way to get clang-format to suppress this.

src/solvers/equation_solver.hpp
```diff
@@ -49,52 +49,93 @@ class EquationSolver : public mfem::Solver {
...
   /**
-   * Returns the underlying linear solver object, even if the class instance
-   * has been configured as a nonlinear solver
-   * @return The underlying linear solver
+   * Returns the underlying linear solver object
+   * @return A non-owning reference to the underlying linear solver
    */
-  mfem::IterativeSolver&       linearSolver() { return *iter_lin_solver_; }
-  const mfem::IterativeSolver& linearSolver() const { return *iter_lin_solver_; }
+  mfem::Solver& linearSolver()
```
jamiebramwell, line 88: Nice!

src/solvers/equation_solver.hpp
```diff
@@ -49,52 +49,93 @@ class EquationSolver : public mfem::Solver {
...
 private:
-  std::unique_ptr<mfem::IterativeSolver>                iter_lin_solver_;
-  std::optional<std::unique_ptr<mfem::IterativeSolver>> nonlin_solver_;
-  std::unique_ptr<mfem::Solver>                         prec_;
+  /**
+   * @brief Builds an iterative solver given a set of linear solver parameters
+   * @param[in] comm The MPI communicator object
+   * @param[in] lin_params The parameters for the linear solver
+   */
+  static std::unique_ptr<mfem::IterativeSolver> buildIterLinSolver(MPI_Comm                      comm,
```
jamiebramwell: Style issue: I generally like verbose names for methods. While this is fairly obvious, I think `buildIterativeLinearSolver` would be a more descriptive name.

src/solvers/equation_solver.cpp
```diff
@@ -13,39 +13,97 @@ namespace serac {
 EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_params,
                                const std::optional<NonlinearSolverParameters>& nonlin_params)
 {
-  // Preconditioner configuration is too varied, maybe a PrecondParams is needed?
-  // Maybe a redesign to better support custom preconditioners as well
+  if (lin_params.lin_solver == LinearSolver::SuperLU) {
```
jamiebramwell, line 16: Breaking this up cleaning things up nicely.

src/solvers/equation_solver.cpp
```diff
@@ -13,39 +13,97 @@ namespace serac {
...
+  newton_solver->SetAbsTol(nonlin_params.abs_tol);
+  newton_solver->SetMaxIter(nonlin_params.max_iter);
+  newton_solver->SetPrintLevel(nonlin_params.print_level);
+  return newton_solver;
+}
+
+void EquationSolver::SetOperator(const mfem::Operator& op)
+{
+  if (nonlin_solver_) {
+    nonlin_solver_->SetOperator(op);
```
jamiebramwell: Since we are implicitly using the `SetOperator` method for the related linear solver in this call, I don't think this would work for a direct linear solver underneath a Newton solver. It's probably a good idea to test this in one of the nonlinear elasticity tests.

jamiebramwell (commented):
I think we should add a test to verify SuperLU works under a nonlinear Newton solver. Otherwise this looks good! Thanks for taking it over.

jamiebramwell:
Adding @salazardetroya since you are interested in this feature.

src/solvers/equation_solver.cpp
```diff
@@ -13,39 +13,97 @@ namespace serac {
 EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_params,
                                const std::optional<NonlinearSolverParameters>& nonlin_params)
 {
-  // Preconditioner configuration is too varied, maybe a PrecondParams is needed?
-  // Maybe a redesign to better support custom preconditioners as well
+  if (lin_params.lin_solver == LinearSolver::SuperLU) {
+    lin_solver_ = std::make_unique<mfem::SuperLUSolver>(comm);
```
salazardetroya, line 17: Adding `std::get<std::unique_ptr<mfem::SuperLUSolver>>(lin_solver_)->SetColumnPermutation(mfem::superlu::PARMETIS);` here significantly increases speed. I just tested it.

src/solvers/equation_solver.cpp
jamiebramwell: Also, if this doesn't work without a bunch of hassle (which I suspect), we can check for this incompatibility and fix this in a later PR.

tests/serac_thermal_solver.cpp
samuelpmishLLNL, line 60: It is a little bit confusing to have parameters that effectively do nothing for a given setting of "lin_solver" (e.g. I assume the value of rel_tol has no effect on SuperLU). Another option is to make each method its own type, e.g.

```cpp
namespace Method {

struct GMRES{
  double relative_tolerance;
  double absolute_tolerance;
  int max_iterations;
  int restart_iterations;
};

struct CG{
  double relative_tolerance;
  double absolute_tolerance;
  int max_iterations;
};

struct SuperLU{};

}
```

```cpp
struct EquationSolver{
...
};
```

jamiebramwell: In the future, I imagine these parameters will live in an Inlet-generated data store.

jamiebramwell: In the interest of keeping PRs small, I'm going to add an issue on this for future work.

src/solvers/equation_solver.hpp
```diff
@@ -49,52 +49,142 @@ class EquationSolver : public mfem::Solver {
...
+   * @param[in] lin_solver The base linear solver object
+   */
+  static std::unique_ptr<mfem::NewtonSolver> buildNewtonSolver(MPI_Comm                         comm,
+                                                               const NonlinearSolverParameters& nonlin_params,
+                                                               mfem::Solver&                    lin_solver);
+
+  /**
+   * @brief A wrapper class for combining a nonlinear solver with a SuperLU direct solver
+   */
+  class SuperLUNonlinearOperatorWrapper : public mfem::Operator {
```
jamiebramwell, line 123: This is kinda gross, but I can't think of a better way out of it.

src/solvers/equation_solver.cpp
```diff
@@ -13,39 +13,113 @@ namespace serac {
...
+  newton_solver->SetMaxIter(nonlin_params.max_iter);
+  newton_solver->SetPrintLevel(nonlin_params.print_level);
+  return newton_solver;
+}
+
+void EquationSolver::SetOperator(const mfem::Operator& op)
+{
+  if (nonlin_solver_) {
+    if (std::holds_alternative<std::unique_ptr<mfem::SuperLUSolver>>(lin_solver_)) {
+      superlu_wrapper_ = std::make_unique<SuperLUNonlinearOperatorWrapper>(op);
```
jamiebramwell, line 73: This is called at every timestep. Will calling `make_unique` instead of `reset` leak memory?

jamiebramwell: My googling shows it doesn't. TIL! There's probably a bunch of `reset`s that we don't need hanging around in the code due to my misunderstanding....

jamiebramwell (approved):

jamiebramwell:
@samuelpmishLLNL is there anything else in this PR that needs work? I moved your comment about breaking up the solverparams into a separate issue for the sake of keeping PRs small.

samuelpmishLLNL (approved):

joshessman-llnl:
LGTM

PR #178, opened 2020-09-04
Delete Azure and Gitlab "build_and_test" scripts

### LINKED ISSUE #176: Unify build/test scripts
We have two scripts that do basically the same thing between gitlab and azure pipelines.  These can be unified into one script.  Another, probably better, option is to remove both and get+modify the build_src.py script in Axom and have it take a host-config as a command line option to keep the single host-config per pipeline functionality.

### LINKED ISSUE #182: GitLab CI: Have lassen and quartz builds run in parallel
Using "needs" it’s possible to have a job starting as soon as possible instead on waiting for unrelated previous stages.

(see https://github.com/LLNL/Umpire/blob/ca5bb88bbd1fbfe2bbb45c184f8dc2255ed5ba69/.gitlab/lassen-templates.yml#L23)

joshessman-llnl:
Using for testing on GL

Resolves #176 
Resolves #182


joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

.gitlab-ci.yml
```diff
@@ -81,26 +73,21 @@ stages:
...
-    #TEST
-    - echo -e "Test ${CI_PROJECT_NAME}"
-    - lalloc 1 -W 5 ${LSF_ACCOUNT} scripts/gitlab/build_and_test.sh --test-only
+    - echo -e "section_start:$(date +%s):dependencies\r\e[0K
+      Build dependencies and generate host-config file (uberenv)"
+    - lalloc 1 -W 20 ${LSF_ACCOUNT} scripts/llnl/build_tpls.py --spec=${SPEC} --directory=${CI_PROJECT_NAME}
+    - echo -e "section_end:$(date +%s):dependencies\r\e[0K"
+    #BUILD + TEST
+    - echo -e "Build and test ${CI_PROJECT_NAME}"
+    - lalloc 1 -W 10 ${LSF_ACCOUNT} scripts/llnl/build_src.py
```
white238: should probably pass the host-config on this line though i guess there is only one at this point,.

scripts/llnl/build_src.py
```diff
@@ -0,0 +1,119 @@
...
+    try:
+        original_wd = os.getcwd()
+        os.chdir(repo_dir)
+        timestamp = get_timestamp()
+        # Command-line arg has highest priority
+        if opts["hostconfig"] != "":
+          hostconfig = opts["hostconfig"]
+        
+        # Then check the environment
+        elif os.environ.get("HOST_CONFIG") != None:
```
white238: my distaste for magic environment variables is high...

.gitlab/build_lassen.yml
```diff
@@ -37,5 +37,7 @@ clang_upstream_gfortran (PR build_on_lassen):
 clang_upstream_gfortran (Main build_with_deps_on_lassen):
   variables:
     COMPILER: "clang@upstream_gfortran"
-    SPEC: "@develop%${COMPILER}"
+    SPEC: "\"@develop+cuda%${COMPILER} cuda_arch=70\""
+    # The "-cuda" at the end of the host-config needs to be hardcoded for now
+    HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-${COMPILER}-cuda.cmake"
```
adrienbernede: It was on purpose that the host-config file name did not appear in CI.

Indeed it’s much easier to change the configuration/spec when not mentioning the host-config file name, which is "generated" at an intermediate step.

To do so, I simply made the assumption that the only host-config file present at the root directory was the one I want, since it has been produced in the "dependencies" step.

I hope it makes sense.

adrienbernede (commented):
I think mentioning the host-config file in CI configuration is counter productive, since it is generated.

scripts/llnl/build_src.py
adrienbernede: Don’t blame him, he ported a "feature" of my script... :)

white238: i know.. but this time it is my script its infecting =) it's my baby...

.gitlab/build_lassen.yml
white238: that is true and a good point if these scripts are only used in one thing but the being that they are used to unify all CI, building public libraries, and even some development.  They need to work in the same way across all of those workflows.

You wouldn't want a script like this "finding" a host-config in the root that you didn't expect it to. when really you wanted it to test all of the host-configs for this platform.

scripts/llnl/build_src.py
```diff
@@ -0,0 +1,119 @@
...
+            hostconfig = os.environ["HOST_CONFIG"]
+        
+        # Otherwise try to build it
+        else:
+            import socket
```
white238: Did you copy this script from axom or make a new one?  The original script built all the present host-configs for the current sys_type that was in the host-configs directory.  Which was a very useful feature that isnt here anymore.

scripts/llnl/build_src.py
adrienbernede: Get a vaccine...
In bash it was a convenience, to avoid dealing with multiple arguments, and variables needed were exposed at the beginning of the script. That’s not a very good defense, but that is to say I won’t fight to keep the magic :)

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

white238 (approved):

.gitlab/build_lassen.yml
adrienbernede: Allow me to insist:
in `.build_with_deps_blueos_3_ppc64le_ib_p9_script:`
The script calls are
```
    - lalloc 1 -W 20 ${LSF_ACCOUNT} scripts/llnl/build_tpls.py --spec=${SPEC} --directory=${CI_PROJECT_NAME}
    - lalloc 1 -W 10 ${LSF_ACCOUNT} scripts/llnl/build_src.py
```
So it seems to me `HOST_CONFIG` is not used (or by environment variable?), otherwise wouldn’t it be with `--host-config ${HOST_CONFIG}` in the second call?

I still think that _even outside of CI_ it would be useful to be able to build the code and its dependencies with a single command and only the "spec" and _not_ the host-config file name.

adrienbernede (commented):
Last attempt :)

.gitlab/build_lassen.yml
joshessman-llnl: It is used by environment variable here, but it makes sense to specify it explicitly - I'll add that in.

.gitlab-ci.yml
```diff
@@ -16,9 +16,13 @@ stages:
   - l_build_with_deps
   - q_allocate_resources
   - q_build
+  - q_build_with_deps
   - q_build_devtools
   - q_release_resources
+<<<<<<< HEAD
+=======
   - q_build_with_deps
+>>>>>>> develop
```
adrienbernede: Merge artifact

adrienbernede (changes requested):

joshessman-llnl:
LGTM

adrienbernede (approved):

.gitlab-ci.yml
```diff
@@ -59,14 +55,10 @@ stages:
...
-    - echo -e "Build ${CI_PROJECT_NAME}"
-    - ${EXEC_PREFIX} ${RESOURCES} scripts/gitlab/build_and_test.sh --build-only
-    #TEST
-    - RESOURCES="-t 5 -N 1"
-    - echo -e "Test ${CI_PROJECT_NAME}"
-    - ${EXEC_PREFIX} ${RESOURCES} scripts/gitlab/build_and_test.sh --test-only
+    #BUILD + TEST
+    - RESOURCES="-t 10 -N 1"
+    - echo -e "Build and test ${CI_PROJECT_NAME}"
+    - ${EXEC_PREFIX} ${RESOURCES} scripts/llnl/build_src.py
```
white238, line 65: host-config here for consistency...

joshessman-llnl: the quartz build_with_deps jobs can figure out their host-configs from the hostname/sys_type and don't have host-configs explicitly specified in the CI - should I add them?

white238: Is it finding it through the environment variable? Because there are two different host-configs on this system we test.

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #184, opened 2020-09-05
Fix the wsl build and clarify the docs

jamiebramwell:
I had to add the pthread library to the link line of hdf5. I'm not sure if there's a better way to fix it. Any ideas?


scripts/uberenv/packages/hdf5/package.py
```diff
@@ -0,0 +1,381 @@
...
+                extra_args.append('CXX=%s' % self.spec['mpi'].mpicxx)
+
+            if '+fortran' in self.spec:
+                extra_args.append('FC=%s' % self.spec['mpi'].mpifc)
+
+        extra_args.append('--with-zlib=%s' % self.spec['zlib'].prefix)
+ 
+        #SERAC EDIT BEGIN
+        #for some reason, it needs -pthread on WSL
+        extra_args.append('LDFLAGS=-pthread')
```
white238: @becker33 is there a better way of doing this? Adding -pthread to the global spack LDFLAGS caused problems.

jamiebramwell: I realized I tried `ldflags` instead of `ldlibs`. The latter works as a compiler-specific fix.

jamiebramwell:
LGTM

scripts/uberenv/packages/hdf5/package.py
white238: I wonder if its an ordering issue between ldflags and ldlibs, since according to the gnu docs they are both given to the compiler for the linker.  Anyways, thanks for the fix.

src/docs/index.rst
```diff
@@ -29,46 +29,31 @@ Serac is hosted on `GitHub <https://github.com/LLNL/serac>`_. Serac uses git sub
...
+cmake-based `BLT HPC build system <https://github.com/LLNL/blt>`_.
 
 Building Serac's Developer Tools
 --------------------------------
 
 .. note::
   This can be skipped if you are not doing Serac development or if you are on an LC machine.
-  We have them installed for you in a public space defined in 
-  ``host-config/<machine name>-<SYS_TYPE>-<compiler>.cmake``
+  They are installed in a public space defined in ``host-config/<machine name>-<SYS_TYPE>-<compiler>.cmake``
```
white238: Note these aren't exactly "public" they are guarded by our group.  We could on the otherhand, use Axom's devtools which are built in /collab/usr/gapps/axom and open to most of anyone that has LC access.

src/docs/index.rst
```diff
@@ -119,30 +105,53 @@ Serac.
...
+you can do so by changing the uberenv command as follows:
 
 .. code-block:: bash
 
    $ python scripts/uberenv/uberenv.py --upstream=</path/to/my/spack>/opt/spack
 
+Building Serac's Dependencies by Hand
+-------------------------------------
+
+To build Serac's dependencies by hand, a ``host-config`` (CMake configuration file) is required by the BLT
```
white238: not technically required.. "strongly encouraged" since you would have to specify all of it on the command line otherwise..

src/docs/index.rst
```diff
@@ -182,6 +194,35 @@ We provide the following useful build targets that can be run from the build dir
...
+and the following commands for Ubuntu 18.04:
+
+.. code-block:: bash
+
+   $ sudo apt-get update
+   $ sudo apt-get upgrade
+   $ sudo apt-get install g++-8 gcc-8
+   $ sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 --slave /usr/bin/g++ g++ /usr/bin/g++-8
+   $ sudo apt-get install cmake libopenblas-dev libopenblas-base mpich mesa-common-dev libglu1-mesa-dev freeglut3-dev cppcheck doxygen libreadline-dev
+   $ sudo ln -s /usr/lib/x86_64-linux-gnu/* /usr/lib
```
white238, line 219: This should probably go back to the spack group so they can handle this behind the scenes for people.  @becker33

white238 (approved):

jamiebramwell:
LGTM

joshessman-llnl (approved):
It's interesting that it broke - I was able to build on WSL ubuntu 20 just last week

joshessman-llnl:
LGTM

jamiebramwell:
> 
> 
> It's interesting that it broke - I was able to build on WSL ubuntu 20 just last week

It's entirely possible that it's some weirdness with my new laptop. 🤷

jamiebramwell:
It looks like the Quartz DAT ends at 8 PST, so I'll kick off the tests then.

PR #189, opened 2020-09-09
Fix: Subtraction is not commutative

joshessman-llnl:
I made an error copy-pasting the operator overloads that only works if subtraction is commutative, which it isn't.... 🤦 

Thanks @samuelpmishLLNL for catching this!

### DIFF
--- a/tests/expr_templates.cpp
+++ b/tests/expr_templates.cpp
@@ -67,8 +67,10 @@ TEST(expr_templates, subtraction_not_commutative)
     c[i] = i * i * i * 7 + 23;
   }
 
-  mfem::Vector result1 = a + c - b;
-  mfem::Vector result2 = c - b + a;
+  // Tests that switching the order of operations
+  // does not change the result
+  mfem::Vector result1 = a + c - b;  // Parsed as (a + c) - b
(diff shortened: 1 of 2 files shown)


samuelpmishLLNL (approved):

jamiebramwell (approved):

tests/expr_templates.cpp
```diff
@@ -54,6 +54,26 @@ TEST(expr_templates, basic_add_lambda)
...
+  mfem::Vector  b(size);
+  mfem::Vector  c(size);
+
+  for (int i = 0; i < size; i++) {
+    a[i] = i * 4 + 1;
+    b[i] = i * i * 3 + 2;
+    c[i] = i * i * i * 7 + 23;
+  }
+
+  mfem::Vector result1 = a + c - b;
```
white238: Shouldnt you need a multiply in here instead of just addition and subtraction to test for non-commutativeness (non-commutativity? non-commutatude?) Since we are testing for order of operations seems like you would need a mix of both.

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

white238 (approved):

PR #193, opened 2020-09-11
Improvements to expression templates (scalar division, removed duplicated overloads, no-alloc eval)

joshessman-llnl:
Thanks to a suggestion from @samuelpmishLLNL, the ownership tags for vector expressions were removed from the template parameter lists and scalar division was added.

To minimize the number of copy-pasted operator overloads (was 9 per op for the binary operations), `mfem::Vector` parameters were converted to templates and perfect forwarding was used to instantiate the expression templates accordingly.  This reduces the work required to add a new operation and reduces the opportunity for error when writing a new op.

Also, the `evaluate` function was overloaded to accept a preallocated `mfem::Vector` that can be used as follows:
```cpp
mfem::Vector result(size);
evaluate(a + b - 0.3 * c, result);
```

New time comparison tests were added that do a single memory allocation at the beginning of the test, and thus avoid potentially expensive `malloc`s in the "kernels".

### DIFF
--- a/src/common/expr_template_internal.hpp
+++ b/src/common/expr_template_internal.hpp
@@ -24,30 +24,31 @@
 namespace serac::internal {
 
 /**
+ * @brief Determines whether a given type should be owned by a vector expression
+ * @tparam vec The vector expression type
+ * @note A vector should be owned if it is not an (optionally const) reference
+ * to an mfem::Vector.  Note that std::remove_const does not apply here because
+ * const is not a top-level qualifier on a const reference
+ */
+template <typename vec>
+inline constexpr bool owns_v = !std::is_same_v<vec, const mfem::Vector&> && !std::is_same_v<vec, mfem::Vector&>;
+
+/**
  * @brief Type alias for what should be stored by a vector expression - an object
  * if ownership is desired, otherwise, a reference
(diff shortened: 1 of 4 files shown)


src/common/expr_template_internal.hpp
```diff
@@ -98,11 +102,40 @@ class ScalarMultOp {
...
+    if constexpr (is_divisor) {
+      return arg / scalar_;
+    } else {
+      return scalar_ / arg;
+    }
+  }
+
+private:
+  double scalar_;
+};
```
samuelpmishLLNL: `is_denominator` is clearer than `is_divisor`

samuelpmishLLNL (approved):
The use of perfect forwarding to eliminate explicitly writing different argument combinations is clever!

samuelpmishLLNL:
I like the `evaluate()` overload idea too-- I don't think we'll need to keep it in the long term, but it should give the same performance as writing an mfem::Vector constructor.

white238 (approved):
This cleans up a lot of code and makes it a lot more readable (as far as multi-level templates can be readable), thanks @joshessman-llnl !

joshessman-llnl:
LGTM

PR #199, opened 2020-09-16
Add an optional clang-tidy check

### LINKED ISSUE #98: Investigate C++ linting tools
Let's try to find a code check tool that can handle context-aware C++ style checking. [cpplint](https://github.com/cpplint/cpplint) is a possibility.

joshessman-llnl:
This updates BLT to the latest version (which has clang-tidy integration), and adds some recommended improvements to demonstrate the kinds of things that clang-tidy will issue warnings for.  The full list of warnings for the serac source is attached.

If we wanted to make this a required CI check, we would have to do some fairly significant refactoring or disable most of the checks, which sort of defeats the purpose.  Any thoughts?

[guidelines_lint.txt](https://github.com/LLNL/serac/files/5233274/guidelines_lint.txt)

Partially addresses #98


jamiebramwell:
I think this is a great idea to at least add as an option. I agree we don't want to enable it in CI until we do significant cleaning, but having the tool available is a good first step.

samuelpmishLLNL:
I think clang-tidy is a great tool, but the output file you attached doesn't seem particularly useful. I agree that we would have to use it to check for specific things, as it's not helpful to be buried in thousands of warnings (most of which we wouldn't or couldn't change)!

src/common/initialize.cpp
```diff
@@ -12,13 +12,26 @@
 
 namespace serac {
 
+std::pair<int, int> getMPIInfo(MPI_Comm comm)
+{
+  int num_procs = 0;
+  int rank      = 0;
+  if (MPI_Comm_size(comm, &num_procs) != MPI_SUCCESS) {
+    serac::exitGracefully(true);
```
white238: This is nice but we need an error message here and below, hopefully this never fails but you never know.

tests/copy_elision.cpp
```diff
@@ -29,10 +29,11 @@ std::size_t ArrayCtr::default_ = 0;
 
 ArrayCtr doubleArrayMaker()
 {
-  ArrayCtr result;
+  constexpr int size = 10;
+  ArrayCtr      result;
   // Just a random loop - if nothing was done to the array before returning,
   // the compiler might be able to do an unnamed returned value optimimization
-  for (auto i = 0; i < 10; i++) {
+  for (auto i = 0; i < size; i++) {
```
white238, line 36: Is this something clang-tidy complained about?

joshessman-llnl: Yeah, it considers this a "magic number" - though there's still a lot more floating around

samuelpmishLLNL: I agree with the idea of removing magic numbers in general, but it seems like there's a lot of cases where they are fine (or even preferable).

white238: I figured it was the case just making sure.  Is it possible to turn off magic numbers on the test directory? (don't put too much effort into this)

joshessman-llnl: Yes, I've set it up as two different check targets with the magic number warnings only enabled for the stuff in `src/`.

jamiebramwell (approved):
I think it's a good idea to have this available to optionally run.

src/infrastructure/initialize.hpp
```diff
@@ -17,6 +17,15 @@
...
 
 namespace serac {
+
+/**
+ * @brief Returns the number of processes and rank for an MPI communicator
+ * @param comm The MPI communicator to initialize with
+ * @return A pair containing the size and rank relative to the provided MPI communicator
+ * @pre The logger must be initialized (to display error messages)
+ */
+std::pair<int, int> getMPIInfo(MPI_Comm comm = MPI_COMM_WORLD);
```
samuelpmishLLNL, line 27: The `@return` comment implies that `std::get<0>(getMPIInfo(...))` is the size and `std::get<1>(getMPIInfo(...))` the rank, but I had to check the implementation to be sure. Maybe we could explicitly state

```
// A pair containing {size, rank} relative to the provided MPI communicator
```

or returning a self-documenting struct like

```
struct MPIInfo{ int size; int rank; };
```

src/physics/base_solver.hpp
```diff
@@ -179,6 +179,16 @@ class BaseSolver {
...
 
+  /**
+   * @brief Number of significant figures to output for floating-point
+   */
+  static constexpr int FLOAT_PRECISION_ = 8;
+
+  /**
+   * @brief Damping amount for ODE integration methods
+   */
+  static constexpr double DAMPING_ = 0.5;
```
samuelpmishLLNL: It looks like this might be a way of silencing a "magic number" warning, but I don't believe the 0.5 values in the related ODESolver constructors are related to damping.

samuelpmishLLNL (approved):
looks good, but I do think the DAMPING_ name is misleading, how about

```cpp
      constexpr double alpha = 0.5;
      ode_solver_ = std::make_unique<mfem::RK2Solver>(alpha);
      break;
.
.
.
      constexpr double rho = 0.5;
      ode_solver_ = std::make_unique<mfem::GeneralizedAlphaSolver>(rho);
      break;
```

or just use the default values in mfem

```cpp
      ode_solver_ = std::make_unique<mfem::RK2Solver>();
      break;
.
.
.
      ode_solver_ = std::make_unique<mfem::GeneralizedAlphaSolver>();
      break;
```

joshessman-llnl:
LGTM

PR #204, opened 2020-09-17
Add a .gitattributes file to force Unix line endings

### LINKED ISSUE #202: Add gitattributes file to control line endings
To avoid weird diffs due to line endings, we should lock it down in a `.gitattributes` file.

joshessman-llnl:
This will force all text files to use LF (Unix-style) line endings regardless of how a developer has configured their `autocrlf`.
The image below demonstrates what will happen if a file is converted to CRLF - note that a switch to CRLF no longer counts as a "change" for the purposes of adding to a commit.  

image

Resolves #202

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,20 @@
+# Enforce LF (Unix) line endings for all text files
+
+# Source code
+*.cpp text eol=lf
+*.hpp text eol=lf
+
+# Plain or formatted text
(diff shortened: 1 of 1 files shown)


samuelpmishLLNL (approved):

jamiebramwell (approved):

jamiebramwell:
LGTM

PR #207, opened 2020-09-18
Switch to Google Benchmark for expression template benchmarking

joshessman-llnl:
This PR switches the expression template timing checks to a separate executable, meaning that the tests are no longer bogged down with timing checks.  A set of utility functions were introduced to reduce the amount of copy-pasting of initialization logic in each test.

For comparison purposes I also attempted a naive MPI implementation of the expression evaluation.

Now that Google Benchmark is enabled, we could explore benchmarking other components of serac.


jamiebramwell (approved):
I think this makes a lot of sense. These tests were really slowing down my debug build setup.

CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ endif()
 set(ENABLE_ASTYLE      OFF CACHE BOOL "")
 set(ENABLE_CLANGQUERY  OFF CACHE BOOL "")
 set(ENABLE_UNCRUSTIFY  OFF CACHE BOOL "")
+set(ENABLE_BENCHMARKS  ON  CACHE BOOL "")
```
jamiebramwell: Should this be off or on by default? Any thoughts @white238 ?

white238: Off by default is better. I have ran into problems before and its just another to break on customers.

white238 (approved):

joshessman-llnl:
LGTM

src/numerics/vector_expression.hpp
```diff
@@ -93,6 +93,46 @@ void evaluate(const VectorExpr<T>& expr, mfem::Vector& result)
...
+  MPI_Comm_size(comm, &num_procs);
+  MPI_Comm_rank(comm, &rank);
+
+  double* result_arr = result;
+
+  // If the array size is not divisible by # elements, add one
+  const long long per_proc = (SIZE / num_procs) + (SIZE % num_procs != 0) ? 1 : 0;
+
+  // Truncate the number of elements for the last process
+  const long long n_entries = (rank == num_procs - 1) ? SIZE - ((num_procs - 1) * per_proc) : per_proc;
```
adrienbernede, line 122: @joshessman-llnl it seems to me `n_entries` can be negative with this sequence. It sure looks like an edge case, but:
- if SIZE = 1001 and num_procs = 100, then per_proc = 11, and the last process would receive n_entries = 1001 - 99*11 = -88.

I found it because usually when partitioning the trick to preserve load-balance is to split the residual among the first "residual"-number of procs.

I _think_ those two lines can be substituted by:
```
const long long per_proc = (SIZE / num_procs) + (SIZE % num_procs >= rank + 1) ? 1 : 0;
```

I wanted to investigate it further for fun, but I’m running out of for-fun-time.

PR #208, opened 2020-09-18
Added preliminary change of variables integrator

jamiebramwell:


src/integrators/wrapper_integrator.hpp
```diff
@@ -156,6 +157,64 @@ class MixedBilinearToNonlinearFormIntegrator : public mfem::NonlinearFormIntegra
   std::shared_ptr<mfem::ParFiniteElementSpace> trial_fes_;
 };
 
+/**
+ * @brief A class to convert NonlinearFormIntegrator to one where the input parameter undergoes a change of variables
+ */
+class SubstitutionNonlinearFormIntegrator : public mfem::NonlinearFormIntegrator {
```
jamiebramwell: I think `TransformedNonlinearFormIntegrator` is more descriptive of what is going on (unless I'm missing something).

tests/serac_wrapper_tests.cpp
```diff
@@ -210,6 +210,133 @@ TEST_F(WrapperTests, nonlinear_linear_thermal)
...
+
+  // Solve Nonlinear
+
+  ConstantCoefficient one(1.);
+
+  auto   diffusion           = std::make_shared<DiffusionIntegrator>(one);
+  auto   nonlinear_diffusion = std::make_unique<BilinearToNonlinearFormIntegrator>(diffusion);
+  double multiplier          = 2.;
+  double offset              = 1.;
+  auto   substitute          = [=](const mfem::Vector& x) {
```
jamiebramwell: Yeah, I think the phrase `substitute` confused me when I first looked at this. I think renaming these fields `transformed` (or something similar) would help a bunch, especially because you're doing a similar thing with the `TransformedScalarCoefficient`s.

jamiebramwell:
I apparently can't review this PR since I opened it. 😄

jamiebramwell:
@jonwong12 , other than merging in develop, is this ready for review?

src/integrators/wrapper_integrator.hpp
```diff
@@ -156,6 +157,72 @@ class MixedBilinearToNonlinearFormIntegrator : public mfem::NonlinearFormIntegra
...
+public:
+  /// alias for prototype of the residual_func
+  using transformed_func = std::shared_ptr<mfem::Vector>(const mfem::FiniteElement& el, mfem::ElementTransformation& Tr,
+                                                         const mfem::Vector&);
+
+  /// alias for prototype of the gradient of residual_func
+  using transformed_gradient_func = std::shared_ptr<mfem::DenseMatrix>(const mfem::FiniteElement&   el,
+                                                                       mfem::ElementTransformation& Tr,
+                                                                       const mfem::DenseMatrix&);
+
```
joshessman-llnl: Do these need to be `shared_ptr`s, or can they just return `mfem::Vector`/`mfem::DenseMatrix`?

src/integrators/wrapper_integrator.cpp
```diff
@@ -81,4 +81,30 @@ void MixedBilinearToNonlinearFormIntegrator::AssembleElementGrad(const mfem::Fin
...
+}
+
+void TransformedNonlinearFormIntegrator::AssembleElementGrad(const mfem::FiniteElement&   el,
+                                                             mfem::ElementTransformation& Tr, const mfem::Vector& elfun,
+                                                             mfem::DenseMatrix& elmat)
+{
+  auto transformed = transformed_function_(el, Tr, elfun);
+  R_->AssembleElementGrad(el, Tr, *transformed, elmat);
+  auto dense_grad_transformed = transformed_function_grad_(el, Tr, elmat);
+  elmat                       = *dense_grad_transformed;
```
joshessman-llnl: Eliminating the shared_ptr + dereference here will also allow for a return value optimization here if/when MFEM chooses to adopt move semantics.

src/integrators/wrapper_integrator.hpp
```diff
@@ -156,6 +157,72 @@ class MixedBilinearToNonlinearFormIntegrator : public mfem::NonlinearFormIntegra
...
+   */
+  virtual void AssembleElementGrad(const mfem::FiniteElement& el, mfem::ElementTransformation& Tr,
+                                   const mfem::Vector& elfun, mfem::DenseMatrix& elmat);
+
+private:
+  /**
+   * @brief The NonlinearFormIntegrator form to wrap
+   *
+   */
+  std::shared_ptr<mfem::NonlinearFormIntegrator> R_;
```
joshessman-llnl, line 210: Could this be a `unique_ptr`?

src/integrators/wrapper_integrator.hpp
jonwong12: I guess there are a few more options that I can think of:
1. I can mutate the `mfem::Vector/mfem::DenseMatrix` being passed in.
2. I could take in the output `mfem::Vector/mfem::DenseMatrix` by reference.
3. I could also return unique_ptr or raw ptr.

Just let me know which one you prefer, or if there is another option you prefer.

src/integrators/wrapper_integrator.hpp
jonwong12: Great question. 

My intention was for it to be shared, so someone can create a base mfem integrator and then one can modify the behavior of this integrator by using the `Transformed*` functions. For example in some of the time stepping algorithms, we might have different `Transformed*` functions for explicit and implicit time integration, but they are based on the same initial integrator.

Currently when the `mfem::NonlinearForm` goes out of scope, it will automatically delete all of it's integrators. In this case the original `NonlinearFormIntegrator` which may be defined and used elsewhere also won't be deleted. This is nice as we can modify different numerical behaviors by modifying the base integrator for different problems and potentially in comparing different time integration methods.

I'm open to other ways of implementing this.

samuelpmishLLNL:
I don't really understand what this code is doing, can you give a short explanation of what it does and where it might be applied to serac?

jonwong12:
> I don't really understand what this code is doing, can you give a short explanation of what it does and where it might be applied to serac?

Sorry, I just noticed it's not in the description for the PR. 

So this PR adds the ability to "transform" any nonlinearintegrator by applying a function to the input of the "residual" term, and applying another function to post-process the "tangent" term. One application of this PR is to provide a "change of variables" for a given problem `R(u) = 0`, for `u`. So if one wanted instead to solve the residual in terms of `v`, this wrapper will allow  nonlinearforms to possibly undergo "change of variables" transparently and solve `R(u(v)) = 0` for `v`. This is used in the newmark beta solver, so the equilbrium PDE can be solved in a general way, while the time integration schemes are supplied elsewhere in the code.

jamiebramwell:
@joshessman-llnl @samuelpmishLLNL what are the outstanding issues on this PR?

samuelpmishLLNL (approved):

joshessman-llnl (approved):

jamiebramwell:
LGTM

joshessman-llnl:
LGTM

Issue #209, opened 2020-09-21
Fix warnings in clang 10 build


jamiebramwell:
After the last expression template update, some warnings popped up in the clang 10 build.

```
/usr/workspace/bramwell/gitspaces/serac/src/physics/../numerics/vector_expression.hpp:119:75: warning: operator '?:' has lower precedence than '+'; '+' will be evaluated first [-Wparentheses]
  const long long per_proc = (SIZE / num_procs) + (SIZE % num_procs != 0) ? 1 : 0;
                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^
/usr/workspace/bramwell/gitspaces/serac/tests/expr_templates.cpp:62:3: note: in instantiation of function template specialization 'serac::evaluate<serac::internal::BinaryVectorExpr<mfem::Vector &, mfem::Vector &, std::plus<double> > >' requested here
  evaluate(lhs + rhs, expr_result, MPI_COMM_WORLD);
  ^
/usr/workspace/bramwell/gitspaces/serac/src/physics/../numerics/vector_expression.hpp:119:75: note: place parentheses around the '+' expression to silence this warning
  const long long per_proc = (SIZE / num_procs) + (SIZE % num_procs != 0) ? 1 : 0;
                                                                          ^
                             (                                           )
/usr/workspace/bramwell/gitspaces/serac/src/physics/../numerics/vector_expression.hpp:119:75: note: place parentheses around the '?:' expression to evaluate it first
  const long long per_proc = (SIZE / num_procs) + (SIZE % num_procs != 0) ? 1 : 0;
                                                                          ^
                                                  (                              )
```

PR #210, opened 2020-09-21
Update vector_expression.hpp

samuelpmishLLNL:
see #209

### DIFF
--- a/src/numerics/vector_expression.hpp
+++ b/src/numerics/vector_expression.hpp
@@ -116,7 +116,7 @@ void evaluate(const VectorExpr<T>& expr, mfem::Vector& result, MPI_Comm comm)
   double* result_arr = result;
 
   // If the array size is not divisible by # elements, add one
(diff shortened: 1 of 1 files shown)


joshessman-llnl:
The intent was actually that the parentheses go the other way - the partitioning was intended to add one if the vector size is not evenly divisible by the number of processes, i.e.
```cpp
const long long per_proc = (SIZE / num_procs) + ((SIZE % num_procs != 0) ? 1 : 0);
```

I was working on another semi-related change to the parallelized evaluation, but I don't have a problem with merging this in first.

samuelpmishLLNL:
Oh, I guess I misunderstood! Can't you just do ` (SIZE / num_procs) + (SIZE % num_procs != 0)` then?
In CUDA we usually write this as `(SIZE + num_procs + 1) / num_procs`. I'm not sure which is the clearest.

samuelpmishLLNL:
if you've got another related change, I'll just close this PR! I was just looking at the repo during an earlier meeting.

PR #212, opened 2020-09-21
Add high order compressed ParaView output type

jandrej:


src/physics/utilities/solver_config.hpp
```diff
@@ -20,6 +20,7 @@ namespace serac {
 enum class OutputType
 {
   GLVis,
+  ParaView,
```
jamiebramwell, line 23: A test for this output type is needed.

src/physics/base_solver.cpp
```diff
@@ -118,6 +118,20 @@ void BaseSolver::initializeOutput(const serac::OutputType output_type, const std
       break;
     }
 
+    case serac::OutputType::ParaView: {
```
jamiebramwell, line 122: Does this require any extra dependencies?

jandrej: No, its self contained in mfem. It just outputs text through a stream.

src/physics/utilities/solver_config.hpp
jandrej: Yes, which example would be good to change the output type on?

jamiebramwell: Hmmm, really any of them is fine. Maybe one of the glvis ones (thermal-explicit?) to help clean out the output in the test directory?

src/physics/base_solver.hpp
```diff
@@ -204,6 +204,11 @@ class BaseSolver {
    */
   int order_;
 
+  /**
+   * @brief ParaView data collection pointer
+   */
+  std::unique_ptr<mfem::ParaViewDataCollection> pv_dc_;
```
joshessman-llnl: Instead of having separate variables for different types of data collection, it might be preferable to have a 
`std::unique_ptr<mfem::DataCollection>`.

I have something similar implemented here: https://github.com/LLNL/serac/pull/88/files (branch is waiting on an Axom PR)

jamiebramwell:
LGTM

white238 (approved):

jamiebramwell:
LGTM

jamiebramwell (approved):

PR #213, opened 2020-09-21
Fix operator precedence warnings, add support for HypreParVector expr templates

### LINKED ISSUE #209: Fix warnings in clang 10 build

joshessman-llnl:
The distribution of expression template evaluation isn't particularly useful outside the context of a distributed memory model, so this PR adds experimental support for using `mfem::HypreParVector` with expression templates.  The Sphinx docs have been updated to reflect this new functionality and its limitations.

Resolves #209

### DIFF
--- a/tests/benchmark_expr_templates.cpp
+++ b/tests/benchmark_expr_templates.cpp
@@ -34,6 +34,29 @@ static std::pair<mfem::DenseMatrix, mfem::Vector> sample_matvec(const int rows, 
   return {matrix, vec_in};
 }
 
+static auto build_partitioning(MPI_Comm comm, const int size)
+{
+  int num_procs = 0;
+  int rank      = 0;
+  MPI_Comm_size(comm, &num_procs);
+  MPI_Comm_rank(comm, &rank);
+  bool assumed_partition = HYPRE_AssumedPartitionCheck();
+  auto partitioning      = std::make_unique<int[]>(assumed_partition ? 2 : (num_procs + 1));
+  auto per_proc          = (size / num_procs) + ((size % num_procs != 0) ? 1 : 0);
+
+  if (assumed_partition) {
+    auto n_entries  = (rank == num_procs - 1) ? size - ((num_procs - 1) * per_proc) : per_proc;
+    partitioning[0] = per_proc * rank;
+    partitioning[1] = (per_proc * rank) + n_entries;
+  } else {
+    for (int i = 0; i < num_procs; i++) {
+      partitioning[i] = per_proc * i;
+    }
+    partitioning[num_procs] = size;
+  }
+  return std::make_pair(std::move(partitioning), per_proc * rank);
+}
+
 static void BM_mixed_expr_MFEM(benchmark::State& state)
 {
   MPI_Barrier(MPI_COMM_WORLD);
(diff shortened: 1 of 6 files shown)


tests/benchmark_expr_templates.cpp
```diff
@@ -34,6 +34,29 @@ static std::pair<mfem::DenseMatrix, mfem::Vector> sample_matvec(const int rows,
...
 }
 
+static auto build_partitioning(MPI_Comm comm, const int size)
+{
+  int num_procs = 0;
+  int rank      = 0;
+  MPI_Comm_size(comm, &num_procs);
+  MPI_Comm_rank(comm, &rank);
+  bool assumed_partition = HYPRE_AssumedPartitionCheck();
+  auto partitioning      = std::make_unique<int[]>(assumed_partition ? 2 : (num_procs + 1));
```
samuelpmishLLNL, line 44: At first glance, I'm not actually sure what `std::make_unique<int[]>(assumed_partition ? 2 : (num_procs + 1));` does. Does this need `std::unique_ptr`, or does `std::vector<int>` work?

tests/benchmark_expr_templates.cpp
```diff
@@ -178,12 +201,38 @@ static void BM_large_expr_single_alloc_par_EXPR(benchmark::State& state)
...
+}
+
 BENCHMARK(BM_mixed_expr_MFEM)->RangeMultiplier(2)->Range(10, 10 << 10);
 BENCHMARK(BM_mixed_expr_EXPR)->RangeMultiplier(2)->Range(10, 10 << 10);
 BENCHMARK(BM_mixed_expr_single_alloc_EXPR)->RangeMultiplier(2)->Range(10, 10 << 10);
 BENCHMARK(BM_large_expr_MFEM)->RangeMultiplier(2)->Range(10, 10 << 10);
 BENCHMARK(BM_large_expr_single_alloc_EXPR)->RangeMultiplier(2)->Range(10, 10 << 10);
-BENCHMARK(BM_large_expr_single_alloc_par_EXPR)->RangeMultiplier(2)->Range(10, 10 << 10);
+
+// Too slow
```
samuelpmishLLNL, line 233: If we changed the large expr examples to use sparse matrices instead of dense, it could reduce the runtime of those tests and would be more representative of our calculations too.

samuelpmishLLNL (approved):

tests/benchmark_expr_templates.cpp
joshessman-llnl: The conditional size check is because Hypre uses one of two schemes for partitioning, depending on how it was compiled.

The constructor for `HypreParVector` only needs a raw integer array.  `std::vector` + `v.data()` would work fine, but `unique_ptr` is slightly less overhead.

white238 (approved):

joshessman-llnl:
LGTM

PR #214, opened 2020-09-21
Add PETSc build option

### LINKED ISSUE #164: Add netcdf and petsc build options

joshessman-llnl:
This PR adds a `+petsc` variant (defaulting to false) to the Spack package which both enables the `+petsc` MFEM variant and exposes PETSc within serac.

A "smoke test" was also added to confirm that PETSc could be linked against.

Spack locks the version of PETSc too early and picks one that is incompatible with our version of SuperLU, so I decided to update SuperLU - some of the bugs associated with the direct solver integration are not present in this newer version of superlu-dist.

Resolves #164


scripts/uberenv/packages/serac/package.py
```diff
@@ -123,7 +130,7 @@ class Serac(CMakePackage, CudaPackage):
     # Libraries that do not have a debug variant
     depends_on("conduit@master~shared~python")
     depends_on("caliper@master~shared+mpi~callpath~adiak~papi", when="+caliper")
-    depends_on("superlu-dist@5.4.0~shared")
+    depends_on("superlu-dist@6.1.1~shared")
```
jamiebramwell, line 137: Can we turn the superlu tests back on with this change?

tests/CMakeLists.txt
```diff
@@ -97,6 +97,17 @@ if (ENABLE_GTEST)
...
 
+    if(SERAC_USE_PETSC)
+        blt_add_executable( NAME        petsc_smoketest
+                            SOURCES     petsc_smoketest.cpp
+                            OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
+                            DEPENDS_ON  PETSc mpi gtest
+                            FOLDER      serac/tests )
+        blt_add_test(NAME          petsc_smoketest
+                    COMMAND       petsc_smoketest
+                    NUM_MPI_TASKS 1 )
```
jamiebramwell: It's probably a good idea to do some parallel thing with the smoke test since that is a possible major source of errors.

jamiebramwell (commented):
I think it just needs a parallel smoke test.

tests/CMakeLists.txt
joshessman-llnl: I went ahead and just used one of PETSc's introductory examples

jamiebramwell (approved):

scripts/uberenv/packages/serac/package.py
joshessman-llnl: I have some changes to the SuperLU in an MFEM branch that probably should be handled first, but I was thinking that they would instead be permanently replaced by Pastix or something else that's more stable.

scripts/uberenv/packages/mfem/package.py
```diff
@@ -479,10 +479,10 @@ def find_optional_library(name, prefix):
                 ld_flags_from_library_list(spec[sun_spec].libs)]
 
         if '+petsc' in spec:
+            # SERAC EDIT BEGIN (TODO: spack PR)
             options += [
-                'PETSC_OPT=%s' % spec['petsc'].headers.cpp_flags,
-                'PETSC_LIB=%s' %
-                ld_flags_from_library_list(spec['petsc'].libs)]
+                'PETSC_DIR=%s' % spec['petsc'].prefix]
```
white238, line 484: I guess this begs the question why the rest of the dependencies build the link and include lines if you can just override the <tpl name>_DIR makefile variable

scripts/uberenv/packages/serac/package.py
```diff
@@ -357,6 +364,9 @@ def hostconfig(self, spec, prefix, py_site_pkgs_dir=None):
         netcdf_dir = get_spec_path(spec, "netcdf-c", path_replacements)
         cfg.write(cmake_cache_entry("NETCDF_DIR", netcdf_dir))
 
+        petsc_dir = get_spec_path(spec, "petsc", path_replacements)
```
white238: You have to guard this with ``if "+petsc" in spec:``

scripts/uberenv/packages/mfem/package.py
joshessman-llnl: I think building the link/include stuff in spack is preferable in case the directory structure assumed in the makefile doesn't match what was installed - it's just in this case that no other option is available (for static linking.)

jamiebramwell:
Does this need anything else @white238 ?

cmake/thirdparty/FindPETSc.cmake
```diff
@@ -0,0 +1,147 @@
...
+string(REPLACE "PETSC_CC_INCLUDES = " "" PETSC_CC_INCLUDES ${PETSC_CC_INCLUDES})
+string(STRIP ${PETSC_CC_INCLUDES} PETSC_CC_INCLUDES)
+
+get_target_property(PETSC_INCLUDE_DIRS PETSc::PETSc INTERFACE_INCLUDE_DIRECTORIES)
+get_target_property(PETSC_LIB PETSc::PETSc INTERFACE_LINK_LIBRARIES)
+
+blt_register_library(
+    NAME          PETSc
+    INCLUDES      ${PETSC_INCLUDE_DIRS} ${PETSC_CC_INCLUDES}
+    LIBRARIES     ${PETSC_LIB} ${PETSC_EXTERNAL_LIB_BASIC}
```
white238, line 144: Is there a reason you don't just use the exported cmake target directly?

white238 (approved):

joshessman-llnl:
LGTM

PR #216, opened 2020-09-22
ODEs with time-varying boundary conditions

samuelpmishLLNL:
The main goal of this PR is to support time dependent boundary conditions, so the user can write something like

```cpp
auto f = std::make_shared< StdFunctionCoefficient>([](mfem::Vector x, double t) { 
  return 1.0 + 6.0 * x[0] * t - 2.0 * x[1] * t + (x[0] - x[1]) * x[0] * x[0];
});

therm_solver.setTemperatureBCs({1}, f);
```

and have the prescribed values change in time accordingly.


jamiebramwell:
I spoke to @samuelpmishLLNL , and he thinks it is an appropriate time for people to start looking at this PR and giving some feedback as this is a fairly major feature development.

scripts/mathematica/ODE_verification.nb
```diff
@@ -0,0 +1,9134 @@
+(* Content-type: application/vnd.wolfram.mathematica *)
```
joshessman-llnl: I think it's a good idea to hang on to these scripts for reproducibility, etc, but I'm not sure that this repo is the right place for it.  Axom uses the https://github.com/LLNL/axom_data repo for storing large stuff, so possibly we could look at a similar approach for notebooks, large mesh files, large matrix files, etc.   git-lfs is also an option.

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -11,18 +11,31 @@
 
 namespace serac {
 
-StdFunctionCoefficient::StdFunctionCoefficient(std::function<double(mfem::Vector&)> func) : func_(func) {}
+StdFunctionCoefficient::StdFunctionCoefficient(std::function<double(mfem::Vector&)> func)
+    : func_([=](mfem::Vector& v, double /* t */) { return func(v); }), is_time_dependent_(false)
```
joshessman-llnl: 👍

samuelpmishLLNL:
it is a little confusing, since there's some different implementations in here. The one I think makes the most sense is in `odes.hpp`, using a finite difference stencil in time to get the required derivatives.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -29,29 +30,55 @@ namespace serac {
...
+    return (y(x, t + dt) - y(x, t - dt)) / (2.0 * dt);
+  });
+}
+
+inline StdFunctionCoefficient d2_dt2(const StdFunctionCoefficient & y, const double dt = 1.0e-4) {
+  return StdFunctionCoefficient([dt, y = y.func_](mfem::Vector& x, double t) {
+    return (y(x, t + dt) - 2 * y(x, t) + y(x, t - dt)) / (dt * dt);
+  });
+}
+
```
joshessman-llnl: Did these need to be defined outside the class definition?  I think ADL applies here but I'm not 100% sure.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -67,24 +94,71 @@ class StdFunctionVectorCoefficient : public mfem::VectorCoefficient {
...
 };
 
+inline StdFunctionVectorCoefficient d_dt(const StdFunctionVectorCoefficient & y, const double dt = 1.0e-8) {
+  return StdFunctionVectorCoefficient(y.vdim, [dt, y = y.func_](mfem::Vector& x, mfem::Vector & return_value, double t) {
+    // if mfem would use return statements instead of 
+    // always using out parameters, we could just write
+    // return (y(x, t + dt) - y(x, t - dt)) / (2.0 * dt);
+
+    // instead, we get this:
+    mfem::Vector yl, yr;
```
joshessman-llnl: Since using `y.func_` directly shortcuts the vector resizing in `Eval`, do these need to be allocated/sized here?

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,263 @@
...
+
+  mutable mfem::Vector                  uf;
+  mutable mfem::Vector                  uc;
+  mutable mfem::Vector                  duc_dt;
+  mutable mfem::Vector                  uc_plus;
+  mutable mfem::Vector                  uc_minus;
+  std::unique_ptr<mfem::HypreParMatrix> T_;
+  std::unique_ptr<mfem::HypreParMatrix> Mf_;
+  double                                previous_dt;
+  double                                epsilon;
```
joshessman-llnl: Can these be made private?

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,263 @@
...
+    uc_minus = 0.0;
+    for (const auto& bc : bcs_.essentials()) {
+      bc.projectBdrToDofs(uc, t);
+      bc.projectBdrToDofs(uc_plus, t + epsilon);
+      bc.projectBdrToDofs(uc_minus, t - epsilon);
+    }
+    duc_dt   = (uc_plus - uc_minus) / (2.0 * epsilon);
+    d2uc_dt2 = (uc_plus - 2 * uc + uc_minus) / (epsilon * epsilon);
+
+    d2u_dt2 = invT_ * (f_ - (M_ * d2uc_dt2 - C_ * duc_dt - K_ * uc) - (C_ * duf_dt - K_ * uf));
```
joshessman-llnl: 👍

src/physics/thermal_solver_rework.cpp
```diff
@@ -0,0 +1,162 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
```
joshessman-llnl: It's kind of difficult to tell what was changed in this thermal solver rework, but this shouldn't be a problem since the plan is to replace the thermal solver entirely.

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,366 @@
...
+#include "mfem.hpp"
+
+template <typename T>
+struct has_print {
+  template <typename C>
+  static float test(decltype(&C::Print));
+  template <typename C>
+  static double         test(...);
+  static constexpr bool value = sizeof(test<T>(0)) == sizeof(float);
+};
```
joshessman-llnl: I don't have a preference either way, but checks for particular member functions/operators pre-C++20 can also be implemented in terms of `std::true_type`/`std::false_type`, which is a little more canonical with respect to the standard library's type trait implementations.

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,366 @@
...
+mfem::Vector fext(double t)
+{
+  mfem::Vector force(3);
+  force[0] = -10 * t;
+  force[1] = 0;
+  force[2] = 10 * t * t;
+  return force;
+}
+
+#if 0
```
joshessman-llnl: Can this section be deleted?

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+};
+#endif
+
+
+double first_order_test_mfem(int num_steps)
+{
+  // clang-format off
+  mfem::DenseMatrix M(3, 3);
+  M(0, 0) = 2; M(0, 1) = 1; M(0, 2) = 0;
```
joshessman-llnl: If only there was an `mfem::DenseMatrix::DenseMatrix(std::initializer_list<std::initializer_list<double>>)`...

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+  double k = 100;
+  mfem::DenseMatrix K(3, 3);
+  K(0, 0) =  k; K(0, 1) =    -k; K(0, 2) =  0;
+  K(1, 0) = -k; K(1, 1) = 2 * k; K(1, 2) = -k;
+  K(2, 0) =  0; K(2, 1) =    -k; K(2, 2) =  k;
+  // clang-format on
+
+  double beta  = 1.0 / 6.0;
+  double gamma = 1.0 / 2.0;
```
joshessman-llnl: `gamma` gets highlighted differently for me because `gamma` is a deprecated C function - I don't think we need to worry about a collision here, though.

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  mfem::Vector solution(3);
+  solution[0] = -0.1443913076373983;
+  solution[1] = 0.0;
+  solution[2] = 0.04968869236260167;
+
+  mfem::Vector error = u - solution;
+
+  return error.Norml2();
```
joshessman-llnl: Can these be converted to Google Test tests that loop over a couple step values and `EXPECT_LT(error.Norml2(), threshold)` for some appropriate set of thresholds?

joshessman-llnl (commented):
Looking good so far!  I think some of the new stuff is missing Doxygen annotations, and once the thermal solver is replaced with the reworked thermal solver it'll be easier to tell what changed.

src/physics/thermal_solver_rework.cpp
samuelpmishLLNL: I kept them separate for the sake of comparing outputs, but you're right that it would be more helpful to actually just apply the changes to the thermal solver to better understand the code changes!

tests/time_integration_with_constraints.cpp
samuelpmishLLNL: Good feedback-- I expect that this test file will be deleted in the final PR, since it was mostly just a way to convince myself that the operators were written correctly.

src/coefficients/stdfunction_coefficient.hpp
samuelpmishLLNL: I prefer the free function syntax here, because it more closely resembles the mathematical notation. That being said, these are not actually being used in the odes.hpp implementation. This is another possible way to get at the time derivatives, but it seemed harder to integrate with the boundary condition manager.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -67,24 +94,71 @@ class StdFunctionVectorCoefficient : public mfem::VectorCoefficient {
...
   /**
    * @brief The function to evaluate for the coefficient
    */
-  std::function<void(mfem::Vector&, mfem::Vector&)> func_;
+  std::function<void(mfem::Vector&, mfem::Vector&, double)> func_;
+  bool is_time_dependent_;
 };
 
+inline StdFunctionVectorCoefficient d_dt(const StdFunctionVectorCoefficient & y, const double dt = 1.0e-8) {
+  return StdFunctionVectorCoefficient(y.vdim, [dt, y = y.func_](mfem::Vector& x, mfem::Vector & return_value, double t) {
```
solberg2: I'm probably missing something here, but why is dt defaulted to 1.0e-8?

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -67,24 +94,71 @@ class StdFunctionVectorCoefficient : public mfem::VectorCoefficient {
...
+
+    // instead, we get this:
+    mfem::Vector yl, yr;
+    y(x, yl, t - dt);
+    y(x, yr, t + dt);
+    return_value = (yr - yl) * (1.0 / (2.0 * dt));
+  });
+}
+
+inline StdFunctionVectorCoefficient d2_dt2(const StdFunctionVectorCoefficient & y, const double dt = 1.0e-4) {
```
solberg2: Again, I'm probably missing something but why is dt defaulted to 1.0e-4?

src/coefficients/stdfunction_coefficient.hpp
samuelpmish: it was based on the assumption that the user is picking appropriate units to get the characteristic time on the order of unity,

first_derivative_error

src/coefficients/stdfunction_coefficient.hpp
samuelpmish: like above, assumes the characteristic time is close to 1. The "sweet spot" for the second derivative appears to be at a larger value of dt than for the first derivative

second_derivative_error

if the user isn't scaling their dimensions, they can provide their own appropriate value of dt.

src/coefficients/stdfunction_coefficient.hpp
solberg2: I don't think that's a good assumption.  People generally use physical units (seconds, microseconds) and don't rescale.

solberg2:
> it is a little confusing, since there's some different implementations in here. The one I think makes the most sense is in odes.hpp, using a finite difference stencil in time to get the required derivatives.

The way this is handled in Diablo, essentially (translating into object-oriented world) is that we created a class that has as one of its function outputs the time derivative(s) necessary.  So basically force the user to do that.  That's probably the best way.  The finite-difference stencil is prone to difficulty.  One could provide as part of the class the option to use a finite-difference stencil, but I think it is way better to make the requirements non-opaque to the user.

Why do you need the second-derivative, by the way?  I cannot think of how that would make sense.

src/coefficients/stdfunction_coefficient.hpp
jonwong12: Looks good to me. Would be nice to have some doxygen comments.

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,263 @@
...
+  }
+  void ImplicitSolve(const double dt0, const double dt1, const mfem::Vector& u, const mfem::Vector& du_dt,
+                     mfem::Vector& d2u_dt2)
+  {
+    implicit_func(t, dt0, dt1, u, du_dt, d2u_dt2);
+  }
+
+  std::function<explicit_signature> explicit_func;
+  std::function<implicit_signature> implicit_func;
+};
```
jonwong12, line 67: Are these used in this PR? Is the main idea to be able to access `Mult` and `ImplicitSolve` via `explicit_func` and `implicit_func`?

jamiebramwell:
@samuelpmishLLNL is this still a WIP or is it ready for review?

src/physics/base_solver_refactor.cpp
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
```
jamiebramwell: Delete this.

src/physics/nonlinear_solid.hpp
```diff
@@ -228,6 +230,40 @@ class NonlinearSolid : public BasePhysics {
    * @brief Nonlinear system solver instance
    */
   EquationSolver nonlin_solver_;
+
+  StdFunctionOperator residual;
```
jamiebramwell: The new member variables need trailing underscores.

src/physics/nonlinear_solid.hpp
```diff
@@ -228,6 +230,40 @@ class NonlinearSolid : public BasePhysics {
...
+
+  // predicted displacements and velocities
+  mfem::Vector x;
```
jamiebramwell: Doxygen style comments are needed.

src/physics/nonlinear_solid.hpp
```diff
@@ -228,6 +230,40 @@ class NonlinearSolid : public BasePhysics {
...
+  mfem::Vector u;
+  mfem::Vector du_dt;
+  mfem::Vector previous;
+
+  // temporary values used to compute finite difference approximations
+  // to the derivatives of constrained degrees of freedom
+  mfem::Vector U_minus;
```
jamiebramwell: I would use more descriptive names. We are currently relying on case sensitivity to distinguish between `u` and `U`.

src/physics/nonlinear_solid.cpp
```diff
@@ -132,47 +148,148 @@ void NonlinearSolid::completeSetup()
...
-  nonlin_solver_.Mult(zero, displacement_.trueVec());
+    // the dynamic case is described by a residual function and a second order 
+    // ordinary differential equation. Here, we define the residual function in
+    // terms of an acceleration.
+    residual.function = [=](const mfem::Vector& d2u_dt2, mfem::Vector& res) mutable {
+
+      // TODO: we should avoid re-assembling this when not required
+      C_mat.reset(C->ParallelAssemble());
+
+      res = (*M_mat) * d2u_dt2 + (*C_mat) * (du_dt + c1 * d2u_dt2) + (*H) * (x + u + c0 * d2u_dt2);
```
jamiebramwell: This is so nice!!!

src/physics/nonlinear_solid.cpp
```diff
@@ -132,47 +148,148 @@ void NonlinearSolid::completeSetup()
...
+              dU_dt   = (U_plus - U_minus) / (2.0 * epsilon) - c1 * d2U_dt2;
+              U       = U - c0 * d2U_dt2;
+            }
+          } else {
+            d2U_dt2 = (U_minus - 2.0 * U + U_plus) / (epsilon * epsilon);
+            dU_dt   = (U_plus - U_minus) / (2.0 * epsilon);
+          }
+
+          auto constrained_dofs = bcs_.allEssentialDofs();
+          u.SetSubVector(constrained_dofs, 0.0);
```
jamiebramwell: Relying on case sensitivity makes me a bit nervous.

src/physics/nonlinear_solid_solver_refactor.cpp
```diff
@@ -0,0 +1,341 @@
+// Copyright (c) 2019, Lawrence Livermore National Security, LLC and
```
jamiebramwell: This can be deleted.

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,61 @@
+#pragma once
```
jamiebramwell, line 1: We should move toward everything using `#pragma once`.

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <variant>
+
+#include "mfem.hpp"
+#include "numerics/expr_template_ops.hpp"
+#include "physics/utilities/boundary_condition_manager.hpp"
+#include "physics/utilities/equation_solver.hpp"
+
+class SecondOrderODE : public mfem::SecondOrderTimeDependentOperator {
+public:
+  using signature = void(const double, const double, const double, const mfem::Vector&, const mfem::Vector&,
```
jamiebramwell: A slightly more descriptive name might be nice. Maybe `ODESignature`?

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  std::function<signature> f;
+};
+
+class FirstOrderODE : public mfem::TimeDependentOperator {
+public:
+  using signature = void(const double, const double, const mfem::Vector&, mfem::Vector&);
+  FirstOrderODE() : mfem::TimeDependentOperator(0, 0.0) {}
+  FirstOrderODE(int n, std::function<signature> f) : mfem::TimeDependentOperator(n, 0.0), f(f) {}
+  FirstOrderODE(SecondOrderODE other_ode) : mfem::TimeDependentOperator(2 * other_ode.Height(), other_ode.GetTime())
```
jamiebramwell: I would rename `other_ode` -> `second_order_ode`.

src/physics/operators/stdfunction_operator.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <functional>
+
+#include "mfem.hpp"
+
+class StdFunctionOperator : public mfem::Operator {
```
jamiebramwell, line 17: Needs doxygen-style comments.

src/physics/operators/stdfunction_operator.hpp
```diff
@@ -0,0 +1,14 @@
...
+public:
+  StdFunctionOperator(int n) : mfem::Operator(n) {}
+  void            Mult(const mfem::Vector& k, mfem::Vector& y) const { function(k, y); };
+  mfem::Operator& GetGradient(const mfem::Vector& k) const { return jacobian(k); };
+  mutable std::function<void(const mfem::Vector&, mfem::Vector&)> function;
```
jamiebramwell: This is so so so nice!

jamiebramwell: Should these be private variables with setters and such? Any thoughts @joshessman-llnl  @white238 ?

src/physics/thermal_conduction.hpp
```diff
@@ -32,20 +34,52 @@ namespace serac {
 class ThermalConduction : public BasePhysics {
 public:
   /**
-   * @brief A timestep method and configs for the M and T solvers
+   * @brief A timestep method and config for the M solver
    */
   struct DynamicSolverParameters {
```
jamiebramwell, line 38: This goes beyond this PR, but it might be nice to have a common clear idiom for how we define and set all of these physics parameters. Maybe renaming things like this `DynamicOptions`?

src/physics/thermal_conduction.hpp
```diff
@@ -32,20 +34,52 @@ namespace serac {
...
+  {
+    return {.rel_tol     = 1.0e-6,
+            .abs_tol     = 1.0e-12,
+            .print_level = 0,
+            .max_iter    = 200,
+            .lin_solver  = LinearSolver::CG,
+            .prec        = HypreSmootherPrec{mfem::HypreSmoother::Jacobi}};
+  }
+
+  static NonlinearSolverParameters defaultNonlinearParameters()
```
jamiebramwell, line 63: This should go away once we inlet-ify the thermal class.

src/physics/thermal_conduction.hpp
```diff
@@ -183,6 +217,27 @@ class ThermalConduction : public BasePhysics {
    */
   std::unique_ptr<DynamicConductionOperator> dyn_oper_;
 
+  FirstOrderODE ode_;
+
+  StdFunctionOperator residual;
```
jamiebramwell: Needs a trailing underscore.

src/physics/thermal_conduction.hpp
```diff
@@ -183,6 +217,27 @@ class ThermalConduction : public BasePhysics {
...
+  EquationSolver nonlin_solver_;
+  EquationSolver mass_solver_;
+  EquationSolver stiffness_solver_;
+
+  double                                dt_, previous_dt_;
+  std::unique_ptr<mfem::HypreParMatrix> J_;
+
+  mfem::Vector zero_;
+
+  mfem::Vector u_;
```
jamiebramwell, line 268: `u` vs. `U` again

src/physics/thermal_conduction.cpp
```diff
@@ -16,24 +16,36 @@ ThermalConduction::ThermalConduction(int order, std::shared_ptr<mfem::ParMesh> m
...
-    K_inv_ = EquationSolver(mesh->GetComm(), *K_params);
-    setTimestepper(TimestepMethod::QuasiStatic);
-  }
-  // Otherwise, two sets of parameters for the dynamic M/T solve
-  else if (auto dyn_params = std::get_if<DynamicSolverParameters>(&params)) {
-    setTimestepper(dyn_params->timestepper);
-    // Save these to initialize the DynamicConductionOperator later
-    dyn_M_params_ = dyn_params->M_params;
-    dyn_T_params_ = dyn_params->T_params;
+  nonlin_solver_ = EquationSolver(mesh->GetComm(), params.T_lin_params, params.T_nonlin_params);
```
jamiebramwell, line 24: I really like upgrading this to a full nonlinear form.

src/physics/thermal_conduction.cpp
```diff
@@ -88,36 +100,142 @@ void ThermalConduction::completeSetup()
...
-  // Eliminate the essential DOFs from the stiffness matrix
+  // Project the essential boundary coefficients
   for (auto& bc : bcs_.essentials()) {
-    bc.eliminateFromMatrix(*K_mat_);
+    // Project the coefficient
+    bc.projectBdr(temperature_, time_);
+
+    auto ids = bc.getTrueDofs();
+
+    for (int i = 0; i < ids.Size(); i++) {
```
jamiebramwell: Delete these debug print statements

src/physics/thermal_conduction.cpp
```diff
@@ -88,36 +100,142 @@ void ThermalConduction::completeSetup()
     *rhs_ = 0.0;
   }
 
-  // Assemble the stiffness matrix
-  K_mat_.reset(K_form_->ParallelAssemble());
+  // Build the dof array lookup tables
+  temperature_.space().BuildDofToArrays();
```
jamiebramwell, line 120: Is this still needed?

src/physics/thermal_conduction.cpp
```diff
@@ -88,36 +100,142 @@ void ThermalConduction::completeSetup()
...
     // If dynamic, assemble the mass matrix
     M_form_ = temperature_.createOnSpace<mfem::ParBilinearForm>();
     M_form_->AddDomainIntegrator(new mfem::MassIntegrator());
     M_form_->Assemble(0);  // keep sparsity pattern of M and K the same
     M_form_->Finalize();
 
-    M_mat_.reset(M_form_->ParallelAssemble());
+    M_.reset(M_form_->ParallelAssemble());
+
+    //// Make the time integration operator and set the appropriate matrices
```
jamiebramwell: Delete this dead code.

src/physics/thermal_solver_rework.hpp
```diff
@@ -0,0 +1,197 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
```
jamiebramwell: This should be deleted.

src/physics/thermal_solver_rework.cpp
```diff
@@ -0,0 +1,203 @@
...
+// other Serac Project Developers. See the top-level LICENSE file for
```
jamiebramwell: This should be deleted.

src/physics/utilities/boundary_condition.cpp
```diff
@@ -48,19 +48,34 @@ void BoundaryCondition::project(FiniteElementState& state) const
...
+      // a single component of a vector-valued function
+      auto scalar_coef = std::get<std::shared_ptr<mfem::Coefficient>>(coef_);
+      if (component_ == -1) {
+        state.gridFunc().ProjectCoefficient(*scalar_coef, dof_list);
+      } else {
+        state.gridFunc().ProjectCoefficient(*scalar_coef, dof_list, component_);
+      }
     }
+
+    // if (component_ == -1) {
```
jamiebramwell: Dead code.

src/physics/utilities/boundary_condition.cpp
```diff
@@ -48,19 +48,34 @@ void BoundaryCondition::project(FiniteElementState& state) const
...
-    if (component_ == -1) {
-      // If it contains all components, project the vector
-      auto vec_coef = std::get_if<std::shared_ptr<mfem::VectorCoefficient>>(&coef_);
-      SLIC_ASSERT_MSG(vec_coef,
-                      "Essential boundary condition contained all components but had a non-vector coefficient.");
-      state.gridFunc().ProjectCoefficient(**vec_coef, dof_list);  // Deref ptr-to-result, then shared_ptr
+    // the only reason to store a VectorCoefficient is to act on all components
+    if (std::holds_alternative<std::shared_ptr<mfem::VectorCoefficient>>(coef_)) {
+      auto vec_coef = std::get<std::shared_ptr<mfem::VectorCoefficient>>(coef_);
+      state.gridFunc().ProjectCoefficient(*vec_coef, dof_list);
```
jamiebramwell, line 54: This function expects full vdofs. We should probably add a TODO comment referencing this.

src/physics/utilities/solver_config.hpp
```diff
@@ -44,7 +44,42 @@ enum class TimestepMethod
   ImplicitMidpoint,
   SDIRK23,
   SDIRK34,
-  QuasiStatic
+  QuasiStatic,
+
+  HHTAlpha,
```
jamiebramwell, line 52: Add a comment that these are second order options.

src/physics/utilities/solver_config.hpp
```diff
@@ -44,7 +44,42 @@ enum class TimestepMethod
...
+  WBZAlpha,
+  AverageAcceleration,
+  LinearAcceleration,
+  CentralDifference,
+  FoxGoodwin
+};
+
+// this enum describes the which way to
```
jamiebramwell: Doxygen style needed.

tests/serac_dynamic_solver.cpp
```diff
@@ -168,8 +168,8 @@ TEST(dynamic_solver, dyn_direct_solve)
   double v_norm = dyn_solver.velocity().gridFunc().ComputeLpError(2.0, zerovec);
   double x_norm = dyn_solver.displacement().gridFunc().ComputeLpError(2.0, zerovec);
 
-  EXPECT_NEAR(12.86733, x_norm, 0.0001);
-  EXPECT_NEAR(0.22298, v_norm, 0.0001);
+  EXPECT_NEAR(1.4225, x_norm, 0.0001);
```
jamiebramwell, line 165: Thank you for unifying the displacement vs. position issue!

tests/serac_dynamic_solver_previous.cpp
```diff
@@ -0,0 +1,238 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
```
jamiebramwell: This should be deleted.

tests/serac_thermal_solver_rework.cpp
```diff
@@ -0,0 +1,432 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
```
jamiebramwell: This should be deleted.

tests/serac_thermal_structural_solver.cpp
```diff
@@ -165,4 +165,4 @@ int main(int argc, char* argv[])
   MPI_Finalize();
 
   return result;
-}
+}
```
jamiebramwell: Add EOF newline.

src/physics/nonlinear_solid.cpp
```diff
@@ -199,6 +197,10 @@ void NonlinearSolid::completeSetup()
     // ordinary differential equation. Here, we define the residual function in
     // terms of an acceleration.
     residual.function = [=](const mfem::Vector& d2u_dt2, mfem::Vector& res) mutable {
+
+      // TODO: we should avoid re-assembling this when not required
+      C_mat.reset(C->ParallelAssemble());
+
       res = (*M_mat) * d2u_dt2 + (*C_mat) * (du_dt + c1 * d2u_dt2) + (*H) * (x + u + c0 * d2u_dt2);
```
solberg2: This is a "nonlinear solid" but it's really being formulated as a linear solid, with C and H.  How does that work?

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+  mfem::Vector error = u - solution;
+
+  std::cout << error << std::endl;
+
+  return error.Norml2();
+}
+#endif
+
+double second_order_test_mfem(int num_steps, int method = 2)
```
jamiebramwell: Additional comments are needed in this function.

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,435 @@
...
+  error.Print(std::cout);
+
+  return error.Norml2();
+}
+
+int main(int argc, char** argv)
+{
+  MPI_Init(&argc, &argv);
+
+#if 0
```
jamiebramwell: dead code.

jamiebramwell (commented):
This is great! The vast majority of my comments deal with style and documentation issues.

src/physics/nonlinear_solid.cpp
solberg2: Maybe that's the function pointer "H".  But you would really want it to be a function of both u and v, to accomodate non-linear viscoelastic/viscoplastic effects

samuelpmish: Most of these changes to the NonlinearSolid class are just reimplementing the ODE with less boilerplate code, rather than making it support more material models. You're right that in the longer term we need to provide a way to pass both displacements and velocities to target more constitutive models, but Serac isn't quite there yet. However, Julian has been doing some important work recently that would give us a more elegant way to define materials involving multiple finite element fields (e.g. pressure/velocity, displacement/velocity, efield/bfield, etc), so stay tuned!

jamiebramwell: Yep! The issue is that the `H` operator is explicitly a hyperelastic model right now. Once we refactor this, we should be able to pass in both displacement and velocity appropriately.

jamiebramwell:
The tests now pass! 🎉

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -11,18 +11,33 @@
...
-    : mfem::VectorCoefficient(dim), func_(func)
+    : mfem::VectorCoefficient(dim),
+      func_([=](mfem::Vector& v, mfem::Vector& w, double /* t */) { return func(v, w); }),
+      is_time_dependent_(false)
+{
+}
+
+StdFunctionVectorCoefficient::StdFunctionVectorCoefficient(
+    int dim, std::function<void(mfem::Vector&, mfem::Vector&, double)> func)
+    : mfem::VectorCoefficient(dim), func_(func), is_time_dependent_(true)
```
joshessman-llnl: Instead of extending `StdfunctionCoef`, potentially you could use MFEM's `Function/VectorFunctionCoefficient`, which already supports time-dependent `std::function`s (though with a slightly different signature, they use `void(const mfem::Vector&, double, mfem::Vector&)`)

src/physics/nonlinear_solid.cpp
```diff
@@ -45,17 +47,32 @@ NonlinearSolid::NonlinearSolid(int order, std::shared_ptr<mfem::ParMesh> mesh, c
...
+  previous = 0.0;
+
+  zero = mfem::Vector(true_size);
+  zero = 0.0;
+
+  U_minus = mfem::Vector(true_size);
+  U       = mfem::Vector(true_size);
+  U_plus  = mfem::Vector(true_size);
+  dU_dt   = mfem::Vector(true_size);
+  d2U_dt2 = mfem::Vector(true_size);
```
joshessman-llnl: Could this be a call to `SetSize`? 
Depending on how large `true_size` gets it might be good to avoid the double allocations due to the lack of a move assignment operator in `mfem::Vector` - as written the default constructor is called, then the temporary is constructed (allocation), and then there's a copy assignment (second allocation).

src/physics/nonlinear_solid.cpp
```diff
@@ -132,49 +148,147 @@ void NonlinearSolid::completeSetup()
...
+  nonlin_solver_.nonlinearSolver().iterative_mode = true;
+  nonlin_solver_.SetOperator(residual);
+
   if (timestepper_ == serac::TimestepMethod::QuasiStatic) {
-    nonlin_solver_.nonlinearSolver().iterative_mode = true;
-    nonlinear_oper_ = std::make_unique<NonlinearSolidQuasiStaticOperator>(std::move(H_form), bcs_);
-    nonlin_solver_.SetOperator(*nonlinear_oper_);
+    // the quasistatic case is entirely described by the residual,
+    // there is no ordinary differential equation
+    residual.function = [=](const mfem::Vector& u, mfem::Vector& r) mutable {
```
joshessman-llnl: The default lambda-capture makes me a little nervous, especially in a member function.  Capturing `[this]` may be more appropriate.  Also, it doesn't look like this lambda needs to be `mutable`.

src/physics/nonlinear_solid.cpp
```diff
@@ -132,49 +148,147 @@ void NonlinearSolid::completeSetup()
...
+      H->Mult(u, r);  // r := H(u)
+      r.SetSubVector(bcs_.allEssentialDofs(), 0.0);
+    };
+
+    residual.jacobian = [=](const mfem::Vector& u) mutable -> mfem::Operator& {
```
joshessman-llnl: At a glance, this lambda and the remaining lambdas can also capture a reference to the calling object + not need `mutable`.

src/physics/nonlinear_solid.cpp
```diff
@@ -132,49 +148,147 @@ void NonlinearSolid::completeSetup()
...
+      // TODO: we should avoid re-assembling this when not required
+      C_mat.reset(C->ParallelAssemble());
+
+      res = (*M_mat) * d2u_dt2 + (*C_mat) * (du_dt + c1 * d2u_dt2) + (*H) * (x + u + c0 * d2u_dt2);
+      res.SetSubVector(bcs_.allEssentialDofs(), 0.0);
+    };
+
+    residual.jacobian = [=](const mfem::Vector& d2u_dt2) mutable -> mfem::Operator& {
+      // J = M + c1 * C + c0 * H(u_predicted)
+      auto localJ = std::unique_ptr<mfem::SparseMatrix>(Add(1.0, M->SpMat(), c1, C->SpMat()));
```
joshessman-llnl: Separate PR of course, but it looks like there may be some utility for matrix expression templates.

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "numerics/expr_template_ops.hpp"
+#include "physics/utilities/boundary_condition_manager.hpp"
+#include "physics/utilities/equation_solver.hpp"
+
+class SecondOrderODE : public mfem::SecondOrderTimeDependentOperator {
+public:
+  using signature = void(const double, const double, const double, const mfem::Vector&, const mfem::Vector&,
+                         mfem::Vector&);
+  SecondOrderODE() : mfem::SecondOrderTimeDependentOperator(0, 0.0) {}
+  SecondOrderODE(int n, std::function<signature> f) : mfem::SecondOrderTimeDependentOperator(n, 0.0), f(f) {}
```
joshessman-llnl: It might be useful to also have an overloaded constructor accepting `std::function&&` and moving it into `f`

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,61 @@
...
+  {
+    f(t, 0.0, 0.0, u, du_dt, d2u_dt2);
+  }
+  void ImplicitSolve(const double c0, const double c1, const mfem::Vector& u, const mfem::Vector& du_dt,
+                     mfem::Vector& d2u_dt2)
+  {
+    f(t, c0, c1, u, du_dt, d2u_dt2);
+  }
+
+  std::function<signature> f;
```
joshessman-llnl: It looks like there's an invariant between the `n` in the constructor and the size of the `u` function argument - can this be private?

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,61 @@
...
+  {
+    int n = other_ode.Height();
+
+    // here, we assume that x and dx_dt have the following block form:
+    //
+    //     ⎡ du_dt ⎤          ⎡ d2u_dt2 ⎤
+    // x = ⎢  ---  ⎢, dx_dt = ⎢   ---   ⎢
+    //     ⎣   u   ⎦          ⎣  du_dt  ⎦
+    //
+    f = [=](const double t, const double dt, const mfem::Vector& x, mfem::Vector& dx_dt) {
```
joshessman-llnl: If `other_ode` and `n` are the only things that need to be captured, they can be listed explicitly in the capture group

src/physics/operators/odes.hpp
joshessman-llnl: As with `SecondOrderODE`, it might be useful to add constructor overloads that take rvalue references to a function/SecondOrderODE, depending on how you envision the constructors being used.

src/physics/operators/stdfunction_operator.hpp
joshessman-llnl: Yeah, it looks like there are class invariants that should maintained with `private`.  Would it be sufficient to just have a `StdFunctionOperator(n, function, jacobian)`?

src/physics/thermal_conduction.cpp
```diff
@@ -16,24 +16,36 @@ ThermalConduction::ThermalConduction(int order, std::shared_ptr<mfem::ParMesh> m
...
   } else {
-    SLIC_ERROR("ThermalCondution::SolverParameters did not contain a value");
+    setTimestepper(TimestepMethod::QuasiStatic);
   }
+
+  dt_          = 0.0;
+  previous_dt_ = -1.0;
+
+  int true_size = temperature_.space().TrueVSize();
+  u_            = mfem::Vector(true_size);
```
joshessman-llnl: `u_.SetSize(true_size);`?

src/physics/utilities/boundary_condition.hpp
```diff
@@ -191,6 +191,35 @@ class BoundaryCondition {
...
+  /**
+   * @brief Allocates an integrator of type "Integrator" on the heap,
+   * constructing it with the boundary condition's coefficient,
+   * intended to be passed to mfem::*LinearForm::Add*Integrator
+   * @return An owning pointer to the new integrator
+   * @pre Requires Integrator::Integrator(mfem::Coefficient&)
+   */
+  template <typename Integrator>
+  std::unique_ptr<Integrator> newIntegrator() const;
+
```
joshessman-llnl: These were removed a little while back - it doesn't look like they're used anywhere so this could be from a merge that didn't go quite right.

tests/time_integration_with_constraints.cpp
```diff
@@ -0,0 +1,435 @@
...
+      }
+
+      if (method == 2) {
+        acceleration[1] = (uc(t) - u[1]) / c0;
+      }
+    }
+
+    // while (norm(r(acceleration) > tolerance) {
+    //   acceleration -= J(acceleration)^{-1} * r(acceleration);
+    // }
```
joshessman-llnl: Can this be deleted?

joshessman-llnl (commented):
Just a few comments on lambda setup/opportunities for move semantics

src/coefficients/stdfunction_coefficient.cpp
samuelpmishLLNL: This PR predated our transition to mfem 4.2, so I used the implementation we already had. However, it would make more sense to use the mfem native options now that they're available.

src/physics/nonlinear_solid.cpp
samuelpmishLLNL: I'd like to speak to you about moving some of this stuff into the boundary condition manager class at some point, so I expect these lines of code to disappear. The time spent in the extra copy is insignificant in the overall scheme of things, but SetSize() is certainly just a better choice!

src/coefficients/stdfunction_coefficient.cpp
joshessman-llnl: Yep, it wasn't merged in until just yesterday.

src/physics/operators/odes.hpp
samuelpmishLLNL: I'll change it to "TypeSignature" to be more in agreement with the cs terminology. I would prefer to not append ODE to the prefix, since it would make the full name redundant
`SecondOrderODE::ODESignature`

I didn't really think about it until now, but I guess `using` statements follow the same naming conventions of other types, so I'll update them accordingly to CamelCase.

src/physics/operators/stdfunction_operator.hpp
samuelpmishLLNL: `mfem::Operator` is also frequently used for just its Mult function, so maybe
```
StdFunctionOperator(n, function)
StdFunctionOperator(n, function, jacobian)
```
is enough

src/physics/nonlinear_solid.cpp
```diff
@@ -140,49 +140,159 @@ void NonlinearSolid::completeSetup()
...
-    nonlinear_oper_                                 = buildQuasistaticOperator(std::move(H_form));
-    nonlin_solver_.SetOperator(*nonlinear_oper_);
+    // the quasistatic case is entirely described by the residual,
+    // there is no ordinary differential equation
+    residual_ = StdFunctionOperator(
+        displacement_.space().TrueVSize(),
+
+        // residual function
+        [this](const mfem::Vector& u, mfem::Vector& r) mutable {
+          H_->Mult(u, r);  // r := H(u)
```
joshessman-llnl, line 181: From what I can tell `H` was promoted to a member variable so these functions wouldn't have a dangling reference later.

For a later PR, it might be worth refining the ownership model here - it looks like the `StdFunctionOperator` can take ownership of `H` but both its functions need to use it so it can't be moved into either lambda.

src/physics/nonlinear_solid.cpp
```diff
@@ -140,49 +140,159 @@ void NonlinearSolid::completeSetup()
...
-    nonlin_solver_.nonlinearSolver().iterative_mode = true;
-    nonlinear_oper_                                 = buildQuasistaticOperator(std::move(H_form));
-    nonlin_solver_.SetOperator(*nonlinear_oper_);
+    // the quasistatic case is entirely described by the residual,
+    // there is no ordinary differential equation
+    residual_ = StdFunctionOperator(
+        displacement_.space().TrueVSize(),
+
+        // residual function
+        [this](const mfem::Vector& u, mfem::Vector& r) mutable {
```
joshessman-llnl: The `mutable` on all four of these lambdas can be removed, I think.

src/physics/nonlinear_solid.hpp
```diff
@@ -33,8 +34,9 @@ class NonlinearSolid : public BasePhysics {
    * @brief A timestep method and config for the M solver
    */
   struct DynamicSolverParameters {
-    TimestepMethod         timestepper;
-    LinearSolverParameters M_params;
+    TimestepMethod             timestepper;
+    DirichletEnforcementMethod enforcement_method;
+    LinearSolverParameters     M_params;
```
joshessman-llnl: This can be removed

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,114 @@
...
+   * @brief Default constructor for creating an uninitialized SecondOrderODE
+   */
+  SecondOrderODE() : mfem::SecondOrderTimeDependentOperator(0, 0.0) {}
+
+  /**
+   * @brief Constructor defining the size and specific system of ordinary differential equations to be solved
+   *
+   * @param[in] n The number of components in each vector of the ODE
+   * @param[in] f The function that describing how to solve for the second derivative, given the current state
+   *    its time derivative. The two functions
```
joshessman-llnl: Should this be
> given the current state **and** its time derivative

src/physics/thermal_conduction.cpp
```diff
@@ -16,25 +16,37 @@ ThermalConduction::ThermalConduction(int order, std::shared_ptr<mfem::ParMesh> m
...
-  if (auto K_params = std::get_if<LinearSolverParameters>(&params)) {
-    K_inv_ = EquationSolver(mesh->GetComm(), *K_params);
-    setTimestepper(TimestepMethod::QuasiStatic);
-  }
-  // Otherwise, two sets of parameters for the dynamic M/T solve
-  else if (auto dyn_params = std::get_if<DynamicSolverParameters>(&params)) {
-    setTimestepper(dyn_params->timestepper);
-    // Save these to initialize the DynamicConductionOperator later
-    dyn_M_params_ = dyn_params->M_params;
-    dyn_T_params_ = dyn_params->T_params;
```
joshessman-llnl, line 33: I think these member variables can be removed

src/physics/thermal_conduction.hpp
```diff
@@ -32,20 +33,52 @@ namespace serac {
...
-   * @brief A timestep method and configs for the M and T solvers
+   * @brief A timestep method and config for the M solver
    */
   struct DynamicSolverParameters {
-    TimestepMethod         timestepper;
-    LinearSolverParameters M_params;
-    LinearSolverParameters T_params;
+    TimestepMethod             timestepper;
+    DirichletEnforcementMethod enforcement_method;
+    LinearSolverParameters     M_params;
```
joshessman-llnl: I think this can be removed

src/physics/utilities/boundary_condition.cpp
```diff
@@ -48,18 +48,19 @@ void BoundaryCondition::project(FiniteElementState& state) const
...
 
-    if (component_ == -1) {
-      // If it contains all components, project the vector
-      auto vec_coef = std::get_if<std::shared_ptr<mfem::VectorCoefficient>>(&coef_);
-      SLIC_ASSERT_MSG(vec_coef,
-                      "Essential boundary condition contained all components but had a non-vector coefficient.");
-      state.gridFunc().ProjectCoefficient(**vec_coef, dof_list);  // Deref ptr-to-result, then shared_ptr
+    // the only reason to store a VectorCoefficient is to act on all components
+    if (std::holds_alternative<std::shared_ptr<mfem::VectorCoefficient>>(coef_)) {
+      auto vec_coef = std::get<std::shared_ptr<mfem::VectorCoefficient>>(coef_);
```
joshessman-llnl:
```cpp
if (auto vec_coef = std::get_if<std::shared_ptr<mfem::VectorCoefficient>>(&coef))
```

src/physics/utilities/boundary_condition.cpp
```diff
@@ -48,18 +48,19 @@ void BoundaryCondition::project(FiniteElementState& state) const
...
-                      "Essential boundary condition contained a single component but had a non-scalar coefficient.");
-      state.gridFunc().ProjectCoefficient(**scalar_coef, dof_list, component_);
+      // an mfem::Coefficient could be used to describe a scalar-valued function, or
+      // a single component of a vector-valued function
+      auto scalar_coef = std::get<std::shared_ptr<mfem::Coefficient>>(coef_);
+      if (component_ == -1) {
+        state.gridFunc().ProjectCoefficient(*scalar_coef, dof_list);
+      } else {
+        state.gridFunc().ProjectCoefficient(*scalar_coef, dof_list, component_);
+      }
```
joshessman-llnl: I ran into this in #287 and solved it [slightly differently](https://github.com/LLNL/serac/pull/287/files#diff-dd9bd49c73ab278f257c5c3c0c11dcf9e60388f04d4484a64bc0ade8a8d55387R57-R70), but I think I like this approach more.  Maybe they could be combined by condensing 59-63 with a ternary operator?

joshessman-llnl (commented):
Looking good, just a few minor things and then I think it's ready to go!

src/physics/nonlinear_solid.cpp
samuelpmish: I'm leaning toward keeping `H` a member variable of the physics module. Off the top of my head, design/topology optimization, eigenvalue problems, and custom preconditioners may want to know something about `H`, and it might get trickier to access if we narrow its scope any more.

src/physics/nonlinear_solid.cpp
samuelpmish: whoops, I got the thermal_conduction ones but missed these!

src/physics/utilities/boundary_condition.cpp
samuelpmish: changed to
```cpp
state.gridFunc().ProjectCoefficient(*scalar_coef, dof_list, std::max(0, component_));
```

src/physics/utilities/boundary_condition.cpp
samuelpmish: how about this? I'd like to keep the symmetry of the if/else that comes with using `std::get<>`
```cpp
    if (is_vector_valued(coef_)) {
      auto vec_coef = std::get<std::shared_ptr<mfem::VectorCoefficient>>(coef_);
      state.gridFunc().ProjectCoefficient(*vec_coef, dof_list);
    } else {
      auto scalar_coef = std::get<std::shared_ptr<mfem::Coefficient>>(coef_);
      state.gridFunc().ProjectCoefficient(*scalar_coef, dof_list, std::max(0, component_));
    }
```
these two functions get added under the definition of `GeneralCoefficient`:
```cpp
inline bool is_scalar_valued(const GeneralCoefficient & coef) {
  return std::holds_alternative<std::shared_ptr<mfem::Coefficient>>(coef);
}

inline bool is_vector_valued(const GeneralCoefficient & coef) {
  return std::holds_alternative<std::shared_ptr<mfem::VectorCoefficient>>(coef);
}
```

src/physics/nonlinear_solid.hpp
samuelpmish: What do you mean?

src/physics/nonlinear_solid.cpp
joshessman-llnl: I think once these are removed the `mutable` on the `StdFunctionOperator` members can also be removed.

src/physics/nonlinear_solid.hpp
joshessman-llnl: It doesn't look like the `M_params` is used anywhere - looks like GitHub displays the previous two lines but I just meant 39.

samuelpmish: ah, that makes more sense!

src/physics/nonlinear_solid.hpp
```diff
@@ -233,14 +224,92 @@ class NonlinearSolid : public BasePhysics {
...
+   * @brief the previous acceleration, used as a starting guess for newton's method
+   */
+  mfem::Vector previous_;
+
+  // TODO: wrap mfem's second order ODE solvers to incorporate this boundary condition info
+  //
+  // temporary values used to compute finite difference approximations
+  // to the derivatives of constrained degrees of freedom
+  mfem::Vector U_minus_;
+  mfem::Vector U_;
```
jamiebramwell, line 304: I'm still worried about collisions with `u_`. Maybe `u_temp_`?

jamiebramwell: These members also still need doxygen-style comments.

src/physics/nonlinear_solid.cpp
jamiebramwell: I agree. There are a lot of times when something other than the operator may want stiffness information.

src/physics/nonlinear_solid.cpp
```diff
@@ -140,49 +140,159 @@ void NonlinearSolid::completeSetup()
...
+              U_       = u_;
+            }
+
+            if (enforcement_method_ == DirichletEnforcementMethod::RateControl) {
+              d2U_dt2_ = (dU_dt_ - du_dt_) / c1_;
+              dU_dt_   = du_dt_;
+              U_       = u_;
+            }
+
+            if (enforcement_method_ == DirichletEnforcementMethod::FullControl) {
```
jamiebramwell, line 261: We should add an error check if the given `enforcement_method_` is not found.

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,114 @@
...
+   *    and its time derivative. The two functions
+   *
+   *      mfem::SecondOrderTimeDependentOperator::Mult and mfem::SecondOrderTimeDependentOperator::ImplicitSolve
+   *      (described in more detail here:
+   * https://mfem.github.io/doxygen/html/classmfem_1_1SecondOrderTimeDependentOperator.html)
+   *
+   *    are consolidated into a single std::function, where
+   *
+   *      mfem::SecondOrderTimeDependentOperator::Mult corresponds to the case where fac0, fac1 are both zero
+   *      mfem::SecondOrderTimeDependentOperator::Mult corresponds to the case where either of fac0, fac1 are nonzero
```
jamiebramwell: It think this should be `ImplicitSolve`.

src/physics/operators/odes.hpp
```diff
@@ -0,0 +1,114 @@
...
+   * @param[in] f The function that describing how to solve for the first derivative, given the current state.
+   *    The two functions
+   *
+   *      mfem::TimeDependentOperator::Mult and mfem::TimeDependentOperator::ImplicitSolve
+   *      (described in more detail here: https://mfem.github.io/doxygen/html/classmfem_1_1TimeDependentOperator.html)
+   *
+   *    are consolidated into a single std::function, where
+   *
+   *      mfem::TimeDependentOperator::Mult corresponds to the case where dt is zero
+   *      mfem::TimeDependentOperator::Mult corresponds to the case where dt is nonzero
```
jamiebramwell: I think this should also be `ImplicitSolve`.

src/physics/operators/stdfunction_operator.hpp
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include <functional>
+
+#include "mfem.hpp"
+
+/**
+ * @brief StdFunctionOperator is a class wrapping mfem::Operator
```
jamiebramwell, line 8: This is such a great functionality! 🎉

src/physics/thermal_conduction.hpp
jamiebramwell: This comment still stands. `U_temp_` would fix it.

src/physics/thermal_conduction.cpp
jamiebramwell: If this is needed, a comment saying why would be useful.

src/physics/thermal_conduction.hpp
```diff
@@ -204,16 +234,24 @@ class ThermalConduction : public BasePhysics {
    */
   std::unique_ptr<mfem::Coefficient> mass_coef_;
 
-  /**
-   * @brief Configuration for dynamic equation solvers
-   */
-  std::optional<LinearSolverParameters> dyn_M_params_;
-  std::optional<LinearSolverParameters> dyn_T_params_;
+  FirstOrderODE ode_;
```
jamiebramwell, line 242: These need doxygen-style comments.

jamiebramwell:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

src/physics/nonlinear_solid.hpp
jamiebramwell: After this PR is merged, these should go away as the MFEM second order ODE solver will be wrapped.

src/physics/operators/odes.hpp
jamiebramwell: This is still outstanding.

src/physics/operators/odes.hpp
jamiebramwell: This is still outstanding.

src/physics/thermal_conduction.cpp
jamiebramwell: Is this array needed?

src/physics/thermal_conduction.hpp
jamiebramwell: These still need doxygen comments.

samuelpmishLLNL:
LGTM

samuelpmishLLNL:
LGTM

jamiebramwell (approved):
I'm super excited to use this!!!

src/physics/thermal_conduction.cpp
```diff
@@ -130,47 +162,106 @@ void ThermalConduction::completeSetup()
...
+
+        [this](const mfem::Vector & /*du_dt*/) -> mfem::Operator& {
+          if (dt_ != previous_dt_) {
+            J_.reset(mfem::Add(1.0, *M_, dt_, *K_));
+            bcs_.eliminateAllEssentialDofsFromMatrix(*J_);
+          }
+          return *J_;
+        });
+
+    ode_ = FirstOrderODE(temperature_.trueVec().Size(), [=](const double t, const double dt, const mfem::Vector& u,
```
joshessman-llnl: Can we capture `[this]` here for clarity?

joshessman-llnl (approved):

src/physics/thermal_conduction.cpp
samuelpmishLLNL: good call, I missed this one

samuelpmishLLNL:
LGTM

PR #217, opened 2020-09-24
Remove global partitioning from our HYPRE builds

jamiebramwell:
This removes the global partitioning option that we had patched into the hypre spack package. This option is generally not scalable and it causes issues with creating Hypre matricies from block systems. The default install directory has also been added to the `.gitignore` file.


jamiebramwell:
LGTM

joshessman-llnl (approved):

jandrej (approved):

jamiebramwell:
For some reason, this is failing the gcc TPL install on quartz. 🤷

jamiebramwell:
LGTM

jamiebramwell:
I believe the azure failure is due to a microsoft outage.

PR #219, opened 2020-09-25
Detect clang-format-10 on Debian/Ubuntu

### LINKED ISSUE #198: Check if clang-format-10 exists in uberenv
On non-LC systems, we should check if clang-format-10 exists, and append it to the host config if it does.

joshessman-llnl:
To my knowledge no current serac users use a distro that is not Ubuntu or Debian (when not on LC).  If/when this changes we can check the version of the binary as described in my comment.

It may also be useful to add an option to the `blt_add_code_checks` macro to require a particular version of `clang-format` and explicit version options for the other formatting tools BLT supports, e.g. `astyle`.

Resolves #198

### DIFF
--- a/scripts/uberenv/packages/serac/package.py
+++ b/scripts/uberenv/packages/serac/package.py
@@ -410,9 +410,14 @@ class Serac(CMakePackage, CudaPackage):
         else:
             cfg.write(cmake_cache_option("ENABLE_DOCS", False))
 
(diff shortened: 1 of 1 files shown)


jamiebramwell (approved):

white238 (approved):

joshessman-llnl:
LGTM

PR #220, opened 2020-09-25
Pass (and store) all coefficients as non-owning references for uniformity

### LINKED ISSUE #192: Move the coefficient interface to value semantics
After looking at the fallout of using pointer semantics to store coefficients in solvers, I think it's better for drivers to own them and pass the coefficients to the solvers by lvalue reference for ease of use improvements.

joshessman-llnl:
Because coefficients are created in the driver and have essentially the same lifetime as the program, there is no need to transfer ownership to the solver classes.  Per #192, passing all coefficients by reference allows the user to not mess with smart pointer ownership semantics which as of right now are not uniform (sometimes `shared_ptr`, other times `unique_ptr`).

See https://github.com/LLNL/serac/compare/feature/essman/coefs_by_ref?expand=1#diff-e3294c291ae5e7defca270eae18097c6 for an example of how these changes allow for a more uniform user interface.

Resolves #192


tests/serac_thermal_solver.cpp
```diff
@@ -105,8 +105,8 @@ TEST(thermal_solver, static_solve_multiple_bcs)
   therm_solver.setTemperatureBCs(marked_2, u_1);
 
   // Set the conductivity of the thermal operator
-  auto kappa = std::make_unique<mfem::ConstantCoefficient>(0.5);
-  therm_solver.setConductivity(std::move(kappa));
+  mfem::ConstantCoefficient kappa(0.5);
```
jamiebramwell, line 108: I think this is much clearer.

jamiebramwell (approved):
I think this provides a much cleaner interface with physics objects. However, it does add the requirement that the lifetime of driver-defined coefficients cannot be extended by the individual physics modules. I think this is a reasonable requirement given how much it cleans up the code, but I look forward to hearing other people's opinions on this. Thanks for putting this together @joshessman-llnl !

jamiebramwell:
Another thought - do the coefficients have copy constructors? If it's just a lightweight wrapper to a data pointer, maybe that's the way to go.

jandrej:
> Another thought - do the coefficients have copy constructors? If it's just a lightweight wrapper to a data pointer, maybe that's the way to go.

Imho the default copy constructor should work for coefficients by design. I usually pass them around by value since the skeleton type allocation is negligible.

joshessman-llnl:
> 
> 
> Another thought - do the coefficients have copy constructors? If it's just a lightweight wrapper to a data pointer, maybe that's the way to go.

The `mfem::Coef/VecCoef` class hierarchy would need to be modified for polymorphic copying, plus the base class is pure virtual.  @samuelpmishLLNL is looking at type erasure as an alternative that allows for ownership to be taken.

jamiebramwell:
Any thoughts on this @white238 @samuelpmish @jandrej ?

samuelpmish:
Josh and I spoke yesterday morning about this PR, and how to manage the lifetime of the coefficient classes. To summarize, it looks like most of the mfem::Coefficient copy constructors are shallow copies, which makes it impossible to take ownership of a coefficient that is passed in. One idea that we discussed was to not actually pass in mfem::Coefficients at all, and instead construct them in the solvers. An example of the difference is given below:

previous implementation
```cpp
// solver class
void MySolver::SetTemperature(const mfem::Coefficient & coef) {
  ... // do something, but we can't really copy what the user passes in, so we're stuck with references or pointers
}

-----------------------

// user-code
mfem::ConstantCoefficient values(3.14); // the user must keep this around or the solver will have a dangling reference
solver.setTemperature(values);

or

serac::StdFunctionCoefficient([](auto x){ return x(1); }); // the user must keep this around or the solver will have a dangling reference
solver.setTemperature(values);
```

proposed implementation
```cpp
struct coef {
  public:
    coef (double x) : self_(std::make_unique<mfem::ConstantCoefficient>(x)) {}
    coef (std::function <double(double)> f) : self_(std::make_unique<serac::StdFunctionCoefficient>(f)) {}
    operator mfem::Coefficient&() { return *self_; }
    operator mfem::Coefficient&() const { return *self_; }
    std::unique_ptr<mfem::Coefficient> self_;
};

-----------------------

// solver class
void MySolver::SetTemperature(const coef & c) {
  ... // doubles and std::functions can be copied, so the solver can own and manage the lifetime directly
}

-----------------------

// user code
solver.setTemperature(3.14);

or

solver.setTemperature([](auto x){ return x(1); });
```

This approach also has the benefit of removing more mfem data types from the solver interface.

This is also a big change to serac though, and probably warrants a discussion from the other developers.

jamiebramwell:
> 
> This is also a big change to serac though, and probably warrants a discussion from the other developers.

I like the idea of hiding the MFEM data structures from users. However, a big benefit of MFEM is the flexibility of the `Coefficient` class. For instance with the proposed interface, a user couldn't set of a `GridFunctionCoefficient` which would be required for some multiphysics applications. Additionally, I know LiDO is looking at coefficient extensions to enable optimizations work flows. I will put this on the schedule for a larger discussion on Oct. 8.

samuelpmish:
if the gridfunction is copiable, then it could have a constructor in the `coef` struct and still fit into the proposed pattern above

joshessman-llnl:
> 
> 
> if the gridfunction is copiable, then it could have a constructor in the `coef` struct and still fit into the proposed pattern above

GridFunctions have a deep copy constructor, but are much more expensive than copying a `double` or a `std::function`.   Implementing a vector equivalent of the `coef` class above (to wrap `VectorConstantCoefficient`, for example) will also require vector copies, though those vectors are probably not very large.

jamiebramwell:
I think after the discussion today, it's best to skip this referenced-based step in favor of the adapter implementation outlined by @samuelpmishLLNL . We should get some use cases together to scope out the desired functionality.

jamiebramwell:
I'm closing this one for now in favor of the adapter solution outlined above. The new issue for this approach is here: https://github.com/LLNL/serac/issues/252

PR #223, opened 2020-09-28
Switch to injected EquationSolvers, add custom mfem::Solver override

### LINKED ISSUE #31: Add sensible defaults for linear and nonlinear solves

### LINKED ISSUE #186: Allow custom MFEM solvers
We should add hooks for user-specific custom linear solvers and preconditioners, e.g. set by a generic `mfem::solver` object.

### LINKED ISSUE #187: Clean up the storage of solver parameters
It is a little bit confusing to have parameters that effectively do nothing for a given setting of "lin_solver" (e.g. I assume the value of rel_tol has no effect on SuperLU). Another option is to make each method its own type, e.g.

```cpp
namespace Method {

struct GMRES{
  double relative_tolerance;
  double absolute_tolerance;
  int max_iterations;
  int restart_iterations;
};

struct CG{
  double relative_tolerance;
  double absolute_tolerance;
  int max_iterations;
};

struct SuperLU{};

}
```

```cpp
struct EquationSolver{
...
};
```

_Originally posted by @samuelpmishLLNL in https://github.com/LLNL/serac/pull/173#discussion_r485050950_

joshessman-llnl:
It didn't make sense to add a `setLinearSolver` function to all the physics modules that just acts as a passthrough to the `EquationSolver`, so I thought this would be a good time to switch to injected EquationSolvers as constructor arguments.

Resolves #186, addresses part of #118 

By removing the dependency of the operator solvers on the "main" solver, this also resolves #187 - each of the three solver types (direct, iterative, custom) have their own set of parameters, that is, iterative parameters will no longer be unconditionally required.

This PR also adds default parameter sets for each of the solvers and resolves #31.

#### Breaking changes and how to fix them

 - Physics modules should now define a `Parameters` member type that encapsulates the `LinearSolverParams` and optionally the `NonlinearSolverParams` required for the `EquationSolver` member, in addition to any solver params used in a separate operator class, and a `TimeStepMethod` for dynamic simulations.  For example, if a physics module supports only quasistatic mode and uses a single nonlinear solver, the type might be defined as `std::tuple<LinearSolverParams, NonlinearSolverParams>`.
A physics module that uses one linear solver in the time-dependent operator and one linear solver in quasistatic mode might use
`std::variant<LinearSolverParams, std::pair<serac::TimestepMethod, LinearSolverParams>>`.
- This `Parameters` member should be added as a constructor argument
- The `solver_` member of the `BasePhysics` class was removed.  Any physics modules inheriting from this class should add an `EquationSolver` member if they used `solver_`
 - Any `set*SolverParams` functions can be removed from physics modules.  If associated member variables are not used after construction (e.g. to initialize an operator), they should also be removed.
- Before constructing an `EquationSolver`, the parameters should be augmented by passing them to `augmentAMGWithSpace`, e.g. `const auto& augmented_params = augmentAMGWithSpace(params, fe_space_to_use_with_AMG);`
- The default solver parameters should be added to the class as `static constexpr` member variables.

Examples of all the above can be found in the `Elasticity`, `ThermalConduction`, and `NonlinearSolid` modules


samuelpmish:
From our discussion earlier, we might also want to introduce some sensible defaults for the solver parameters. One approach might be to define physics-specific defaults as const static member variables, e.g.

```cpp
LinearSolverParams lin_params = PhysicsModule::DefaultLinearSolverParams;
NonlinearSolverParams nonlin_params  = PhysicsModule::DefaultNonlinearSolverParams;
EquationSolver eqn_solver(MPI_COMM_WORLD, lin_params, nonlin_params);
PhysicsModule m(/*prev params */, eqn_solver);
```

or take the parameters as constructor arguments, and initialize the EquationSolver inside the Physics module:

```cpp
LinearSolverParams lin_params = PhysicsModule::DefaultLinearSolverParams;
NonlinearSolverParams nonlin_params  = PhysicsModule::DefaultNonlinearSolverParams;
PhysicsModule m(/*prev params */, lin_params, nonlin_params);
```

src/physics/thermal_solid.hpp
```diff
@@ -30,8 +30,12 @@ class ThermalSolid : public BasePhysics {
    *
    * @param[in] order The order of the temperature and displacement discretizations
    * @param[in] mesh The parallel mesh object on which to solve
+   * @param[in] therm_system_solver The system solver instance for the conduction physics
```
white238: typo on this line and next.

src/physics/base_physics.hpp
```diff
@@ -42,6 +42,7 @@ class BasePhysics {
    * @param[in] mesh The primary mesh
    * @param[in] n Number of state variables
    * @param[in] p Order of the solver
+   * @param[in] solver The system solver instance
    */
   BasePhysics(std::shared_ptr<mfem::ParMesh> mesh, int n, int p);
```
samuelpmishLLNL: comment doesn't match function signature

src/physics/thermal_conduction.hpp
```diff
@@ -31,13 +31,26 @@ namespace serac {
  */
 class ThermalConduction : public BasePhysics {
 public:
+  /**
+   * @brief A timestep method and configs for the M and T solvers (in that order)
+   */
+  using DynamicParameters = std::tuple<TimestepMethod, LinearSolverParameters, LinearSolverParameters>;
```
jamiebramwell: I think a struct instead of a tuple would improve readability.

src/physics/utilities/equation_solver.cpp
```diff
@@ -14,23 +14,57 @@ namespace serac {
...
+        using Param = std::decay_t<decltype(lin_params)>;
+        if constexpr (std::is_same_v<Param, IterativeSolverParameters>) {
+          lin_solver_ = buildIterativeLinearSolver(comm, lin_params);
+
+          // Set the preconditioner if it's an iterative solver
+          std::visit(
+              [this, &lin_params](const auto& prec) {
+                using Prec = std::decay_t<decltype(prec)>;
+                if constexpr (std::is_same_v<Prec, HypreBoomerAMGPrec>) {
+                  SLIC_ERROR_IF(prec.pfes == nullptr, "FESpace is required to use the HypreBoomerAMG preconditioner.");
```
white238: These will not actually exit the program due to us turning off abort on error here:

https://github.com/LLNL/serac/blob/f460cd95b6dcee8207ad52e9968eba20cd93385b/src/infrastructure/logger.cpp#L70

Hopefully we can move forward in axom with providing an exit function to slic but that hasnt happened.

src/physics/nonlinear_solid.hpp
```diff
@@ -28,13 +28,26 @@ namespace serac {
...
 public:
+  /**
+   * @brief A timestep method and config for the M solver (in that order)
+   */
+  using DynamicParameters = std::tuple<TimestepMethod, LinearSolverParameters>;
+  /**
+   * @brief A configuration variant for the various solves
+   * Either quasistatic, or time-dependent with timestep and M params
+   */
+  using NonlinearSolidParameters =
```
jamiebramwell: I'm a bit concerned about using this as a user-facing data type. We now expect admittedly non-expert C++ users to understand nested `tuples` with an embedded `std::optional`. I think a plain parameter struct might be more straightforward.

src/physics/nonlinear_solid.hpp
```diff
@@ -132,61 +137,97 @@ class NonlinearSolid : public BasePhysics {
    */
   virtual ~NonlinearSolid();
 
-protected:
   /**
-   * @brief Velocity field
+   * @brief The default parameters for an iterative linear solver
    */
-  std::shared_ptr<FiniteElementState> velocity_;
+  constexpr static IterativeSolverParameters default_qs_linear_params = {
```
jamiebramwell, line 157: @white238 should we define default this way or through the upcoming Inlet PR?

jamiebramwell: Would it be better to wait for that one to go in first?

src/physics/utilities/equation_solver.hpp
```diff
@@ -167,9 +153,9 @@ class EquationSolver : public mfem::Solver {
   std::unique_ptr<mfem::Solver> prec_;
 
   /**
-   * @brief The linear solver object, either direct (SuperLU) or iterative
+   * @brief The linear solver object, either custom, direct (SuperLU), or iterative
```
jamiebramwell, line 160: 👍

tests/serac_quasistatic_solver.cpp
```diff
@@ -136,17 +116,80 @@ TEST(nonlinear_solid_solver, qs_direct_solve)
...
+  custom_solver->SetMaxIter(custom_params.max_iter);
+  custom_solver->SetPrintLevel(custom_params.print_level);
+
+  // Define the solver object
+  NonlinearSolid solid_solver(
+      1, pmesh,
+      // Compiler cannot deduce through two layers of indirection so explicitly initialize the CustomParams here
+      // Note that a non-owning pointer is used and that the lifetime of the physics module is dependent on that
+      // of the driver
+      {CustomSolverParameters{custom_solver.get()}, NonlinearSolid::default_qs_nonlinear_params, std::nullopt});
```
jamiebramwell: I think this is a bit to convoluted for user-facing code. A struct of parameters (or more likely an inlet object) and a solver pointer would be a cleaner interface.

jamiebramwell (commented):
My biggest concern is the exposure of `std::optional` and nested `std::tuple`s in the driver code. Additionally, physics developers will probably look into the `physics` source file to see how to implement a new one, and they would be expected to learn the `std::variant` logic. I think a bunch of these concerns will go away with @white238 's inlet PR, so it may be best to wait for that one to go in first.

Is it possible to separate the custom solver work from the parameter stuff? Or should that wait as well? Thoughts?

tests/serac_quasistatic_solver.cpp
joshessman-llnl: I agree - this has been changed to a struct:
```cpp
NonlinearSolid::NonlinearSolidParameters params;
params.H_lin_params    = CustomSolverParameters{custom_solver.get()};
params.H_nonlin_params = NonlinearSolid::default_qs_nonlinear_params;

// Define the solver object
NonlinearSolid solid_solver(1, pmesh, params);
```

src/physics/nonlinear_solid.hpp
joshessman-llnl: Fortunately, the compiler bug I encountered doesn't affect user-facing code, so I was able to remove the remaining `std::tuple` and switch to a struct with named fields.

src/physics/nonlinear_solid.hpp
joshessman-llnl: Currently, the inlet integration supports a single set of defaults (as opposed to different sets for quasistatic/dynamic), so there will be more duplication (at least in the tests) with an inlet-defaults-only approach.

In the tests though, we could just use actual input files.

jamiebramwell:
After develop gets merged in, is this ready for review again?

joshessman-llnl:
> 
> 
> After develop gets merged in, is this ready for review again?

The new stdlib (introduced as part of the inlet PR) is slightly broken.  Instead of trying to work around that directly, I think it makes more sense to switch to input files for the defaults, but reading structs in from inlet is still in development.  Possibly I could just remove all the default solver parameters that were introduced per https://github.com/LLNL/serac/pull/223#issuecomment-700271819

joshessman-llnl:
Closing (at least temporarily), extended by #243

PR #225, opened 2020-09-30
Add basic input deck functionality

white238:
This adds basic input deck functionality via Inlet.  I only implemented what the serac driver used as defaults.  It required a few other changes to get to this point and unfortunately will require some spack version locking and rebuilding all the host-configs everywhere so that is coming.

Other random changes:
* Rename serac back to serac (this required a BLT fix and updating our BLT submodule)
* Move meshes into a subdirectory of data
* Add new data/decks directory
* Build Axom with C++14 so that Sol is built (required by LuaReader)
* Remove [INFO] from the start of all SLIC_INFO messages
* Move CLI to one place and greatly reduce it since it should be in the input file
* Add verbose option to the build scripts
* Fix some paths in the spack configs for the docker containers.
* Add verbose flag to build scripts
* Fix some paths in the spack configs for the docker containers


scripts/uberenv/spack_configs/toss_3_x86_64_ib/packages.yaml
```diff
@@ -26,9 +26,9 @@ packages:
     buildable: false
   mvapich2:
     externals:
-    - spec: mvapich2@2.3%clang@10.0.0 arch=linux-rhel7-ivybridge
+    - spec: mvapich2@2.3%clang@10.0.0 process_managers=slurm arch=linux-rhel7-ivybridge
```
white238, line 29: This missing causes very subtle problems if you are trying to find the mpi wrapper via the process manager which was suggested by Greg for compatibility issues.  Axom does this and we should eventually.

src/drivers/serac.cpp
```diff
@@ -15,113 +15,97 @@
...
+  // Initialize Inlet and read input file
+  auto inlet = serac::input::initialize(datastore, input_file_path);
+  serac::defineInputFileSchema(inlet, rank);
+
+  // Save input values to file
+  datastore->getRoot()->save("serac_input", "json");
+
+  // serial and parallel refinement levels
+  int ser_ref_levels;
+  inlet->get("ser_ref_levels", ser_ref_levels);  // has default value
```
white238, line 86: Normally you would check if this was found (Inlet::get returns a boolean), but most of these either have default values or are required in the input file.  I didn't really want to put a bunch of boiler plate code around all these calls so I added a comment why they should not fail.  Thoughts?

src/drivers/serac.cpp
```diff
@@ -158,16 +148,42 @@ int main(int argc, char* argv[])
...
+  inlet->get("nonlinear_solid/K", K);    // has default value
+
   // Set the material parameters
   solid_solver.setHyperelasticMaterialParameters(mu, K);
 
-  // Set the linear solver parameters
-  if (gmres_solver == true) {
+  // Solver parameters
+  serac::NonlinearSolverParameters nonlin_params;
+  inlet->get("nonlinear_solid/solver/nonlinear/rel_tol", nonlin_params.rel_tol);          // has default value
```
white238, line 157: This kind of shows a needed functionality in Inlet to get a Table then find child values from that given table.  Don't know how common a very nested value would be but I'm going to create an issue on this for future work.

Created https://github.com/LLNL/axom/issues/361

white238:
Incoming change that moves us completely away from using git branches in our dependencies which is causing us problems of chasing moving targets and the git mirror issue.

src/drivers/serac.cpp
jamiebramwell: I think it would be extremely common.

src/drivers/serac.cpp
joshessman-llnl: I don't have a problem with this, but additionally perhaps inlet could have a mode of operation where it aborts the program if a variable is not found and has no default, i.e. one mode where the user should be checking the return values, and another where the user wants to "assert" that all `get`s succeed.

src/drivers/serac.cpp
```diff
@@ -158,16 +148,42 @@ int main(int argc, char* argv[])
...
+  inlet->get("nonlinear_solid/solver/nonlinear/abs_tol", nonlin_params.abs_tol);          // has default value
+  inlet->get("nonlinear_solid/solver/nonlinear/max_iter", nonlin_params.max_iter);        // has default value
+  inlet->get("nonlinear_solid/solver/nonlinear/print_level", nonlin_params.print_level);  // has default value
+
+  serac::LinearSolverParameters lin_params;
+  inlet->get("nonlinear_solid/solver/linear/rel_tol", lin_params.rel_tol);          // has default value
+  inlet->get("nonlinear_solid/solver/linear/abs_tol", lin_params.abs_tol);          // has default value
+  inlet->get("nonlinear_solid/solver/linear/max_iter", lin_params.max_iter);        // has default value
+  inlet->get("nonlinear_solid/solver/linear/print_level", lin_params.print_level);  // has default value
```
joshessman-llnl, line 166: Not sure if it belongs in this PR, but maybe it's possible to have a a constructor for `LinearSolverParams` that accepts the relevant Sidre subtree (in this case `nonlinear_solid/solver/linear`)

data/decks/default.lua
```diff
@@ -0,0 +1,40 @@
+mesh = "../meshes/beam-hex.mesh"
+
+-- serial and parallel refinement levels
+ser_ref_levels = 0
+par_ref_levels = 0
+
+-- polynomial interpolation order
+order = 1
```
jamiebramwell: This is related with `nonlinear_solid`. We could have multiple physics modules with different polynomial orders.

src/drivers/serac.cpp
```diff
@@ -158,16 +148,42 @@ int main(int argc, char* argv[])
...
+
+  // solver input args
+  std::string solver_type;
+  inlet->get("nonlinear_solid/solver/linear/solver_type", solver_type);  // has default value
+  std::cout << "~~ Solver Type = " << solver_type << std::endl;
+  if (solver_type == "gmres") {
     lin_params.prec       = serac::Preconditioner::BoomerAMG;
     lin_params.lin_solver = serac::LinearSolver::GMRES;
-  } else {
+  } else if (solver_type == "slu") {
```
joshessman-llnl: I think this should be "minres"

data/decks/default.lua
```diff
@@ -0,0 +1,40 @@
+mesh = "../meshes/beam-hex.mesh"
```
jamiebramwell: Do we want to logically connect this with the refinement levels in a `mesh` table?

jamiebramwell: (I'm thinking these lua tables are kind of like input deck blocks)

scripts/uberenv/spack_configs/toss_3_x86_64_ib/packages.yaml
jamiebramwell: yes, I understand some of those words.... 😄 I trust you on this one.

src/infrastructure/cli.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "infrastructure/logger.hpp"
+#include "infrastructure/terminator.hpp"
+
+namespace serac {
+
+namespace cli {
+
+//------- Command Line Interface -------
+
+std::shared_ptr<std::unordered_map<std::string, std::string>> defineAndParse(int argc, char* argv[], int rank)
```
joshessman-llnl: Do we need shared ownership of this and the Inlet object?  From what I can see, the lifetime of the driver and the lifetime of the inlet object are closely coupled enough that the driver can be responsible for it, in which case both this `unordered_map` and the `inlet` object can be on the stack (declared in `main`, returned by value from these utility functions).

src/drivers/serac.cpp
```diff
@@ -15,113 +15,97 @@
...
 #include "serac_config.hpp"
 
-int main(int argc, char* argv[])
-{
-  auto [num_procs, rank] = serac::initialize(argc, argv);
+namespace serac {
 
+//------- Input file -------
+
+void defineInputFileSchema(std::shared_ptr<axom::inlet::Inlet> inlet, int rank)
```
jamiebramwell, line 38: If I understand, this is where a driver code would add any input toggles that are not in the physics. It is probably a good idea to move the mesh stuff to another helper function in `numerics/mesh_utils.cpp`

src/drivers/serac.cpp
```diff
@@ -15,113 +15,97 @@
...
-  lin_params.print_level = 0;
+  // Physics
+  serac::NonlinearSolid::defineInputFileSchema(inlet);
 
-  // solver input args
-  bool gmres_solver = true;
-  bool slu_solver   = false;
+  // Verify input file
+  if (!inlet->verify()) {
+    SLIC_ERROR_ROOT(rank, "Input deck failed to verify.");
```
jamiebramwell: Should we call them input files or input decks? Decks is the historical term, but it is quickly becoming obsolete.

src/drivers/serac.cpp
```diff
@@ -15,113 +15,97 @@
...
 
-  // neo-Hookean material parameters
-  double mu = 0.25;
-  double K  = 5.0;
+}  // namespace serac
 
-  // loading parameters
-  double tx = 0.0;
-  double ty = 1.0e-3;
-  double tz = 0.0;
```
jamiebramwell, line 68: This fills my heart with joy!

src/drivers/serac.cpp
```diff
@@ -158,16 +148,42 @@ int main(int argc, char* argv[])
...
+  std::string solver_type;
+  inlet->get("nonlinear_solid/solver/linear/solver_type", solver_type);  // has default value
+  std::cout << "~~ Solver Type = " << solver_type << std::endl;
+  if (solver_type == "gmres") {
     lin_params.prec       = serac::Preconditioner::BoomerAMG;
     lin_params.lin_solver = serac::LinearSolver::GMRES;
-  } else {
+  } else if (solver_type == "slu") {
     lin_params.prec       = serac::Preconditioner::Jacobi;
     lin_params.lin_solver = serac::LinearSolver::MINRES;
```
jamiebramwell, line 176: This should be `serac::LinearSolver::SuperLU`.

src/drivers/serac.cpp
samuelpmishLLNL: do you think it's possible to support a syntax like

```cpp
auto nonlin_params = inlet->get<serac::NonlinearSolverParameters>("nonlinear_solid/solver/nonlinear");
```

it's similar to what https://github.com/nlohmann/json does, and is reminiscent of `std::get<T>` in the standard library.

(the user provides a definition of the conversion to a specified type somewhere)

src/infrastructure/input.cpp
```diff
@@ -0,0 +1,75 @@
...
+  luareader->parseFile(input_file_path);
+
+  // Store inlet data under its own group
+  axom::sidre::Group* inlet_root  = datastore->getRoot()->createGroup("input_file");
+  auto                serac_inlet = std::make_shared<axom::inlet::Inlet>(luareader, inlet_root);
+
+  return serac_inlet;
+}
+
+std::string findMeshFile(const std::string& mesh_path, const std::string& input_file_path)
```
jamiebramwell: I think a better name for this function is needed. `findMeshFilePath`?

src/physics/utilities/equation_solver.cpp
```diff
@@ -126,4 +126,21 @@ mfem::Operator& EquationSolver::SuperLUNonlinearOperatorWrapper::GetGradient(con
...
+  nonlinear_table->addInt("max_iter", "Maximum iterations for the Newton solve.")->defaultValue(500);
+  nonlinear_table->addInt("print_level", "Nonlinear print level.")->defaultValue(0);
+  nonlinear_table->addString("solver_type", "Not currently used.")->defaultValue("");
+
+  auto linear_table = schema_creator->addTable("linear", "Linear Equation Solver Parameters")->required(true);
+  linear_table->addDouble("rel_tol", "Relative tolerance for the linear solve.")->defaultValue(1.0e-6);
+  linear_table->addDouble("abs_tol", "Absolute tolerance for the linear solve.")->defaultValue(1.0e-8);
+  linear_table->addInt("max_iter", "Maximum iterations for the linear solve.")->defaultValue(5000);
+  linear_table->addInt("print_level", "Linear print level.")->defaultValue(0);
+  linear_table->addString("solver_type", "Solver type (gmres|slu).")->defaultValue("gmres");
```
joshessman-llnl: `slu` is on its way out, but this should at least include `minres` since that's currently exposed in the driver

src/drivers/serac.cpp
joshessman-llnl: The json-analogous way of doing this is defining two free functions `void to_inlet(const T& t, sidre::group& g)` and `void from_inlet(const sidre::group& g, T& t)`, but this could also be done with constructors and conversion operators

samuelpmishLLNL: I'm not familiar with the way inlet is implemented, but the approach Josh and I are describing makes it really easy to define arbitrarily nested data types too.

src/drivers/serac.cpp
samuelpmishLLNL: Does the term input "deck" actually from a collection of (punch) cards?

scripts/uberenv/packages/axom/package.py
```diff
@@ -53,6 +53,12 @@ class Axom(CMakePackage, CudaPackage):
 
     version('main', branch='main', submodules=True)
     version('develop', branch='develop', submodules=True)
+
+    # SERAC EDIT START
+    version('0.4.0p1', SHA='fd2edc420f1ab3504c4890654266039b8966bb37', submodules="True")
+    # SERAC EDIT END
```
adrienbernede: He did it!

src/drivers/serac.cpp
white238: It will be common that most codes will have optional values in their input file.  Possibly another idea is to add an optional parameter to `get` that makes it fail if not found.

src/drivers/serac.cpp
white238: I am open to suggestions.  The next big push on inlet is to better support nested Tables and arrays of Tables.

src/drivers/serac.cpp
white238: This PR was to create the basic framework and then move forward with a cleaner solution to this.  Providing either inlet to the constructor or using the idea I proposed in my Axom issue to grab a specific Inlet table and add get to that level would be next.

src/drivers/serac.cpp
white238: yes it is.. i try to catch myself whenever i call it a deck and move forward with the less lab specific name (input file).  I'll do another round and switch them.

scripts/uberenv/packages/axom/package.py
white238: kicking and screaming i will be brought into the future...

src/drivers/serac.cpp
white238: Yea that was a remnant of my long standing branch and grabbing the wrong conflict.  Should i just remove superlu for minres or keep both?

data/decks/default.lua
white238: do you want something like:

```
mesh = {
  path = "../meshes/beam-hex.mesh",
  
  -- serial and parallel refinement levels
  ser_ref_levels = 0,
  par_ref_levels = 0,
}
```

src/infrastructure/cli.cpp
white238: I removed the `shared_ptr` from the `datastore` and `cli_opts` but due to some possibly ill-advised decisions in Inlet i couldn't remove those.  We can talk about better ways to handle this tomorrow and implement them in a follow up PR.

white238:
Currently running into an extremely frustrating error with the installed using-with-cmake example. This only occurs on the toss3 gcc@8.3.1 when running both specs through the build_tpls.py.  It acts almost like a timing issue but the `ls` command works just fine and I added a `sleep`.

Error:

```
[log file: /usr/WS2/smithdev/libs/toss_3_x86_64_ib/2020_09_30_20_48_20/_serac_build_and_test_2020_09_30_20_48_20/build-rzgenie-toss_3_x86_64_ib-gcc@8.3.1/output.log.make.examples.txt]
[exe: cd /usr/WS2/smithdev/libs/toss_3_x86_64_ib/2020_09_30_20_48_20/_serac_build_and_test_2020_09_30_20_48_20/install-rzgenie-toss_3_x86_64_ib-gcc@8.3.1/examples/using-with-cmake/build && make && ls -al && sleep 5 && ./serac_example]
[ERROR: Make examples for host-config: /usr/WS2/smithdev/libs/toss_3_x86_64_ib/2020_09_30_20_48_20/rzgenie-toss_3_x86_64_ib-gcc@8.3.1.cmake failed]
```

Log file:

```
Scanning dependencies of target serac_example
[ 50%] Building CXX object CMakeFiles/serac_example.dir/serac_example.cpp.o
[100%] Linking CXX executable serac_example
[100%] Built target serac_example
total 3556
drwx--S--- 3 asmith smithdev    4096 Sep 30 21:23 .
drwx--S--- 3 asmith smithdev    4096 Sep 30 21:23 ..
-rw------- 1 asmith smithdev   16644 Sep 30 21:23 CMakeCache.txt
drwx--S--- 6 asmith smithdev    4096 Sep 30 21:23 CMakeFiles
-rw------- 1 asmith smithdev    5867 Sep 30 21:23 Makefile
-rw------- 1 asmith smithdev    1792 Sep 30 21:23 cmake_install.cmake
-rw------- 1 asmith smithdev    2110 Sep 30 21:23 compile_commands.json
-rwx------ 1 asmith smithdev 3573848 Sep 30 21:23 serac_example
```

It is supposed to end with the output of serac_example (which you can run by hand):

```
[asmith@rzgenie2:build]$ ./serac_example 
Logger activated: serac_serial_logger
MFEM version: MFEM v4.1.1 (development)
Axom information:

AXOM_VERSION_FULL: v0.4.0
AXOM_VERSION_MAJOR: 0
AXOM_VERSION_MINOR: 4
AXOM_VERSION_PATCH: 0
Compiler Settings: 
   C++ Standard: c++14
   OpenMP support: DISABLED
Available components: 
 { core; inlet; mint; lumberjack; primal; quest; sidre; slam; slic; }
Active Dependencies: 
 { CLI11; conduit; fmt; hdf5; lua; mpi; sol; sparsehash; }
Serac loaded successfully.
```

I'm done for the night.

scripts/uberenv/spack_configs/docker/ubuntu18/compilers.yaml
```diff
@@ -2,8 +2,7 @@ compilers:
   - compiler:
       environment: {}
       extra_rpaths: []
-      flags:
-        cxxflags: -stdlib=libc++
+      flags: {}
```
white238, line 5: This was the source of my Sol segfault.

jamiebramwell: 😢 At least you found it! Take a vacation day!

white238:
LGTM

scripts/docker/dockerfile_clang-10
```diff
@@ -12,9 +12,11 @@ WORKDIR "/home/serac"
 USER serac
 
 RUN git clone --recursive https://github.com/LLNL/serac.git
+RUN cd serac && git checkout feature/white238/addinlet && cd ..
```
jamiebramwell, line 15: Should we leave a hard call to this branch name here?

scripts/docker/dockerfile_gcc-8
```diff
@@ -2,21 +2,20 @@ FROM axom/compilers:gcc-8
...
+RUN sudo ln -f -s python3 /usr/bin/python # Compatibility with legacy Python
 
 RUN sudo useradd -m -s /bin/bash -G sudo serac
 
 WORKDIR "/home/serac"
 USER serac
 
-RUN git clone --recursive https://github.com/LLNL/serac.git
-RUN cd serac && git checkout ci/essman/docker_build_fixes && cd ..
+RUN git clone -b feature/white238/addinlet --recursive https://github.com/LLNL/serac.git
```
jamiebramwell, line 14: Same comment about hard reference to this branch name.

scripts/docker/dockerfile_clang-10
white238: These are generated by the build_new_image script.  They probably shouldn't be checked in.

src/infrastructure/logger.cpp
```diff
@@ -31,37 +31,43 @@ bool initialize(MPI_Comm comm)
...
 
-  std::string fmt_id = "[<LEVEL>]: <MESSAGE>\n";
+  std::string fmt_i  = "<MESSAGE>\n";
+  std::string fmt_d  = "[<LEVEL>]: <MESSAGE>\n";
   std::string fmt_we = "[<LEVEL> (<FILE>:<LINE>)]\n<MESSAGE>\n\n";
 
   // Only create a parallel logger when there is more than one rank
   if (numRanks > 1) {
-    fmt_id = "[<RANK>]" + fmt_id;
+    fmt_i  = "[<RANK>] " + fmt_i;
```
jamiebramwell, line 45: Why does this one have a space and the other two don't?

scripts/docker/dockerfile_clang-10
joshessman-llnl: This is fixed by #233

jamiebramwell (approved):
Looks good modulo a few fiddly comments.

jamiebramwell:
Anyone else have outstanding comments @joshessman-llnl @samuelpmishLLNL ?

src/infrastructure/logger.cpp
white238: So I can provide horrible inconsistency.. i'll fix it.

white238: Oh no it's right.  The other messages have prefix's for their messages.  fmt_i is the only one that doesn't. So the others have a previous space or newline.

joshessman-llnl (approved):
I would anticipate this changing a fair amount as inlet evolves but this is a good first start

white238:
> I would anticipate this changing a fair amount as inlet evolves but this is a good first start

That was the plan.  Serac is the first use-case and the idea was to get the basic functionality in quickly so we improve it and get Serac up and going with input files. As well as find implementation holes. (like get only being at the Inlet class level)

PR #226, opened 2020-10-01
Removed no-longer-needed #includes

joshessman-llnl:
I went through and removed most of the dead includes, and also augmented the signal handler to include the name of the signal - a Ctrl-C now prints `[SIGNAL]: Received signal 2 (Interrupt), exiting` instead of `[SIGNAL]: Received signal 2, exiting`.

### DIFF
--- a/src/infrastructure/logger.cpp
+++ b/src/infrastructure/logger.cpp
@@ -6,8 +6,6 @@
 
 #include "infrastructure/logger.hpp"
 
-#include "infrastructure/terminator.hpp"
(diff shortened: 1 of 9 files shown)


jamiebramwell (approved):
Should we add something like [include what you use](https://include-what-you-use.org/) in our toolchain to avoid this in the future?

samuelpmishLLNL (approved):

jamiebramwell:
LGTM

PR #237, opened 2020-10-06
Refactor the boundary condition manager to support arbitrary types

joshessman-llnl:
This PR adds a variadic template parameter to the `BoundaryConditionManager` class intended to be populated with strongly aliased basic (dirichlet/neumann) BCs and BCs of user-defined type.

The primary user (incl. physics implementer) changes are the addition of a new structure describing the boundary conditions for the problem:
```cpp
struct ThermalConductionBC {
  struct Temperature : ScalarBoundaryCondition {
  };
  struct Flux : ScalarBoundaryCondition {
  };
  using Manager = BoundaryConditionManager<EssentialAlias<Temperature>, NaturalAlias<Flux>>;
};
```

Boundary conditions are added and accessed with an explicit tag:
```cpp
using ThermalConductionBC::Temperature;
bcs.addEssential<Temperature>(/*args*/);
for (auto& bc : bcs.essentials<Temperature>())
{
  // stuff
}
bcs.eliminateAllEssentialDofsFromMatrix<Temperature>(K_mat);
```

This adds a small amount of boilerplate for solvers with only one type of essential/natural boundary condition, but may improve readability somewhat.  It requires the recompilation of the `BoundaryConditionManager` class for each physics module.

This variadic template class will help with the inlet integration - if each BC type (including user-defined BC types) implements schema definition/fromInlet functions, a hierarchical input structure is possible as the types can be iterated over.

@samuelpmishLLNL had proposed a slightly different approach (https://godbolt.org/z/h7v515) that also templates the `BoundaryCondition` class(es) on physics-specific tags instead of using a passthrough `StrongAlias` type.  Any thoughts on the comparison between these two approaches are welcome.

This PR also removes the need for specifying within the solver logic whether a BC should be a scalar or vector coefficient - the implementer can specify this by having their tags inherit from `ScalarBoundaryCondition` or `VectorBoundaryCondition`.  

The behavior of this is as follows (I am not sure this is completely correct):
* If a user builds a Dirichlet BC with a VecCoef and no component, the component is set to -1
* If a user builds a Dirichlet BC with a ScalarCoef and no component:
   * If the module implementer has specified that the BC coef should be vector-valued (see above), a compiler error is thrown.  If it was designated as scalar-valued, the program compiles.
   * Otherwise (type not specified), the program fails to compile with the following message: 
       > Boundary condition does not specify whether it is scalar- or vector-valued.  If this is a scalar-valued BC, 
        pass a component of zero, otherwise, pass the vector component to which this scalar coef should apply.
* Behavior when a component is specified by the user is unchanged


samuelpmishLLNL:
I think I misunderstood the motivation for `StrongAlias<>` when making my godbolt example, what is its purpose?

joshessman-llnl:
> 
> 
> I think I misunderstood the motivation for `StrongAlias<>` when making my godbolt example, what is its purpose?

It's intended to be a generic way of introducing strong aliases (i.e., an alias not interchangeable with its underlying type) without modifying the aliased type.  `using TemperatureBC = EssentialBoundaryCondition;` wouldn't prevent a user from passing a different boundary condition to a function that takes a `TemperatureBC`.  Boost does this with [a macro](https://www.boost.org/doc/libs/1_61_0/libs/serialization/doc/strong_typedef.html), for example, but templates are sufficient here.

jamiebramwell:
After talking with @joshessman-llnl , I have a fundamental question about the cost-benefit analysis of the `BoundaryConditionManager`. Does the benefit of grouping all of the boundary condition data together outweigh the cost of verbose template errors and physics module developers needing to understand the variadic structure? An alternative approach would be to have each physics module define their own vectors of boundary conditions of various physics-specific types. I would appreciate any thoughts on this subject. We can also discuss this in a Thursday meeting.

jamiebramwell:
Closing this for now since it's pretty stale at this point.

PR #241, opened 2020-10-10
Implements a second order time integrator that uses mfem::NewmarkSolver.

jonwong12:
Updated:
This PR implements a simple 2D cantilever floppy beam problem using the `serac::SecondOrderODE`. Additionally `serac::SecondOrderODE` was extend such that first-order time stepping algorithms (e.g. `mfem::BackwardEulerSolver`) can be used automatically. This should allow for straightforward integration with Sundials' first order time-integration methods (e.g. CVODES and etc.).

The PR includes:
- A `PointerNonlinearFormIntegrator` allows `mfem::NonlinearformIntegrators` to be shared and reused and protects the underlying integrator from automatic `mfem::ParNonlinearForm` deletion

The `tests/serac_newmark_test` performs 3 tests: (needs to be updated)
1. `Simple` - Is a rigid body test. The time-step solver values are checked after one time step.
2. `Equilibrium` - Is a cantilever beam fixed at one end. A constant gravity load is applied and is activated at t=0. A `mfem::NewmarkSolver` and `NewmarkBetaSecondOrder` operator is built from a `SecondOrderResidual`.
3. `Equilibrium_firstorder` - Is the same problem as `Equilibrium` except a `mfem::BackwardEulerSolver` and `FirstOrderSystem` operator is built from the same `SecondOrderResidual` class as test 2.


jamiebramwell:
@samuelpmishLLNL could you take a look at this and comment how it would work with your PR?

samuelpmishLLNL:
> 
> 
> @samuelpmishLLNL could you take a look at this and comment how it would work with your PR?

I don't know enough about mfem to understand what `PointerNonlinearFormIntegrator` and `SubstitutionNonlinearFormIntegrator` are doing, but the other parts of this PR appear to be a way to refactor the way that serac lets us specify ODEs and the methods to solve them (which is part of the goal of #216.

In that sense, the two PRs are conflicting-- they both offer a new way to express residuals and ODEs, but it wouldn't make sense to have both implementations in serac. I don't like the way that this is currently done in serac (with the externally-defined companion operator classes for each physics module), because it introduces _a lot_ of boiler plate code. 

Some of the key differences between the two PRs:

|   |  this PR | #216 |
| --| -- | -- |
| residuals | defined by registering integrators | defined by a lambda function that uses the members of a physics module |
| ODEs | constructed from the residual? | defined by a lambda function that uses the members of a physics module |
| Methods for Solving ODEs | unclear | uses existing mfem::ODESolver options |

In the example in this PR, `NewmarkBetaSecondOrder` inherits from `mfem::SecondOrderTimeDependentOperator`, which is basically the mfem way of saying "`NewmarkBetaSecondOrder` is a second order differential equation". I think this is a mistake since Newmark is not a second order ODE, but a _method_ for solving second order ODEs. In my mind, it is important to not combine these two concepts: the definition of an ODE should be independent of the method used to solve it.

jonwong12:
> In that sense, the two PRs are conflicting-- they both offer a new way to express residuals and ODEs, but it wouldn't make sense to have both implementations in serac. I don't like the way that this is currently done in serac (with the externally-defined companion operator classes for each physics module), because it introduces a lot of boiler plate code.

I personally don't have a strong opinion about the ODEs and I certainly like having the lambdas available generally. I would like to propose the following:
- Use #216 operator definition (it seems general enough).
- When we have nonlinear integrators available, we can use them to "fill in" portions of the lambdas in #216 for us.
- I'm not a big fan of `NewmarkBetaSecondOrder` but it does run through a dynamic cantilever beam example, so I'm proposing to just keep it in the test. When the definition for the operator goes in in #216 we can update the test to use a combination of the above.

samuelpmishLLNL:
I think another benefit of Jonathan's way of defining a residual is it may be easier to differentiate, since the integrators already have gradients defined. In my PR, the user is responsible for ensuring the residual and its gradient are consistent.

jonwong12:
> I don't know enough about mfem to understand what `PointerNonlinearFormIntegrator` and `SubstitutionNonlinearFormIntegrator` are doing,

I need to update the name changes from `SubsittutionNonlinearformIntegrator` to `TransformedNonlinearFormIntegrator` with master. So I'll do that in a moment today.

`PointerNonlinearFormIntegrator` is just a way to get around mfem's ownership convention where they delete all the integrators when a NonlinearForm goes out of scope. However, one may want to set boundary conditions on the different portions of the state (displacement, velocity, acceleration) when solving an ODE in different situations. Since mfem::NonlinearIntegrators were written to have a specific input in mind, my reasoning was that a "residual" object can keep the overall definition of the residual made up of different nonlinearform operations (sams's general #216  or built from mfem and stored in a `PointerNonlinearFormIntegrator` in this PR) and use `TransformedNonlinearFormIntegrator` to take the mfem::integrators and solve the problem we want over (displacement, velocity, or acceleration) depending on the time integration scheme used (uses lambdas on the element level).

solberg2:
So should we merge these 2 MR's?  It seems having the lambdas is good, it seems having the residual and the gradient linked is good, it seems having the ODE and the Integrator differentiated is good.

jamiebramwell:
> 
> 
> So should we merge these 2 MR's? It seems having the lambdas is good, it seems having the residual and the gradient linked is good, it seems having the ODE and the Integrator differentiated is good.

Yes, my idea is that these would both get merged after the compatibility issues are settled.

jamiebramwell:
I think the flow should be https://github.com/LLNL/serac/pull/216 and then this PR for the reasons outlined by @jonwong12 , so we should push on getting that merged.

src/serac/physics/operators/odes.cpp
```diff
@@ -42,10 +42,73 @@ void SecondOrderODE::SetTimestepper(const serac::TimestepMethod timestepper)
...
+
+    // Copy back
+    x    = bx.GetBlock(0);
+    dxdt = bx.GetBlock(1);
+  }
+}
+
+void SecondOrderODE::ImplicitSolve(const double dt, const mfem::Vector& u, mfem::Vector& du_dt)
+{
+  /* A second order p.d.e can be recast as a first order system
```
samuelpmish: typo: should be `o.d.e.`

samuelpmish:
Looks promising! 

It's making me think that some of the refactoring of the ODE classes Josh and I did might have been premature, though. The interface that makes the most sense to me is to do support something like

```
// user sets up their description of the second order ode
SecondOrderODE dynamics(...);

// if the user wants to treat the dynamics as a first order ODE, we should
// just implement FirstOrderODE(SecondOrderODE) to perform the conversion
FirstOrderODE equivalent_dynamics(dynamics);

// and now we can pass a block vector in
equivalent_dynamics.step(x_and_v, t, dt);
```

But in the refactor, we made that sort of impossible to achieve. As a result, your implementation has to make SecondOrderODE juggle two separate implementations for treating it as either first or second order!

These comments are meant as more of an apology than a criticism-- we reworked the interface of these classes to remove some unused stuff, but that seems to have made your work here more challenging.


edit: this comment is related to the interface for converting second order -> first order

jonwong12:
> 
> 
> Looks promising!
> 
> It's making me think that some of the refactoring of the ODE classes Josh and I did might have been premature, though. The interface that makes the most sense to me is to do support something like
> 
> ```
> // user sets up their description of the second order ode
> SecondOrderODE dynamics(...);
> 
> // if the user wants to treat the dynamics as a first order ODE, we should
> // just implement FirstOrderODE(SecondOrderODE) to perform the conversion
> FirstOrderODE equivalent_dynamics(dynamics);
> 
> // and now we can pass a block vector in
> equivalent_dynamics.step(x_and_v, t, dt);
> ```
> 
> But in the refactor, we made that sort of impossible to achieve. As a result, your implementation has to make SecondOrderODE juggle two separate implementations for treating it as either first or second order!
> 
> These comments are meant as more of an apology than a criticism-- we reworked the interface of these classes to remove some unused stuff, but that seems to have made your work here more challenging.
> 
> edit: this comment is related to the interface for converting second order -> first order

Interesting. I actually think the refactor changed the ownership mechanism from having the ODESolvers own the ODEs to the ODES owning the ODESolvers in a good way? I think that actually allows me to utilize the fact that the SecondOrderODEs inherit from FirstOrder `mfem::TimeDependentOperators` and just directly implement the underlying calls that mfem's default implementation errors out. Instead depending on the `TimeStepperMethod` the refactor allows us to use different underlying ODESolvers and then apply them to the same SecondOrderODE class.

I think the way it's done now, while fundamentally different from the `mfem-way`, is cleaner and more straightforward.

samuelpmish:
I do think the ODESolvers inside the ODEs is good, but by taking an EquationSolver, it's not obvious to me how we could support having a conversion constructor of the form `FirstOrderODE(SecondOrderODE)`.

src/serac/physics/operators/odes.cpp
jonwong12:
```suggestion
  /* A second order o.d.e can be recast as a first order system
```

data/input_files/tests/nonlinear_solid/dyn_newmark_solve_bending.lua
```diff
@@ -0,0 +1,99 @@
+-- Comparison information
+expected_x_l2norm = 1.4225
+expected_v_l2norm = 0.2252
+epsilon = 0.0001
+
+
+main_mesh = {
+    type = "generate",
```
jamiebramwell, line 8: I'm a bit confused why this is labeled as a generated mesh, but it also lists a mesh file.

src/serac/infrastructure/input.cpp
```diff
@@ -19,6 +19,7 @@ axom::inlet::Inlet initialize(axom::sidre::DataStore& datastore, const std::stri
 {
   // Initialize Inlet
   auto luareader = std::make_unique<axom::inlet::LuaReader>();
+  luareader->solState().open_libraries(sol::lib::math);
```
jamiebramwell: @white238 @joshessman-llnl was there a conclusion as to whether this should just be enabled by default in inlet?

src/serac/integrators/wrapper_integrator.hpp
```diff
@@ -221,4 +221,44 @@ class TransformedNonlinearFormIntegrator : public mfem::NonlinearFormIntegrator
   std::function<transformed_gradient_func> transformed_function_grad_;
 };
 
+/**
+ * @brief A class to hold a shared_ptr to an NonlinearformIntegrator that can be deleted without deleting the shared
+ * pointer improperly
+ */
+
+class PointerNonlinearFormIntegrator : public mfem::NonlinearFormIntegrator {
```
jamiebramwell, line 229: When do you need this structure?

src/serac/physics/operators/odes.hpp
```diff
@@ -160,6 +165,12 @@ class SecondOrderODE : public mfem::SecondOrderTimeDependentOperator {
    * @brief MFEM solver object for second-order ODEs
    */
   std::unique_ptr<mfem::SecondOrderODESolver> ode_solver_;
```
jamiebramwell: Maybe rename these as `second_order_ode_solver_` and `first_order_system_ode_solver_`? I know the names are long, but I think it would minimize confusion.

src/serac/physics/operators/odes.cpp
```diff
@@ -42,10 +42,73 @@ void SecondOrderODE::SetTimestepper(const serac::TimestepMethod timestepper)
...
+  if (ode_system_solver_) {
+    // we need to adjust the width of this operator
+    width *= 2;
+    ode_system_solver_->Init(*this);
+  }
+}
+
+void SecondOrderODE::Step(mfem::Vector& x, mfem::Vector& dxdt, double& t, double& dt)
+{
+  if (ode_solver_) {  // if we used a 2nd order method
```
jamiebramwell: Do we ever have both an `ode_system_solver` and an `ode_solver`?

jamiebramwell:
> 
> 
> > Looks promising!
> > It's making me think that some of the refactoring of the ODE classes Josh and I did might have been premature, though. The interface that makes the most sense to me is to do support something like
> > ```
> > // user sets up their description of the second order ode
> > SecondOrderODE dynamics(...);
> > 
> > // if the user wants to treat the dynamics as a first order ODE, we should
> > // just implement FirstOrderODE(SecondOrderODE) to perform the conversion
> > FirstOrderODE equivalent_dynamics(dynamics);
> > 
> > // and now we can pass a block vector in
> > equivalent_dynamics.step(x_and_v, t, dt);
> > ```
> > 
> > 
> > But in the refactor, we made that sort of impossible to achieve. As a result, your implementation has to make SecondOrderODE juggle two separate implementations for treating it as either first or second order!
> > These comments are meant as more of an apology than a criticism-- we reworked the interface of these classes to remove some unused stuff, but that seems to have made your work here more challenging.
> > edit: this comment is related to the interface for converting second order -> first order
> 
> Interesting. I actually think the refactor changed the ownership mechanism from having the ODESolvers own the ODEs to the ODES owning the ODESolvers in a good way? I think that actually allows me to utilize the fact that the SecondOrderODEs inherit from FirstOrder `mfem::TimeDependentOperators` and just directly implement the underlying calls that mfem's default implementation errors out. Instead depending on the `TimeStepperMethod` the refactor allows us to use different underlying ODESolvers and then apply them to the same SecondOrderODE class.
> 
> I think the way it's done now, while fundamentally different from the `mfem-way`, is cleaner and more straightforward.

I agree with this. While it doesn't have the conversion syntax we originally envisioned, the same functionality is achieved by having a single physics running as either a first order system or a full second order solver.

src/serac/infrastructure/input.cpp
white238: https://github.com/LLNL/axom/blob/494fc6429ac217b9d6aff5c9bc967dd3ee8aa5b9/src/axom/inlet/LuaReader.hpp#L44-L47

Added 23 days ago to Axom and the TPLs were updated Dec 17th.  So if we rebuild the TPLs it will be there.  Do you want me to handle rebuilding them today? Easy to do during meetings...

jamiebramwell: Eh, it's not a huge rush. Just making sure it went in.

white238: The new PR i just merged has the a LuaReader w/ this enabled so you can remove this line(s).

data/input_files/tests/nonlinear_solid/dyn_newmark_solve_bending.lua
jonwong12: Sorry. The `type` field from #334 hasn't been merged yet. I was in the middle of updating this.

src/serac/physics/operators/odes.cpp
jonwong12: I have not run into a case like that yet. I think `SecondODESolver` and `OdeSolver` are completely separate, so usually they would never interact.

src/serac/integrators/wrapper_integrator.hpp
jonwong12: It was originally used in the original newmark example to allow an `Integrator` to be shared by different `mfem::NonlinearForms` [earlier comment](https://github.com/LLNL/serac/pull/241#issuecomment-718139842). When the forms go out-of-scope, it deletes all attached interrogators. This class is a light-weight wrapper that keeps `mfem::NonlinearForm`s happy while also properly ensuring that integrators can be shared (using `std::shared_ptr`) amongst `-Forms` or different solvers.

The original use of this structure in the newmark beta example allowed one to reuse the same integrator in combination with a `serac::TransformedNonlinearIntegrator` to do some variable substitutions depending on how you want to solve the 2nd order problem (solve for acceleration vs displacements). This could move some of the integrator-related logic that is currently coded in the rebuilding the `H_form` during updates.

data/input_files/tests/nonlinear_solid/dyn_newmark_solve_bending.lua
jonwong12: The `mesh` field is required at the moment as part of the nonlinear schema. I hope to adjust it in the lua_mesh PR (#334).

src/serac/integrators/wrapper_integrator.hpp
jamiebramwell: Gotcha. So we don't need it any more, but it's good to get in the repo for future use?

src/serac/physics/operators/odes.cpp
```diff
@@ -25,27 +25,91 @@ void SecondOrderODE::SetTimestepper(const serac::TimestepMethod timestepper)
...
   }
-  ode_solver_->Init(*this);
+
+  if (second_order_ode_solver_) {
+    second_order_ode_solver_->Init(*this);
+  } else if (first_order_system_ode_solver_) {
+    // we need to adjust the width of this operator
+    width *= 2;
+    first_order_system_ode_solver_->Init(*this);
+  }
```
jamiebramwell, line 74: A else clause (or an extra one with a slic error) would prevent this from ever failing to define an ODE solver.

src/serac/physics/operators/odes.cpp
```diff
@@ -25,27 +25,91 @@ void SecondOrderODE::SetTimestepper(const serac::TimestepMethod timestepper)
...
+}
+
+void SecondOrderODE::Step(mfem::Vector& x, mfem::Vector& dxdt, double& t, double& dt)
+{
+  if (second_order_ode_solver_) {
+    // if we used a 2nd order method
+    second_order_ode_solver_->Step(x, dxdt, t, dt);
+  } else if (first_order_system_ode_solver_) {
```
jamiebramwell, line 100: A else clause (or an extra one with a slic error) would prevent this from ever failing to define an ODE solver.

jamiebramwell (approved):

jamiebramwell:
@samuelpmishLLNL , any issues outstanding on this PR?

solberg2 (approved):

src/serac/integrators/wrapper_integrator.hpp
jonwong12: Yes. It would be useful to keep for adjoint problems. If we don't use them, once we start doing adjoint problems we can throw it away or deprecate it.

data/input_files/tests/nonlinear_solid/dyn_newmark_solve_bending.lua
jonwong12: I'll update this now that #334 has been merged.

src/serac/physics/operators/odes.cpp
jonwong12: I've added the SLIC_ERROR because I think it might help catch any developer implementation errors. It shouldn't be possible for users to muck with this.

src/serac/physics/operators/odes.cpp
jonwong12: I'll add a similar one for `FirstOrderODE`, I believe we don't check if `SetTimeStepper` has been called before running `Step`.

jamiebramwell:
LGTM

PR #243, opened 2020-10-13
Add custom solver override, set solver params in module constructors

### LINKED ISSUE #186: Allow custom MFEM solvers

### LINKED ISSUE #187: Clean up the storage of solver parameters

joshessman-llnl:
This is a slight modification of #223 that omits the module-provided solver defaults in order to sidestep a standard library bug.  The new module-defined solver parameters structures are still useful as they will be populated by Inlet when those changes are ready in Axom.

Resolves #186
Resolves #187


src/physics/utilities/solver_config.hpp
```diff
@@ -64,14 +68,24 @@ enum class NonlinearSolver
...
+ * @brief Stores the information required to configure a HypreBoomerAMG preconditioner
+ */
+struct HypreBoomerAMGPrec {
+  mfem::ParFiniteElementSpace* pfes = nullptr;
 };
 
+/**
+ * @brief Preconditioning method
+ */
+using Preconditioner = std::variant<HypreSmootherPrec, HypreBoomerAMGPrec>;
```
jamiebramwell, line 87: Is it considered better to use variants instead of base class pointers now?

jamiebramwell: I found [this discussion](https://stackoverflow.com/questions/59784261/stdvariant-vs-pointer-to-base-class-for-heterogeneous-containers-in-c) on stack overflow. One thing to keep in mind is variant could grow quite a bit as we add more non-hypre preconditioners.

src/physics/utilities/equation_solver.hpp
```diff
@@ -171,9 +157,9 @@ class EquationSolver : public mfem::Solver {
   std::unique_ptr<mfem::Solver> prec_;
 
   /**
-   * @brief The linear solver object, either direct (SuperLU) or iterative
+   * @brief The linear solver object, either custom, direct (SuperLU), or iterative
    */
-  std::variant<std::unique_ptr<mfem::IterativeSolver>, std::unique_ptr<mfem::SuperLUSolver>> lin_solver_;
+  std::variant<std::unique_ptr<mfem::IterativeSolver>, std::unique_ptr<mfem::SuperLUSolver>, mfem::Solver*> lin_solver_;
```
jamiebramwell, line 162: I think it goes beyond this PR, but will this `variant` get huge as we have various solver options?

src/drivers/serac.cpp
```diff
@@ -93,10 +93,39 @@ int main(int argc, char* argv[])
...
+    std::string msg = fmt::format("Unknown Linear solver type given: {0}", solver_type);
+    SLIC_ERROR_ROOT(rank, msg);
+    serac::exitGracefully(true);
+  }
+
   // Define the solid solver object
   int order;
   inlet->get("nonlinear_solid/order", order);  // has default value
-  serac::NonlinearSolid solid_solver(order, mesh);
+  serac::NonlinearSolid solid_solver(order, mesh, {lin_params, nonlin_params});
```
jamiebramwell, line 128: 👍

jamiebramwell (approved):
This looks great! My only concern is that the internal `std::variant` containing the actual solver objects may get huge once we start adding options. This can be addressed in a future PR if it becomes an issue. I don't think the user-facing solver parameters variant will grow too much as this should cover the vast majority of use cases.

src/physics/utilities/solver_config.hpp
joshessman-llnl: In my opinion it depends on the situation - I think interfaces can be useful when there is lots of commonality of behavior, but in this case I would expect variant members to just be small structs/dataclasses with little commonality of state or behavior.

The lack of commonality could also require a fair amount of dynamic_casting if this were implemented with inheritance as the exact preconditioner type must be known for the AMG prec to be set up correctly.

jamiebramwell:
@white238 @samuelpmishLLNL Any comments on this PR?

src/physics/utilities/equation_solver.cpp
```diff
@@ -54,6 +64,24 @@ std::unique_ptr<mfem::IterativeSolver> EquationSolver::buildIterativeLinearSolve
   iter_lin_solver->SetMaxIter(lin_params.max_iter);
   iter_lin_solver->SetPrintLevel(lin_params.print_level);
 
+  // Handle the preconditioner - currently just BoomerAMG and HypreSmoother are supported
+  if (std::holds_alternative<HypreBoomerAMGPrec>(lin_params.prec)) {
+    auto par_fes = std::get<HypreBoomerAMGPrec>(lin_params.prec).pfes;
+    SLIC_ERROR_IF(par_fes == nullptr, "FESpace is required to use the HypreBoomerAMG preconditioner.");
```
white238, line 70: This won't exit until we change the logger to abortOnError and register the exit gracefully function with SLIC.

src/physics/utilities/equation_solver.hpp
white238: How many solver options does a production code normally have?

src/physics/nonlinear_solid.hpp
```diff
@@ -29,13 +29,31 @@ namespace serac {
...
+  /**
+   * @brief A configuration variant for the various solves
+   * Either quasistatic, or time-dependent with timestep and M params
+   */
+  struct NonlinearSolidParameters {
+    LinearSolverParameters           H_lin_params;
+    NonlinearSolverParameters        H_nonlin_params;
+    std::optional<DynamicParameters> dyn_params = std::nullopt;
+  };
+
```
samuelpmishLLNL: I would like clearer names here: `NonlinearSolid::DynamicParameters` sounds like it's related to the physics, not the method of solving an ODE. `NonlinearSolid::NonlinearSolidParameters` has a redundant prefix and also sounds like it should be contain the elasticity coefficients or something.

The might be combined by something like

```
class NonlinearSolid : public BasePhysics {
  struct SolverParameters {
    LinearSolverParameters                   linear;
    std::optional<NonlinearSolverParameters> nonlinear = std::nullopt;
    std::optional<TimestepMethod>            method = std::nullopt;
  };
  ...
};
```
or renamed to something like `{Dynamic,Quasistatic}-SolverParameters`

src/physics/utilities/equation_solver.hpp
samuelpmishLLNL: if the custom solvers inherit from `mfem::Solver` or `mfem::IterativeSolver` it shouldn't need to grow, right?

src/physics/utilities/equation_solver.cpp
joshessman-llnl: I've gone ahead and changed the logger initialization to register exitGracefully so this will terminate as expected.

src/physics/nonlinear_solid.hpp
joshessman-llnl: I think this is a good idea - I've renamed to `<physics type>::SolverParameters` per your suggestion and renamed the extra dynamic solver parameters, when needed, to `<physics type>::DynamicSolverParameters`.  I think it's useful to keep the dynamic stuff together in an optional struct instead of optional fields to eliminate errors where one dynamic parameter is set but not the other.

samuelpmishLLNL (approved):

joshessman-llnl:
LGTM

white238 (approved):

PR #248, opened 2020-10-14
Begin utilizing inlet to read in structs

joshessman-llnl:
This PR bumps the axom version (and SuperLU for some reason - it may be worth trying to re-enable the direct solver tests).

It also allows the driver to be simplified by delegating the reading of data to struct-specific template specializations.

To solve the issue we discussed without waiting on an Axom PR, a line was added to the driver to add the input file path to the Inlet structure.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,12 +11,12 @@ jobs:
     matrix:
       linux_gcc8:
           VM_ImageName: 'ubuntu-16.04'
-          Compiler_ImageName: 'seracllnl/tpls:gcc-8'
+          Compiler_ImageName: 'seracllnl/tpls:gcc-8_10-14-20_19h-51m'
           TEST_TARGET: 'linux_gcc8'
           HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake'
       linux_clang10:
           VM_ImageName: 'ubuntu-18.04'
-          Compiler_ImageName: 'seracllnl/tpls:clang-10'
+          Compiler_ImageName: 'seracllnl/tpls:clang-10_10-14-20_19h-51m'
           TEST_TARGET: 'linux_clang10'
           HOST_CONFIG: 'docker-linux-ubuntu18.04-x86_64-clang@10.0.0.cmake'
 
@@ -43,7 +43,7 @@ jobs:
   variables:
     DO_STYLE_CHECK: 'yes'
     VM_ImageName: 'ubuntu-18.04'
-    Compiler_ImageName: 'seracllnl/tpls:clang-10'
+    Compiler_ImageName: 'seracllnl/tpls:clang-10_10-14-20_19h-51m'
     TEST_TARGET: 'linux_clang10'
     HOST_CONFIG: 'docker-linux-ubuntu18.04-x86_64-clang@10.0.0'
   pool:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -11,12 +11,12 @@ jobs:
     matrix:
       linux_gcc8:
           VM_ImageName: 'ubuntu-16.04'
-          Compiler_ImageName: 'seracllnl/tpls:gcc-8_10-14-20_19h-51m'
+          Compiler_ImageName: 'seracllnl/tpls:gcc-8_10-19-20_18h-06m'
           TEST_TARGET: 'linux_gcc8'
           HOST_CONFIG: 'docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake'
       linux_clang10:
           VM_ImageName: 'ubuntu-18.04'
-          Compiler_ImageName: 'seracllnl/tpls:clang-10_10-14-20_19h-51m'
+          Compiler_ImageName: 'seracllnl/tpls:clang-10_10-19-20_22h-40m'
           TEST_TARGET: 'linux_clang10'
           HOST_CONFIG: 'docker-linux-ubuntu18.04-x86_64-clang@10.0.0.cmake'
 
@@ -43,7 +43,7 @@ jobs:
   variables:
     DO_STYLE_CHECK: 'yes'
     VM_ImageName: 'ubuntu-18.04'
-    Compiler_ImageName: 'seracllnl/tpls:clang-10_10-14-20_19h-51m'
+    Compiler_ImageName: 'seracllnl/tpls:clang-10_10-19-20_22h-40m'
     TEST_TARGET: 'linux_clang10'
     HOST_CONFIG: 'docker-linux-ubuntu18.04-x86_64-clang@10.0.0'
   pool:
--- a/data/input_files/default.lua
+++ b/data/input_files/default.lua
@@ -2,7 +2,7 @@
 t_final = 1.0
 dt      = 0.25
 
-primary_mesh = {
+main_mesh = {
     -- mesh file
     mesh = "../meshes/beam-hex.mesh",
     -- serial and parallel refinement levels
@@ -12,7 +12,7 @@ primary_mesh = {
 
 -- Solver parameters
 nonlinear_solid = {
-    mesh_info = primary_mesh,
+    mesh_info = main_mesh,
 
     solver = {
         nonlinear = {
--- a/data/input_files/default.lua
+++ b/data/input_files/default.lua
@@ -39,7 +39,9 @@ nonlinear_solid = {
     K  = 5.0,
 
     -- loading parameters
-    tx = 0.0,
-    ty = 1.0e-3,
-    tz = 0.0,
+    traction = {
+        x = 0.0,
+        y = 1.0e-3,
+        z = 0.0,
+    },
 }
--- a/data/input_files/default.lua
+++ b/data/input_files/default.lua
@@ -1,15 +1,17 @@
-mesh = "../meshes/beam-hex.mesh"
-
--- serial and parallel refinement levels
-ser_ref_levels = 0
-par_ref_levels = 0
-
 -- Simulation time parameters
 t_final = 1.0
 dt      = 0.25
 
 -- Solver parameters
 nonlinear_solid = {
+    mesh_info = {
+        -- mesh file
+        mesh = "../data/meshes/beam-hex.mesh",
+        -- serial and parallel refinement levels
+        ser_ref_levels = 0,
+        par_ref_levels = 0,
+    },
+
     solver = {
         nonlinear = {
             rel_tol     = 1.0e-2,
--- a/data/input_files/default.lua
+++ b/data/input_files/default.lua
@@ -2,15 +2,17 @@
 t_final = 1.0
 dt      = 0.25
 
+primary_mesh = {
+    -- mesh file
+    mesh = "../meshes/beam-hex.mesh",
+    -- serial and parallel refinement levels
+    ser_ref_levels = 0,
+    par_ref_levels = 0,
+}
+
 -- Solver parameters
 nonlinear_solid = {
-    mesh_info = {
-        -- mesh file
-        mesh = "../data/meshes/beam-hex.mesh",
-        -- serial and parallel refinement levels
-        ser_ref_levels = 0,
-        par_ref_levels = 0,
-    },
+    mesh_info = primary_mesh,
 
     solver = {
         nonlinear = {
--- a/data/input_files/default.lua
+++ b/data/input_files/default.lua
@@ -12,8 +12,6 @@ main_mesh = {
 
 -- Solver parameters
 nonlinear_solid = {
-    mesh_info = main_mesh,
-
     solver = {
         nonlinear = {
             rel_tol     = 1.0e-2,
--- a/scripts/uberenv/packages/axom/package.py
+++ b/scripts/uberenv/packages/axom/package.py
@@ -55,7 +55,7 @@ class Axom(CMakePackage, CudaPackage):
     version('develop', branch='develop', submodules=True)
 
     # SERAC EDIT START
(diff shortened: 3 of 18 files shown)


azure-pipelines.yml
```diff
@@ -43,7 +43,7 @@ jobs:
   variables:
     DO_STYLE_CHECK: 'yes'
     VM_ImageName: 'ubuntu-18.04'
-    Compiler_ImageName: 'seracllnl/tpls:clang-10'
+    Compiler_ImageName: 'seracllnl/tpls:clang-10_10-14-20_19h-51m'
```
jamiebramwell: Awesome!

data/input_files/default.lua
```diff
@@ -1,15 +1,19 @@
...
-
--- serial and parallel refinement levels
-ser_ref_levels = 0
-par_ref_levels = 0
-
 -- Simulation time parameters
 t_final = 1.0
 dt      = 0.25
 
+primary_mesh = {
```
jamiebramwell: Maybe `default_mesh` or `main_mesh`? I want to avoid collision with the primary/secondary names for contact surfaces.

src/drivers/serac.cpp
```diff
@@ -37,15 +37,12 @@ namespace serac {
...
-  // mesh
-  serac::mesh::defineInputFileSchema(inlet);
-
   // Simulation time parameters
   inlet->addDouble("t_final", "Final time for simulation.")->defaultValue(1.0);
   inlet->addDouble("dt", "Time step.")->defaultValue(0.25);
 
   // Physics
-  serac::NonlinearSolid::defineInputFileSchema(inlet);
+  serac::NonlinearSolid::defineInputFileSchema(*inlet->getGlobalTable());
```
jamiebramwell: This is the dereferencing stuff that's going to get cleaned up in a future inlet PR right?

scripts/uberenv/packages/axom/package.py
```diff
@@ -55,7 +55,7 @@ class Axom(CMakePackage, CudaPackage):
     version('develop', branch='develop', submodules=True)
 
     # SERAC EDIT START
-    version('0.4.0p1', commit='d74c5d971c62a2263efa438736440ff28620dc3d', submodules="True")
```
white238, line 58: Are you sure this caused axom to get incremented?  I don't know if they track it in the mirror based on version number or commit...

joshessman-llnl: I was worried about that too, but it must have worked because the new Inlet features compile and run on Azure.

src/physics/nonlinear_solid.cpp
```diff
@@ -211,9 +212,12 @@ void NonlinearSolid::advanceTimestep(double& dt)
 
 NonlinearSolid::~NonlinearSolid() {}
 
-void NonlinearSolid::defineInputFileSchema(std::shared_ptr<axom::inlet::SchemaCreator> schema_creator)
+void NonlinearSolid::defineInputFileSchema(axom::inlet::Table& schema_creator)
```
white238: Should probably rename schema_creator

scripts/uberenv/packages/axom/package.py
white238: Maybe in the future we should rename these to like 0.4.0serac so it implies its a moving target specific to us.

src/drivers/serac.cpp
```diff
@@ -81,26 +78,15 @@ int main(int argc, char* argv[])
...
-  int par_ref_levels;
-  inlet->get("par_ref_levels", par_ref_levels);  // has default value
-
-  // Build mesh
-  std::string mesh_file_path;
-  inlet->get("mesh", mesh_file_path);  // required in input file
-  mesh_file_path = serac::input::findMeshFilePath(mesh_file_path, input_file_path);
-  auto mesh      = serac::buildMeshFromFile(mesh_file_path, ser_ref_levels, par_ref_levels);
+  // Annotate the mesh with the input filename
+  inlet->addTable("nonlinear_solid/mesh_info")->addString("input_file_path")->defaultValue(input_file_path);
```
jamiebramwell: I think the file path belongs at the root level instead of the `nonlinear_solid/mesh_info level`.

src/numerics/mesh_utils.cpp
```diff
@@ -170,16 +170,23 @@ std::shared_ptr<mfem::ParMesh> buildCuboidMesh(int elements_in_x, int elements_i
...
-  schema_creator->addInt("par_ref_levels", "Number of times to refine the mesh uniformly in parallel.")
-      ->defaultValue(0);
+  schema_creator.addInt("ser_ref_levels", "Number of times to refine the mesh uniformly in serial.")->defaultValue(0);
+  schema_creator.addInt("par_ref_levels", "Number of times to refine the mesh uniformly in parallel.")->defaultValue(0);
 }
 
 }  // namespace mesh
 }  // namespace serac
+
+std::shared_ptr<mfem::ParMesh> FromInlet<std::shared_ptr<mfem::ParMesh>>::operator()(axom::inlet::Table& base)
```
jamiebramwell: I was a bit confused at at first figuring out what this is doing. I know this really belongs in Axom, but renaming it `BuildFromInlet` might help.

src/physics/nonlinear_solid.cpp
```diff
@@ -228,7 +232,28 @@ void NonlinearSolid::defineInputFileSchema(std::shared_ptr<axom::inlet::SchemaCr
...
   auto solver_table = table->addTable("solver", "Linear and Nonlinear Solver Parameters.");
-  serac::EquationSolver::defineInputFileSchema(solver_table);
+  serac::EquationSolver::defineInputFileSchema(*solver_table);
 }
 
 }  // namespace serac
+
+using serac::NonlinearSolid;
+
+NonlinearSolid FromInlet<NonlinearSolid>::operator()(axom::inlet::Table& base)
```
jamiebramwell: What is the argument for using these templated builder methods as opposed to passing an inlet table to the constructor?

src/drivers/serac.cpp
joshessman-llnl: Yes, once https://github.com/LLNL/axom/pull/369 goes in I will start working on that

src/drivers/serac.cpp
joshessman-llnl: For encapsulation reasons Inlet currently only passes the table corresponding to the mesh object (in this case, `/nonlinear_solid/mesh_info`) instead of the global table to the function that reads an inlet table to create a mesh object, so that subtable needs to contain all the information required to construct a mesh.  One option that Chris and I discussed is adding unrestricted access to global variables from subtables.

src/physics/nonlinear_solid.cpp
joshessman-llnl: They are less intrusive (require no changes to the class definition) which makes it possible to define "deserialization" from third-party types (MFEM, STL, etc).  `nlohman::json` uses a similar approach, as does the STL for `std::hash`.

src/drivers/serac.cpp
jamiebramwell: I think that makes sense

jamiebramwell (approved):

src/drivers/serac.cpp
```diff
@@ -154,10 +112,9 @@ int main(int argc, char* argv[])
   std::set<int> trac_bdr = {2};
 
   // loading parameters
-  double tx, ty, tz;
-  inlet->get("nonlinear_solid/tx", tx);  // has default value
-  inlet->get("nonlinear_solid/ty", ty);  // has default value
-  inlet->get("nonlinear_solid/tz", tz);  // has default value
+  double tx = (*inlet)["nonlinear_solid/tx"];  // has default value
```
white238: These should be a struct.  Though I don't know what to name it, do we need a generic `Point` struct? @jamiebramwell ?

src/numerics/mesh_utils.cpp
white238: This is a good point. I've been mulling over the name of the `FromInlet` struct for a while and it's name is not as helpful as it could be.  `BuildFromInlet` is better for sure. Throwing out another idea `ConstructFromInlet`?

src/physics/nonlinear_solid.cpp
```diff
@@ -200,24 +201,45 @@ void NonlinearSolid::advanceTimestep(double& dt)
...
+{
+  auto mesh = base["mesh_info"].get<std::shared_ptr<mfem::ParMesh>>();
```
white238: Once we have multiple physics using the same mesh, this will need to be a shared mesh instance across them all as opposed to each creating a mesh in each of them.

src/physics/utilities/equation_solver.cpp
```diff
@@ -163,16 +163,18 @@ mfem::Operator& EquationSolver::SuperLUNonlinearOperatorWrapper::GetGradient(con
   return *superlu_grad_mat_;
 }
 
-void EquationSolver::defineInputFileSchema(std::shared_ptr<axom::inlet::SchemaCreator> schema_creator)
+void EquationSolver::defineInputFileSchema(axom::inlet::Table& table)
 {
-  auto nonlinear_table = schema_creator->addTable("nonlinear", "Newton Equation Solver Parameters")->required(true);
+  auto nonlinear_table = table.addTable("nonlinear", "Newton Equation Solver Parameters");
+  nonlinear_table->required(false);
```
white238, line 172: Did you mean to flip this to required=false?

src/physics/utilities/equation_solver.cpp
```diff
@@ -181,3 +183,50 @@ void EquationSolver::defineInputFileSchema(std::shared_ptr<axom::inlet::SchemaCr
...
+    params.prec       = serac::HypreBoomerAMGPrec{};
+    params.lin_solver = serac::LinearSolver::GMRES;
+  } else if (solver_type == "minres") {
+    params.prec       = serac::HypreSmootherPrec{mfem::HypreSmoother::l1Jacobi};
+    params.lin_solver = serac::LinearSolver::MINRES;
+  } else {
+    serac::logger::flush();
+    std::string msg = fmt::format("Unknown Linear solver type given: {0}", solver_type);
+    SLIC_ERROR(msg);
+    serac::exitGracefully(true);
```
white238: With your previous fix this exitGracefully can be removed.

src/numerics/mesh_utils.cpp
white238: I've got it! `TemplateCreateAndFillMagicStruct`

src/physics/utilities/equation_solver.cpp
joshessman-llnl: Yeah, in general an `EquationSolver` doesn't needed nonlinearparameters - it's just the solid solver that needs it.

src/physics/nonlinear_solid.cpp
joshessman-llnl: Currently we don't inject the member solvers (i.e., `ThermalSolid::ThermalSolid(NonlinearSolid, ThermalCondution)`), so this shouldn't require any changes to the Inlet stuff - we won't be calling `FromInlet<NonlinearSolid>` from `FromInlet<ThermalSolid>`.

src/numerics/mesh_utils.cpp
jamiebramwell: I think either `Build` or `Construct` would be good. No comment on `MagicStruct`. 😄

src/drivers/serac.cpp
jamiebramwell: This is really a generic vector of length dimension of the problem. I think a generic vector struct of three (or two) doubles makes sense.

src/physics/nonlinear_solid.cpp
jamiebramwell: I think the problem is if you have something like this:
```
main_mesh = {some mesh stuff}

nonlinear_solid = {
    mesh = main_mesh
}

thermal_conduction = {
    mesh = main_mesh 
}
```

This can happen when you have two non-coupled physics modules operating on the same mesh. For example, the solid module could deform the mesh and the thermal solver then operates in a non-coupled fashion on the deformed geometry. This can't happen if we create two separate mesh objects in `FromInlet<NonlinearSolid>` and `FromInlet<ThermalConduction>`.

joshessman-llnl: Hmm, we don't have a way of checking referential equality in Inlet, but could there ever be a case where the user would set up a problem as you've described and *not* want to have the mesh be shared?  If not, we could "cache" constructed mesh objects and assume that if a mesh is requested with the same filename, refinement, etc, then the cached mesh should be returned.  Even if that was needed, maybe we could add an optional `unique` flag to the mesh parameters.

src/drivers/serac.cpp
joshessman-llnl: I went ahead and added a conversion from inlet to `mfem::Vector`, though I'm not 100% sure where it belongs - I put it in `infrastructure/input.*pp` for now because it's somewhat input related.

src/physics/nonlinear_solid.cpp
jamiebramwell: Probably not. However, this cache option seems a bit convoluted (particularly for a user-facing option).

white238: There are going to be complicated objects that just can't be constructed directly from Inlet without jumping through so many hoops that make the program flow and user input convoluted.  This is one of them.

Unless it is a very basic type that can be constructed easily from the template we should be pulling the info out of Inlet as a struct then passing that into the constructor along with the other required information.

In this example, something like this:

```
auto physicsInputInfo = myInlet.get<SomePhysicsInputInfo>(physicsTable);
SomePhysics somePhysics(physicsInputInfo, mesh);
```

white238 (approved):

joshessman-llnl:
LGTM

PR #249, opened 2020-10-20
Add block incomplete LU preconditioner option

### LINKED ISSUE #185: Add BlockILU as preconditioner option
For high contrast DG diffusion problems, we should add a block ILU preconditioner option. It is [available in MFEM](https://github.com/mfem/mfem/blob/1cf578ea0b039b6e5e32431d1b8080663ef0ed20/linalg/solvers.hpp#L624).

joshessman-llnl:
Resolves #185 

The actual changes required to expose mfem's block ILU class were minimal, but to check for regressions I also pasted in mfem's ex9p and made some small modifications (will annotate) to use `serac::EquationSolver` and to test against the results obtained with the unmodified version.  Ideally this test can be removed once the analogous Smith module is open-sourced.


tests/mfem_ex9p_blockilu.cpp
```diff
@@ -0,0 +1,758 @@
...
+      : M(M_),
+        K(K_),
+        A(NULL),
+        linear_solver(M.GetComm(),
+                      serac::IterativeSolverParameters{.rel_tol     = 1e-9,
+                                                       .abs_tol     = 0.0,
+                                                       .print_level = 0,
+                                                       .max_iter    = 100,
+                                                       .lin_solver  = serac::LinearSolver::GMRES,
+                                                       .prec        = serac::BlockILUPrec{fes.GetFE(0)->GetDof()}}),
```
joshessman-llnl, line 125: The only functional change was replacing the GMRES/BlockILU members with a `serac::EquationSolver`

tests/mfem_ex9p_blockilu.cpp
```diff
@@ -0,0 +1,758 @@
...
+        cout << "time step: " << ti << ", time: " << t << endl;
+      }
+
+      // 11. Extract the parallel grid function corresponding to the finite
+      //     element approximation U (the local solution on each processor).
+      *u = *U;
+
+      if (myid == 0) {
+        EXPECT_FLOAT_EQ(u->Norml2(), UNMODIFIED_SOLN.at((ti / vis_steps) - 1));
+      }
```
joshessman-llnl, line 506: I also hardcoded the results from the unmodified version of the example and added a comparison.

jamiebramwell (approved):
Looks good!

tests/mfem_ex9p_blockilu.cpp
jamiebramwell: This should probably be moved to an integration test with a binary comparison (along with most of the current tests) once that capability comes online.

white238 (approved):

tests/mfem_ex9p_blockilu.cpp
white238: Also moved to a more Serac way of programming at the same time but not now.. in the future

joshessman-llnl:
LGTM

PR #251, opened 2020-10-23
hotfix: explicit dependency on mfem

joshessman-llnl:
Having the code that reads in a vector from Inlet in the infrastructure directory as part of input.cpp/input.hpp required an explicit dependency on MFEM - not sure why this didn't cause any problems until the full TPL builds.

### DIFF
--- a/src/infrastructure/CMakeLists.txt
+++ b/src/infrastructure/CMakeLists.txt
@@ -22,7 +22,7 @@ set(infrastructure_sources
     terminator.cpp
     )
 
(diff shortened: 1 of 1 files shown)


joshessman-llnl:
LGTM

jamiebramwell (approved):

jamiebramwell:
@white238 @joshessman-llnl could you take a quick look at this?

samuelpmishLLNL (approved):

white238:
Was it possibly being brought in as a part of the Axom target?

PR #254, opened 2020-10-27
Integrate MFEM's CUDA support into serac + AMGX, merge in changes from MFEM spack package

joshessman-llnl:
This PR now builds `mfem+cuda` when Serac's `+cuda` variant is enabled, and adds a new `accelerator` "component" that wraps MFEM's `Device` object.

Actual performance improvements are probably negligible at best as the relevant integrators used in the driver are not (to my knowledge) GPU-enabled.  It would probably be worthwhile to test this with a large mesh if more precise information is desired.

### EDIT

This PR also adds an AMGX preconditioner option (and an associated test).  This required a version bump to MFEM 4.2.


cmake/thirdparty/FindMFEM.cmake
```diff
@@ -125,8 +125,18 @@ endif()
...
-    INCLUDES      ${MFEM_INCLUDE_DIRS}
-    LIBRARIES     ${MFEM_LIBRARIES}
-    TREAT_INCLUDES_AS_SYSTEM ON)
+if(ENABLE_CUDA)
+    message(STATUS "MFEM: Adding required CUDA includes/libraries")
+    blt_register_library(
+        NAME          mfem
+        INCLUDES      ${MFEM_INCLUDE_DIRS} ${CUDA_INCLUDE_DIRS}
+        # FIXME: How does this work with first-class CMake CUDA support
+        LIBRARIES     ${MFEM_LIBRARIES} ${CMAKE_CUDA_LINK_FLAGS} ${CUDA_LIBRARIES}
```
white238: Could you not use the BLT CUDA registered libraries? `cuda` or `cuda_runtime`

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@upstream_gfortran-cuda.cmake
```diff
@@ -57,9 +57,9 @@ set(MPIEXEC_EXECUTABLE "/usr/tce/packages/spectrum-mpi/spectrum-mpi-rolling-rele
 #---------------------------------------
 # Library Dependencies
 #---------------------------------------
-set(TPL_ROOT "/usr/WS2/smithdev/libs/blueos_3_ppc64le_ib_p9/2020_10_19_15_04_12/clang-upstream_gfortran" CACHE PATH "")
+set(TPL_ROOT "/usr/WS2/smithdev/libs/blueos_3_ppc64le_ib_p9/2020_10_27_08_37_57/clang-upstream_gfortran" CACHE PATH "")
 
-set(AXOM_DIR "${TPL_ROOT}/axom-0.4.0serac" CACHE PATH "")
+set(AXOM_DIR "${TPL_ROOT}/axom-0.4.0p1" CACHE PATH "")
```
white238: was this reverted on purpose?

joshessman-llnl: It was actually an oversight when I was juggling images around - the change to our version of axom's package.py is actually in https://github.com/LLNL/serac/pull/88/files#diff-f2386102a0ae90f37dd97db98adcfd7e856a8bf99811ffe7649e55245e2368e9R58 (which hasn't been merged).

That PR needs to be merged in before this one, so there will be a rebuild with the correct tag name.

cmake/thirdparty/FindMFEM.cmake
joshessman-llnl: I'm not sure that BLT-registered libraries can be used when registering a BLT library - I could be wrong though.

Either way, the runtime libraries alone are not sufficient because MFEM also requires `cusparse` but doesn't reference it in the `MFEM_EXT_LIBS` variable that we use to get the paths to the rest of its libraries.

white238: You can, just list it on the LIBRARIES line.  You should also be able to just add `cusparse`.

white238: Just to correct my previous statement... blt registered libraries should go on the DEPENDS_ON line.  /hangHeadInShame

joshessman-llnl:
This won't work properly until Axom can be built against a version of MFEM built with CUDA support.

scripts/uberenv/packages/mfem/package.py
```diff
@@ -44,6 +44,7 @@ class Mfem(Package):
     # SERAC EDIT BEGIN
     # Version spec used in serac based on mfem@develop commit SHA
     version('4.1.0p1', commit='bb3c788a05f430bcfed03065ba39868974964924')
+    version('4.2', tag='v4.2')
```
jamiebramwell: We should pull in the official tarball version of 4.2.0 from their spack package. These package updates should also probably get incorporated: https://github.com/spack/spack/pull/19748 .

cmake/thirdparty/FindMFEM.cmake
joshessman-llnl: This will be handled in a cleaner way once https://github.com/LLNL/blt/pull/421 is merged

scripts/uberenv/packages/mfem/package.py
joshessman-llnl: The Spack changes have been merged in, and once this PR is merged I can open a PR to add the AMGX integration to the official Spack package (which isn't part of that PR)

joshessman-llnl:
LGTM

src/physics/utilities/solver_config.hpp
```diff
@@ -82,6 +82,12 @@ struct HypreBoomerAMGPrec {
   mfem::ParFiniteElementSpace* pfes = nullptr;
 };
 
+/**
+ * @brief Stores the information required to configure a NVIDIA AMGX preconditioner
+ */
+struct AMGXPrec {
```
white238, line 88: Why is this just an empty struct?

white238 (approved):

src/physics/utilities/solver_config.hpp
joshessman-llnl: Right now it's only being used as the "tag" in the preconditioner variant, as the default settings appeared to be sufficient for the test I added, but in the future I think we'll want to configure AMGX as we do Hypre's AMG preconditioner implementation.

joshessman-llnl:
@jamiebramwell Now that the "official" 4.2 updates to our version of the MFEM spack package have been merged in, was there anything else you were looking to see as part of this PR?

jamiebramwell:
I'll give this another look, but I think it should be good to go as long as the TPLs are built on LC and in docker.

host-configs/docker/docker-linux-ubuntu16.04-x86_64-gcc@8.1.0.cmake
```diff
@@ -59,7 +59,7 @@ set(PARMETIS_DIR "${TPL_ROOT}/parmetis-4.0.3" CACHE PATH "")
 
 set(SUPERLUDIST_DIR "${TPL_ROOT}/superlu-dist-6.1.1" CACHE PATH "")
 
-set(MFEM_DIR "${TPL_ROOT}/mfem-4.1.0p1" CACHE PATH "")
+set(MFEM_DIR "${TPL_ROOT}/mfem-4.2.0" CACHE PATH "")
```
jamiebramwell, line 62: 👍

jamiebramwell (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #256, opened 2020-10-28
Unambiguous ownership of FiniteElementState objects

joshessman-llnl:
Assigning ownership of the `FEState` objects to just the "base" solver allows them to be moved from the heap to the stack.
That is, instead of `ThermalSolid` sharing ownership of the temperature field with its `ThermalConduction` module, the conduction module has exclusive ownership and the `ThermalSolid` just holds a reference.  

The `BaseSolver::state_` vector now also holds non-owning references.  This adds a little bit of extra code to dereference the `reference_wrapper` with `get()`, but this is internal to the `BaseSolver` only.  Alternatively, a vector of raw pointers could be used, though the absence of ownership might not be as clear as with a vector of `ref_wrapper`s.

### DIFF
--- a/src/physics/base_physics.cpp
+++ b/src/physics/base_physics.cpp
@@ -26,7 +26,6 @@ BasePhysics::BasePhysics(std::shared_ptr<mfem::ParMesh> mesh)
 BasePhysics::BasePhysics(std::shared_ptr<mfem::ParMesh> mesh, int n, int p) : BasePhysics(mesh)
 {
   order_ = p;
-  // state_.resize(n);
   gf_initialized_.assign(n, false);
 }
 
--- a/src/physics/base_physics.cpp
+++ b/src/physics/base_physics.cpp
@@ -26,7 +26,7 @@ BasePhysics::BasePhysics(std::shared_ptr<mfem::ParMesh> mesh)
 BasePhysics::BasePhysics(std::shared_ptr<mfem::ParMesh> mesh, int n, int p) : BasePhysics(mesh)
 {
   order_ = p;
-  state_.resize(n);
+  // state_.resize(n);
   gf_initialized_.assign(n, false);
 }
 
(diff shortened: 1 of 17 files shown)


src/physics/base_physics.cpp
```diff
@@ -40,17 +39,22 @@ void BasePhysics::setState(const std::vector<serac::GeneralCoefficient>& state_c
...
-void BasePhysics::setState(const std::vector<std::shared_ptr<serac::FiniteElementState> >& state)
+void BasePhysics::setState(std::vector<serac::FiniteElementState>&& state)
 {
   SLIC_ASSERT_MSG(state.size() > 0, "State vector array of size 0.");
-  state_ = state;
+  // To avoid making a shallow copy of the references, we move individually
+  for (std::size_t i = 0; i < state.size(); i++) {
+    // Assigning to the FiniteElementState being referenced,
+    // not the reference itself
+    state_[i].get() = std::move(state[i]);
```
jamiebramwell, line 53: Could you explain a bit why this is `std::move` instead of just assignment?

jamiebramwell (commented):
This is great! I just have a small question about using `std::move` in the `setState` method.

src/physics/base_physics.cpp
joshessman-llnl: This function isn't currently used anywhere, but I had envisioned it being used in a context where the vector of FEState objects would be constructed by the user only for passing to this function, which is why I required that the vector be moved into this function.  Since the vector is an rvalue, we can use move assignment on its elements instead of copy assignment, which isn't possible for `FiniteElementState` as things like a pointer to `FiniteElementCollection` can't be copied.

Let me know if I've misunderstood the intent of this function though - without seeing it used I wasn't 100% sure.

jamiebramwell: I think I added this just because I thought it would be useful to have for building state from stuff we don't control. I think this assumption is fine as long as it's noted in the doxygen comments that this steals the data.

jamiebramwell (approved):

white238 (approved):

jamiebramwell:
LGTM

jamiebramwell:
LGTM

PR #257, opened 2020-10-28
Re-enable direct solver tests

joshessman-llnl:
In #248 we upgraded our SuperLU version to 6.1.1 from 5.4.0 - the memory safety issues that we were running into appear to have been resolved between those versions, so it should be safe to re-enable the direct solver tests.  I've ran the tests on a couple machines and none of them report any memory issues as they did previously.


jamiebramwell (approved):

white238 (approved):

joshessman-llnl:
LGTM

PR #262, opened 2020-10-30
Update Inlet version to remove shared_ptr

joshessman-llnl:
Inlet was recently updated to not rely on shared pointers, meaning that we can remove all the messy deferencing when accessing with `operator[]`, for example.

The API should *hopefully* be more stable from here on out, so we can start look at extending inlet to the other modules.

### DIFF
--- a/src/drivers/serac.cpp
+++ b/src/drivers/serac.cpp
@@ -35,21 +35,21 @@ namespace serac {
 
 //------- Input file -------
 
-void defineInputFileSchema(std::shared_ptr<axom::inlet::Inlet> inlet, int rank)
+void defineInputFileSchema(axom::inlet::Inlet& inlet, int rank)
 {
   // Simulation time parameters
-  inlet->addDouble("t_final", "Final time for simulation.")->defaultValue(1.0);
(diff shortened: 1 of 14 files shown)


src/drivers/serac.cpp
```diff
@@ -83,12 +83,12 @@ int main(int argc, char* argv[])
   datastore.getRoot()->save("serac_input.json", "json");
 
   // Build the mesh
-  auto mesh_info      = (*inlet)["main_mesh"].get<serac::mesh::InputInfo>();
+  auto mesh_info      = inlet["main_mesh"].get<serac::mesh::InputInfo>();
```
jamiebramwell, line 86: 👍

jamiebramwell (approved):
Looks good!

jamiebramwell:
LGTM

white238 (approved):

joshessman-llnl:
Ran into an issue building TPLs on LC so the LC hostconfigs haven't been updated yet

joshessman-llnl:
LGTM

Issue #252, opened 2020-10-23
Create adapter class for cleaner coefficient interface


jamiebramwell:
Josh and I spoke yesterday morning about this PR, and how to manage the lifetime of the coefficient classes. To summarize, it looks like most of the mfem::Coefficient copy constructors are shallow copies, which makes it impossible to take ownership of a coefficient that is passed in. One idea that we discussed was to not actually pass in mfem::Coefficients at all, and instead construct them in the solvers. An example of the difference is given below:

previous implementation
```cpp
// solver class
void MySolver::SetTemperature(const mfem::Coefficient & coef) {
  ... // do something, but we can't really copy what the user passes in, so we're stuck with references or pointers
}

-----------------------

// user-code
mfem::ConstantCoefficient values(3.14); // the user must keep this around or the solver will have a dangling reference
solver.setTemperature(values);

or

serac::StdFunctionCoefficient([](auto x){ return x(1); }); // the user must keep this around or the solver will have a dangling reference
solver.setTemperature(values);
```

proposed implementation
```cpp
struct coef {
  public:
    coef (double x) : self_(std::make_unique<mfem::ConstantCoefficient>(x)) {}
    coef (std::function <double(double)> f) : self_(std::make_unique<serac::StdFunctionCoefficient>(f)) {}
    operator mfem::Coefficient&() { return *self_; }
    operator mfem::Coefficient&() const { return *self_; }
    std::unique_ptr<mfem::Coefficient> self_;
};

-----------------------

// solver class
void MySolver::SetTemperature(const coef & c) {
  ... // doubles and std::functions can be copied, so the solver can own and manage the lifetime directly
}

-----------------------

// user code
solver.setTemperature(3.14);

or

solver.setTemperature([](auto x){ return x(1); });
```

This approach also has the benefit of removing more mfem data types from the solver interface.

This is also a big change to serac though, and probably warrants a discussion from the other developers.

_Originally posted by @samuelpmish in https://github.com/LLNL/serac/pull/220#issuecomment-701544878_

jamiebramwell:
Closing this due to the pending weak form refactoring.

PR #263, opened 2020-10-30
Possibly improve std::variant readability

joshessman-llnl:
`std::variant` is kind of ugly to use without true pattern matching (hopefully in C++23), but `get_if` at least allows the type name to not be specified twice, which might be a little nicer to read.

There's some extra dereferencing involved with the coefficients, but #252 should clean that up.


src/physics/thermal_conduction.cpp
```diff
@@ -21,18 +21,16 @@ ThermalConduction::ThermalConduction(int order, std::shared_ptr<mfem::ParMesh> m
...
-    auto dyn_params = std::get<DynamicSolverParameters>(params);
-    setTimestepper(dyn_params.timestepper);
+  else if (auto dyn_params = std::get_if<DynamicSolverParameters>(&params)) {
+    setTimestepper(dyn_params->timestepper);
     // Save these to initialize the DynamicConductionOperator later
-    dyn_M_params_ = dyn_params.M_params;
-    dyn_T_params_ = dyn_params.T_params;
+    dyn_M_params_ = dyn_params->M_params;
+    dyn_T_params_ = dyn_params->T_params;
   }
```
jamiebramwell, line 36: Should we error if this block totally falls through or is that being too cautious?

src/physics/thermal_conduction.cpp
```diff
@@ -21,18 +21,16 @@ ThermalConduction::ThermalConduction(int order, std::shared_ptr<mfem::ParMesh> m
   state_.push_back(temperature_);
 
   // If it's just the single set of params for a quasistatic K solve...
-  if (std::holds_alternative<LinearSolverParameters>(params)) {
-    auto K_params = std::get<LinearSolverParameters>(params);
-    K_inv_        = EquationSolver(mesh->GetComm(), K_params);
+  if (auto K_params = std::get_if<LinearSolverParameters>(&params)) {
```
jamiebramwell, line 24: I like this syntax more.

jamiebramwell (approved):
I think this is a good improvement for readability.

jamiebramwell:
LGTM

src/physics/thermal_conduction.cpp
white238: i like being cautious

white238 (approved):
Much better! thanks @joshessman-llnl !

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #264, opened 2020-11-03
Move boundary attribute information to input file, use input files in tests

joshessman-llnl:
This PR moves some boundary condition information (the attribute set) to the input file via a new `BoundaryConditionInputInfo` struct, which just contains its name and attribute set for now.  

This allows tests to be described (mostly) in the input file instead of the C++ source.  I've converted the `serac_component_bc` test as an example.

The only big remaining test-specific information is the coefficients, so the next logical step after this PR is determining how coefficients should be described in inlet (and by extension Lua).  The rest is boilerplate that could possibly be abstracted into a function - this would make the tests more readable.

### DIFF
--- a/data/input_files/default.lua
+++ b/data/input_files/default.lua
@@ -13,19 +13,20 @@ main_mesh = {
 -- Solver parameters
 nonlinear_solid = {
     solver = {
-        nonlinear = {
-            rel_tol     = 1.0e-2,
-            abs_tol     = 1.0e-4,
-            max_iter    = 500,
-            print_level = 0,
-        },
-
         linear = {
             rel_tol     = 1.0e-6,
             abs_tol     = 1.0e-8,
             max_iter    = 5000,
             print_level = 0,
             solver_type = "gmres",
+            prec_type   = "AMG",
+        },
+
+        nonlinear = {
+            rel_tol     = 1.0e-2,
+            abs_tol     = 1.0e-4,
+            max_iter    = 500,
+            print_level = 0,
         },
     },
 
@@ -41,5 +42,17 @@ nonlinear_solid = {
         x = 0.0,
         y = 1.0e-3,
         z = 0.0,
+    },
+
+    -- boundary condition parameters
+    boundary_conds = {
+        [1] = {
+            name = "displacement",
+            attrs = {1},
+        },
+        [2] = {
+            name = "traction",
+            attrs = {2},
+        },
     },
 }
--- a/data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
+++ b/data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
@@ -1,0 +1,63 @@
+-- Comparison information
+expected_l2norm = 0.03330115
+epsilon = 0.0001
+
+-- Simulation time parameters
+dt      = 1.0
+
+main_mesh = {
+    -- mesh file
+    mesh = "../../../../meshes/square.mesh",
+    -- serial and parallel refinement levels
+    ser_ref_levels = 2,
+    par_ref_levels = 0,
+}
+
+-- Solver parameters
+nonlinear_solid = {
+    solver = {
+        linear = {
+            rel_tol     = 1.0e-8,
+            abs_tol     = 1.0e-12,
+            max_iter    = 5000,
+            print_level = 0,
+            solver_type = "minres",
+            prec_type   = "L1JacobiSmoother",
+        },
+
+        nonlinear = {
+            rel_tol     = 1.0e-6,
+            abs_tol     = 1.0e-8,
+            max_iter    = 5000,
+            print_level = 1,
+        },
+    },
+
+    -- polynomial interpolation order
+    order = 2,
+
+    -- neo-Hookean material parameters
+    mu = 0.25,
+    K  = 10.0,
+
+    -- loading parameters
+    traction = {
+        x = 0.0,
+        y = 1.0e-3,
+        z = 0.0,
+    },
+
+    -- boundary condition parameters
+    boundary_conds = {
+        {
+            name = "displacement_x",
+            -- boundary attribute 1 (index 0) is fixed (Dirichlet) in the x direction
+            attrs = {1},
+        },
+        {
+            name = "displacement_y",
+            -- boundary attribute 1 (index 0) is fixed (Dirichlet) in the x direction
+            attrs = {2},
+        },
+    },
+}
--- a/data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
+++ b/data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
@@ -40,13 +40,6 @@ nonlinear_solid = {
     mu = 0.25,
     K  = 10.0,
 
-    -- loading parameters
-    traction = {
-        x = 0.0,
-        y = 1.0e-3,
-        z = 0.0,
-    },
-
     -- boundary condition parameters
     boundary_conds = {
         {
@@ -56,7 +49,6 @@ nonlinear_solid = {
         },
         {
             name = "displacement_y",
-            -- boundary attribute 1 (index 0) is fixed (Dirichlet) in the x direction
             attrs = {2},
         },
     },
--- a/src/drivers/serac.cpp
+++ b/src/drivers/serac.cpp
@@ -119,13 +119,15 @@ int main(int argc, char* argv[])
   auto traction_coef = std::make_shared<serac::VectorScaledConstantCoefficient>(traction);
 
   // Set the boundary condition information
-  const auto disp_info = solid_solver_info.boundary_conditions.find("displacement");
-  SLIC_ERROR_ROOT_IF(disp_info == solid_solver_info.boundary_conditions.end(), rank, "No displacement BC specified");
-  solid_solver.setDisplacementBCs(disp_info->second.attrs, disp_coef);
-
-  const auto trac_info = solid_solver_info.boundary_conditions.find("traction");
-  SLIC_ERROR_ROOT_IF(trac_info == solid_solver_info.boundary_conditions.end(), rank, "No traction BC specified");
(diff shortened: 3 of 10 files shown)


data/input_files/default.lua
```diff
@@ -42,4 +43,16 @@ nonlinear_solid = {
         y = 1.0e-3,
         z = 0.0,
     },
+
+    -- boundary condition parameters
+    boundary_conds = {
```
jamiebramwell, line 48: 👍

data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
```diff
@@ -0,0 +1,55 @@
+-- Comparison information
+expected_l2norm = 0.03330115
```
jamiebramwell, line 2: I can see this eventually moving to an integrated test solution. @white238 do you think it's best to put it in an input deck for now? It definitely improves readability.

src/infrastructure/input.cpp
```diff
@@ -102,3 +108,16 @@ mfem::Vector FromInlet<mfem::Vector>::operator()(const axom::inlet::Table& base)
   }
   return result;
 }
+
+serac::input::BoundaryConditionInputInfo FromInlet<serac::input::BoundaryConditionInputInfo>::operator()(
+    const axom::inlet::Table& base)
+{
+  serac::input::BoundaryConditionInputInfo result;
+  // Build a set with just the values of the map
+  auto bdr_attr_map = base["attrs"].get<std::unordered_map<int, int>>();
```
jamiebramwell, line 117: I'm confused why this is an unordered map instead of a `std::vector`. Is it because lua stuff is implicitly maps instead of vectors?

src/infrastructure/input.cpp
```diff
@@ -102,3 +108,16 @@ mfem::Vector FromInlet<mfem::Vector>::operator()(const axom::inlet::Table& base)
...
+  for (const auto& [_, val] : bdr_attr_map) {
```
jamiebramwell, line 118: Is the underscore a special character for an ignored variable? I can't find a reference on this.

src/physics/nonlinear_solid.hpp
```diff
@@ -64,6 +64,9 @@ class NonlinearSolid : public BasePhysics {
     // Lame parameters
     double mu;
     double K;
+
+    // Boundary condition information
+    std::unordered_map<std::string, input::BoundaryConditionInputInfo> boundary_conditions;
```
jamiebramwell: What does the string in this map represent?

data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
```diff
@@ -0,0 +1,55 @@
...
+    order = 2,
+
+    -- neo-Hookean material parameters
+    mu = 0.25,
+    K  = 10.0,
+
+    -- boundary condition parameters
+    boundary_conds = {
+        {
+            name = "displacement_x",
```
jamiebramwell, line 46: Do we want separate boundary condition names instead of a set of active component directions? I lean toward the latter, but I could be persuaded. Thoughts @samuelpmishLLNL @solberg2 ?

src/infrastructure/input.cpp
joshessman-llnl: Yeah, everything in Lua is a table, and indices can be arbitrary, e.g.:
```lua
map = { [7] = "a", [13] = "c" }
```
which doesn't conform to the contiguity implied by `std::vector`.

src/infrastructure/input.cpp
joshessman-llnl: It's a popular convention in Python for ignoring something when unpacking a tuple, and it's used here as `std::ignore` doesn't work in structured bindings as it does in `std::tie`, which is a similar context.  For reasons I do not understand a standardized way of ignoring in a structured binding was not part of the C++17 standard.

src/physics/nonlinear_solid.hpp
joshessman-llnl: The keys correspond to names of boundary conditions - I'm not sure if that's the best way to disambiguate between them, though.

data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
white238: Yes we should be moving as many of these to input decks as we can.  It will assist in the integrated tests as well.

src/infrastructure/input.cpp
jamiebramwell: Since this smells like a language feature (that I agree probably should have been implemented), I think a name for this variable would be useful to keep modern C++ newbies like myself from getting confused.

src/physics/nonlinear_solid.hpp
jamiebramwell: How is this different from the name in `BoundaryConditionInputInfo`?

joshessman-llnl: It's duplicated, my thinking was that with name-identified BCs it would be useful to have a map for easier lookup as opposed to just `std::vector<BoundaryConditionInputInfo>`.

It would probably be more natural to allow for string-indexed arrays in Inlet but currently Inlet considers an array to be integer-indexed only.  If that ends up being needed we could add it and then do something like:
```lua
boundary_conds = {
   ["displacement"] = { attrs = {1}, },
   ["traction"] = { attrs = {2}, },
}
```

jamiebramwell: I think the string index is a good thing to work toward, but for now I think reducing the duplication makes sense for clarity. I can imagine each physics module looping through the boundary condition input and setting them up in a  like this:
```cpp
for (const auto& bc : boundary_conditions) {
  if (bc.name == "displacement") {
    // add a displacement bc...
  } else if (bc.name == "traction") {
    // add a traction bc...
  }
}

File: src/drivers/serac.cpp [hunk 7]
```diff
@@ -108,23 +108,24 @@ int main(int argc, char* argv[])
...
   auto traction_coef = std::make_shared<serac::VectorScaledConstantCoefficient>(traction);
 
   // Set the boundary condition information
-  solid_solver.setDisplacementBCs(ess_bdr, disp_coef);
-  solid_solver.setTractionBCs(trac_bdr, traction_coef);
+  const auto disp_info = solid_solver_info.boundary_conditions.find("displacement");
+  SLIC_ERROR_ROOT_IF(disp_info == solid_solver_info.boundary_conditions.end(), rank, "No displacement BC specified");
+  solid_solver.setDisplacementBCs(disp_info->second.attrs, disp_coef);
+
+  const auto trac_info = solid_solver_info.boundary_conditions.find("traction");
```
jamiebramwell: Is this the only place where you're using the duplicate names field? If so, I think this could be changed to a loop over the boundary conditions instead.

tests/serac_component_bc.cpp
```diff
@@ -9,49 +9,67 @@
...
-  // boundary attribute 1 (index 0) is fixed (Dirichlet) in the x direction
-  std::set<int> ess_bdr = {1};
+  int dim = mesh->Dimension();
 
   // define the displacement vector
   auto disp_coef = std::make_shared<StdFunctionCoefficient>([](mfem::Vector& x) { return x[0] * -1.0e-1; });
 
   // Pass the BC information to the solver object setting only the z direction
-  solid_solver.setDisplacementBCs(ess_bdr, disp_coef, 0);
+  solid_solver.setDisplacementBCs(solid_solver_info.boundary_conditions.at("displacement").attrs, disp_coef, 0);
```
jamiebramwell: While the `at` lookup instead of a loop is nice, I think it adds unnecessary complexity and duplication, especially since this will change in the near future anyway.

data/input_files/tests/nonlinear_solid/serac_component_bc/qs_attribute_solve.lua
jonwong12: That's a good question. I think for certain b.c.'s "tangent" and "normal" are also good descriptions vs `x-,y-,z-`.

solberg2: "Tangent" is a pretty difficult thing to handle in 3D, because of course there are 2 orthogonal tangent directions, and they have to be defined somehow.

For many verification problems simple component directions are good.  What we do is allow the user to specify a set of local component directions "X_LOCAL", "Y_LOCAL", which then we provide a facility for defining these.

jamiebramwell: I think this is good for now, but we probably want to have a larger design discussion about this. I'll create an issue.

src/infrastructure/input.cpp
jamiebramwell: I think this is my only outstanding issue. It's not a big deal so I could be convinced otherwise. Any thoughts @white238 ?

white238: `_` is a pretty common convention. I am fine with `_` or `unused` if that's more clear.

jamiebramwell (approved):

samuelpmishLLNL:
Would it make sense to just add new tests with Inlet input files, as opposed to removing the existing ones?

Is it intended that a user would only ever use Serac through Inlet + Lua (or similar)?

jamiebramwell:
> 
> 
> Would it make sense to just add new tests with Inlet input files, as opposed to removing the existing ones?
> 
> Is it intended that a user would only ever use Serac through Inlet + Lua (or similar)?

I'm not sure if there's much added benefit for having these tests duplicated for in source magic numbers vs. lua decks. In my mind, these are really integration tests packaged as unit ones, so it makes sense to drive them from an input file using the yet-to-be-developed integration test framework. I think real unit tests can stay as-is (and we should definitely add more).

white238 (approved):

white238:
LGTM

joshessman-llnl:
LGTM

PR #267, opened 2020-11-05
Moving to the new spack approved llvm package for clang-format / clang-tidy

white238:
This moves our hardcoded path to a use the `llvm+clang` spack package

PR #268, opened 2020-11-05
Update MFEM to 4.2.0

jamiebramwell:
This updates our MFEM dependency to v4.2.0.

### DIFF
--- a/scripts/uberenv/packages/mfem/package.py
+++ b/scripts/uberenv/packages/mfem/package.py
@@ -50,6 +50,11 @@ class Mfem(Package):
     # other version.
     version('develop', branch='master')
 
+    version('4.2.0',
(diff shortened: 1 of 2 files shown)


jamiebramwell:
I created this off of the clang format branch, so it should get merged after https://github.com/LLNL/serac/pull/267 .

joshessman-llnl:
This is also part of #254, so I'm not sure how we want to interleave these PRs that require TPL rebuilds

jamiebramwell:
> 
> 
> This is also part of #254, so I'm not sure how we want to interleave these PRs that require TPL rebuilds

Ahhhh I missed that. I would just update the MFEM pointer in that PR to the official 4.2.0 release and call it good.

PR #270, opened 2020-11-09
Bugfix/white238/turnoffdocs

white238:
LC's Sphinx and Doxygen keep failing with random incompatibilities.  This turns off docs for the build scripts until we rebuild a new set of Devtools with our own Sphinx/Doxygen.

### DIFF
--- a/cmake/serac-config.cmake.in
+++ b/cmake/serac-config.cmake.in
@@ -88,13 +88,16 @@ if(NOT SERAC_FOUND)
     endif()
   endif()
 
-   # tribol
+  # tribol
   if(SERAC_USE_TRIBOL)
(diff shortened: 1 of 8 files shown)


white238:
LGTM

jamiebramwell (approved):

joshessman-llnl (approved):
We may also be able to turn off the generation of the PDF doxygen documentation and do just the HTML

white238:
> We may also be able to turn off the generation of the PDF doxygen documentation and do just the HTML

Yea.  I think i would prefer to have a working doxygen though.  If this doesn't work or proves more annoying, we can go down that route.

white238:
LGTM

PR #273, opened 2020-11-12
Add dynamic thermal conduction parameters

jamiebramwell:
This adds the ability to model variable density and heat conduction capacity in the dynamic thermal conduction physics module according to the equation below.

Tex2Img_1605209684

### DIFF
--- a/src/coefficients/stdfunction_coefficient.cpp
+++ b/src/coefficients/stdfunction_coefficient.cpp
@@ -149,7 +149,7 @@ void TransformedVectorCoefficient::Eval(mfem::Vector& V, mfem::ElementTransforma
 
 TransformedScalarCoefficient::TransformedScalarCoefficient(std::shared_ptr<mfem::Coefficient>  s1,
                                                            std::function<double(const double)> func)
-    : mfem::Coefficient(), s1_(s1), mono_function_(func), bi_function_(nullptr)
+    : mfem::Coefficient(), s1_(s1), s2_(nullptr), mono_function_(func), bi_function_(nullptr)
 {
 }
 
@@ -160,39 +160,15 @@ TransformedScalarCoefficient::TransformedScalarCoefficient(std::shared_ptr<mfem:
 {
 }
 
-TransformedScalarCoefficient::TransformedScalarCoefficient(std::unique_ptr<mfem::Coefficient>&& u1,
-                                                           std::function<double(const double)>  func)
-    : mfem::Coefficient(), u1_(std::move(u1)), mono_function_(func), bi_function_(nullptr)
-{
-}
-
-TransformedScalarCoefficient::TransformedScalarCoefficient(std::unique_ptr<mfem::Coefficient>&&              u1,
-                                                           std::unique_ptr<mfem::Coefficient>&&              u2,
-                                                           std::function<double(const double, const double)> func)
-    : mfem::Coefficient(), u1_(std::move(u1)), u2_(std::move(u2)), mono_function_(nullptr), bi_function_(func)
-{
-}
-
 double TransformedScalarCoefficient::Eval(mfem::ElementTransformation& T, const mfem::IntegrationPoint& ip)
 {
-  if (s1_ != nullptr) {
-    double temp = s1_->Eval(T, ip);
+  double temp = s1_->Eval(T, ip);
 
-    if (mono_function_) {
-      return mono_function_(temp);
-    } else {
-      double temp2 = s2_->Eval(T, ip);
-      return bi_function_(temp, temp2);
-    }
+  if (mono_function_) {
+    return mono_function_(temp);
   } else {
-    double temp = u1_->Eval(T, ip);
-
-    if (mono_function_) {
-      return mono_function_(temp);
-    } else {
-      double temp2 = u2_->Eval(T, ip);
-      return bi_function_(temp, temp2);
-    }
+    double temp2 = s2_->Eval(T, ip);
+    return bi_function_(temp, temp2);
   }
 }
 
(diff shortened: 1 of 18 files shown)


src/physics/thermal_conduction.cpp
```diff
@@ -106,7 +119,25 @@ void ThermalConduction::completeSetup()
...
+    // If both a specific heat capacity and density have been specified, use their product as the mass matrix
+    // coefficient
+    if (cp_ != nullptr && rho_ != nullptr) {
+      auto product_function = [](double a, double b) { return a * b; };
+      mass_coef_ = std::make_unique<TransformedScalarCoefficient>(std::move(rho_), std::move(cp_), product_function);
+    }
+    // otherwise, just use the one that is specified
+    else if (cp_ != nullptr) {
+      mass_coef_ = std::move(cp_);
+    } else if (rho_ != nullptr) {
```
joshessman-llnl: Minor: `std::unique_ptr`s have an implicit `operator bool()`, so this could just be
```cpp
if (cp_ && rho_)
```
etc - personally, I find it a little more readable that way.

src/coefficients/stdfunction_coefficient.hpp
```diff
@@ -290,11 +310,13 @@ class TransformedScalarCoefficient : public mfem::Coefficient {
    * @brief The first scalar coefficient in the transformation
    */
   std::shared_ptr<mfem::Coefficient> s1_;
+  std::unique_ptr<mfem::Coefficient> u1_;
```
joshessman-llnl: While I definitely appreciate the avoidance of `shared_ptr`s here, I'm wondering if it's worth leaving everything as `shared_ptr`s until we develop a better way of managing coefficient ownership in #252.  `unique_ptr`s can be "promoted" via a converting constructor, so there wouldn't need to be any changes to the `TransformedCoef` interface.

src/physics/thermal_conduction.cpp
```diff
@@ -106,7 +119,25 @@ void ThermalConduction::completeSetup()
   if (timestepper_ != serac::TimestepMethod::QuasiStatic) {
     // If dynamic, assemble the mass matrix
     M_form_ = temperature_.createOnSpace<mfem::ParBilinearForm>();
-    M_form_->AddDomainIntegrator(new mfem::MassIntegrator());
+
+    // If both a specific heat capacity and density have been specified, use their product as the mass matrix
+    // coefficient
+    if (cp_ != nullptr && rho_ != nullptr) {
+      auto product_function = [](double a, double b) { return a * b; };
+      mass_coef_ = std::make_unique<TransformedScalarCoefficient>(std::move(rho_), std::move(cp_), product_function);
```
joshessman-llnl: Out of curiosity, is this any different than `mfem::ProductCoefficient`? If not, that might be preferable if performance is a concern - `TransformedCoef` is of course more flexible but incurs additional `std::function` overhead.

jamiebramwell: Nope - I simply forgot product coefficient existed!

src/physics/thermal_conduction.cpp
```diff
@@ -106,7 +118,24 @@ void ThermalConduction::completeSetup()
...
+    // otherwise, just use the one that is specified
+    else if (cp_) {
+      mass_coef_ = std::move(cp_);
+    } else if (rho_) {
+      mass_coef_ = std::move(rho_);
+    }
+    // if neither is specified, use the default coefficient of 1.0
+    else {
+      mass_coef_ = std::make_unique<mfem::ConstantCoefficient>(1.0);
+    }
```
samuelpmish: is this equivalent to putting
```cpp
cp_ = std::make_unique<mfem::ConstantCoefficient>(1.0);
rho_ = std::make_unique<mfem::ConstantCoefficient>(1.0);
```
somewhere in the constructor?

src/physics/thermal_conduction.hpp
```diff
@@ -172,6 +186,24 @@ class ThermalConduction : public BasePhysics {
...
+   *
+   */
+  std::unique_ptr<mfem::Coefficient> cp_;
+
+  /**
+   * @brief Combined mass matrix coefficient (rho * cp)
+   *
+   */
+  std::unique_ptr<mfem::Coefficient> mass_coef_;
+
```
samuelpmish, line 206: This is a funny quantity to name, because it appears in the mass matrix calculation, but actually has no dimension of mass in it. wikipedia says this is named "volumetic heat capacity" and associates the symbol "s" with it, but I don't know if that's better or worse than the current name.

src/coefficients/stdfunction_coefficient.cpp
```diff
@@ -160,15 +160,39 @@ TransformedScalarCoefficient::TransformedScalarCoefficient(std::shared_ptr<mfem:
...
+    double temp = u1_->Eval(T, ip);
+
+    if (mono_function_) {
+      return mono_function_(temp);
+    } else {
+      double temp2 = u2_->Eval(T, ip);
+      return bi_function_(temp, temp2);
+    }
   }
 }
```
samuelpmish: The coefficient class seems like such a clunky way to shoehorn simple calculations into the integrators, I look forward to Julian's work making this sort of thing obsolete!

src/physics/thermal_conduction.cpp
jamiebramwell: ....that's a much better way to do this. I'll go change it!

src/coefficients/stdfunction_coefficient.cpp
jamiebramwell: me too!!!

src/physics/thermal_conduction.hpp
jamiebramwell: Yeah, I wasn't really sure what to call this, but I figured this could be interpreted as "mass matrix coefficient".

joshessman-llnl (approved):
Looks great!

samuelpmish (approved):

jamiebramwell:
LGTM

samuelpmishLLNL (approved):
oops, I approved from the wrong account before!

Issue #222, opened 2020-09-25
Organize the unit tests


jamiebramwell:
At some point we should revisit the unit tests.  either move them into a test directory next to their source they go with or rename and put them in subdirectories so they more match the src dir.

_Originally posted by @white238 in https://github.com/LLNL/serac/pull/221#issuecomment-699121003_

white238:
Though maybe ignore this until inlet is done and we can move some of them out of there an into the integration tests.

white238:
Outside LLNL test suites from @adrienbernede :

https://github.com/hpc/pavilion2
https://jchronoss.hpcframework.com/
https://github.com/cea-hpc/PCVS

PR #274, opened 2020-11-16
Convert remaining NonlinearSolid tests to use scripts

joshessman-llnl:
This PR converts the `serac_quasistatic_solver` and `serac_dynamic_solver` tests to use input files instead of hardcoding information into the `.cpp` files.

This required a little bit of duct tape pending some changes in https://github.com/LLNL/axom/pull/385 and https://github.com/LLNL/axom/pull/391, but once those goes in we should be able to move the coefficient definitions to Lua as well.

To cut down on duplicated code I added a library (just one function right now) to the tests directory that does some of the inlet-related setup for the tests.  This utility logic will have to be moved as part of #222.


data/input_files/tests/nonlinear_solid/serac_dynamic_solver/dyn_linesearch_solve.lua
```diff
@@ -0,0 +1,66 @@
...
+
+        nonlinear = {
+            rel_tol     = 1.0e-4,
+            abs_tol     = 1.0e-8,
+            max_iter    = 500,
+            print_level = 1,
+            solver_type = "KINLineSearch",
+        },
+
+        dynamic = {
```
jamiebramwell: I think these are the parameters for the mass matrix inversion. It might be clearer to name it `mass_solver` or something like that. The other parameters could be `stiffness_solver` for additional clarity.

src/physics/nonlinear_solid.hpp
```diff
@@ -55,8 +55,11 @@ class NonlinearSolid : public BasePhysics {
      * @brief Input file parameters specific to this class
      *
      * @param[in] table Inlet's SchemaCreator that input files will be added to
+     * @param[in] dynamic Whether the solid solver will be dynamic
      **/
-    static void defineInputFileSchema(axom::inlet::Table& table);
+    // FIXME: The dynamic parameter can be removed once the enhanced axom::inlet::Proxy::contains
```
jamiebramwell, line 61: 👍

data/input_files/tests/nonlinear_solid/serac_dynamic_solver/dyn_linesearch_solve.lua
jamiebramwell: Essentially, the user should know that these parameters are for the mass matrix solve and the other parameters are for the stiffness solve.

data/input_files/tests/nonlinear_solid/serac_dynamic_solver/dyn_direct_solve.lua
```diff
@@ -33,18 +33,18 @@ nonlinear_solid = {
...
-            linear = {
-                rel_tol     = 1.0e-4,
-                abs_tol     = 1.0e-8,
-                max_iter    = 500,
-                print_level = 0,
-                solver_type = "gmres",
-                prec_type   = "JacobiSmoother",
-            },
-            timestepper = "SDIRK33",
+    mass_solver = {
```
jamiebramwell, line 38: This is good for now, but I need to think about what an ideal input syntax for dynamic simulations would be.

jamiebramwell (approved):

src/physics/nonlinear_solid.cpp
```diff
@@ -227,8 +227,15 @@ void NonlinearSolid::InputInfo::defineInputFileSchema(axom::inlet::Table& table)
   traction_table.getField("y").defaultValue(1.0e-3);
   traction_table.getField("z").defaultValue(0.0);
 
-  auto& solver_table = table.addTable("solver", "Linear and Nonlinear Solver Parameters.");
-  serac::EquationSolver::defineInputFileSchema(solver_table);
+  auto& stiffness_solver_table =
+      table.addTable("stiffness_solver", "Linear and Nonlinear stiffness Solver Parameters.");
+  serac::EquationSolver::defineInputFileSchema(stiffness_solver_table);
+
+  if (dynamic) {
```
white238, line 351: Do we want to make this required based on `dynamic` instead of not defining the schema at all? If we toggle this way, it won't get written out to the documentation or if in the future we recreate a fully blown out input file.

joshessman-llnl: The problem is that Inlet's implementation of `contains` on a Table only checks if the child table or field exists, not whether it contains a value.  This is fixed in one of my Inlet PRs but for right now this appears to be the only way to check if an optional parameter was defined.

joshessman-llnl:
I've added an additional comment on the conditionally enabled schema definition - this will be removed when https://github.com/LLNL/axom/pull/385 is merged.

white238 (approved):

joshessman-llnl:
LGTM

PR #287, opened 2020-11-24
Incorporate initial code coverage support

### LINKED ISSUE #279: Add code coverage tools
We should add code coverage tools (probably [gcov](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html)) to make sure we are testing appropriately.

joshessman-llnl:
This PR adds support for "interactive" code coverage analysis (recently updated in BLT) via LCOV with the `serac_coverage` target and for coverage report uploading to [codecov](https://codecov.io/gh/LLNL/serac/) (which is what Umpire uses).  It also adds a few sample error handling unit tests - these will get moved as part of #222.  These unit tests revealed an opportunity for inconsistency with coefficients that is partially addressed by #237 but is fixed here for now.

Instead of death tests (which don't play nice with the threading used by some of our TPLs, even in threadsafe mode), SLIC was configured to raise an exception on error which can be caught by GTest.

Resolves #279


src/physics/thermal_conduction.cpp
```diff
@@ -51,7 +51,7 @@ void ThermalConduction::setTemperature(mfem::Coefficient& temp)
 void ThermalConduction::setTemperatureBCs(const std::set<int>&               temp_bdr,
                                           std::shared_ptr<mfem::Coefficient> temp_bdr_coef)
 {
-  bcs_.addEssential(temp_bdr, temp_bdr_coef, temperature_);
+  bcs_.addEssential(temp_bdr, temp_bdr_coef, temperature_, 0);
```
jamiebramwell: I'm not sure we want to require this. It makes sense that a scalar coefficient on a scalar finite element state would not need a component index.

tests/serac_error_handling.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "numerics/mesh_utils.hpp"
+#include "physics/utilities/boundary_condition.hpp"
+#include "physics/utilities/equation_solver.hpp"
+
+class SlicErrorException : public std::exception {
+};
+
+namespace serac {
+
+TEST(serac_error_handling, equationsolver_bad_lin_solver)
```
jamiebramwell, line 20: Nice tests!

jamiebramwell (commented):
Looks good! My only concern is requiring that scalar coefficients on scalar FE states require a component.

src/physics/thermal_conduction.cpp
joshessman-llnl: I think that's reasonable, it looks like the condition that should be changed then is here:
https://github.com/LLNL/serac/blob/9fc81ff51ce6a4b85a82e88e593f3bdc88eba771/src/physics/utilities/boundary_condition.cpp#L51-L63

I think the correct modification is to always assume scalar coef when the FEState's dimension is zero - does that sound right to you?

jamiebramwell: Sounds good to me!

src/physics/utilities/boundary_condition.cpp
```diff
@@ -59,7 +66,8 @@ void BoundaryCondition::project(FiniteElementState& state) const
       auto scalar_coef = std::get_if<std::shared_ptr<mfem::Coefficient>>(&coef_);
       SLIC_ASSERT_MSG(scalar_coef,
                       "Essential boundary condition contained a single component but had a non-scalar coefficient.");
-      state.gridFunc().ProjectCoefficient(**scalar_coef, dof_list, component_);
+      const int component = (vdim == 1) ? 0 : component_;
```
jamiebramwell: 👍

jamiebramwell (approved):

white238 (approved):

joshessman-llnl:
LGTM

PR #290, opened 2020-11-26
Guard against optional, but required by us, axom headers

white238:
It is confusing when these fail to be present.


white238:
LGTM

jamiebramwell (approved):

joshessman-llnl (approved):

PR #292, opened 2020-12-02
discussion: is it worth it to use a helper class to construct/store different types of mfem::Coefficient?

samuelpmishLLNL:
See https://github.com/LLNL/serac/issues/252

Pros:
- usage is more succinct than previous implementation
- convenience constructors select appropriate mfem::Coefficient type for arguments passed in (e.g. `double -> mfem::ConstantCoefficient`)
- clearer ownership (owned by the physics module)

Cons:
- move-only semantics for serac::CoefficientWrapper, to work around memory ownership inconsistencies in mfem::Coefficient
- `std::shared_ptr<mfem::Coefficient>` still appears in `SetTemperature`, `SetTemperatureBCs`, etc. 
  - Do we want these to be shared? 
  - Having some of the interface be smart pointers and other parts use these convenience constructors makes it less consistent.


joshessman-llnl:
I implemented a `GeneralCoefficientWrapper` class in #237 that was intended to abstract away the Coefficient/VectorCoefficient duality a bit more than just the raw `std::variant`:
https://github.com/LLNL/serac/blob/24af67c30d20bab511f9574b2cbbe489bd6bead2/src/physics/utilities/boundary_condition.hpp#L46-L103

I wonder if it would be useful for this `CoefficientWrapper` class to have similar functionality (even if by composition) so the user could initialize it with arguments that correspond to either scalar or vector coefs.  That would keep the variant out of the user interface and possibly simplify some of the logic in the physics modules and boundary condition utilities.

jamiebramwell:
Closing this as it will become obsolete with the upcoming variational form rewrite.

PR #296, opened 2020-12-08
Add fake mpi target to using-with-cmake example

white238:
Slightly cleaner so we don't have to keep adding targets to this list.


jamiebramwell (approved):

white238:
LGTM

joshessman-llnl (approved):
For context: this is waiting on https://github.com/LLNL/blt/pull/436

PR #297, opened 2020-12-08
Move ODE solver into ODE class

### LINKED ISSUE #291: Wrap the MFEM second order ODE solver
After this PR is merged, these should go away as the MFEM second order ODE solver will be wrapped (similar to `EquationSolver`).

_Originally posted by @jamiebramwell in https://github.com/LLNL/serac/pull/216#r532969519_

joshessman-llnl:
After some discussion with @samuelpmishLLNL it was determined that the functionality of `mfem::TimeDependentOperator` and `mfem::ODESolver` (and their second-order counterparts) can be merged into one class for simplicity.

Specifically, there does not appear to be a need to have a serac-wrapped object that conforms to the `mfem::ODESolver` interface (as opposed to `EquationSolver` which needed to implement the `mfem::Solver` interface.

This PR adds a `Step` method to our ODE classes and moves the TimestepMethod -> ODESolver logic into the ODE classes.  `setTimestepper` was removed from the `BaseSolver` and replaced with a `is_quasistatic_` flag.

Resolves #291

### DIFF
--- a/src/drivers/serac.cpp
+++ b/src/drivers/serac.cpp
@@ -130,9 +130,6 @@ int main(int argc, char* argv[])
     }
   }
 
-  // Set the time step method
-  solid_solver.setTimestepper(serac::TimestepMethod::QuasiStatic);
-
   // Complete the solver setup
   solid_solver.completeSetup();
 
--- a/src/physics/base_physics.cpp
+++ b/src/physics/base_physics.cpp
@@ -19,8 +19,7 @@ BasePhysics::BasePhysics(std::shared_ptr<mfem::ParMesh> mesh)
     : comm_(mesh->GetComm()), mesh_(mesh), output_type_(serac::OutputType::VisIt), time_(0.0), cycle_(0), bcs_(*mesh)
 {
   std::tie(mpi_size_, mpi_rank_) = getMPIInfo(comm_);
-  BasePhysics::setTimestepper(serac::TimestepMethod::ForwardEuler);
-  order_ = 1;
+  order_                         = 1;
 }
 
 BasePhysics::BasePhysics(std::shared_ptr<mfem::ParMesh> mesh, int n, int p) : BasePhysics(mesh)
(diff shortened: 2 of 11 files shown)


samuelpmish (approved):
Looks great, Josh! 

I wonder what we should do regarding class method naming style when inheriting from mfem::Something but add our own methods. Since our style guides are different, the serac parts start with lowercase, but the mfem parts start with Uppercase, so the interface is appears to have an inconsistent style. 

I'm not sure what the right decision is here. Maybe things that are derived from an mfem class should adopt their naming convention entirely?

samuelpmishLLNL (approved):
oops, approvals from my personal account don't carry any weight.

joshessman-llnl:
> I wonder what we should do regarding class method naming style when inheriting from mfem::Something but add our own methods. Since our style guides are different, the serac parts start with lowercase, but the mfem parts start with Uppercase, so the interface is appears to have an inconsistent style.
> 
> I'm not sure what the right decision is here. Maybe things that are derived from an mfem class should adopt their naming convention entirely?

I had thought about this as well.  I think you're right though - if we can't be consistent with the rest of serac (b/c MFEM uses a different naming convention), we can at least be consistent within a class.

jamiebramwell (approved):
This cleans up the interface nicely!

joshessman-llnl:
LGTM

jamiebramwell:
> 
> 
> > I wonder what we should do regarding class method naming style when inheriting from mfem::Something but add our own methods. Since our style guides are different, the serac parts start with lowercase, but the mfem parts start with Uppercase, so the interface is appears to have an inconsistent style.
> > I'm not sure what the right decision is here. Maybe things that are derived from an mfem class should adopt their naming convention entirely?
> 
> I had thought about this as well. I think you're right though - if we can't be consistent with the rest of serac (b/c MFEM uses a different naming convention), we can at least be consistent within a class.

I think following the MFEM convention for derived classes makes sense, but I don't have a strong opinion either way. Any thoughts @white238 ?

white238:
> > > I wonder what we should do regarding class method naming style when inheriting from mfem::Something but add our own methods. Since our style guides are different, the serac parts start with lowercase, but the mfem parts start with Uppercase, so the interface is appears to have an inconsistent style.
> > > I'm not sure what the right decision is here. Maybe things that are derived from an mfem class should adopt their naming convention entirely?
> > 
> > 
> > I had thought about this as well. I think you're right though - if we can't be consistent with the rest of serac (b/c MFEM uses a different naming convention), we can at least be consistent within a class.
> 
> I think following the MFEM convention for derived classes makes sense, but I don't have a strong opinion either way. Any thoughts @white238 ?

Just submit a patch to MFEM to go with our naming scheme =)

First, I would err on being consistent per class. Mixing a single classes convention would be confusing.  Since this is just an extension of an MFEM class, I would go with theirs for the whole class.

We should probably denote these are MFEM extension classes in some way?  `serac::mfem::SecondOrderODE` or `serac::MFEMSecondOrderODE`? I think i prefer the former it is more C++...

jamiebramwell:
I like the idea of appending the `mfem` namespace to any mfem extensions. It would also make it easier to eventually upstream if it makes sense.

PR #301, opened 2020-12-08
Move MFEM extensions into mfem_extensions namespace

### LINKED ISSUE #298: Add mfem namespace to mfem extension objects
I like the idea of appending the `mfem` namespace to any mfem extensions. It would also make it easier to eventually upstream if it makes sense.

_Originally posted by @jamiebramwell in https://github.com/LLNL/serac/issues/297#issuecomment-740827113_

joshessman-llnl:
All classes within serac implementing an MFEM interface have been moved into the `serac::mfem_extensions` namespace.

Resolves #298


src/coefficients/coefficient_extensions.hpp
```diff
@@ -243,6 +243,6 @@ class TransformedScalarCoefficient : public mfem::Coefficient {
   std::function<double(const double, const double)> bi_function_;
```
jamiebramwell: This isn't really an MFEM extension, I.e. it is not derived from an MFEM class. Do we still want to label it as an MFEM extension?

src/physics/nonlinear_solid.hpp
```diff
@@ -278,12 +278,12 @@ class NonlinearSolid : public BasePhysics {
   /**
    * @brief Nonlinear system solver instance
    */
-  EquationSolver nonlin_solver_;
+  mfem_extensions::EquationSolver nonlin_solver_;
```
jamiebramwell: Same question. Do we want to mark things which could be upstreamed to mfem or only derived objects?

src/coefficients/coefficient_extensions.hpp
jamiebramwell: Thoughts @white238 ?

src/physics/nonlinear_solid.hpp
joshessman-llnl: Hmm, I'm not sure what would qualify as potentially upstreamable to MFEM.  My thinking was that part of what the `mfem_extensions` namespace would indicate is that things like function names follow the MFEM instead of serac convention.

src/coefficients/coefficient_extensions.hpp
```diff
@@ -19,7 +19,7 @@
 #include "mfem.hpp"
 #include "numerics/expr_template_ops.hpp"
 
-namespace serac {
+namespace serac::mfem_extensions {
 
 /**
  * @brief MakeTrueEssList takes in a FESpace, a vector coefficient, and produces a list
```
white238: Should we only do this for the class extensions and code we hope to get back into MFEM?  Do these functions fall under that? they seem like helper functions...

Clearly not from this PR but the documentation case doesn't match the actual function name case (`MakeTrueEssList vs `makeTrueEssList`)

jamiebramwell: Let's talk about this at the meeting tomorrow.

jamiebramwell:
Per our meeting discussion, anything that could feasibly get upstreamed to mfem (i.e. not physics related and closely coupled with mfem types) should go under the `mfem_extensions` namespace and use the mfem casing style (`Foo foo.Bar()`).

src/coefficients/coefficient_extensions.hpp
white238: Just to log the output of our meeting on this topic the `serac::mfem_extensions` namespace will cover two cases:

1) Extended MFEM classes
2) Functionality we hope will go into MFEM

The idea is to keep their case style so it is less of a rework when they get moved into MFEM and also a class with mixed casing is very awkward.

white238:
@samuelpmishLLNL mentioned the `serac::mfem_extensions` is a rather long name.  Does anyone have another suggestion? I usually fall under the verbose category but I can be persuaded.

`serac::mfem` - I would say not this because it could end up being just `mfem` in our code and that leads to confusion
`serac::mfem_ext` - not opposed, though I prefer the longer for no real reason

jamiebramwell:
> 
> 
> @samuelpmishLLNL mentioned the `serac::mfem_extensions` is a rather long name. Does anyone have another suggestion? I usually fall under the verbose category but I can be persuaded.
> 
> `serac::mfem` - I would say not this because it could end up being just `mfem` in our code and that leads to confusion
> `serac::mfem_ext` - not opposed, though I prefer the longer for no real reason

I think I have a slight preference for `serac::mfem_ext` because I don't like long namespaces, but I could go either way.

src/serac/physics/utilities/equation_solver.hpp
```diff
@@ -21,7 +21,7 @@
 #include "serac/infrastructure/input.hpp"
 #include "serac/physics/utilities/solver_config.hpp"
 
-namespace serac {
+namespace serac::mfem_ext {
```
jamiebramwell, line 24: I'm not sure `EquationSolver` should be in this namespace. I can't imagine this getting upststreamed to MFEM as it is a convenient wrapper of MFEM stuff for us.

tests/serac_component_bc.cpp
```diff
@@ -65,7 +65,7 @@ TEST(component_bc, qs_solve)
       }
   });
 
-  mfem::Array<int> ess_corner_bc_list = makeTrueEssList(solid_solver.displacement().space(), zero_bc);
+  mfem::Array<int> ess_corner_bc_list = mfem_ext::makeTrueEssList(solid_solver.displacement().space(), zero_bc);
```
jamiebramwell: `makeTrueEssList` should be capitalized.

tests/serac_wrapper_tests.cpp
```diff
@@ -160,8 +160,8 @@ TEST_F(WrapperTests, nonlinear_linear_thermal)
     return 0.;
   });
 
-  Array<int> bdr_attr_list_zero = serac::makeBdrAttributeList(*pmesh_, x_zero);
-  Array<int> bdr_attr_list_one  = serac::makeBdrAttributeList(*pmesh_, x_one);
+  Array<int> bdr_attr_list_zero = mfem_ext::makeBdrAttributeList(*pmesh_, x_zero);
```
jamiebramwell: `makeBdrAttributeList` should be capitalized.

jamiebramwell (commented):
Looks good! My main comment is I'm not sure `EquationSolver` should be in this new namespace. Any thoughts?

jamiebramwell:
We should also add a blurb in the style guide about these new rules.

src/serac/physics/utilities/equation_solver.hpp
joshessman-llnl: My thinking was that because it extends an MFEM class it uses MFEM naming conventions and that we should keep it in the same namespace for consistency, i.e., MFEM naming **iff** in `mfem_ext`.

jamiebramwell: It is a convenience wrapper around MFEM types that I doubt will get upstreamed, so I'm on the side of it shouldn't be in `mfem_ext`. However, I don't feel too strongly about this and will defer to the group consensus. Any thoughts @white238 @samuelpmishLLNL @jonwong12 ?

white238 (approved):

jamiebramwell:
... I forgot `EqationSolver` inherits from `mfem::Solver`...

jamiebramwell (approved):

joshessman-llnl:
LGTM

Issue #289, opened 2020-11-25
AMGX and `serac::buildBallMesh` on a second order H1 space do not place nice.


jonwong12:
There seems to be a strange behavior where AMGX does not play nice with a mesh generated from `serac::buildBallMesh(10000)` when an order (2+) H1 space is used.

I've dumped of the mesh generated by `serac::buildBallMesh(10000)` and also one where I ran `pmesh->ReorientTetMesh()` and they are in `<root>/data/meshes/` in the branch [`feature/jonwong12/amgx_playground`](SHA: b4e165ab0682e0246dc1eb9a23968076b1250f2a).

I added a modified `mfem_ex1p` example that can be run to see the observed behavior. Some of the more interesting defaults are: `-no-amgx`, `order=2`, and , `-m data/meshes/star.mesh`.
```
./tests/mfem_ex1p -o <order> [-amgx | -no-amgx] [-m ../data/meshes/ballmesh.mesh]
```
Observations: 
* `order=1` should work on `star.mesh`, `ballmesh.mesh` and `ballmesh.reorder.mesh` with both `-amgx` and without (`-no-amgx`).
* `order=2` should work on `star.mesh`, `ballmesh.mesh` and `ballmesh.reorder.mesh` without (`-no-amgx`), but only `star.mesh` with `-amgx`.

jonwong12:
It seems the default configuration isn't guaranteed to work. So it may take some experimentation to use a different amgx smoother ([mfem conversation and resolution](https://github.com/mfem/mfem/issues/1900#issuecomment-740200100))

The following json file did work with order 2 and `ballmesh.mesh`.
```
{                                                                                                                                                                    
    "config_version": 2,                                                                                                                                             
    "solver": {                                                                                                                                                      
        "solver": "AMG",                                                                                                                                             
        "smoother": "JACOBI_L1",                                                                                                                                     
        "presweeps": 1,                                                                                                                                              
        "interpolator": "D2",                                                                                                                                        
        "max_iters": 2,                                                                                                                                             
        "scope": "main",                                                                                                                                             
        "max_row_sum": 0.9,                                                                                                                                          
        "strength_threshold": 0.25,                                                                                                                                  
        "max_levels": 1000,                                                                                                                                          
        "postsweeps": 1,                                                                                                                                             
        "tolerance": 0.0,                                                                                                                                            
        "norm": "L1",                                                                                                                                                
        "cycle": "V"                                                                                                                                                 
    }                                                                                                                                                                
}    
```

PR #302, opened 2020-12-09
Begin adding configuration options to AMGX

### LINKED ISSUE #300: Add configuration options for AMGX
It seems the default configuration isn't guaranteed to work. So it may take some experimentation to use a different amgx smoother ([mfem conversation and resolution](https://github.com/mfem/mfem/issues/1900#issuecomment-740200100))

The following json file did work with order 2 and `ballmesh.mesh`.
```
{                                                                                                                                                                    
    "config_version": 2,                                                                                                                                             
    "solver": {                                                                                                                                                      
        "solver": "AMG",                                                                                                                                             
        "smoother": "JACOBI_L1",                                                                                                                                     
        "presweeps": 1,                                                                                                                                              
        "interpolator": "D2",                                                                                                                                        
        "max_iters": 2,                                                                                                                                             
        "scope": "main",                                                                                                                                             
        "max_row_sum": 0.9,                                                                                                                                          
        "strength_threshold": 0.25,                                                                                                                                  
        "max_levels": 1000,                                                                                                                                          
        "postsweeps": 1,                                                                                                                                             
        "tolerance": 0.0,                                                                                                                                            
        "norm": "L1",                                                                                                                                                
        "cycle": "V"                                                                                                                                                 
    }                                                                                                                                                                
}    
```

_Originally posted by @jonwong12 in https://github.com/LLNL/serac/issues/289#issuecomment-741092435_

joshessman-llnl:
This PR adds configuration options to what was just a placeholder `AMGXPrec` structure.  The default options were insufficient for an AMGX version of the static thermal solve test (see #289), which is added in this PR.

AMGX uses a JSON string for configuration but I did not think it was necessary to pull in a full JSON library - I was able to implement the required functionality in < 35 LLOC.

Thanks to @jonwong12 and @artv3 for tracking down the required configuration changes - AMGX has a [fair number of options](https://github.com/NVIDIA/AMGX/blob/919fb92a2c02ea3d8d3d0f0660e3b507d84705bc/core/src/core.cu#L331-L560), so we can look into exposing more of them in future PRs, or possibly exposing my minimal `JSONTable` interface which is a little friendlier than having the user pass a full JSON string.  I'm not sure if any of these changes might be worth propagating upstream to MFEM.

Resolves #300

### DIFF
--- a/src/physics/utilities/equation_solver.cpp
+++ b/src/physics/utilities/equation_solver.cpp
@@ -114,13 +114,12 @@ std::ostream& operator<<(std::ostream& out, const JSONTable& table)
   char        sep = ' ';                      // Start with empty separator to avoid trailing comma
   for (const auto& [key, val] : table.literals_) {
     out << sep << "\n" << indent << "\"" << key << "\": ";
-    // Strings need to be quoted with escaped strings
+    // Strings need to be wrapped in escaped quotes
     if (auto str_ptr = std::get_if<std::string>(&val)) {
       out << "\"" << *str_ptr << "\"";
     } else {
       std::visit([&out](const auto contained_val) { out << contained_val; }, val);
     }
-
     sep = ',';
   }
 
--- a/src/physics/utilities/equation_solver.cpp
+++ b/src/physics/utilities/equation_solver.cpp
@@ -192,7 +192,6 @@ std::unique_ptr<mfem::AmgXSolver> configureAMGX(const MPI_Comm comm, const AMGXP
 
   std::ostringstream oss;
   oss << options_table;
-  SLIC_INFO(options_table);
   // Treat the string as the config (not a filename)
   amgx->ReadParameters(oss.str(), mfem::AmgXSolver::INTERNAL);
   amgx->InitExclusiveGPU(comm);
--- a/src/physics/utilities/equation_solver.cpp
+++ b/src/physics/utilities/equation_solver.cpp
@@ -40,55 +40,138 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_
 
 namespace detail {
 #ifdef MFEM_USE_AMGX
-std::unique_ptr<mfem::AmgXSolver> configureAMGX(const MPI_Comm comm, const AMGXPrec& options)
-{
-  auto amgx                              = std::make_unique<mfem::AmgXSolver>();
-  using JSONType                         = std::variant<std::string, double, int>;
-  static const auto default_prec_options = []() {
-    std::unordered_map<std::string, JSONType> result;
-    result["solver"]       = "AMG";
-    result["presweeps"]    = 1;
-    result["postsweeps"]   = 2;
-    result["interpolator"] = "D2";
-    result["max_iters"]    = 2;
-    result["convergence"]  = "ABSOLUTE";
-    result["cycle"]        = "V";
-    return result;
-  }();
+using JSONLiteral  = std::variant<std::string, double, int>;
+using JSONLiterals = std::unordered_map<std::string, JSONLiteral>;
 
-  auto options_table = default_prec_options;
-  if (options.verbose) {
-    static const auto default_verbose_options = []() {
-      std::unordered_map<std::string, JSONType> result;
-      result["obtain_timings"]    = 1;
-      result["monitor_residual"]  = 1;
-      result["print_solve_stats"] = 1;
-      result["print_solve_stats"] = 1;
-      return result;
-    }();
-    options_table.insert(default_verbose_options.begin(), default_verbose_options.end());
+/**
+ * @brief A structure for storing simple JSON data
+ *
+ * Used to produce the configuration for AMGX
+ */
+class JSONTable {
+public:
+  /**
+   * @brief Creates a new table
+   * @param[in] depth The depth of the created table
+   */
+  JSONTable(const int depth = 0) : depth_(depth) {}
+  /**
+   * @brief Returns the subtable at the given index, creating a new table if necessary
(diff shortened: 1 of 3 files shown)


tests/serac_thermal_solver.cpp
```diff
@@ -345,6 +345,47 @@ TEST(thermal_solver_rework, dyn_imp_solve_time_varying)
...
 
+#ifdef MFEM_USE_AMGX
+TEST(thermal_solver, static_amgx_solve)
+{
+  MPI_Barrier(MPI_COMM_WORLD);
+
+  auto pmesh = buildBallMesh(10000);
+
+  auto params                                                   = ThermalConduction::defaultQuasistaticParameters();
+  std::get<IterativeSolverParameters>(params.T_lin_params).prec = AMGXPrec{.smoother = AMGXSolver::JACOBI_L1};
```
joshessman-llnl, line 356: Just setting the smoothing algorithm here was enough to have it converge in 1 iteration, vs the default block Jacobi smoother hitting the 500 iter cap and still being off by ~2 orders of magnitude over the tolerance

src/physics/utilities/equation_solver.cpp
```diff
@@ -38,6 +38,169 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_
   }
 }
 
+namespace detail {
```
jamiebramwell, line 41: I think this is fine for now, but it might be worthwhile to pull in a real JSON parser sooner rather than later. I can imagine needing one for other applications. I think there is already one in either Conduit or Axom. Thoughts @white238 ?

jamiebramwell: One is actually already included in mfem: https://github.com/mfem/mfem/blob/master/fem/picojson.h

tests/serac_thermal_solver.cpp
jamiebramwell: Out of curiosity, what was the default smoother?

src/physics/utilities/solver_config.hpp
```diff
@@ -121,10 +121,43 @@ struct HypreBoomerAMGPrec {
...
+  KPZ_POLYNOMIAL,
+  BLOCK_JACOBI,
+  MULTICOLOR_GS,
+  MULTICOLOR_DILU
+};
+
 /**
  * @brief Stores the information required to configure a NVIDIA AMGX preconditioner
  */
 struct AMGXPrec {
```
jamiebramwell, line 148: This goes beyond this PR, but we should append these option-type structs with `Option` so it's clearer what these things are doing.

artv3:
Nice work @joshessman-llnl ! I think the JSON table interface would be awesome to have at the MFEM level.

src/physics/utilities/equation_solver.cpp
```diff
@@ -38,6 +38,169 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_
...
+      result.literal("monitor_residual")  = 1;
+      result.literal("print_solve_stats") = 1;
+      result.literal("print_solve_stats") = 1;
+      return result;
+    }();
+    options_table["solver"].add(default_verbose_options);
+  }
+
+  // FIXME: magic_enum?
+  static const auto solver_names = []() {
```
jamiebramwell, line 170: I agree this is a bit gross. magic_enum seems like a good way out of it.

jamiebramwell (approved):
Thanks for figuring this out!

src/physics/utilities/equation_solver.cpp
joshessman-llnl: I'm not sure if picojson is flexible enough for us, it seems mostly focused on parsing/accessing and not the ability to add/manipulate objects.

Conduit has a JSON library that we may not want to use right now, but Axom (I think) will have a different one soon for Inlet.

src/physics/utilities/equation_solver.cpp
joshessman-llnl: Unfortunately I don't think this is an option for us because it requires GCC 9+.  There's quite a few instances where we'd benefit from this (mostly other enums in `solver_config.hpp`) so if/when we're able to upgrade our GCC we can look into this.

joshessman-llnl:
> Out of curiosity, what was the default smoother?

The default is a block Jacobi method, I think.  I've left that as the default in `solver_config.hpp`.

tests/serac_thermal_solver.cpp
joshessman-llnl: The default is a block Jacobi method, I think. I've left that as the default in `solver_config.hpp`.

src/physics/utilities/equation_solver.cpp
```diff
@@ -38,6 +38,169 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverParameters& lin_
...
+  }
+  out << "\n" << indent << "}";
+  return out;
+}
+
+std::unique_ptr<mfem::AmgXSolver> configureAMGX(const MPI_Comm comm, const AMGXPrec& options)
+{
+  auto              amgx                 = std::make_unique<mfem::AmgXSolver>();
+  static const auto default_prec_options = []() {
+    JSONTable result;
```
white238: `result` is an odd name for the thing you aren't returning.

white238 (approved):

src/physics/utilities/equation_solver.cpp
joshessman-llnl: Good catch, I had initially written this a little differently

joshessman-llnl:
LGTM

PR #303, opened 2020-12-09
Remove unnecessary exitGracefullys, add tests for error handling and coefficients

joshessman-llnl:
This PR cleans up the `exitGracefully`s that follow `SLIC_ERROR`s, which now call `exitGracefully` automatically.

It also adds a set of tests that check for proper error handling, and for coefficient extensions that are not tested elsewhere.

### DIFF
--- a/src/drivers/serac.cpp
+++ b/src/drivers/serac.cpp
@@ -51,7 +51,6 @@ void defineInputFileSchema(axom::inlet::Inlet& inlet, int rank)
   // Verify input file
   if (!inlet.verify()) {
     SLIC_ERROR_ROOT(rank, "Input file failed to verify.");
-    serac::exitGracefully(true);
   }
 }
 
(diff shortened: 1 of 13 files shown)


src/drivers/serac.cpp
```diff
@@ -51,7 +51,6 @@ void defineInputFileSchema(axom::inlet::Inlet& inlet, int rank)
   // Verify input file
   if (!inlet.verify()) {
     SLIC_ERROR_ROOT(rank, "Input file failed to verify.");
```
white238, line 53: You confirmed this calls exit on all ranks?

white238 (approved):

jamiebramwell (approved):

src/drivers/serac.cpp
joshessman-llnl: Yep, when the calling (root) process terminates [all the others will as well](https://www.open-mpi.org/doc/v4.0/man1/mpirun.1.php#sect16).

joshessman-llnl:
LGTM

PR #304, opened 2020-12-09
Actually use MFEM's CMake config when it exists

### LINKED ISSUE #293: MFEM import logic never uses the CMake logic
The logic to load the exported CMake config file from MFEM is never used.  The following line is not the location of the `MFEMConfig.cmake` file:

https://github.com/LLNL/serac/blob/bac04a96dd534a4518f78bd06020aa226a873d1c/cmake/thirdparty/FindMFEM.cmake#L22

It is actually here: `<MFEM install prefix>/lib/cmake/mfem/MFEMConfig.cmake`

This could be further guarded against possible locations like the following:

https://github.com/LLNL/axom/blob/d390eec1e2221fd2b63ef5d4fab92bf36d448dbe/src/cmake/thirdparty/FindMFEM.cmake#L26-L32

### LINKED ISSUE #429: use blt_import_library
Just noticed that mfem is still a registered library.  This should be converted to blt_import_library

joshessman-llnl:
Resolves #293
Resolves #429 

MFEM uses the uppercase convention for its CMake config (but still the lowercase target name) - the `NAMES` keyword was insufficient here.


cmake/thirdparty/FindMFEM.cmake
```diff
@@ -19,18 +19,15 @@ endif()
 message(STATUS "Looking for MFEM using MFEM_DIR = ${MFEM_DIR}")
 serac_assert_is_directory(VARIABLE_NAME MFEM_DIR)
 
-set(_mfem_cmake_config "${MFEM_DIR}/MFEMConfig.cmake")
+set(_MFEM_DIR ${MFEM_DIR}) # Save MFEM_DIR as a non-cache variable
+find_package(MFEM CONFIG NO_DEFAULT_PATH PATHS "${MFEM_DIR}/lib/cmake/mfem")
+# find_package will overwrite MFEM_DIR, so restore it here
```
white238, line 24: why does their find_package overwrite MFEM_DIR?  What does it get overwritten to?

white238 (approved):

cmake/thirdparty/FindMFEM.cmake
joshessman-llnl: I think it's the default behavior of `find_package` - if the config is found it'll be overwritten with the directory containing the `MFEMConfig.cmake`, otherwise it'll be set to `NOTFOUND`.

white238:
We should bite the bullet and merge this after the branch gets updated.

jamiebramwell (approved):

joshessman-llnl:
LGTM

white238 (approved):

Issue #280, opened 2020-11-18
Prefix CMake targets with `serac_`


white238:
This is needed to avoid common target name collision when Serac is used via CMake's `add_subdirectory` command.

PR #308, opened 2020-12-10
Prefix CMake targets with serac_

white238:
#280

Also limits duplication of targets that are inherited through other CMake targets
Also stops clang-format from sorting includes (didn't fix order in this PR)


white238:
LGTM

joshessman-llnl (approved):

jamiebramwell (approved):

PR #310, opened 2020-12-15
Move to uberenv as a submodule

white238:
* Updates us to the newest uberenv* with our specific fixes
* Move project specific files out of `scripts/uberenv` and into `scripts/spack`
* Updates various paths in build scripts to new paths

*Now with the power of submodules!*

PR #312, opened 2020-12-16
Enable sundials by default

jamiebramwell:
This turns on the sundials dependency by default. However, I had to disable the no sundials variant to get it to work due to concretization errors.


white238:
I'll handle rebuilding in my #310 PR

white238:
#327

PR #315, opened 2020-12-18
Fix the WSL/Ubuntu/GCC-9 TPL builds

jamiebramwell:
For some reason when updating spack, the `-ldl -lz` link flags got moved to the front of the link line for netcdf. I don't like this fix, so I am open to any other suggestions.

### DIFF
--- a/scripts/spack/packages/netcdf-c/package.py
+++ b/scripts/spack/packages/netcdf-c/package.py
@@ -184,6 +184,11 @@ class NetcdfC(AutotoolsPackage):
         if hdf5_hl.satisfies('~shared'):
             libs.append(hdf5_hl.libs.link_flags)
 
+            # SERAC EDIT BEGIN
+            # WSL gcc 9 needs ldl and lz at the end of the link line
+            libs.append('-ldl -lz')
+            # SERAC EDIT END
+
         if '+parallel-netcdf' in self.spec:
             config_args.append('--enable-pnetcdf')
             pnetcdf = self.spec['parallel-netcdf']


scripts/spack/packages/netcdf-c/package.py
```diff
@@ -184,6 +184,11 @@ def configure_args(self):
         if hdf5_hl.satisfies('~shared'):
             libs.append(hdf5_hl.libs.link_flags)
 
+            # SERAC EDIT BEGIN
+            # WSL gcc 9 needs ldl and lz at the end of the link line
+            libs.append('-ldl -lz')
```
white238, line 189: I can't even find where these were coming from hdf5 before.. @becker33 do you have an idea why this changed in the newest Spack?  This worked prior to updating to a newer develop.

white238: So I don't forget this before I get to it.  Per @becker33, we should fix hdf5 to properly add `-ldl` to its libs when `~shared`.   I assume `-lz` is missing as well and should be added.

We think this was accidentally working before and the ordering changed and this issue was exposed. 

There is a fix I put into mfem for this as well that should be removed if this is fixed properly.

joshessman-llnl: I have also run into issues where liblua is missing `-ldl` on WSL.  I imagine this is probably related.

jamiebramwell:
Should we merge this as a stopgap fix until https://github.com/LLNL/serac/issues/324 gets resolved?

white238 (approved):

white238:
> Should we merge this as a stopgap fix until #324 gets resolved?

Yea I'll try to get to it this week.

jamiebramwell:
@joshessman-llnl could you take a look at this?

joshessman-llnl (approved):

jamiebramwell:
LGTM

PR #317, opened 2020-12-21
Initial Sidre DataStore integration

### LINKED ISSUE #29: Move to using a SidreDataCollection for state variable storage

joshessman-llnl:
We won't be able to take advantage of the "restore" functionality until https://github.com/LLNL/axom/pull/445 is merged in, but I thought it might be worthwhile to discuss the proposed integration.

The primary change is that instead of calling the `FiniteElementState` constructor directly, a `StateManager` intermediary is inserted to populate the state object with existing data from the datastore (in the case of a restart).

A couple possible discussion points:
- Should any part of the input file be preserved in the event of a restart? Currently the previous input data is wiped out.
- Currently (in a restart) the mesh is built from the mesh file and then ignored in favor of the mesh in the datastore.  Should the user be able to query if a mesh already exists and then pass a `nullptr` to the constructor?
- Is specifying the cycle/step to restart from on the command line an ideal user interface?

Resolves #29


jamiebramwell:
> A couple possible discussion points:
> 
> Should any part of the input file be preserved in the event of a restart? Currently the previous input data is wiped out.
>

Ideally, the entire input file is contained in the restart (in my production code, well call this "restart reproducibility"). Essentially this means the only thing needed to continue a run is the restart file. For many reasons, this often isn't the case but it is worth striving toward.

> Currently (in a restart) the mesh is built from the mesh file and then ignored in favor of the mesh in the datastore.  Should the user be able to query if a mesh already exists and then pass a `nullptr` to the constructor?
> 

I think so. I don't want to force a user to make a dummy mesh if one already exists.

>     * Is specifying the cycle/step to restart from on the command line an ideal user interface?

We typically restart from the last timestep available in the restart file, and at least in the past each restart has contained exactly one time step worth of data. Does the sidre structure allow for more flexibility?

tests/serac_restart.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  // Simulate a restart
+  {
+    MPI_Barrier(MPI_COMM_WORLD);
+    // Create DataStore
+    axom::sidre::DataStore datastore;
+    // Load in from the saved cycle
+    serac::StateManager::initialize(datastore, restart_cycle);
+
+    // FIXME: We don't need to build the mesh here if it's a restart
```
jamiebramwell: Is it possible to extract the mesh pointer from the datastore?

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -172,17 +190,85 @@ class FiniteElementState {
...
+  /**
+   * @brief Resets the underlying global datacollection object
+   */
+  static void reset() { datacoll_.reset(); };
+
+private:
+  /**
+   * @brief The datacollection instance
+   */
+  static std::optional<axom::sidre::MFEMSidreDataCollection> datacoll_;
```
jamiebramwell, line 282: I don't think this should be optional.

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -172,17 +190,85 @@ class FiniteElementState {
...
-    return std::make_unique<Tensor>(&space_);
+    return std::make_unique<Tensor>(&retrieve(space_));
   }
 
 private:
+  /**
+   * @brief A helper type for uniform semantics over owning/non-owning pointers
+   */
+  template <typename T>
+  using MaybeOwner = std::variant<T*, std::unique_ptr<T>>;
```
jamiebramwell: `MaybeOwnedPointer`?

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -172,17 +190,85 @@ class FiniteElementState {
...
+  }
+
   // Allows for copy/move assignment
-  std::reference_wrapper<mfem::ParMesh>          mesh_;
-  std::unique_ptr<mfem::FiniteElementCollection> coll_;
-  mfem::ParFiniteElementSpace                    space_;
-  std::unique_ptr<mfem::ParGridFunction>         gf_;
-  mfem::HypreParVector                           true_vec_;
-  std::string                                    name_ = "";
+  MaybeOwner<mfem::ParMesh> mesh_;
```
jamiebramwell: When are these owned vs. not owned?

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -9,16 +9,94 @@
...
+    : mesh_(&mesh),
       coll_(options.coll ? std::move(options.coll)
                          : std::make_unique<mfem::H1_FECollection>(options.order, mesh.Dimension())),
-      space_(&mesh, coll_.get(), options.space_dim ? *options.space_dim : mesh.Dimension(), options.ordering),
-      gf_(std::make_unique<mfem::ParGridFunction>(&space_)),
-      true_vec_(&space_),
+      space_(std::make_unique<mfem::ParFiniteElementSpace>(
+          &mesh, &retrieve(coll_), options.space_dim ? *options.space_dim : mesh.Dimension(), options.ordering)),
+      gf_(options.allocate_gf
+              ? std::make_unique<mfem::ParGridFunction>(&retrieve(space_))
```
jamiebramwell: This is a bit confusing. It seems like it is being allocated in both cases?

src/serac/physics/utilities/finite_element_state.hpp
joshessman-llnl: They are owned by the FiniteElementState in a "normal" run, and non-owned (owned by the DataCollection) in a restart run

tests/serac_restart.cpp
joshessman-llnl: Yes, I can add an accessor for that

src/serac/physics/utilities/finite_element_state.hpp
joshessman-llnl: The use of `optional` here is just a "trick" to defer construction of the global DataCollection object (specifically, until `StateManager::initialize` is called) without the need to keep the object on the heap (via `unique_ptr`)

src/serac/physics/utilities/finite_element_state.cpp
joshessman-llnl: This is a current restriction on how the DataCollection is used - the `nullptr` parameter leaves the underlying data unallocated, to be filled in with a DataCollection-allocated array later.  This process is needed if we want Sidre to own the data (which is a prerequisite for it to be written to a savefile).

joshessman-llnl:
> 
> 
> > A couple possible discussion points:
> > Should any part of the input file be preserved in the event of a restart? Currently the previous input data is wiped out.
> 
> Ideally, the entire input file is contained in the restart (in my production code, well call this "restart reproducibility"). Essentially this means the only thing needed to continue a run is the restart file. For many reasons, this often isn't the case but it is worth striving toward.

Hmm, so would we not need/want any kind of "new" input file in a restart run?  I was under the impression that in some circumstances a user would want to change simulation parameters as part of a restart (e.g., tweaked boundary conditions, material parameters, etc).

tests/serac_restart.cpp
joshessman-llnl: Though this will require some refactoring as physics objects currently require ownership of the mesh.

jamiebramwell:
@joshessman-llnl , is this still waiting on an axom PR or should I take another look?

src/drivers/serac.cpp
```diff
@@ -80,33 +80,44 @@ int main(int argc, char* argv[])
...
-  if (const auto file_opts = std::get_if<serac::mesh::FileInputOptions>(&mesh_options.extra_options)) {
-    auto full_mesh_path = serac::input::findMeshFilePath(file_opts->relative_mesh_file_name, input_file_path);
-    mesh = serac::buildMeshFromFile(full_mesh_path, mesh_options.ser_ref_levels, mesh_options.par_ref_levels);
+  if (!restart_cycle) {
+    // Build the mesh
+    auto mesh_options = inlet["main_mesh"].get<serac::mesh::InputOptions>();
+    if (const auto file_opts = std::get_if<serac::mesh::FileInputOptions>(&mesh_options.extra_options)) {
+      auto full_mesh_path = serac::input::findMeshFilePath(file_opts->relative_mesh_file_name, input_file_path);
+      auto mesh = serac::buildMeshFromFile(full_mesh_path, mesh_options.ser_ref_levels, mesh_options.par_ref_levels);
+      serac::StateManager::setMesh(std::move(mesh));
```
jamiebramwell: I think this is fine for now, but we should consider in the future what multiple meshes look like. I imagine you would append a tag to each mesh and send the appropriate mesh tag to the solvers. Something like this: 

```c++
serac::StateManager::setMesh(mesh_1, "thermal_mesh");
serac::StateManager::setMesh(mesh_2, "solid_mesh");

ThermalConduction(*thermal_solver_options, "thermal_mesh");
Solid(*solid_solver_options, "solid_mesh");
```

src/serac/infrastructure/input.cpp
```diff
@@ -24,8 +24,11 @@ axom::inlet::Inlet initialize(axom::sidre::DataStore& datastore, const std::stri
   luareader->parseFile(input_file_path);
 
   // Store inlet data under its own group
+  if (datastore.getRoot()->hasGroup("input_file")) {
+    // If this is a restart, wipe out the previous input file
+    datastore.getRoot()->destroyGroup("input_file");
```
jamiebramwell, line 29: Is it possible to restart without also including a restart file or do we require a user also include it on restart?

src/serac/numerics/mesh_utils.cpp
```diff
@@ -16,7 +16,7 @@
 
 namespace serac {
 
-std::shared_ptr<mfem::ParMesh> buildMeshFromFile(const std::string& mesh_file, const int refine_serial,
+std::unique_ptr<mfem::ParMesh> buildMeshFromFile(const std::string& mesh_file, const int refine_serial,
```
jamiebramwell: 🎉

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -163,17 +173,92 @@ class FiniteElementState {
...
+   */
+  template <typename T>
+  using MaybeOwner = std::variant<T*, std::unique_ptr<T>>;
+
+  /**
+   * @brief Retrieves a reference to the underlying object in a MaybeOwner
+   * @param[in] obj The object to dereference
+   */
+  template <typename T>
+  static T& retrieve(MaybeOwner<T>& obj)
```
jamiebramwell: Is it possible to overload the `*` operator for this?

src/serac/physics/utilities/finite_element_state.hpp
jamiebramwell: Another option, `GenericPointer` ? I think it would be nice to know this is a pointer type at first glance.

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -163,17 +173,92 @@ class FiniteElementState {
...
   // Allows for copy/move assignment
-  std::reference_wrapper<mfem::ParMesh>          mesh_;
-  std::unique_ptr<mfem::FiniteElementCollection> coll_;
-  mfem::ParFiniteElementSpace                    space_;
-  std::unique_ptr<mfem::ParGridFunction>         gf_;
-  mfem::HypreParVector                           true_vec_;
-  std::string                                    name_ = "";
+  mfem::ParMesh* mesh_;
+  // Must be const as FESpaces store a const reference to their FEColls
+  MaybeOwner<const mfem::FiniteElementCollection> coll_;
```
jamiebramwell: Some comments describing why these need to be `MaybeOwner` would be useful.

src/serac/physics/utilities/finite_element_state.hpp
jamiebramwell: A comment stating this would be appreciated.

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -163,17 +173,92 @@ class FiniteElementState {
...
+   * @note If this is a restart then the options (except for the name) will be ignored
+   */
+  static FiniteElementState newState(FiniteElementState::Options&& options = {});
+
+  /**
+   * @brief Updates the Conduit Blueprint state in the datastore and saves to a file
+   * @param[in] t The current sim time
+   * @param[in] cycle The current iteration number of the simulation
+   */
+  static void step(const double t, const int cycle);
```
jamiebramwell: It's a little confusing to me why a method named `step` actually saves to a file. Maybe `save`?

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -9,16 +9,97 @@
...
 }
 
+FiniteElementState::FiniteElementState(mfem::ParMesh& mesh, mfem::ParGridFunction& gf, const std::string& name)
+    : mesh_(&mesh), space_(gf.ParFESpace()), gf_(&gf), true_vec_(&retrieve(space_)), name_(name)
+{
+  coll_     = retrieve(space_).FEColl();
+  true_vec_ = 0.0;
+}
+
+std::optional<axom::sidre::MFEMSidreDataCollection> StateManager::datacoll_;
```
jamiebramwell, line 36: Add a comment about initializing these static members.

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -9,16 +9,97 @@
...
+      // Use a raw pointer here, lifetime will be managed by the DataCollection
+      gf_(new mfem::ParGridFunction(&retrieve(space_), static_cast<double*>(nullptr))),
+      true_vec_(&retrieve(space_)),
       name_(options.name)
 {
-  *gf_      = 0.0;
   true_vec_ = 0.0;
 }
 
+FiniteElementState::FiniteElementState(mfem::ParMesh& mesh, mfem::ParGridFunction& gf, const std::string& name)
```
jamiebramwell, line 28: A large comment describing why the `MaybeOwned` type and the general memory management strategy of restart vs. non-restart runs would be useful.

jamiebramwell (commented):
Looks good! As is my standard practice, most of my issues are around needing more comments. 😄

src/serac/infrastructure/input.cpp
joshessman-llnl: If I'm understanding this correctly, Inlet doesn't have the capability to reconstruct its internal hierarchy from the corresponding Sidre data (so we'd need a "fresh" input file for everything).  I imagine this would be a useful feature for Inlet, though, since restarts were one of the primary motivations for using Sidre for data storage.

src/serac/physics/utilities/finite_element_state.hpp
joshessman-llnl: To my knowledge it's not possible to implement nonmember operator overloads for nested template classes (the compiler just can't figure out the lookups, I think).

src/serac/physics/utilities/finite_element_state.hpp
joshessman-llnl: I ended up going with `MaybeOwningPointer`, but happy to change it if the consensus is something else

src/drivers/serac.cpp
```diff
@@ -80,33 +80,44 @@ int main(int argc, char* argv[])
...
   // Save input values to file
   datastore.getRoot()->save("serac_input.json", "json");
 
-  std::shared_ptr<mfem::ParMesh> mesh;
-  // Build the mesh
-  auto mesh_options = inlet["main_mesh"].get<serac::mesh::InputOptions>();
-  if (const auto file_opts = std::get_if<serac::mesh::FileInputOptions>(&mesh_options.extra_options)) {
-    auto full_mesh_path = serac::input::findMeshFilePath(file_opts->relative_mesh_file_name, input_file_path);
-    mesh = serac::buildMeshFromFile(full_mesh_path, mesh_options.ser_ref_levels, mesh_options.par_ref_levels);
+  if (!restart_cycle) {
```
white238, line 125: Add a comment here something like:

// Not restarting, mesh must be created and added to the StateManager

src/drivers/serac.cpp
```diff
@@ -80,33 +80,44 @@ int main(int argc, char* argv[])
...
+  }
+
   // Check for the doc creation command line argument
   bool create_input_file_docs = cli_opts.find("create_input_file_docs") != cli_opts.end();
 
   // Create DataStore
   axom::sidre::DataStore datastore;
 
+  // Intialize MFEMSidreDataCollection
+  serac::StateManager::initialize(datastore, restart_cycle);
```
white238, line 107: Add comment here about alternate behavior between restarting/regular run

src/drivers/serac.cpp
```diff
@@ -80,33 +80,44 @@ int main(int argc, char* argv[])
...
-    auto writer = std::make_unique<axom::inlet::SphinxDocWriter>("serac_input.rst", inlet.sidreGroup());
-    inlet.registerDocWriter(std::move(writer));
+    auto writer = std::make_unique<axom::inlet::SphinxWriter>("serac_input.rst");
+    inlet.registerWriter(std::move(writer));
     inlet.writeDoc();
     serac::exitGracefully();
   }
 
   // Save input values to file
   datastore.getRoot()->save("serac_input.json", "json");
```
white238: This originally was "hacked" in to show the input values in a quick and dirty way.  This PR will add extra values to the sidre tree on restarts.  Can you either move this up or probably better yet, only print out the input part of the tree instead of the whole root.

src/serac/infrastructure/cli.cpp
```diff
@@ -44,6 +47,9 @@ std::unordered_map<std::string, std::string> defineAndParse(int argc, char* argv
   // Store found values
   std::unordered_map<std::string, std::string> cli_opts;
   cli_opts.insert({std::string("input_file"), input_file_path});
+  if (restart_opt->count()) {
```
white238: What is count?

src/serac/infrastructure/input.cpp
jamiebramwell: Yeah, I think this is fine for now but we should document this restriction and log a feature request in Inlet.

jamiebramwell (approved):
Awesome stuff!

src/serac/infrastructure/input.cpp
joshessman-llnl: Logged as https://github.com/LLNL/axom/issues/487

white238: Should we save the path of the input file in sidre and load from that so the user doesn't have to specify?

src/serac/physics/base_physics.cpp
```diff
@@ -72,20 +77,17 @@ void BasePhysics::initializeOutput(const serac::OutputType output_type, const st
...
-      dc_ = std::make_unique<axom::sidre::MFEMSidreDataCollection>(root_name_, &state_.front().get().mesh());
       break;
     }
 
     default:
       SLIC_ERROR_ROOT("OutputType not recognized!");
   }
 
-  if ((output_type_ == OutputType::VisIt) || (output_type_ == OutputType::SidreVisIt)) {
+  if (output_type_ == OutputType::VisIt) {
```
white238, line 90: Can you add some comments explaining this function? At first glance, it seems weird there is two case statements of basically the same thing.

white238 (approved):

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #322, opened 2020-12-29
Switch to pragma once, conform to google style guide for header ordering

### LINKED ISSUE #161: Investigate pragma once for header files
Now that we can compile on lassen, we should test for convenience C++ feature such as pragma once.

### LINKED ISSUE #306: Investigate turning off re-ordering of header includes
`clang-format` reordering header files can be problematic if we rely on a redefining something.  Per the Google Style Guide we should have it in the following order:

https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

joshessman-llnl:
Also switches to nested namespaces (e.g. `namespace serac::logger` instead of `namespace serac` then `namespace logger`).

Resolves #161 
Resolves #306

### DIFF
--- a/src/serac/numerics/expr_template_impl.hpp
+++ b/src/serac/numerics/expr_template_impl.hpp
@@ -5,14 +5,13 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 
 /**
- * \@file expr_templates_internal.hpp
+ * \@file expr_templates_impl.hpp
  *
  * @brief The internal implementation for a set of template classes used to
  * represent the evaluation of unary and binary operations on vectors
  */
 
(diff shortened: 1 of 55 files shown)


src/serac/numerics/expr_template_impl.hpp
```diff
@@ -21,7 +20,7 @@
 #include "serac/infrastructure/logger.hpp"
 #include "serac/numerics/vector_expression.hpp"
 
-namespace serac::internal {
+namespace detail {
```
white238: serac::detail?

jamiebramwell (approved):

white238 (approved):

joshessman-llnl:
LGTM

PR #323, opened 2020-12-30
Add input schemas/test files for thermal solver

joshessman-llnl:
This PR moves the thermal solver tests to input files.

Most of this PR is analogous to the similar PR(s) for the `NonlinearSolid` module, so I have templated the test utility functions on the physics module to reduce duplication.

### DIFF
--- a/data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
+++ b/data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
@@ -6,6 +6,9 @@ epsilon = 0.00001
 dt      = 0.0001
 t_final = 0.001
 
+-- Simulation output format
+output_type = "ParaView"
+
 main_mesh = {
     -- mesh file
     mesh = "../../../meshes/star.mesh",
--- a/data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
+++ b/data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
@@ -17,6 +17,11 @@ main_mesh = {
     par_ref_levels = 1,
 }
 
+temp_func = function (x, y, z)
+    local norm = math.sqrt(x ^ 2 + y ^ 2 + z ^ 2)
+    if norm < 0.5 then return 2.0 else return 1.0 end
+end
+
 -- Solver parameters
 thermal_conduction = {
     stiffness_solver = {
@@ -53,20 +58,14 @@ thermal_conduction = {
 
     -- initial conditions
     initial_temperature = {
-        coef = function (x, y, z)
-            local norm = math.sqrt(x ^ 2 + y ^ 2 + z ^ 2)
-            if norm < 0.5 then return 2.0 else return 1.0 end
-        end
+        coef = temp_func
     },
 
     -- boundary condition parameters
     boundary_conds = {
         ['temperature'] = {
             attrs = {1},
-            coef = function (x, y, z)
-                local norm = math.sqrt(x ^ 2 + y ^ 2 + z ^ 2)
-                if norm < 0.5 then return 2.0 else return 1.0 end
-            end
+            coef = temp_func
         },
     },
 }
--- a/data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
+++ b/data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
@@ -1,13 +1,14 @@
 -- Comparison information
-expected_t_l2norm = 2.56980679
+expected_t_l2norm = 2.6493029
 epsilon = 0.00001
 
 -- Simulation time parameters
-dt      = 1.0
+dt      = 0.0001
+t_final = 0.001
 
 main_mesh = {
     -- mesh file
-    mesh = "../../../meshes/star_with_2_bdr_attributes.mesh",
+    mesh = "../../../meshes/star.mesh",
     -- serial and parallel refinement levels
     ser_ref_levels = 1,
     par_ref_levels = 1,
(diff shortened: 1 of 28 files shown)


data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
```diff
@@ -0,0 +1,72 @@
...
+    },
+
+    -- polynomial interpolation order
+    order = 2,
+
+    -- material parameters
+    kappa = 0.5,
+
+    -- initial conditions
+    initial_temperature = {
```
jamiebramwell, line 60: Is there any way to pull this function out for a common definition between the initial condition and the essential boundary? I can see this being a source of errors.

data/input_files/tests/thermal_conduction/static_solve_multiple_bcs.lua
```diff
@@ -0,0 +1,63 @@
+-- Comparison information
+expected_t_l2norm = 2.56980679
+epsilon = 0.00001
+
+-- Simulation time parameters
+dt      = 1.0
+
+-- Simulation output format
+output_type = "GLVis"
```
jamiebramwell, line 9: We should probably add this to the solid module as well.

src/serac/physics/nonlinear_solid.hpp
```diff
@@ -68,6 +68,8 @@ class NonlinearSolid : public BasePhysics {
     double mu;
     double K;
 
+    std::optional<double> viscosity;
```
jamiebramwell: It may be clearer if this is non-optional with a default value of zero. Any thoughts?

data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
jamiebramwell: ...I now see that you already did that!

data/input_files/tests/thermal_conduction/static_solve_multiple_bcs.lua
jamiebramwell: ...I now see that you already did that!

src/serac/physics/nonlinear_solid.hpp
joshessman-llnl: I think that's fine, it cleans up some of the messy-looking dereferencing too.

src/serac/infrastructure/input.cpp
```diff
@@ -21,6 +21,7 @@ axom::inlet::Inlet initialize(axom::sidre::DataStore& datastore, const std::stri
   // Initialize Inlet
   auto luareader = std::make_unique<axom::inlet::LuaReader>();
   luareader->parseFile(input_file_path);
+  luareader->solState().open_libraries(sol::lib::math);
```
white238, line 22: This can be removed after we update axom

jamiebramwell (approved):

jamiebramwell:
@samuelpmishLLNL  @jonwong12 can you take a look at this?

src/serac/physics/nonlinear_solid.cpp
```diff
@@ -351,6 +355,8 @@ NonlinearSolid::InputOptions FromInlet<NonlinearSolid::InputOptions>::operator()
   result.mu = base["mu"];
   result.K  = base["K"];
 
+  result.viscosity = base["viscosity"];
```
jonwong12, line 358: Should we automatically make this zero if it's not defined or does the underlying c++ code methods do that?

jonwong12:
Would it be worth leaving one example as C++ code? I know when I was moving towards using the lua scripting interface for the newmark beta work it took a little bit of time to figure out how to add/modify options and how the lua stuff is integrated. I think if there was one 1-to-1 example between a lua test and the templated test that might be beneficial if users have a test that runs in serac and want to move it towards being fully scripted from lua.

jonwong12 (approved):
Overall it looks good. I just had a comment about moving the test logic over to `test_utilities.cpp`. I'd prefer if the "physics" helper methods could be found more easily. e.g. if I want to add a new test that doesn't quite use the same logic in the helper method and requires adding something new like a body force or treats the same information differently, what might be the workflow for adding these new tests.

src/serac/physics/nonlinear_solid.cpp
joshessman-llnl: I have it defaulted to zero:
https://github.com/LLNL/serac/blob/cd1a05753c091eb0b4d48c9c4875fab3fee9b132/src/serac/physics/nonlinear_solid.cpp#L295

jamiebramwell:
> 
> 
> Would it be worth leaving one example as C++ code? I know when I was moving towards using the lua scripting interface for the newmark beta work it took a little bit of time to figure out how to add/modify options and how the lua stuff is integrated. I think if there was one 1-to-1 example between a lua test and the templated test that might be beneficial if users have a test that runs in serac and want to move it towards being fully scripted from lua.

I think this is a good idea, but maybe not a duplicate test. Maybe in an examples directory or something?

jamiebramwell:
> 
> 
> Overall it looks good. I just had a comment about moving the test logic over to `test_utilities.cpp`. I'd prefer if the "physics" helper methods could be found more easily. e.g. if I want to add a new test that doesn't quite use the same logic in the helper method and requires adding something new like a body force or treats the same information differently, what might be the workflow for adding these new tests.

I think this is a fair point. Maybe having separate files for the physics-specific instantiations (for example `solid_test_utils.cpp` and `thermal_test_utils.cpp`) would make it a bit clearer.

joshessman-llnl:
> 
> 
> > Overall it looks good. I just had a comment about moving the test logic over to `test_utilities.cpp`. I'd prefer if the "physics" helper methods could be found more easily. e.g. if I want to add a new test that doesn't quite use the same logic in the helper method and requires adding something new like a body force or treats the same information differently, what might be the workflow for adding these new tests.
> 
> I think this is a fair point. Maybe having separate files for the physics-specific instantiations (for example `solid_test_utils.cpp` and `thermal_test_utils.cpp`) would make it a bit clearer.

Hmm, the test driver is currently physics-independent.  In the current design, physics-specific setup is delegated to the physics module constructor that processes input from the file.

For example, if you wanted to incorporate the `NonlinearSolid::setExternalForces` method into a test, ideally it would be possible to define the relevant coefficients in an input file and add logic to `NonlinearSolid::NonlinearSolid(std::shared_ptr<mfem::ParMesh>, const NonlinearSolid::InputOptions&)` that checks if any external forces are defined in the input file and calls the method accordingly.

The hope is that eventually an input file would completely describe a simulation, without the need to short-circuit any of the input file processing logic.

jonwong12:
> 
> 
> > > Overall it looks good. I just had a comment about moving the test logic over to `test_utilities.cpp`. I'd prefer if the "physics" helper methods could be found more easily. e.g. if I want to add a new test that doesn't quite use the same logic in the helper method and requires adding something new like a body force or treats the same information differently, what might be the workflow for adding these new tests.
> > 
> > 
> > I think this is a fair point. Maybe having separate files for the physics-specific instantiations (for example `solid_test_utils.cpp` and `thermal_test_utils.cpp`) would make it a bit clearer.
> 
> Hmm, the test driver is currently physics-independent. In the current design, physics-specific setup is delegated to the physics module constructor that processes input from the file.
> 
> For example, if you wanted to incorporate the `NonlinearSolid::setExternalForces` method into a test, ideally it would be possible to define the relevant coefficients in an input file and add logic to `NonlinearSolid::NonlinearSolid(std::shared_ptr<mfem::ParMesh>, const NonlinearSolid::InputOptions&)` that checks if any external forces are defined in the input file and calls the method accordingly.
> 
> The hope is that eventually an input file would completely describe a simulation, without the need to short-circuit any of the input file processing logic.

Ahh I see. I think the name `defineNonlinSolidInputFileSchema` threw me off. At closer inspection it's really just for the nonlinear tests, so I retract my comment.

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #326, opened 2021-01-06
Add setExternalForces method to NonlinearSolid

jonwong12:
This PR allows users to specify a `vector` of external forces (`VectorCoefficients`) to apply to a `NonlinearSolid`.

The forces are added to `NonlinearSolid.H_` because the `H_` operator is used both in the quasistatic and dynamic modes and will be consistent.

Instead of writing yet another nonlinearformintegrator, a `mfem::LinearformIntegrator` was wrapped as an `mfem::NonlinearformIntegrator`.

### DIFF
--- a/src/serac/physics/nonlinear_solid.cpp
+++ b/src/serac/physics/nonlinear_solid.cpp
@@ -165,9 +165,9 @@ void NonlinearSolid::completeSetup()
 
   // Add external forces
   for (auto& force : ext_force_coefs_) {
-    H_->AddDomainIntegrator(
-        new LinearToNonlinearFormIntegrator(std::make_shared<mfem::VectorDomainLFIntegrator>(*force),
-                                            std::make_shared<mfem::ParFiniteElementSpace>(*H_->ParFESpace())));
+    H_->AddDomainIntegrator(new serac::mfem_ext::LinearToNonlinearFormIntegrator(
+        std::make_shared<mfem::VectorDomainLFIntegrator>(*force),
+        std::make_shared<mfem::ParFiniteElementSpace>(*H_->ParFESpace())));
   }
 
   // Build the dof array lookup tables
--- a/src/serac/physics/nonlinear_solid.cpp
+++ b/src/serac/physics/nonlinear_solid.cpp
@@ -119,9 +119,9 @@ void NonlinearSolid::setTractionBCs(const std::set<int>&                     tra
   bcs_.addNatural(trac_bdr, trac_bdr_coef, component);
 }
 
-void NonlinearSolid::setExternalForces(std::vector<std::shared_ptr<mfem::VectorCoefficient>> ext_force_coefs)
+void NonlinearSolid::addBodyForce(std::shared_ptr<mfem::VectorCoefficient> ext_force_coef)
 {
-  ext_force_coefs_ = ext_force_coefs;
+  ext_force_coefs_.push_back(ext_force_coef);
 }
 
 void NonlinearSolid::setHyperelasticMaterialParameters(const double mu, const double K)
--- a/src/serac/physics/nonlinear_solid.cpp
+++ b/src/serac/physics/nonlinear_solid.cpp
@@ -9,6 +9,7 @@
 #include "serac/infrastructure/logger.hpp"
 #include "serac/integrators/hyperelastic_traction_integrator.hpp"
 #include "serac/integrators/inc_hyperelastic_integrator.hpp"
+#include "serac/integrators/wrapper_integrator.hpp"
 #include "serac/numerics/expr_template_ops.hpp"
 #include "serac/numerics/mesh_utils.hpp"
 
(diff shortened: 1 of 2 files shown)


src/serac/physics/nonlinear_solid.cpp
```diff
@@ -118,6 +119,11 @@ void NonlinearSolid::setTractionBCs(const std::set<int>&                     tra
   bcs_.addNatural(trac_bdr, trac_bdr_coef, component);
 }
 
+void NonlinearSolid::setExternalForces(std::vector<std::shared_ptr<mfem::VectorCoefficient>> ext_force_coefs)
```
white238: Who are these vector coefficients shared with?

jonwong12: The vector coefficients might be created by the user (or sidre) and shared with the `NonlinearSolid`. The user or another solver might change the vector coefficients over time. 

This was one way to guarantee the VectorCoefficient's don't get out of scope. It's similar to the displacement and tractionBCs.

jamiebramwell: I know these aren't really boundary conditions, but it does describe an external loading and a lot of production codes treat it as such. Should we add this as another boundary condition type? Any thoughts @samuelpmishLLNL @joshessman-llnl @solberg2 ?

solberg2: These are body forces, I think?  So yes it makes sense to treat them as a boundary condition.

Sometimes people will have body forces coming from an external source, basically electromagnetic forces.

Sometimes people will have body forces coming from an imposed external acceleration (gravity, centripetal acceleration)

Sometimes people will have body forces coming from the method of manufactured solutions.

src/serac/physics/nonlinear_solid.hpp
```diff
@@ -121,6 +121,13 @@ class NonlinearSolid : public BasePhysics {
   void setTractionBCs(const std::set<int>& trac_bdr, std::shared_ptr<mfem::VectorCoefficient> trac_bdr_coef,
                       int component = -1);
 
+  /**
+   * @brief Set the external force vectors on the domain
+   *
+   * @param[in] ext_force_coefs The sd::vector of vector-valued external force coefficients applied to the domain
+   */
+  void setExternalForces(std::vector<std::shared_ptr<mfem::VectorCoefficient>> ext_force_coefs);
```
jamiebramwell: I think `setBodyForces` is a bit clearer for what this is actually doing.

jamiebramwell: Also, I think having this with a single coefficient and then pushing them to a vector internally would be easier to use in the common gravity load case. This is similar to how we handle the boundary conditions.

src/serac/physics/nonlinear_solid.cpp
```diff
@@ -157,6 +163,13 @@ void NonlinearSolid::completeSetup()
                              nat_bc_data.markers());
   }
 
+  // Add external forces
+  for (auto& force : ext_force_coefs_) {
+    H_->AddDomainIntegrator(
+        new LinearToNonlinearFormIntegrator(std::make_shared<mfem::VectorDomainLFIntegrator>(*force),
```
jamiebramwell: The convenience wrappers are so nice!

src/serac/physics/nonlinear_solid.hpp
jonwong12: Hmm. I just wanted to clarify if you think we should internally have vectorcoefficient internally, or to take as input only one vectorcoefficient.

I was envisioning the case of two loads (gravity + applied torque) and how we might want to accommodate that. In other codes, the gravity would have a density term and that might be modified, where the applied torque might not. Someone else would then have to combine them, whereas if they are store separately, they can be mutated individually and we can combine them in any way we like.

I do see that there would be a performance benefit of internally having one vector coefficient since we wouldn't have to loop over multiple domain integrators.

jamiebramwell: Oh, input as only one vector coefficient. So internally, it's is still a `std::vector<VectorCoefficient>`, but the function call sets them one at a time. This means to have multiple body loads, a user would do this
```
solid.setBodyForce(force1);
solid.setBodyForce(force2);
```
instead of constructing a `std::vector` themselves.

jamiebramwell: I wouldn't worry too much about the performance hit as it will probably change significantly with the upcoming `VariationalForm` anyway.

jamiebramwell: maybe `addBodyForce` for clarity.

jamiebramwell (approved):

jamiebramwell:
@samuelpmishLLNL could you take a look at this? Thanks!

samuelpmishLLNL (approved):

samuelpmishLLNL:
As we discussed in last week's meeting, there are some additional load types we may want to support as well but this seems like a reasonable start for for small deformations

jamiebramwell:
LGTM

PR #334, opened 2021-01-20
WIP/DISCUSS: Add cuboid and rectangle mesh generation options to meshing schema.

jonwong12:
This PR adds extra meshing schema options to distinguish meshes read in from files and meshes that are generated using serac mesh generators.

Some discussion topics: 
1. Does it make sense to have various mesh input options structs tied to different meshing paths?
2. Is there a mechanism to tell inlet that if `mesh.elements` is defined, then `mesh.elements.x` should exist?


src/serac/numerics/mesh_utils.cpp
```diff
@@ -172,20 +173,59 @@ namespace mesh {
...
+  table.addString("mesh", "Path to Mesh file");
 
   // Refinement levels
   table.addInt("ser_ref_levels", "Number of times to refine the mesh uniformly in serial.").defaultValue(0);
   table.addInt("par_ref_levels", "Number of times to refine the mesh uniformly in parallel.").defaultValue(0);
+
+  // mesh generation options
+  auto& elements = table.addTable("elements");
+  // JW: Can these be specified as requierd if elements is defined?
+  elements.addInt("x", "x-dimension");
```
white238, line 199: Did you try  `elements.addInt("x", "x-dimension").required(true);" (This may not work due to requiring it to be there all the time)

Otherwise you can add a custom verifier via a lambda given to the `elements` table.

white238: Example of verifier, that is actually pretty close to your use case:

https://github.com/LLNL/axom/blob/494fc6429ac217b9d6aff5c9bc967dd3ee8aa5b9/src/axom/inlet/examples/verification.cpp#L32-L53

jonwong12: Yes, when I added `.required()` to "x", it required it even when `elements` was not defined (e.g. in the case when we're building a mesh from file).

I will try the custom verifier.

white238: I wonder if we could support this by setting the table to not be required and then field being required.  Something like:

```
elements = table.addTable("elements").required(false);
elements.addInt("x").required(true);
```

This may be a bad idea... since we would have to track you explicitly set the table to false instead of just default?

Better idea: required only counts when the parent exists? So it would only be fully required like this:

```
elements = table.addTable("elements").required(true);
elements.addInt("x").required(true);
```

but the following is required only if the elements table is defined:

```
elements = table.addTable("elements");
elements.addInt("x").required(true);
```

data/input_files/tests/meshing/meshing.lua
```diff
@@ -0,0 +1,22 @@
...
+    ser_ref_levels = 1,
+    par_ref_levels = 0,
+}
+
+main_mesh_rect = {
+    -- mesh file
+    elements = {x = 3, y = 3},
+    ser_ref_levels = 1,
+    par_ref_levels = 0,
+}
```
joshessman-llnl: In the test input files, we could also look at using the approach we use elsewhere in our input files for variants (e.g., for solver options):
```lua
main_mesh = {
  type = "from_file",
  from_file_options = {
    file = "/path/to/mesh"
  },
  generated_options = {
    shape = "cuboid",
    elements = {x = 3, y = 3, z = 3},
    size = {x = 1, y = 2, z = 3},
  }
  ser_ref_levels = 1,
  par_ref_levels = 0,
}
```

src/drivers/serac.cpp
```diff
@@ -89,10 +89,14 @@ int main(int argc, char* argv[])
...
   // Build the mesh
-  auto mesh_options   = inlet["main_mesh"].get<serac::mesh::InputOptions>();
-  auto full_mesh_path = serac::input::findMeshFilePath(mesh_options.relative_mesh_file_name, input_file_path);
-  auto mesh = serac::buildMeshFromFile(full_mesh_path, mesh_options.ser_ref_levels, mesh_options.par_ref_levels);
+  auto mesh_options = inlet["main_mesh"].get<serac::mesh::InputOptions>();
+  if (std::holds_alternative<serac::mesh::FileInputOptions>(mesh_options.extra_options)) {
+    auto full_mesh_path = serac::input::findMeshFilePath(
+        std::get<serac::mesh::FileInputOptions>(mesh_options.extra_options).relative_mesh_file_name, input_file_path);
+    mesh = serac::buildMeshFromFile(full_mesh_path, mesh_options.ser_ref_levels, mesh_options.par_ref_levels);
+  }
```
joshessman-llnl: Could we use `get_if` here?

```suggestion
  if (const auto file_opts = std::get_if<serac::mesh::FileInputOptions>(&mesh_options.extra_options)) {
    auto full_mesh_path = serac::input::findMeshFilePath(file_opts->relative_mesh_file_name, input_file_path);
    mesh = serac::buildMeshFromFile(full_mesh_path, mesh_options.ser_ref_levels, mesh_options.par_ref_levels);
  }
```

joshessman-llnl (commented):
Looking good so far, it'll be useful in #323 (depending on which one goes in first)

src/drivers/serac.cpp
jonwong12: Thanks! I'm still getting up to speed with variants, so I was hoping to get feedback on how to best use them =).

data/input_files/tests/meshing/meshing.lua
jonwong12: I see. Now that I think about it, it's better to have a `type` parameter because it forces the user to choose either `file` or `generated`. Otherwise it might be unclear which is active.

src/serac/numerics/mesh_utils.cpp
joshessman-llnl: I've started to look into this, and recalled a relevant comment from a recent inlet PR:
https://github.com/LLNL/axom/pull/427#discussion_r556825102

I think the idea of only checking sub-objects when the parent object is defined makes sense, but it might be a little tricky to determine when the parent table is defined.  Currently this is based on the existence of at least one expected subobject, so `elements = { q = 1 }` would not count as "defined", but `elements = { x = 1 }` would.  

I suppose we could add logic to just check if the `elements` table is a valid "path" in the input file and go from there.

jamiebramwell (approved):
This is really nice! Thanks!

jamiebramwell:
@joshessman-llnl Do you have any outstanding issues?

src/serac/numerics/mesh_utils.hpp
```diff
@@ -93,12 +100,57 @@ struct InputOptions {
...
+
+/**
+ * @brief Constructs a 3D MFEM mesh of a cuboid
+ *
+ * @param[in] extra_options Cuboid Mesh Options
+ * @param[in] MPI_Comm MPI Communicator
+ * @return A shared_ptr containing the constructed mesh
+ */
+std::shared_ptr<mfem::ParMesh> buildCuboidMesh(serac::mesh::GenerateInputOptions& options,
+                                               const MPI_Comm = MPI_COMM_WORLD);
```
joshessman-llnl: Can these be named (just `comm` or something)?  I think Doxygen likes to have the parameters named in the prototype/header.

joshessman-llnl (approved):

src/serac/numerics/mesh_utils.hpp
jonwong12: Sure. I just copied it from the original buildCuboidMesh. I'll change them.

jamiebramwell:
LGTM

PR #335, opened 2021-01-20
Better ODE Testing + Sphinx Documentation

samuelpmish:
- adds a new test that tries different combinations of  {linear, nonlinear}, {first order, second order}, {constrained, unconstrained} ODEs with the different supported time integration methods and Dirichlet Enforcement methods. A few tests that are known to be unstable are omitted from the suite.

- corrects a missing step when using DirichletEnforcementMethod::FullControl with SecondOrderODE

- removes support for mfem::AverageAcceleration method, as it appears to not actually be implemented directly in terms of mfem's Newmark method (note: we can still support the average acceleration method by setting Newmark parameters accordingly). This is motivated by the fact that we have a good understanding of stability requirements for the DirichletEnforcementMethod options + Newmark, but additional work would have to be done to investigate the stability of mfem::AverageAcceleration + DirichletEnforcementMethods. Alternatively, we could leave it enabled, and just try to warn users that its stability is not well understood.

- adds an additional "Theory Reference" section to the Sphinx/readthedocs documentation that includes an entry explaining the basic theory behind the different Dirichlet Enforcement Methods. 
  - Note: the LaTeX formatting on readthedocs seems to be unusual, see: 
[dirichlet_enforcement_readthedocs.pdf](https://github.com/LLNL/serac/files/5845797/dirichlet_enforcement_readthedocs.pdf)
versus
[dirichlet_enforcement.pdf](https://github.com/LLNL/serac/files/5845800/dirichlet_enforcement.pdf)
  - weird right-alignment is being added when not present in the .rst files, strange whitespace is being added. I'm not sure why, but if there's a workaround, I would like to find it!
 
- wolfram language script generates exact solutions to the small ODEs being solved in the different test cases. These sort of scripts should be interoperable with Mathematica, but are plain text files, so they are smaller and better diffed by version control tools.


samuelpmishLLNL:
LGTM

joshessman-llnl:
> weird right-alignment is being added when not present in the .rst files, strange whitespace is being added. I'm not sure why, but if there's a workaround, I would like to find it!

You might try commenting out https://github.com/LLNL/serac/blob/develop/src/docs/_static/theme_overrides.css, I would be a little surprised if that makes any difference but then again I know very little about CSS.

joshessman-llnl:
Docs viewable at https://serac.readthedocs.io/en/feature-mish2-better_ode_testing/

samuelpmishLLNL:
> Docs viewable at https://serac.readthedocs.io/en/feature-mish2-better_ode_testing/

Thanks!

src/docs/sphinx/theory_reference/odes.rst
```diff
@@ -0,0 +1,171 @@
...
+constraints. The different Dirichlet Enforcement Methods in serac relate
+to the following different interpretations of the constraint equations.
+
+.. _header-n97:
+
+"Direct Control"
+----------------
+
+Direct control proceeds to solve for accelerations by effectively
+replacing the constrained rows the differential equation by the
```
joshessman-llnl:
```suggestion
replacing the constrained rows of the differential equation by the
```

src/serac/physics/operators/odes.cpp
```diff
@@ -233,6 +226,8 @@ void FirstOrderODE::Solve(const double dt, const mfem::Vector& u, mfem::Vector&
   dU_dt_.SetSubVectorComplement(constrained_dofs, 0.0);
   du_dt += dU_dt_;
 
+  // std::cout << "inside: " << t << " " << du_dt(0) << std::endl;
```
joshessman-llnl: Can this be removed?

samuelpmishLLNL:
LGTM

src/docs/sphinx/theory_reference/odes.rst
```diff
@@ -0,0 +1,171 @@
...
+--------------
+
+This last option starts by additively decomposing the solution vector
+into two parts: constrained and unconstrained:
+
+.. math::
+
+   \begin{cases}
+
+   \bf{x}(t) = \bf{x}_c(t) + \bf{x}_u(t) = \bf{G} \; \bf{x}(t) + (\bf{1} - \bf{G}) \; \bf{x}(t) \\
```
joshessman-llnl: Minor: It might be clearer if the distinction between constrained/unconstrained is preserved in the third interpretation of x(t) and its derivatives.  To me, the actual implementation looks like `G * x_c(t) + (1 - G) * x_u(t)` aka `G * g(t) + (1 - G) * x_u(t)`.

src/serac/physics/operators/odes.cpp
```diff
@@ -24,13 +24,24 @@ SecondOrderODE::SecondOrderODE(int n, State&& state, const EquationSolver& solve
...
-    case serac::TimestepMethod::AverageAcceleration:
+     case serac::TimestepMethod::AverageAcceleration:
+      // WARNING: apparently mfem's implementation of AverageAccelerationSolver
+      // is NOT equivalent to Newmark (beta = 0.25, gamma = 0.5), so the
+      // stability analysis of using DirichletEnforcementMethod::DirectControl
+      // with Newmark methods does not apply.
+      //
+      // TODO: do a more thorough stability analysis for mfem::GeneralizedAlpha2Solver
+      // to characterize which parameter combinations work with time-varying
+      // dirichlet constraints
```
joshessman-llnl, line 44: One option would be to just put a `SLIC_WARNING` here

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+
+  M(1, 0) = 1.0;
+  M(1, 1) = 4.0;
+  M(1, 2) = 1.0;
+
+  M(2, 0) = 0.0;
+  M(2, 1) = 1.0;
+  M(2, 2) = 2.0;
+  return M;
+}();
```
joshessman-llnl: I like this use of immediately-invoked lambdas.

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+#include "serac/numerics/expr_template_ops.hpp"
+#include "serac/physics/operators/odes.hpp"
+#include "serac/physics/operators/stdfunction_operator.hpp"
+
+using namespace serac;
+using namespace serac::mfem_ext;
+
+bool with_print_statements = true;
+
+DirichletEnforcementMethod enforcement_methods[3] = {DirichletEnforcementMethod::DirectControl,
```
joshessman-llnl: Could these (and the other globals you use here for configuration) be `const`?

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+enum constraint_type
+{
+  UNCONSTRAINED,
+  SINE_WAVE,
+  TRIANGLE_WAVE
+};
+
+std::string to_string(constraint_type c)
+{
+  if (c == UNCONSTRAINED) return std::string("unconstrained");
```
joshessman-llnl: Out of curiosity, did you run into something weird when you omitted the explicit constructor call?

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+  if (c == SINE_WAVE) return std::string("sine wave");
+  if (c == TRIANGLE_WAVE) return std::string("triangle wave");
+  return "unknown";
+}
+
+std::string to_string(serac::TimestepMethod m)
+{
+  // for first order odes
+  if (m == serac::TimestepMethod::BackwardEuler) return std::string("BackwardEuler");
```
joshessman-llnl: Hopefully we'll be able to switch to GCC 9 soon so we can start using `magic_enum`

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+  void                       SetUp() override { std::tie(type, constraint, timestepper, enforcement) = GetParam(); }
+  ode_type                   type;
+  constraint_type            constraint;
+  TimestepMethod             timestepper;
+  DirichletEnforcementMethod enforcement;
+};
+
+TEST_P(FirstOrderODE_suite, all)
+{
+  if (with_print_statements) {
```
joshessman-llnl: Can these be converted to `SLIC_INFO` messages?

joshessman-llnl (commented):
This is looking really nice - these new tests are really thorough and the theory reference you put together is quite accessible

scripts/wolfram/serac_odes_solutions.wls
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env wolframscript
```
jamiebramwell, line 1: Can this script include the copyright?

src/docs/sphinx/theory_reference/solid_mech.rst
```diff
@@ -0,0 +1,20 @@
...
+-----------
+
+.. math::
+
+   \frac{ \sum_{t=0}^{N}f(t,k) }{N}
+
+inline math :math:`\frac{ \sum_{t=0}^{N}f(t,k) }{N}`
+
+
+This is a placeholder to show how to add additional pages to the Theory Reference Section
```
jamiebramwell: I can add a small section here in a follow-on docs reorganization PR.

src/docs/sphinx/theory_reference/odes.rst
```diff
@@ -0,0 +1,171 @@
+.. ## Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
```
jamiebramwell: Update to 2021.

jamiebramwell: (the other files will be covered by https://github.com/LLNL/serac/pull/337)

src/docs/sphinx/theory_reference/solid_mech.rst
```diff
@@ -0,0 +1,20 @@
+.. ## Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
```
jamiebramwell: Update to 2021.

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
+// Copyright (c) 2019-2020, Lawrence Livermore National Security, LLC and
```
jamiebramwell: Update to 2021.

tests/serac_odes.cpp
jamiebramwell: Agreed!

tests/serac_odes.cpp
jamiebramwell: @white238 is there anything other than will and time keeping us from GCC9?

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+  if (m == serac::TimestepMethod::WBZAlpha) return 2;             // ?
+  if (m == serac::TimestepMethod::AverageAcceleration) return 2;  // ?
+  if (m == serac::TimestepMethod::LinearAcceleration) return 2;   // ?
+  if (m == serac::TimestepMethod::CentralDifference) return 2;    // ?
+  if (m == serac::TimestepMethod::FoxGoodwin) return 2;           // ?
+
+  return -1;
+}
+
+int which_kind_of_ode(serac::TimestepMethod m)
```
jamiebramwell, line 213: Maybe `order_of_ode_solver`?

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+
+  mfem::DenseMatrix J(3, 3);
+
+  mfem::Mesh               mesh1D(2);
+  mfem::ParMesh            mesh(MPI_COMM_WORLD, mesh1D);
+  BoundaryConditionManager bcs(mesh);
+
+  // although these test problems don't really apply to a finite element mesh,
+  // the tools in serac require that these finite element data structures exist
+  serac::FiniteElementState dummy(mesh, FiniteElementState::Options{.order = 1, .name = "dummy"});
```
jamiebramwell, line 259: I believe this defaults to a vector valued FEstate. Is that what you want? I am reading this correctly @joshessman-llnl ?

jamiebramwell: Also I wonder if we should default to a scalar field.

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+  soln[0] = 1.0;
+  soln[1] = 2.0;
+  soln[2] = 3.0;
+
+  for (int i = 0; i < nsteps; i++) {
+    ode.Step(soln, t, dt);
+  }
+
+  // clang-format off
+  double exact_solutions[2][3][3] = {
```
jamiebramwell, line 327: A comment about how you derived these exact solutions would be welcome.

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+
+  mfem::DenseMatrix J(3, 3);
+
+  mfem::Mesh               mesh1D(2);
+  mfem::ParMesh            mesh(MPI_COMM_WORLD, mesh1D);
+  BoundaryConditionManager bcs(mesh);
+
+  // although these test problems don't really apply to a finite element mesh,
+  // the tools in serac require that these finite element data structures exist
+  serac::FiniteElementState dummy(mesh, FiniteElementState::Options{.order = 1, .name = "dummy"});
```
jamiebramwell, line 367: Same question re: vector-valued finite element function.

tests/serac_odes.cpp
```diff
@@ -0,0 +1,645 @@
...
+    }
+  };
+  // clang-format on
+
+  mfem::Vector exact_displacement(exact_displacements[type][constraint], 3);
+  mfem::Vector exact_velocity(exact_velocities[type][constraint], 3);
+  mfem::Vector error_displacement = (exact_displacement - displacement) / exact_displacement.Norml2();
+  mfem::Vector error_velocity     = (exact_velocity - velocity) / exact_velocity.Norml2();
+
+  // std::cout << "displacement constraint error: " << error_displacement(0) << std::endl;
```
jamiebramwell: dead code.

tests/serac_odes.cpp
samuelpmishLLNL: I did not, I will remove the explicit `std::string(****)` constructors

tests/serac_odes.cpp
samuelpmishLLNL: I've never used `magic_enum` before, is that included as part of axom?

tests/serac_odes.cpp
samuelpmishLLNL: I'm trying to avoid "order" since it's ambiguous (order of convergence or first/second order), but `kind` isn't great either! I'll leave a descriptive comment for now.

tests/serac_odes.cpp
white238: GCC 9 is not available on LC yet.  8.3.1 is the newest.  We can always request they install it.

tests/serac_odes.cpp
samuelpmishLLNL: I intended to create a scalar-valued FEstate. I would assume omitting a number of components per node would imply scalar-valued.

joshessman-llnl: It defaults to the dimension of the mesh, which in this case is 1.

tests/serac_odes.cpp
samuelpmish: it might be worth making a PR to mfem to add some convenience constructors of the form:

```cpp
template < int m, int n >
DenseMatrix(const double (&values)[m][n]) { ... }

template < int n >
Vector(const double (&values)[n]) { ... }
```

so we can just write `mfem::DenseMatrix M = {{2.0, 1.0, 0.0}, {1.0, 4.0, 1.0}, {0.0, 1.0, 2.0}}` and avoid having to use the lambda workaround

tests/serac_odes.cpp
jamiebramwell: Agreed. I'll bring this up in a meeting and see if anyone has ideas.

tests/serac_odes.cpp
jamiebramwell: I'll make another PR to change this default. I think it's a different behavior than most FEM libraries so it could be confusing.

jamiebramwell:
LGTM

jamiebramwell (approved):

jamiebramwell:
any outstanding issues @joshessman-llnl ?

joshessman-llnl (approved):

PR #337, opened 2021-01-21
Update copyright year to 2021

### LINKED ISSUE #336: Update copyright year to 2021
There may be scripts we can use for this, but it shouldn't be too difficult to do it with `sed` otherwise.

joshessman-llnl:
Resolves #336

Do we want to keep the slightly modified copy of axom's update script in our repo?


jamiebramwell:
> 
> 
> Resolves #336
> 
> Do we want to keep the slightly modified copy of axom's update script in our repo?

I didn't know that existed! That seems like a good idea. Thoughts @white238 ?

jamiebramwell (approved):

white238 (approved):

jamiebramwell:
LGTM

PR #351, opened 2021-01-27
Adds shadow errors to the build system

### LINKED ISSUE #348: Enable shadow warnings by default
To save us from bugs, we should turn `-Wshadow` on by default. https://interrupt.memfault.com/blog/best-and-worst-gcc-clang-compiler-flags#-wshadow

jamiebramwell:
Adds the `-Wshadow` build flags to guard against possible future bugs. 

Resolves https://github.com/LLNL/serac/issues/348 .

### DIFF
--- a/cmake/SeracCompilerFlags.cmake
+++ b/cmake/SeracCompilerFlags.cmake
@@ -13,3 +13,6 @@ endif()
 
 # Need to add symbols to dynamic symtab in order to be visible from stacktraces
 string(APPEND CMAKE_EXE_LINKER_FLAGS " -rdynamic")
+
+# Enable warnings for overshadowed variable definitions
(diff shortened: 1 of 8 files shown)


jamiebramwell:
This found one real issue (a `t` member in the underlying mfem class) and a bunch of lambda ones. Here's a discussion on it: https://stackoverflow.com/questions/42088015/lambda-capture-and-parameter-with-same-name-who-shadows-the-other-clang-vs-g

white238 (approved):

jamiebramwell:
LGTM

joshessman-llnl (approved):
We should look into enabling `-Wall -Wextra` as well.  I think our use of designated initializers would prevent us from using `Wpedantic`, though.

PR #353, opened 2021-01-28
Consolidate the elasticity and nonlinear solid modules

### LINKED ISSUE #20: Improve material models
Add anisotropic generic hyperelastic materials to help with Seth Watts's project.

### LINKED ISSUE #104: Improve the integrators
We should re-evaluate the hand-rolled spatial integrators and make them more readable instead of adopting the MFEM style.

jamiebramwell:
This is a first attempt at cleaning up the solid mechanics module. It does the following:

Physics Improvements:

* Eliminates the extra elasticity physics module
* Rewrites the solid mechanics integrators to make more sense to mechanics experts
* Changes the dynamic option to use a displacement formulation
* Adds pressure and traction boundary conditions in both the deformed and reference configurations (Cauchy and Piola definitions). Note that the deformed configuration versions still use finite difference for the stiffness calculation.
* Added hooks to set the mass density to the solid mechanics module
* Adds a linear solid mechanics (elasticity) test

Material Improvements:

* Creates a new material directory with neo-Hookean and linear elastic material models
* Removes the legacy constant double material input option in favor of the more flexible coefficient interface

Input Changes:

* Adds a geometric nonlinearity toggle for residual and stiffness calculations
* Adds a material nonlinearity toggle for material evaluations (to be used for testing)
* Renames the solid mechanics input options to `solid` (instead of `nonlinear_solid`)
* Renames the solver option name from `stiffness_solver` to `equation_solver` in the solid and thermal conduction blocks

Code infrastructure:

* Adds a 4D array type for tangent moduli tensors as a placeholder for a real tensor class
* Moves the integrators to the physics directory
* Fixes the style in the thermal structural solver 
* Changes all the input file tests to output as their test name

Resolves https://github.com/LLNL/serac/issues/104.
Resolves https://github.com/LLNL/serac/issues/20.


jamiebramwell:
LGTM

data/input_files/tests/solid/qs_linear.lua
```diff
@@ -0,0 +1,70 @@
...
+    -- Turn the material nonlinearities off
+    -- TODO: this should be replaced with a proper material definition
+    material_nonlin = false,
+
+    -- boundary condition parameters
+    boundary_conds = {
+        ['displacement'] = {
+            -- boundary attribute 1 is fixed (Dirichlet)
+            attrs = {1},
+            vec_coef = function (v)
```
white238: ooo look at that sexy Vector class.. so elegant.

src/drivers/serac.cpp
```diff
@@ -98,11 +98,11 @@ int main(int argc, char* argv[])
   }
 
   // Define the solid solver object
-  auto                  solid_solver_options = inlet["nonlinear_solid"].get<serac::NonlinearSolid::InputOptions>();
-  serac::NonlinearSolid solid_solver(mesh, solid_solver_options);
+  auto         solid_solver_options = inlet["solid"].get<serac::Solid::InputOptions>();
+  serac::Solid solid_solver(mesh, solid_solver_options);
 
   // FIXME: Move time-scaling logic to Lua once arbitrary function signatures are allowed
```
white238: Pretty sure this can be fixed at this point.

src/serac/physics/integrators/CMakeLists.txt
```diff
@@ -6,17 +6,21 @@
...
 
-set(integrators_depends serac_infrastructure)
+set(integrators_depends
+    serac_infrastructure
+    serac_numerics
+    serac_materials
+    )
 
 blt_add_library(
     NAME        serac_integrators
```
white238: To be consistent we should rename this `serac_physics_integrators` like we have named `serac_physics_utilities`

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
```diff
@@ -0,0 +1,166 @@
...
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file inc_hyperelastic_integrator.hpp
+ *
+ * @brief The MFEM integrators for the incremental hyperelastic formulation
+ */
+
+#ifndef DISPLACEMENT_INTEGRATOR_HPP
```
white238: Didn't we switch to `#pragma once`?

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
```diff
@@ -0,0 +1,200 @@
...
+      el1.CalcDShape(eip, DSh_u_);
+      Mult(DSh_u_, J0i_, DS_u_);
+      MultAtB(PMatI_u_, DS_u_, J_);
+
+      CalcInverse(J_, Jinv_);
+
+      Jinv_.MultTranspose(nor_, fnor_);
+
+      el1.CalcShape(eip, shape_);
+      fnor_ *= pressure_.Eval(*Tr.Face, ip) * J_.Det();
```
jamiebramwell: - [x] scale by the pressure in the reference configuration case.

src/drivers/serac.cpp
joshessman-llnl: I can take care of this in a separate PR

tests/serac_newmark_test.cpp
```diff
@@ -10,14 +10,14 @@
 #include <memory>
 
 #include "serac/coefficients/coefficient_extensions.hpp"
-#include "../src/serac/integrators/wrapper_integrator.hpp"
+#include "../src/serac/physics/integrators/wrapper_integrator.hpp"
```
joshessman-llnl: Not strictly in the scope of this PR but this shouldn't have the `../src/` in it

tests/serac_thermal_solver.cpp
```diff
@@ -49,7 +49,7 @@ TEST_P(InputFileTest, thermal_conduction)
   MPI_Barrier(MPI_COMM_WORLD);
   std::string input_file_path =
       std::string(SERAC_REPO_DIR) + "/data/input_files/tests/thermal_conduction/" + GetParam() + ".lua";
-  test_utils::runModuleTest<ThermalConduction>(input_file_path);
+  test_utils::runModuleTest<ThermalConduction>(input_file_path, GetParam());
```
joshessman-llnl, line 39: 👍

src/drivers/serac.cpp
joshessman-llnl: Separately addressed in #363

tests/serac_newmark_test.cpp
jonwong12: Oops. 
```suggestion
#include "serac/physics/integrators/wrapper_integrator.hpp"
```

jonwong12 (approved):
Thanks the material and geometric toggles and the boundary work will be quite handy.

src/serac/physics/materials/hyperelastic_material.cpp
```diff
@@ -0,0 +1,175 @@
...
+        for (int l = 0; l < dim; ++l) {
+          if (i == j && k == l) {
+            C(i, j, k, l) += bulk_ - (2.0 / dim) * mu_;
+          }
+          if (i == k && j == l) {
+            C(i, j, k, l) += mu_;
+          }
+          if (i == l && j == k) {
+            C(i, j, k, l) += mu_;
+          }
```
samuelpmishLLNL: this can be done branchlessly by observing that `int(i==j)` is the same as `kronecker_delta(i,j)`, so you get
```
C(i,j,k,l) += lambda * (i==j) * (k==l) + mu * ((i==k) * (j==l)  + (i==l) * (j==k));
```

src/serac/physics/materials/hyperelastic_material.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+  int dim = F.Width();
+  B_.SetSize(dim);
+  sigma.SetSize(dim);
+
+  if (c_mu_) {
+    EvalCoeffs();
+  }
+
+  // See http://solidmechanics.org/Text/Chapter3_5/Chapter3_5.php for a sample derivation
```
samuelpmishLLNL, line 49: links are good, but it would be good to have our own description of the theory in the future

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
```diff
@@ -65,24 +67,97 @@ class IncrementalHyperelasticIntegrator : public mfem::NonlinearFormIntegrator {
...
+   * @brief The displacement gradient
+   *
+   */
+  mfem::DenseMatrix H_;
+
+  /**
+   * @brief The determinant of the deformation gradient
+   *
+   */
+  double J_;
```
samuelpmishLLNL: there's some naming inconsistencies: here `J` is just the determinant, but in other places it's the entire matrix

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
```diff
@@ -65,24 +67,97 @@ class IncrementalHyperelasticIntegrator : public mfem::NonlinearFormIntegrator {
...
+   * @brief gradients of reference shape functions (dof x dim).
+   *
+   */
+  mfem::DenseMatrix DSh_;
+
+  /**
+   * @brief gradients of the shape functions in the target configuration (DN_i, DX_j)
+   *
+   */
+  mfem::DenseMatrix DS_;
```
samuelpmishLLNL: I'm used to "reference" being the undeformed mesh, but it looks like it might be referring to the master/parent element here?

I know mfem uses names like `DSh` and `DS` but those are pretty opaque. I'm trying out `dN_dx` for shape function gradient w.r.t. current configuration and `dN_dX` for shape function gradient w.r.t. reference configuration, and `dN_dxi` w.r.t. master/parent element in the variationalform branch.

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
```diff
@@ -65,24 +67,97 @@ class IncrementalHyperelasticIntegrator : public mfem::NonlinearFormIntegrator {
...
+   * @brief the Cauchy stress
+   *
+   */
+  mfem::DenseMatrix sigma_;
+
+  /**
+   * @brief Current input state dofs (dof x dim)
+   *
+   */
+  mfem::DenseMatrix PMatI_;
```
samuelpmishLLNL: these are nodal displacements? what does the name `PMatI_` mean?

samuelpmishLLNL: I'm assuming `dofs` is the number of dofs per element, so it's an element-local thing

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
```diff
@@ -0,0 +1,201 @@
...
+      Tr.Elem1->SetIntPoint(&eip);
+      CalcInverse(Tr.Elem1->Jacobian(), J0i_);
+
+      el1.CalcDShape(eip, DSh_u_);
+      Mult(DSh_u_, J0i_, DS_u_);
+      MultAtB(PMatI_u_, DS_u_, F_);
+
+      for (int d = 0; d < dim; d++) {
+        F_(d, d) += 1.0;
+      }
```
samuelpmishLLNL: might be less of an issue for finite deformation, but there can a considerable loss of precision by actually adding 1.0 to the diagonal here. We should be able to calculate any strain measures from `(F - I)` directly, avoiding any loss of precision.

src/serac/physics/integrators/hyperelastic_traction_integrator.hpp
```diff
@@ -0,0 +1,173 @@
...
+  mutable mfem::Vector Sh_p_;
+  mutable mfem::Vector Sh_u_;
+
+  bool compute_on_reference_;
+};
+
+/**
+ * @brief Custom MFEM integrator for pressure loads
+ */
+class HyperelasticPressureIntegrator : public mfem::NonlinearFormIntegrator {
```
samuelpmishLLNL: what does "Hyperelastic" mean in the context of tractions/pressures?

data/input_files/default.lua
```diff
@@ -12,7 +12,7 @@ main_mesh = {
 }
 
 -- Solver parameters
-nonlinear_solid = {
+solid = {
     stiffness_solver = {
```
solberg2: So, when one is doing dynamics, isn't the "stiffness solver" really solving for a combination of mass+stiffness?

Maybe.  "solid mechanics solver" or simply "linear algebra" or "LHS_solver"

data/input_files/tests/solid/qs_linear.lua
```diff
@@ -0,0 +1,70 @@
...
+
+-- Simulation time parameters
+dt      = 1.0
+
+main_mesh = {
+    type = "file",
+    -- mesh file
+    mesh = "../../../meshes/beam-hex.mesh",
+    -- serial and parallel refinement levels
+    ser_ref_levels = 1,
```
solberg2, line 13: what happens if these are both non-zero?

data/input_files/tests/solid/qs_linear.lua
```diff
@@ -0,0 +1,70 @@
...
+            max_iter    = 5000,
+            print_level = 1,
+        },
+    },
+
+    -- polynomial interpolation order
+    order = 1,
+
+    -- neo-Hookean material parameters
+    mu = 0.25,
```
solberg2, line 43: Presumably when we actually have a "material model module" this will get fleshed out...

src/serac/physics/integrators/hyperelastic_traction_integrator.hpp
solberg2: Sam makes a good point.  I think the only two possibilities for the RHS are "linear (normals/areas fixed) and nonlinear".  In the latter case one can specify variations on the tangent, whether you want to include it at all (noting that the consistent version is nonsymmetric)

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
solberg2: Actually, it turns out it is best to calculate du/dx and add I to it than to calculate dx/DX

When u is small and x is large, then x~X so dx/DX looks like I, but you could still have du/dX large

src/serac/physics/materials/hyperelastic_material.cpp
samuelpmishLLNL: it's interesting to look at the generated assembly for different ways of populating `C`, here
https://godbolt.org/z/1oW674

1. your original approach generates a lot of branching for control flow and accumulation
2. the snippet above has a few branches for loop control flow (I think), but is pretty well inlined
3. a for_constexpr produces zero branching and 81 assignments, as the loop indices are evaluated at compile time
4. for_constexpr + `if constexpr` has zero branching and actually only sparsely updates the nonzero components!

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
```diff
@@ -0,0 +1,201 @@
...
+        if (!compute_on_reference_) {
+          residual_contribution *= -1.0 * F_.Det() * fnor.Norml2();
+        }
+        elvec(dof * k + j) += residual_contribution;
+      }
+    }
+  }
+}
+
+void HyperelasticTractionIntegrator::AssembleFaceGrad(const mfem::FiniteElement& el1, const mfem::FiniteElement& el2,
```
solberg2: Probably there should be a class SolidMechanics where the TractionIntegrator can be a subclass, as well as Hyperelastic and Hypoelastic (note MSLib would use a different version of Hypoelasticity than e.g. MDG material models).  One would mix-and-match material models but they should all use the same traction integrator

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+      Finv_.MultTranspose(nor, fnor);
+    }
+
+    el1.CalcShape(eip, shape_);
+    for (int j = 0; j < dof; j++) {
+      for (int k = 0; k < dim; k++) {
+        double residual_contribution = trac(k) * shape_(j) * ip.weight * Tr.Face->Weight();
+        if (!compute_on_reference_) {
+          residual_contribution *= -1.0 * F_.Det() * fnor.Norml2();
```
solberg2: Should get rid of the -1.0

src/serac/physics/materials/hyperelastic_material.cpp
solberg2: isn't better really to use the 6x6 form?

jamiebramwell: Yes - I originally did this but had a typo somewhere so I reverted to a way that was easier to eyeball debug.

src/serac/physics/materials/hyperelastic_material.cpp
jamiebramwell: Agreed. I have a ticket on this here: https://github.com/LLNL/serac/issues/364

src/serac/physics/materials/hyperelastic_material.cpp
jamiebramwell: > 
> 
> it's interesting to look at the generated assembly for different ways of populating `C`, here
> https://godbolt.org/z/1oW674
> 
>     1. your original approach generates a lot of branching for control flow and accumulation
> 
>     2. the snippet above has a few branches for loop control flow (I think), but is pretty well inlined
> 
>     3. a for_constexpr produces zero branching and 81 assignments, as the loop indices are evaluated at compile time
> 
>     4. for_constexpr + `if constexpr` has zero branching and actually only sparsely updates the nonzero components!

Awesome! I figured there were a bunch of ways to optimize this really dumb implementation.

data/input_files/tests/solid/qs_linear.lua
jamiebramwell: It first refines the mesh on processor 0 `ser_ref_levels` times and then distributes the mesh. It then refines the mesh `par_ref_levels` times in a fully distributed fashion.

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
jamiebramwell: Yeah, I think I kept the material interface with the deformation gradient instead of the displacement gradient (F-I) to be consistent with the existing MFEM material models. @solberg2 , what do the material models in your codes use?

src/serac/physics/integrators/hyperelastic_traction_integrator.hpp
jamiebramwell: It means nothing other than I developed them at the same time.  😄 I'll remove it.

data/input_files/default.lua
jamiebramwell: Yes. I remember thinking about this when it initially went in, and we couldn't think of a good name. Maybe `equation_solver` to match the internal class?

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
jamiebramwell: If I recall, there's a weird sign factor in one of the MFEM linear algebra routines. I'll write a test to verify this.

data/input_files/tests/solid/qs_linear.lua
jamiebramwell: Sometimes you need to do this if you have a very coarse mesh and the number of initial elements < number of MPI ranks.

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
solberg2: So what we do is Calculate du/dX and then calculate F = du/dX + I.  That's more accurate than simply calculating dx/DX
Also it means you still have du/dX lying around if you want it.  For hypoelastic materials you want du/dx of course, but with x calculated at the midpoint configuration.

data/input_files/default.lua
solberg2: Yeah equation_solver is a better name, because it is exactly what it is

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
solberg2: We usually call it DetJ

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
samuelpmish: I'm just observing that you rarely ever _need_ F. When u is small, actually forming F will always result in a loss of precision. If a constitutive model needs a strain measure, we can just calculate those directly from du_dx without losing precision by forming F as an intermediate (even though most strain measures are typically expressed most succinctly in terms of F, with exact arithmetic).

solberg2: Yeah.  That's why we essentially form both and pass them down to the material model.  Hypoelastic models based on the Green-Naghdi formulation need F_{n+1} and F_{n} because they calculate R_{n+1} and R_{n} so they can rotate back and forth to the reference configuration

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
jamiebramwell: I'm a bit worried about having both `dN_dx` and `dN_dX` as variables since the only difference is case, but I agree that `DS` and `DSh` are not descriptive so I'll change those.

src/serac/physics/materials/hyperelastic_material.hpp
```diff
@@ -0,0 +1,293 @@
...
+  /**
+   * @brief Evaluate the strain energy density function, W = W(F).
+   *
+   * @param[in] F The deformation gradient
+   * @return double Strain energy density
+   */
+  virtual double EvalStrainEnergy(const mfem::DenseMatrix& F) const = 0;
+
+  /**
+   * @brief Evaluate the Cauchy stress sigma = sigma(F).
```
sethwatts, line 55: In some cases, say for anisotropic response, it might be more natural to define the stress in the reference configuration, i.e. to provide the PK1 stress rather than the Cauchy. Returning the PK1 stress would also make this function more consistent with the `AssembleTangentModuli` method. I understand there might be other cases where using the Cauchy stress is more natural. Could the `HyperelasticMaterial` class provide methods to map between these two stress tensors (and possibly others)?

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
jamiebramwell: Yes - this old form of the nonlinear form integrator is called on an element-by-element basis.

src/serac/physics/materials/hyperelastic_material.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+  /**
+   * @brief Evaluate the derivative of the 1st Piola-Kirchhoff stress tensor in Voigt notation
+   * and assemble its contribution to the local gradient matrix 'A'.
+
+   * @param[in] F The deformation gradient
+   * @param[out] C Tangent moduli 4D Array in spatial form (C^e_ijkl=(d tau_ij)/(d F_km) * F_lm = J * sigma_ij delta_kl
+   + J * (d sigma_ij)/(d F_km) F_lm )
+   */
+  virtual void AssembleTangentModuli(const mfem::DenseMatrix& F, mfem_ext::Array4D<double>& C) const = 0;
```
sethwatts: This name looks like it might be a carryover from MFEM. Since this function defines the tangent stiffness but doesn't integrate it like the MFEM version makes you do, what about `EvalTangentModuli` or `EvalTangentStiffness` (my preference)?

data/input_files/default.lua
jamiebramwell: Ahhh I remember the problem. We don't use a lumped mass matrix, so we can also specify a mass matrix solver for explicit problems.

src/serac/physics/materials/hyperelastic_material.hpp
sethwatts: Related to the above and to the discussion [here](https://github.com/LLNL/serac/pull/353/files#r570538970), would it be possible to provide efficient and accurate "building blocks" like F, C, E, etc. that could be used to implement different material models?

src/serac/physics/materials/hyperelastic_material.cpp
sethwatts: > isn't better really to use the 6x6 form?

I'm sure it is more efficient from a computational standpoint, but Voigt notation always seems to cause bugs due to the factor of 2 on the shear terms. I'd like to avoid it, if possible.

src/serac/physics/solid.hpp
```diff
@@ -138,12 +170,13 @@ class NonlinearSolid : public BasePhysics {
...
+   * @brief Set the material parameters
    *
-   * @param[in] mu Set the mu Lame parameter for the hyperelastic solid
-   * @param[in] K Set the K Lame parameter for the hyperelastic solid
+   * @param[in] mu Set the shear modulus for the solid
+   * @param[in] K Set the bulk modulus for the solid
+   * @param[in] material_nonlin Flag to include material nonlinearities (linear elastic vs. neo-Hookean model)
    */
-  void setHyperelasticMaterialParameters(double mu, double K);
+  void setMaterialParameters(double mu, double K, bool material_nonlin = true);
```
sethwatts: No complaints about this for now, and I assume this will be resolved when material models are fleshed out (as per Jerome's [comment](https://github.com/LLNL/serac/pull/353/files#r570553189)), but I'm curious what the plan is for handling general material models. Something like a `std::vector<double>` of properties that each model can parse, some type of material property container specialized for each model, something else?

src/serac/physics/thermal_solid.hpp
```diff
@@ -63,21 +63,21 @@ class ThermalSolid : public BasePhysics {
...
-  void SetConductivity(std::unique_ptr<mfem::Coefficient>&& kappa) { therm_solver_.setConductivity(std::move(kappa)); };
+  void setConductivity(std::unique_ptr<mfem::Coefficient>&& kappa) { therm_solver_.setConductivity(std::move(kappa)); };
 
   /**
    * @brief Set the density
    *
    * @param[in] rho The density coefficient
    */
-  void SetDensity(std::unique_ptr<mfem::Coefficient>&& rho) { therm_solver_.setDensity(std::move(rho)); };
+  void setDensity(std::unique_ptr<mfem::Coefficient>&& rho) { therm_solver_.setDensity(std::move(rho)); };
```
sethwatts: Would you consider "mass density" instead of "density"? I could see nomenclature collisions with topology optimization (which to be extra confusing also uses rho), although within LiDO I'm trying to either call it design density or local volume fraction.

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
solberg2: So what Diablo does is create different "kinematics" modules

   Hyperelastic
   Hypoelastic
   
 That's kinda of a low-level approach of what we do with "fields", where we let each element level request which fields it wants
 
 To some extent I think what would make sense is to have a kinematics module that has the routines to calculate a bunch of different kinematic variables  (F_{n+1}, du/dx, R_{n+1}, )

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
solberg2: So what we do is 

dN_dx0, dN_dxn, dN_dxnp1, dudXnphalf, etc.

src/serac/physics/thermal_solid.hpp
sethwatts, line 73: I'm a little surprised that this method isn't defined in the `Solid` class.

data/input_files/default.lua
jamiebramwell: However, they're not separate options after @samuelpmishLLNL 's ODE work, so I'll change it to `equation_solver`.

src/serac/physics/solid.hpp
jamiebramwell: This will come in @samuelpmishLLNL 's upcoming overhaul. I imagine it will look something like a specialized parameter packet for each material model, but he can speak to it better than I can. Stay tuned!

src/serac/physics/thermal_solid.hpp
jamiebramwell: Sounds reasonable to me!

src/serac/physics/materials/hyperelastic_material.hpp
jamiebramwell: I'll add a kinematic helper function file to contain these transformations.

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
jamiebramwell: Maybe put it in the `Solid` namespace? it's not really inheriting from anything (other than the appropriate MFEM class), but showing this relation is a good idea.

solberg2: Yes that is a good namespace for that

src/serac/physics/materials/hyperelastic_material.hpp
sethwatts: It occurred to me that a cleaner way to address the stress definition might be to generalize the `compute_on_reference` `bool`s used in e.g. the `TractionIntegrator`. The `EvalStress` method could take two additional arguments, `bool area_on_reference = false, bool force_on_reference = false`. This would let the default return be the Cauchy stress, but allow an implementation to use the PK1 or PK2 stress if that makes more sense; depending on the value of the bools, the returned stress could be transformed into whatever stress measure the solver actually uses. This would also mean someone implementing something other than the Cauchy stress would not need to also implement the transforms themselves, removing a possible source of bugs.

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
jamiebramwell: I think this approach will be useful for @samuelpmishLLNL 's upcoming rewrite.

src/serac/physics/integrators/hyperelastic_traction_integrator.cpp
jamiebramwell: Yep! You were right. Thanks for the catch!

src/serac/physics/materials/hyperelastic_material.hpp
jamiebramwell: I worry about having a list of bools to define behavior in a function call as it is difficult to tell what `foo(bar, false, true)` does at a glance. An easier method is probably having some kind of annotation on the return type. Any thoughts @white238  @joshessman-llnl ?

sethwatts: That's a fair point. Another possibility would be an enumeration of configurations, with default values for `EvalStress`. I think `foo(bar, ReferenceConfig, CurrentConfig)` would at least be easier to parse than the case with `bool`s. Annotating the return type would make it even easier to parse since you wouldn't have to remember the argument ordering. On the other hand, the `enum` could also be in the signature for defining BCs, which might improve their readability. Any solution works for me.

joshessman-llnl: If we introduce two bools/enum flags to the user interface for both the area and force, then this opens up the possibility for the user to request area in terms of the current config and force in terms of the reference config - is that something we'd want to allow?  

If not, maybe it'd be preferable to use an `enum StressType { Cauchy, PK1, PK2 };` in the interface and translate it into `area_on_ref`, `force_on_ref` internally if needed.

jamiebramwell: After talking with @solberg2 and discussing this at the meeting, the consensus was we should keep the Cauchy stress interface for now as it is easier to debug issues at a material level than an assembly level. With @samuelpmishLLNL 's upcoming rewrite, adding options like this will be much more feasible.

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+  mfem::Vector temp_out_1;
+  mfem::Vector temp_out_2;
+  mfem::Vector temp(input_state_vector.GetData(), input_state_vector.Size());
+
+  stiffness_matrix.SetSize(input_state_vector.Size(), input_state_vector.Size());
+
+  stiffness_matrix = 0.0;
+
+  // If computing on the deformed configuration calculate the stiffness contributions via finite difference
+  // Otherwise, the contributuion is zero
```
samuelpmishLLNL:
```suggestion
  // Otherwise, the contribution is zero
```

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+  if (!compute_on_reference_) {
```
samuelpmishLLNL, line 237: Is this code called in a loop over face elements? If so, is it possible to perform this conditional outside the loop instead?

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+    for (int j = 0; j < temp.Size(); j++) {
+      temp[j] += diff_step;
+      AssembleFaceVector(element_1, element_2, parent_to_reference_face_transformation, temp, temp_out_1);
+      temp[j] -= 2.0 * diff_step;
+      AssembleFaceVector(element_1, element_2, parent_to_reference_face_transformation, temp, temp_out_2);
+
```
samuelpmishLLNL, line 243: what does element_2 do, out of curiosity?

jamiebramwell: It is typically the other element connected to the face that the integrator is being applied to. For example, it is the opposite volumetric element when calculating interior face contributions in DG methods. However, since these integrators are only called on boundary faces as they are associated with boundary conditions, we never use it.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: Yes - but it would involve making a different integrator for the reference configuration application case.

src/serac/physics/integrators/traction_integrator.hpp
```diff
@@ -0,0 +1,255 @@
...
+  virtual ~TractionIntegrator() {}
+
+private:
+  /**
+   * @brief The vector coefficient for the traction load
+   */
+  mfem::VectorCoefficient& traction_;
+
+  /**
+   * @brief gradients of shape functions on the parent element (dof x dim).
```
samuelpmishLLNL, line 76: ah, much better

src/serac/physics/integrators/traction_integrator.hpp
```diff
@@ -0,0 +1,255 @@
...
+   * @brief The normal vector in the reference configuration
+   *
+   */
+  mutable mfem::Vector nor_;
+
+  /**
+   * @brief The normal vector in the deformed configuration
+   *
+   */
+  mutable mfem::Vector fnor_;
```
samuelpmishLLNL: the other names use Capital Symbols for reference configuration and lower case for current configuration. `N` and `n` for normals, respectively, might be more consistent with that notation, but are overused symbols.

src/serac/physics/integrators/traction_integrator.cpp
samuelpmishLLNL: sorry, I didn't see the doxygen comments until after writing my question!

src/serac/physics/materials/hyperelastic_material.hpp
```diff
@@ -0,0 +1,308 @@
...
+   * @param[in] du_dX the displacement gradient (du_dX)
+   * @param[out] F the deformation gradient (dx_dX)
+   */
+  static void calcDeformationGradient(const mfem::DenseMatrix& du_dX, mfem::DenseMatrix& F);
+
+  /**
+   * @brief Calculate the linearized strain tensor (epsilon = 1/2 * (du_dX + du_dX^T))
+   *
+   * @param[in] du_dX the displacement gradient (du_dX)
+   * @param[out] epsilon the linearized strain tensor epsilon = 1/2 * (du_dX + du_dX^T)
```
samuelpmishLLNL: these are useful conversions, but aren't really specific to Hyperelasticity. To your point earlier, I wonder where we should put things like this. Maybe `solid_mechanics::calcDeformationGradient`?

samuelpmishLLNL (commented):
There's quite a bit of repeated code between the PressureIntegrator and TractionIntegrator objects, we might be able to consolidate them later on to deduplicate.

samuelpmishLLNL:
> There's quite a bit of repeated code between the PressureIntegrator and TractionIntegrator objects, we might be able to consolidate them later on to deduplicate.

That being said, I think that's beyond the scope of this PR!

src/serac/physics/integrators/traction_integrator.hpp
jamiebramwell: I'm hesitant to distinguish variables by case, but I agree these names can be better.

samuelpmishLLNL (approved):

src/serac/physics/materials/hyperelastic_material.hpp
jamiebramwell: Yeah, I needed to put them here instead of the solid class to avoid a circular dependency. Any ideas @joshessman-llnl @white238 ?

jamiebramwell:
> 
> 
> > There's quite a bit of repeated code between the PressureIntegrator and TractionIntegrator objects, we might be able to consolidate them later on to deduplicate.
> 
> That being said, I think that's beyond the scope of this PR!

Yeah, I'm hesitant to code smith this much more as I expect all of this to be thrown out when the `weak_form` comes online. This might be a project while I'm bored in a meeting. 😄

src/serac/physics/materials/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+   SOURCES     ${materials_sources}
+   DEPENDS_ON  ${materials_depends}
+   )
+
+install(FILES ${materials_headers} DESTINATION include/serac/physics/materials )
+
+install(TARGETS              serac_physics_materials
+        EXPORT               serac-targets
+        DESTINATION          lib
+        )
```
white238: new line here

src/serac/physics/materials/hyperelastic_material.hpp
white238: Another directory for useful calculations/conversions? physics_utilities?

jamiebramwell: Then the problem is how do we connect them with a specific physics. In other words, we have the `serac::Solid` class, so we would use the `serac::solid` namespace. Alternatively we can lump all the physics together and use `serac::physics_utils`.

jamiebramwell: (I tried to add them as static members to the `Solid` class, but that's when I noticed the circular dependency.)

jamiebramwell:
I believe this is ready for a round 2 pass. All feedback is appreciated!

solberg2:
I’ll work on it.  I’ll also take a look at your theory stuff

jamiebramwell:
LGTM

jamiebramwell:
LGTM

jamiebramwell:
LGTM

src/serac/physics/materials/hyperelastic_material.hpp
jamiebramwell: I ended up putting them in a `solid_util` namespace. Suggestions welcome!

jamiebramwell:
LGTM

jamiebramwell:
LGTM

src/serac/physics/integrators/displacement_hyperelastic_integrator.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include "serac/physics/integrators/displacement_hyperelastic_integrator.hpp"
+
+#include "serac/infrastructure/profiling.hpp"
+#include "serac/numerics/expr_template_ops.hpp"
+#include "serac/numerics/array_4D.hpp"
+#include "serac/physics/utilities/physics_utils.hpp"
+
+namespace serac::mfem_ext {
+
+void DisplacementHyperelasticIntegrator::CalcKinematics(const mfem::FiniteElement&    element,
```
solberg2, line 16: All these comments may be useless as they will be OBE when Sam's changes go in.

But probably the better way to do the classes here is:

displacementsolidintegrator:hyperelastickinematics

Then you can mix material models with different kinematics within the same problem

src/serac/physics/integrators/displacement_hyperelastic_integrator.cpp
```diff
@@ -0,0 +1,217 @@
...
+    mfem::Mult(dN_dX_, Finv_, B_);
+    det_J_ = F_.Det();
+  } else {
+    // If not, we integrate on the undeformed stress-free configuration
+    B_     = dN_dX_;
+    det_J_ = 1.0;
+  }
+}
+
+double DisplacementHyperelasticIntegrator::GetElementEnergy(
```
solberg2, line 50: Again, probably better to say "GetHyperelasticElementEnergy", then you could mix material models in the same mesh, by using a mask

src/serac/physics/integrators/displacement_hyperelastic_integrator.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+    // Calculate the deformation gradent and accumulate the strain energy at the current integration point
+    CalcKinematics(element, int_point, parent_to_reference_transformation);
+    energy += det_J_ * int_point.weight * parent_to_reference_transformation.Weight() * material_.evalStrainEnergy(F_);
+  }
+
+  return energy;
+}
+
+void DisplacementHyperelasticIntegrator::AssembleElementVector(
```
solberg2, line 84: I guess if you had a mask, there are a number of levels you could operate the mask at, so one way would just to "run two top-level integrators", in which case my comments about turning "DisplacementHyperelasticIntegrator::CalcKinematics" into "displacementsolidmechanicsintegrator::CalcHyperelasticKinematics" might not be relevant.  Again probably OBE by whatever Sam comes up with

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    // Calculate the shape functions
+    element_1.CalcShape(eip, shape_);
+    for (int j = 0; j < dof; j++) {
+      for (int k = 0; k < dim; k++) {
+        double residual_contribution =
+            -1.0 * traction_vector_(k) * shape_(j) * ip.weight * parent_to_reference_face_transformation.Face->Weight();
+        if (!compute_on_reference_) {
+          residual_contribution *= F_.Det() * current_normal_.Norml2();
```
solberg2, line 93: If I am correct, there is no residual output if compute_on_reference=true

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+      // Calculate F inverse time the normal vector
+      Finv_.MultTranspose(reference_normal_, current_normal_);
+    }
+
+    // Calculate the shape functions
+    element_1.CalcShape(eip, shape_);
+    for (int j = 0; j < dof; j++) {
+      for (int k = 0; k < dim; k++) {
+        double residual_contribution =
+            -1.0 * traction_vector_(k) * shape_(j) * ip.weight * parent_to_reference_face_transformation.Face->Weight();
```
solberg2, line 91: Do you want the -1.0 here?

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "serac/physics/integrators/traction_integrator.hpp"
+
+namespace serac::mfem_ext {
+
+void TractionIntegrator::AssembleFaceVector(const mfem::FiniteElement&        element_1, const mfem::FiniteElement&,
```
solberg2, line 12: Another way to do this would be "AssembleFaceResidual" and "AssembleFaceTangent", which might be better going forward

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+  mfem::Vector temp_out_1;
+  mfem::Vector temp_out_2;
+  mfem::Vector temp(input_state_vector.GetData(), input_state_vector.Size());
+
+  stiffness_matrix.SetSize(input_state_vector.Size(), input_state_vector.Size());
+  stiffness_matrix = 0.0;
+
+  // If computing on the deformed configuration calculate the stiffness contributions via finite difference
+  // Otherwise, the contributuion is zero
+  if (!compute_on_reference_) {
```
solberg2, line 117: Finite Difference!!!!  :-(

Also, this is going to be unsymmetric.  In our codes if we detect a symmetric solver we would symmetrize it.

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+      for (int k = 0; k < temp.Size(); k++) {
+        stiffness_matrix(k, j) = (temp_out_1[k] - temp_out_2[k]) / (2.0 * diff_step);
+      }
+      temp[j] = input_state_vector[j];
+    }
+  }
+}
+
+void PressureIntegrator::AssembleFaceVector(const mfem::FiniteElement&        element_1, const mfem::FiniteElement&,
```
solberg2, line 132: again, maybe better "AssembleFaceResidual"

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+  // Otherwise, the contributuion is zero
+  if (!compute_on_reference_) {
+    for (int j = 0; j < temp.Size(); j++) {
+      temp[j] += diff_step;
+      AssembleFaceVector(element_1, element_2, parent_to_reference_face_transformation, temp, temp_out_1);
+      temp[j] -= 2.0 * diff_step;
+      AssembleFaceVector(element_1, element_2, parent_to_reference_face_transformation, temp, temp_out_2);
+
+      for (int k = 0; k < temp.Size(); k++) {
+        stiffness_matrix(k, j) = (temp_out_1[k] - temp_out_2[k]) / (2.0 * diff_step);
```
solberg2, line 125: The routine is called "facegrad" but here is the stiffness matrix.  But maybe better call it the tangent matrix

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+    } else {
+      // Calculate the deformation gradient and its inverse
+      parent_to_reference_face_transformation.Elem1->SetIntPoint(&eip);
+      CalcInverse(parent_to_reference_face_transformation.Elem1->Jacobian(), dxi_dX_);
+      element_1.CalcDShape(eip, dN_dxi_);
+      Mult(dN_dxi_, dxi_dX_, dN_dX_);
+      MultAtB(input_state_matrix_, dN_dX_, F_);
+
+      for (int d = 0; d < dim; d++) {
+        F_(d, d) += 1.0;
```
solberg2: Two bad there isn't a way to do this without a loop.  But maybe GPUs like loops

src/serac/physics/integrators/traction_integrator.hpp
```diff
@@ -0,0 +1,255 @@
...
+   * @param[in] compute_on_reference Flag for computing the residual in the stress-free configuration vs. the deformed
+   * configuration
+   */
+  explicit TractionIntegrator(mfem::VectorCoefficient& f, bool compute_on_reference = false)
+      : traction_(f), compute_on_reference_(compute_on_reference)
+  {
+  }
+
+  /**
+   * @brief Assemble the nonlinear residual on a boundary element at a current state
```
solberg2: "boundary facet" is maybe better than "boundary element", because the latter means something a bit different

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+      }
+
+      CalcInverse(F_, Finv_);
+
+      Finv_.MultTranspose(reference_normal_, current_normal_);
+      current_normal_ *= F_.Det();
+    }
+
+    element_1.CalcShape(eip, shape_);
+    current_normal_ *= pressure_.Eval(*parent_to_reference_face_transformation.Face, ip);
```
solberg2, line 208: Good to have a comment here that "for !calc_on_reference current_normal has Nanson's formula applied and is therefore not normalized

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+    for (int j = 0; j < dof; j++) {
+      for (int k = 0; k < dim; k++) {
+        output_residual_vector(dof * k + j) +=
+            ip.weight * parent_to_reference_face_transformation.Face->Weight() * current_normal_(k) * shape_(j);
```
solberg2, line 214: Good to have a comment "current_normal has area transformation and includes the applied pressure, and is therefore not normalized"

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+  mfem::Vector temp_out_2;
+  mfem::Vector temp(input_state_vector.GetData(), input_state_vector.Size());
+
+  stiffness_matrix.SetSize(input_state_vector.Size(), input_state_vector.Size());
+
+  stiffness_matrix = 0.0;
+
+  // If computing on the deformed configuration calculate the stiffness contributions via finite difference
+  // Otherwise, the contribution is zero
+  if (!compute_on_reference_) {
```
solberg2, line 237: Finite Difference!!!

src/serac/physics/integrators/traction_integrator.hpp
```diff
@@ -0,0 +1,255 @@
...
+   *
+   * @param[in] element_1 The first element attached to the face
+   * @param[in] element_2 The second element attached to the face
```
solberg2, line 40: Why are there two elements?

src/serac/physics/materials/hyperelastic_material.hpp
solberg2: Realistically, any material that can be written in terms of P_K2, etc. could be calculated as such and then transformed "at the end".  You could even output the P_K2 stresses and strains as plot/history variables for debugging.  But best to communicate to the rest of the code in a consistent manner (e.g. the Cauchy Stress)

src/serac/physics/integrators/traction_integrator.cpp
```diff
@@ -0,0 +1,226 @@
...
+      current_normal_ = reference_normal_;
+    } else {
+      // Calculate the deformation gradient and its inverse
+      parent_to_reference_face_transformation.Elem1->SetIntPoint(&eip);
+      CalcInverse(parent_to_reference_face_transformation.Elem1->Jacobian(), dxi_dX_);
+      element_1.CalcDShape(eip, dN_dxi_);
+      Mult(dN_dxi_, dxi_dX_, dN_dX_);
+      MultAtB(input_state_matrix_, dN_dX_, F_);
+
+      for (int d = 0; d < dim; d++) {
```
solberg2: There is a utility for this, right?

solberg2 (approved):

src/serac/physics/integrators/displacement_hyperelastic_integrator.cpp
jamiebramwell: I called it `HyperelasticIntegrator` as it can only use hyperelastic materials at the moment. I think your naming scheme is better though.

src/serac/physics/integrators/displacement_hyperelastic_integrator.cpp
jamiebramwell: This name is unfortunately fixed as it plugs into the MFEM inheritance structure. This is partially why I tagged this integrator as specific to hyperelasticity.

src/serac/physics/integrators/displacement_hyperelastic_integrator.cpp
jamiebramwell: Agreed. That's probably how we would do it if we didn't have the incoming weak form rewrite.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: If `compute_on_reference = false`, it still accumulates the traction at line 82.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: Yes - this is due to a sign factor the MFEM assembly routine. This is the contribution to the residual (R(u) = K(u) + t(u) = 0) vs the RHS (K(u) = t(u)).

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: This is another MFEM-ism that we unfortunately inherited through the NonlinearForm/NonlinearIntegrator class.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: I have been properly shamed. I will go work out an analytic solution.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: Same comment about MFEM integrators (unfortunately).

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: MFEM-ism.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: Stuff like this will go away when we have a real tensor class (with @samuelpmishLLNL 's update)

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: Once again, I am properly shamed. I will add an analytic derivative.

src/serac/physics/integrators/traction_integrator.cpp
solberg2: Oh yes.  Probably better to put the if statement at the top instead of all the way down at the bottom, for the GPUS

jamiebramwell: Yeah, this implementation won't work at all on GPUs as it heavily relies on virtual function calls. To avoid branching, I would really need to have separate integrators for both cases as these are called on an element-by-element basis.

src/serac/physics/integrators/traction_integrator.hpp
jamiebramwell: This is an MFEM-ism because these face integrators are also used for DG methods. the second element is the optional other element attached to this face if it is calculated on the interior of the domain. It is MFEM convention that this element is null on the boundary.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: Yes! I missed this one when I put it in. Thanks for the catch!

solberg2:
I didn't see where the viscoelastic stuff is done, but, I wanted to point out that it shoud be like \neta *C, where C is the elasticity tensor and \neta has units of time.

src/serac/physics/integrators/traction_integrator.cpp
jamiebramwell: As we discussed in a meeting, I'll add a disclaimer to the code that this is a stop gap.

jamiebramwell:
> 
> 
> I didn't see where the viscoelastic stuff is done, but, I wanted to point out that it shoud be like \neta *C, where C is the elasticity tensor and \neta has units of time.

It's added here: https://github.com/LLNL/serac/blob/853ee20670ba017c776aeb0c7ce5150f9a9a32a8/src/serac/physics/solid.cpp#L246-L251  Since we are reusing the standard MFEM integrator, it doesn't have the elasticity tensor included. I'll make an issue so we don't forget to do this in the rewrite.

jamiebramwell:
LGTM

jamiebramwell:
@white238 @joshessman-llnl , if one of you could do a last check for CS issues, I think this is good to go! 🎉

data/input_files/tests/solid/qs_linear.lua
white238: Is this documented somewhere?

src/serac/physics/materials/hyperelastic_material.hpp
```diff
@@ -0,0 +1,277 @@
...
+
+   * @param[in] du_dX the displacement gradient
+   * @param[out] C Tangent moduli 4D Array in spatial form (C^e_ijkl=(d tau_ij)/(d F_km) * F_lm = J * sigma_ij delta_kl
+   + J * (d sigma_ij)/(d F_km) F_lm )
+   */
+  virtual void evalTangentStiffness(const mfem::DenseMatrix& du_dX, serac::mfem_ext::Array4D<double>& C) const = 0;
+
+protected:
+  /**
+   * @brief Reference element to stree-free configuration (target) transformation
```
joshessman-llnl: Can this be annotated as non-owning?

src/serac/physics/solid.cpp
```diff
@@ -303,9 +348,9 @@ void NonlinearSolid::advanceTimestep(double& dt)
   cycle_ += 1;
 }
 
-NonlinearSolid::~NonlinearSolid() {}
+Solid::~Solid() {}
```
joshessman-llnl: This can just be `~Solid() = default;` in the header

tests/serac_newmark_test.cpp
```diff
@@ -10,14 +10,14 @@
 #include <memory>
 
 #include "serac/coefficients/coefficient_extensions.hpp"
-#include "../src/serac/integrators/wrapper_integrator.hpp"
-#include "../src/serac/numerics/expr_template_ops.hpp"
+#include "serac/physics/integrators/wrapper_integrator.hpp"
```
joshessman-llnl, line 13: 👍

joshessman-llnl (approved):

jamiebramwell:
LGTM

PR #356, opened 2021-01-29
Add a hook to grab the tangent to the last residual

jonwong12:
This PR adds a hook to `serac::BasePhysics` to output the residual and tangent operator. Currently only NonlinearSolid and elasticity integrate these methods. This might be useful for post-processing results or for computing adjoint solutions.

Also cleaned up some leftover method declarations and variables that were not being used.

Waiting for #353 before updating.


jamiebramwell:
I'll push on getting #353 in so we can merge this. @jandrej is also interested in it.

src/serac/physics/elasticity.cpp
```diff
@@ -124,6 +124,8 @@ void Elasticity::QuasiStaticSolve()
   K_inv_.Mult(*bc_rhs_, displacement_.trueVec());
 }
 
+mfem::Operator& Elasticity::getOutputGradient() { return *K_mat_; }
```
jandrej: Any specific reason for the "Output" naming? Same with the Residual.

src/serac/physics/nonlinear_solid.hpp
```diff
@@ -188,6 +188,11 @@ class NonlinearSolid : public BasePhysics {
    */
   void advanceTimestep(double& dt) override;
 
+  /**
+   * @brief Get an mfem::Operator that calculates the residual at the current outputState
+   */
+  mfem::Vector getOutputResidual();
```
jandrej: What do you think about adding the two methods to the `BasePhysics` class? I think it's valuable to keep consistent naming in the physics modules. I'm not sure though if they should be required in an implementation.

src/serac/physics/elasticity.cpp
jonwong12: I was trying to match the  `outputState()` notion. I'm open to better naming.

jonwong12: I changed the name just to `outputGradient` and `outputResidual` just to better match `outputState()`.

src/serac/physics/nonlinear_solid.hpp
jonwong12: I added it to BasePhysics and then implemented stubs/errors the "mfem-way" for non-implemented physics.

jamiebramwell:
Should I remove the WIP tag?

src/serac/physics/thermal_conduction.hpp
```diff
@@ -199,6 +199,24 @@ class ThermalConduction : public BasePhysics {
    */
   virtual ~ThermalConduction() = default;
 
+  /**
+   * @brief Get an mfem::Operator that calculates the residual at the current outputState
+   */
+  [[noreturn]] mfem::Vector outputResidual() override
```
jamiebramwell: tagging @white238 and @joshessman-llnl , Any thoughts on attributes and throwing errors?

src/serac/physics/base_physics.hpp
```diff
@@ -130,6 +130,16 @@ class BasePhysics {
    */
   virtual void outputState() const;
 
+  /**
+   * @brief Get an mfem::Operator that calculates the residual at the current outputState
+   */
+  virtual mfem::Vector outputResidual() = 0;
```
jamiebramwell: What would this be used for? Solving adjoint equations or something else? It may be easier to swap implementations if we just have something like `solveAdjoint` or `solveEigensystem` instead of giving direct access to the underlying objects.

jonwong12: I was mainly using it to test the `outputGradient` method, but I agree it might only be used in checking that a solution is "converged" enough. 

I think it's good to have something that is like `outputGradient()` if that's what you actually want (e.g. you want to specifically grab a tangent matrix). It's also good for checking when adding new linear algebra solvers. 

I think something like `solveAdjoint` would be useful for the adjoint case as you could specifically better handle how dof's are ordered and so forth.

src/serac/physics/base_physics.hpp
```diff
@@ -130,6 +130,16 @@ class BasePhysics {
...
+
+  /**
+   * @brief Get an mfem::Operator that calculates the gradient of the residual at the current outputState
+   */
+  virtual mfem::Operator& outputGradient() = 0;
```
jamiebramwell: Per our conversation yesterday, I think these should go in a `detail` or `impl` namespace with a note that they could be deprecated.

jonwong12: I ended up `friend`ing the implementation. This gets rid of the implementation inheritance which is a little brittle at the moment.

src/serac/physics/thermal_conduction.hpp
jonwong12: This should be removed with the `detail` suggestion changes.

src/serac/physics/elasticity.hpp
```diff
@@ -18,6 +18,16 @@
 
 namespace serac {
 
+class Elasticity;
+
+namespace detail {
```
jamiebramwell: 👍 @joshessman-llnl @white238 , does this look OK to you?

src/serac/physics/operators/odes.hpp
```diff
@@ -147,6 +147,11 @@ class SecondOrderODE : public mfem::SecondOrderTimeDependentOperator {
    */
   void Step(mfem::Vector& x, mfem::Vector& dxdt, double& time, double& dt);
 
+  /**
+   * @brief Get a reference to the current state
+   */
+  const State GetState() { return state_; }
```
jamiebramwell: I think this is missing an &.

src/serac/physics/operators/odes.hpp
```diff
@@ -265,17 +270,6 @@ class FirstOrderODE : public mfem::TimeDependentOperator {
    */
   FirstOrderODE(int n, FirstOrderODE::State&& state, const EquationSolver& solver, const BoundaryConditionManager& bcs);
 
-  /**
```
jamiebramwell, line 268: Was there a reason this was deleted?

jamiebramwell:
@jandrej , will this work for want you want to do?

src/serac/physics/elasticity.hpp
white238: After speaking with @jamiebramwell .  Exposing this is okay as long as we make it very clear you can shoot yourself in the foot.  Confusing the interface is not adding anything.

We should just expose what we need to and put a note in the documentation for the buyer-beware nature...  Please reword to be correct.

```
/**
  Get the current gradient MFEM operator

  Note:  This is for expert users only, changing any values inside of the returned data structures can have drastic and unrecoverable runtime consequences.
**/
const mfem::operator& current_gradient();
```

jandrej:
> @jandrej , will this work for want you want to do?

Looks fine. I'm still not sure why it has the "output" naming scheme.

src/serac/physics/operators/odes.hpp
jonwong12: I think it was a vestigial prototype that I used before the `SecondOrderODE` PR. All `SecondOrderODEs` by default simultaneous implement both `mfem::TimeDependentOperator` and `mfem::SecondOrderTimeDependentOperator`, so the `SecondOrderODE` is a valid mfem object that can be used with solvers.

src/serac/physics/elasticity.hpp
jamiebramwell: Sorry for flip-flopping! After kicking it around in my head, I agree with @white238 that returning a `const mfem::Operator&` is safe enough and it is clear what is going on.

white238 (approved):

jamiebramwell (approved):
Thanks for changing this back and sorry for the confusing input!

jamiebramwell:
I can merge this and add a similar capability to the new solid PR.

src/serac/physics/nonlinear_solid.cpp
```diff
@@ -333,6 +333,38 @@ void NonlinearSolid::InputOptions::defineInputFileSchema(axom::inlet::Table& tab
...
+  } else {
+    // Currently the residual constructed uses d2u_dt2 as input,
+    // but this could change?
+
+    residual_->Mult(ode2_.GetState().d2u_dt2, eval);
+  }
+  return eval;
+}
+
+// Get an Operator that computes
```
white238: This sentence needs to be finished.

jamiebramwell:
LGTM

PR #363, opened 2021-02-04
Re-enable time-dependent scaling of traction coefficient in driver

joshessman-llnl:
I'm not sure if this is exactly the canonical MFEM way of doing this, but I think time-dependent coefficients require "manual" updating of simulation time.

We don't validate the output of the driver, so it might be useful to add some time-dependent tests now that it's easy to define time-dependent coefficients in input files.


data/input_files/default.lua
```diff
@@ -75,13 +75,9 @@ nonlinear_solid = {
         },
         ['traction'] = {
             attrs = {2},
-            vec_coef = function (v)
-                return Vector.new(0, 1.0e-3, 0)
+            vec_coef = function (v, t)
```
jamiebramwell, line 78: This is so nice!

jamiebramwell (approved):
This is really great. Thanks!

jamiebramwell:
@samuelpmishLLNL , can you take a look at this?

tests/benchmark_expr_templates.cpp
```diff
@@ -62,7 +62,7 @@ static void BM_mixed_expr_MFEM(benchmark::State& state)
   MPI_Barrier(MPI_COMM_WORLD);
 
   // Number of rows is the argument that varies
-  const int rows  = state.range(0);
+  const int rows  = static_cast<int>(state.range(0));
```
samuelpmishLLNL, line 65: is it worth it to just define `sample_vectors` to take a `size_t` or whatever `state.range(0)` is, rather than static casting everything to `int`?

samuelpmishLLNL (approved):

tests/benchmark_expr_templates.cpp
jamiebramwell: I'll get this merged but leave this comment as an issue.

jamiebramwell:
LGTM

PR #373, opened 2021-02-07
Add a nonlinear reaction option to the thermal module

jamiebramwell:
This adds a nonlinear reaction term to the thermal conduction module. It allows us to solve equations of the form:

CodeCogsEqn

where s(x,t), f(T), and g(x,t) are a user-defined functions. To enable this, I changed the thermal conduction to use a `NonlinearForm` instead of a `BilinearForm` to handle the nonlinearities that are now possible.

It also changes the `LinearToNonlinearFormIntegrator` to use a const reference to the underlying FESpace instead of a shared pointer as an integrator doesn't make sense separate from a FEState.

Note: this is branched off of my solid mechanics re-write to avoid merge errors, so that should get merged first. I've also added the `addState` PR as a work around for a production problem for now.


data/input_files/tests/thermal_conduction/static_solve.lua
```diff
@@ -43,6 +43,16 @@ thermal_conduction = {
     -- material parameters
     kappa = 0.5,
 
+    -- add a nonlinear source
+    nonlinear_reaction = {
+        reaction_function = function (temp)
```
white238, line 48: heh.. I guess this answers the question @joshessman-llnl  and I talked about a while ago. Whether nested functions worked (Josh was pretty sure they would) but more importantly whether people would want it and if we should advertise it.

jamiebramwell: This isn't nested, it's two different functions. BUT NOW I WANT TO TRY IT.

data/input_files/tests/thermal_conduction/static_reaction_exact.lua
```diff
@@ -0,0 +1,66 @@
+-- Comparison information
```
jamiebramwell, line 1: This was verified to be correct by taking the L2 error against the exact solution in a C++ driver code.

data/input_files/tests/thermal_conduction/static_solve.lua
white238: by nested i mean not at the global scope.  glad it worked!

src/serac/physics/thermal_conduction.hpp
```diff
@@ -70,6 +70,13 @@ class ThermalConduction : public BasePhysics {
     double cp;
     double rho;
 
+    // Nonlinear reaction information
+    std::optional<std::function<double(double)>> reaction_func;
+    std::optional<std::function<double(double)>> d_reaction_func;
```
joshessman-llnl: `std::function` is already nullable (it has an `operator bool` etc) - I think we can leave off the `std::optional` here

src/serac/physics/thermal_conduction.hpp
```diff
@@ -306,6 +301,18 @@ class ThermalConduction : public BasePhysics {
    * nonlinear solver
    */
   mfem::Vector previous_;
+
+  /**
+   * @brief the nonlinear reaction function
+   *
+   */
+  std::optional<std::function<double(double)>> reaction_;
```
joshessman-llnl: See above on `std::optional<std::function>>`

joshessman-llnl (approved):
Just one minor thing - I'm glad to see these Inlet callbacks being used in more places!

src/serac/physics/thermal_conduction.hpp
jamiebramwell: TIL!

jonwong12 (approved):

jamiebramwell:
LGTM

PR #375, opened 2021-02-08
Add OutputType to driver input file schema

### LINKED ISSUE #362: Use `output_type` properly
I was looking into if we required a specific capitalization for the `output_type` input variable and it turns out we are just hard coding the output_type to visit.  We should fix that and also lower case the string before converting to the enum.

https://github.com/LLNL/serac/blob/bd957565001c35e8bece9c8bc2958e049f803be5/src/drivers/serac.cpp#L118

joshessman-llnl:
This exposed a gap in Inlet that I had to work around a bit.

I also renamed `module` to `phys_module` in the test driver as it is a reserved word in C++20.

Resolves #362


jamiebramwell (approved):

white238 (approved):

joshessman-llnl:
LGTM

PR #377, opened 2021-02-09
Adding extra caliper markers and metadata methods

jonwong12:
Added caliper region tags for profiling general regions, as well as some metdata commands.


jamiebramwell:
This will be super handy!

jamiebramwell:
LGTM

src/serac/infrastructure/profiling.hpp
```diff
@@ -71,4 +85,37 @@ void initializeCaliper(const std::string& options = "");
...
+template <>
+void setCaliperMetaData<const char*>(const std::string& name, const char* data);
+
+/**
+ * @brief Adds unsigned int with given name to caliper metadata
+ */
+
+template <>
+void setCaliperMetaData<unsigned int>(const std::string& name, unsigned int data);
+
```
samuelpmishLLNL: I'm not sure we need these forward declarations for each specific type, unless it's just for doxygen.

It seems like setCaliperMetaData could be an overload set, e.g.
```cpp
void setCaliperMetaData(const std::string& name, double data);
void setCaliperMetaData(const std::string& name, int data);
void setCaliperMetaData(const std::string& name, unsigned int data);
...
```
or a function template (without forward declaring the explicit specializations)
```cpp
template <typename T>
void setCaliperMetaData(const std::string& name, T data);
```

I also think metadata is a single word, so the capital D looks out of place to me.

jamiebramwell:
@joshessman-llnl , could you take a look at this?

joshessman-llnl:
In addition to Sam's comment, I think it would be good to see an example of these new metadata annotations.

src/docs/sphinx/dev_guide/profiling.rst
```diff
@@ -45,6 +51,9 @@ To enable Caliper for a program, call ``serac::profiling::initializeCaliper()``
 Optionally, a Caliper `ConfigManager configuration string <https://software.llnl.gov/Caliper/ConfigManagerAPI.html#configmanager-configuration-string-syntax>`_
 can be passed to configure Caliper.
 
+Additionally, you can use ``serac::profiling::setCaliperMetadata(name, type)`` to add extra metadata into the ``.cali`` file. Supported metadata types are ``char *, int, unsigned int, double``.
```
joshessman-llnl: Instead of `type` maybe `data` or `value` would be better (and consistent with `name` here) so you could say something like "supported `value` types are..."

src/serac/infrastructure/profiling.hpp
```diff
@@ -71,4 +85,10 @@ void initializeCaliper(const std::string& options = "");
  */
 void terminateCaliper();
 
+/**
+ * @brief Caliper metadata methods cali_set_global_(double|int|string|uint)_byname()
+ */
+template <typename T>
+void setCaliperMetadata(const std::string& name, T data);
```
joshessman-llnl: For interface consistency it might be worth considering using a macro for this like all the other profiling helpers - this would also guarantee zero overhead on non-profiling runs (since the current approach is still a function call).

Maybe something like `SERAC_ADD_PROFILING_METADATA`?

tests/serac_profiling.cpp
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
```
joshessman-llnl, line 1: :+1:

tests/serac_profiling.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#include "serac/infrastructure/cli.hpp"
+#include "serac/infrastructure/initialize.hpp"
+#include "serac/infrastructure/profiling.hpp"
+#include "serac/numerics/mesh_utils.hpp"
+#include <cstring>
+
+namespace serac {
+  
+TEST(serac_profiling, mesh_refinement)
```
joshessman-llnl, line 19: Since we're not using gtest anywhere in this, it might be better to have this be an example?

joshessman-llnl (commented):
Thanks for taking a look at this - hopefully we'll have caliper-enabled builds on LC soon!

tests/serac_profiling.cpp
jonwong12: I added a simple smoke check on the number of elements after refinement.

src/serac/infrastructure/profiling.hpp
jonwong12: I think I just put them within the macro ifdefs at the top of `profiling.hpp`. I think the calls will now be no-oped if caliper is not enabled.

src/docs/sphinx/dev_guide/profiling.rst
```diff
@@ -24,11 +24,15 @@ to instrument your code:
 
 Use ``SERAC_MARK_FUNCTION`` at the very top of a function to mark it for profiling.
 
+Use ``SERAC_MARK_START(name)`` at the begining of a region and ``SERAC_MARK_END(name)`` at the end of the region.
```
white238: beginning

src/docs/sphinx/dev_guide/profiling.rst
```diff
@@ -45,6 +51,9 @@ To enable Caliper for a program, call ``serac::profiling::initializeCaliper()``
 Optionally, a Caliper `ConfigManager configuration string <https://software.llnl.gov/Caliper/ConfigManagerAPI.html#configmanager-configuration-string-syntax>`_
 can be passed to configure Caliper.
 
+Additionally, you can use ``SERAC_SET_METADATA(name, data)`` to add extra metadata into the ``.cali`` file. Supported metadata are ``string, int, unsigned int, double``.
```
white238, line 54: Example of this?

src/serac/infrastructure/profiling.hpp
```diff
@@ -55,10 +73,16 @@
 #define SERAC_MARK_LOOP_START(id, name)
 #define SERAC_MARK_LOOP_ITER(id, i)
 #define SERAC_MARK_LOOP_END(id)
+#define SERAC_MARK_START(name)
+#define SERAC_MARK_END(name)
+#define SERAC_SET_METADATA(name, data) 
 
 #endif
 
+// profiling namespace
```
white238: Should this be /// or 98 be //?  Seems to be inconsistent

tests/CMakeLists.txt
```diff
@@ -28,7 +29,7 @@ if (ENABLE_GTEST)
         serac_boundary_cond.cpp
         serac_mesh.cpp
         mfem_ex9p_blockilu.cpp
-	    serac_newmark_test.cpp)
+	serac_newmark_test.cpp)
```
white238: spacing is wrong here and on line 73

tests/serac_profiling.cpp
```diff
@@ -0,0 +1,90 @@
...
+    }
+  SERAC_MARK_LOOP_END(refinement_loop);
+  
+  SERAC_SET_METADATA("mesh_file", mesh_file.c_str());
+  SERAC_SET_METADATA("number_mesh_elements", pmesh->GetNE());
+  
+  unsigned int magic_uint = 1819176044;
+  SERAC_SET_METADATA("magic_uint", magic_uint);
+
+  std::array<char, sizeof(magic_uint)+1> uint_string;
```
white238: Some comments to this function would be appreciated.

src/serac/infrastructure/profiling.hpp
jonwong12: Good catch! It just depends on what is necessary to get doxygen to recognize namespace functions.

tests/CMakeLists.txt
white238: Approved on the contingency you fix this.

white238 (approved):

tests/CMakeLists.txt
jonwong12: Weird... Does make style fix CmakeLists.txt files?

white238: Not unless we install cmakeformat (I tried through spack, it didn’t build)

src/serac/infrastructure/profiling.hpp
```diff
@@ -71,4 +94,29 @@ void initializeCaliper(const std::string& options = "");
...
+
+/*!
+  @overload void serac::profiling::setCaliperMetadata(const std::string & name, int data)
+*/
+void setCaliperMetadata(const std::string& name, int data);
+
+/*!
+  @overload void serac::profiling::setCaliperMetadata(const std::string & name, std::string data)
+*/
+void setCaliperMetadata(const std::string& name, std::string data);
```
joshessman-llnl: Can this be `const string&` (or `string_view`)?

src/serac/infrastructure/profiling.hpp
```diff
@@ -71,4 +94,29 @@ void initializeCaliper(const std::string& options = "");
  */
 void terminateCaliper();
 
+/// detail namespace
+namespace detail {
+/**
+ * @fn void serac::profiling::setCaliperMetadata(const std::string & name, double data)
```
joshessman-llnl: The doxygen comment for the first overload should have `brief` and `param` tags - I think all the remaining overloads need *only* an `overload` tag without the function name - I think it usually gets picked up automatically?

tests/serac_profiling.cpp
```diff
@@ -0,0 +1,91 @@
...
+  std::memcpy(&magic_double, "llnl", 4);
+  SERAC_SET_METADATA("magic_double", magic_double);
+
+  // decode the double and print
+  std::array<char, sizeof(magic_double) + 1> double_string;
+  std::fill(std::begin(double_string), std::end(double_string), 0);
+  std::memcpy(double_string.data(), &magic_double, 4);
+  std::cout << double_string.data() << std::endl;
+
+  EXPECT_EQ(2112, pmesh->GetNE());
```
joshessman-llnl: It is a little confusing to me to have this be the only thing actually "tested" in this file - I think it's fine to have this be a "caliper smoketest" but I'm not clear on why this particular check is here.

joshessman-llnl (commented):
Looks good, just a few more small comments/suggestions.

tests/serac_profiling.cpp
jonwong12: Yes, maybe it's better to just make this a smoke test.

src/serac/infrastructure/profiling.hpp
jonwong12: It sort of picks it up. But it doesn't pick up the arguments. Instead it just fills in the following which I think is suitable.
> This is an overloaded member function, provided for convenience. It differs from the above function only in what argument(s) it accepts.

joshessman-llnl: Yep, that's what I would expect

joshessman-llnl (approved):

PR #379, opened 2021-02-09
Re-add the accidentally deleted cylinder tests

jamiebramwell:
While resolving the merge conflicts in the cylinder mesh tools branch, I did a whoopsie and accidentally deleted the tests. Luckily I noticed it in codecov.

PR #380, opened 2021-02-10
HollowCylinder modifications to do arcs

jonwong12:
This PR extends Sam's hollow cylinder work and uses arbitrary sector angles, and allows for extruded arc segments in addition to a cylinder.

TODO: Probably needs to be updated after #379.


src/serac/numerics/mesh_utils.hpp
```diff
@@ -110,12 +110,14 @@ std::shared_ptr<mfem::ParMesh> buildCylinderMesh(int radial_refinement, int elem
  * @param[in] inner inner radius the radius of the cylindrical shell
  * @param[in] outer ouer radius the radius of the cylindrical shell
  * @param[in] height the number of elements in the z-direction
+ * @param[in] sectors the number of starting sectors in the hollow cylinder
```
samuelpmishLLNL: what is "angle"? angle per sector, total angle?
I'd prefer specifying a total angle, personally.

jonwong12: Hmm. I think this implementation assumes that the elements need to be integer multiples of the "sector angle". I agree that total angle is more convenient from a specification standpoint, but not sure what the best way to check the required condition (floating point considerations).

samuelpmishLLNL: hmm, I don't understand, isn't the sector angle just total_angle / num_angular_divisions?

samuelpmishLLNL: one option is to make total_angle the last argument, and default it to 2 * pi. The body of the function can do something like
```cpp
if (min(total_angle, 2 * M_PI) == 2 * M_PI) {
  // full cylinder
} else {
  // partial cylinder
}
```

jonwong12: Sorry, I think I understand what you are suggesting now. Specify `total_angle` and `num_angular_sectors` instead of `sector_angle` and `num_angular_sectors`, correct? I think that's a good idea.

jamiebramwell (approved):

src/serac/numerics/mesh_utils.cpp
```diff
@@ -269,94 +269,135 @@ std::shared_ptr<mfem::ParMesh> buildCylinderMesh(int radial_refinement, int elem
...
+  auto num_boundary_elements = num_elems * 2;
 
   SLIC_ERROR_IF(outer_radius <= inner_radius,
                 "Outer radius is smaller than inner radius while building a cylinder mesh.");
 
-  double vertices[num_vertices][dim];
-  for (int i = 0; i < 8; i++) {
-    double s       = sin(M_PI_4 * i);
-    double c       = cos(M_PI_4 * i);
+  std::vector<std::vector<double>> vertices(static_cast<size_type>(num_vertices), std::vector<double>(dim, 0.));
```
jamiebramwell, line 295: @joshessman-llnl Any way out of this grossness? I know there's a fundamental mismatch between mfem indices and size types.

joshessman-llnl: I'd have to mess around with it to be sure but doing some of the calculations as `size_t` might help, e.g., starting with changing the `sectors` parameter to `size_t` and going from there.

src/serac/numerics/mesh_utils.hpp
```diff
@@ -110,13 +110,33 @@ std::shared_ptr<mfem::ParMesh> buildCylinderMesh(int radial_refinement, int elem
...
+ * @param[in] radial_refinement the number of times to apply uniform mesh refinement to the cross section
+ * @param[in] inner inner radius the radius of the cylindrical shell
+ * @param[in] outer ouer radius the radius of the cylindrical shell
+ * @param[in] total_angle the angle in radians over which to generate the portion of an extruded cylinder
+ * @param[in] sectors the number of starting sectors in the hollow cylinder
+ *
+ * @return A shared_ptr containing the constructed mesh
+ */
+  std::shared_ptr<mfem::ParMesh> buildRingMesh(int radial_refinement,
+					       double inner_radius, double outer_radius,
```
white238, line 135: Did you run clangformat on this?

white238: Apparently it passed... odd.

white238 (approved):

src/serac/numerics/mesh_utils.cpp
```diff
@@ -269,94 +269,135 @@ std::shared_ptr<mfem::ParMesh> buildCylinderMesh(int radial_refinement, int elem
...
       for (int d = 0; d < dim; d++) {
         vertex(d) = new_vertices[d * n + i];
       }
 
-      // stretch the octagonal shape into a circle of the appropriate radius
-      double phi = fmod(atan2(vertex(1), vertex(0)) + M_PI, M_PI_4);
-      vertex *= (cos(phi) + (-1.0 + sqrt(2.0)) * sin(phi));
+      // stretch the polygonal shape into a cylinder
+      // phi is the angle to the closest multiple of a sector angle
+      double theta = atan2(vertex(1), vertex(0));
```
joshessman-llnl, line 357: 👍

joshessman-llnl (approved):

jamiebramwell:
LGTM

PR #383, opened 2021-02-10
WIP: Add an extra finite element state to the output files

jamiebramwell:
This adds a way to add an extra finite element state to the output files. I don't really like this solution. Any ideas @white238 @joshessman-llnl ?


jamiebramwell:
I'm going to close this now in favor of waiting for the sidre data store.

PR #386, opened 2021-02-13
Adding constant coefficient input options

jamiebramwell:
This adds constant coefficients as an option through inlet. It also changes us to the `addStruct` interface instead of the deprecated `addTable`. This allows coefficient definitions of the following forms:

```lua
--Constant scalar coefficients
coef_opt_scalar = { 
    constant = 5.0
}

--Piecewise constant scalar coefficients defined by MFEM mesh attribute
coef_opt_pw_scalar = { 
    piecewise_constant = { 
        [1] = 2.5, 
        [3] = 3.0
    }

--Constant vector coefficients
coef_opt_constant_vector = { 
    vector_constant = {
        x = 3.0,
        y = 0.25
    }
}

--Piecewise constant vector coefficients defined by MFEM mesh attribute
coef_opt_pw_constant_vector = {
    vector_piecewise_constant = {
        [1] = {
            x = 4.0,
            y = -1.0
        },
        [4] = {
            x = 0.0,
            y = 3.5
        }
    }
}
```


jamiebramwell:
I think I have it down to a minimum reproducer. It looks like the `coef` sub-table in the boundary condition table is not getting saved to sidre. You can see it in the nonlinear_solid/cutom_solver test. It uses `/data/inputfiles/tests/nonlinear_solid/qs_solve.lua`. Any ideas?

white238:
> I think I have it down to a minimum reproducer. It looks like the `coef` sub-table in the boundary condition table is not getting saved to sidre. You can see it in the nonlinear_solid/cutom_solver test. It uses `/data/inputfiles/tests/nonlinear_solid/qs_solve.lua`. Any ideas?

@joshessman-llnl fixed nested structs in an Axom PR that has already been brought into Serac.  Replacing `addTable` with `addStruct` will fix this in Serac.  In the newest Axom develop, `addTable` is now private (it is needed internally) and `addStruct` should be used in its place.  Also `addGenericDictionary` was renamed to `addStructDictionary` but they are equivalent in Serac's Axom but `addGenericDictionary`disappears in the next update.

Sorry, we are trying to have a more uniform naming convention on all the add functions and you caught us mid-fix.

jamiebramwell:
No problem! I'm just shocked it wasn't a dumb typo on my part. 😆 Thanks for finding it!

src/serac/physics/nonlinear_solid.cpp
```diff
@@ -77,27 +77,27 @@ NonlinearSolid::NonlinearSolid(std::shared_ptr<mfem::ParMesh> mesh, const Nonlin
...
+    setVelocity(*velo);
   }
   setViscosity(std::make_unique<mfem::ConstantCoefficient>(options.viscosity));
 
   for (const auto& [name, bc] : options.boundary_conditions) {
     // FIXME: Better naming for boundary conditions?
     if (name.find("displacement") != std::string::npos) {
       if (bc.coef_opts.isVector()) {
-        auto disp_coef = std::make_shared<mfem::VectorFunctionCoefficient>(bc.coef_opts.constructVector(dim));
+        std::shared_ptr<mfem::VectorCoefficient> disp_coef(bc.coef_opts.constructVector(dim));
```
joshessman-llnl, line 93: I'm a little surprised that this works - `shared_ptr` doesn't have a constructor that takes a `T&` (https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr)

src/serac/physics/nonlinear_solid.cpp
```diff
@@ -314,19 +314,19 @@ void NonlinearSolid::InputOptions::defineInputFileSchema(axom::inlet::Table& tab
...
-      table.addTable("stiffness_solver", "Linear and Nonlinear stiffness Solver Parameters.");
+      table.addStruct("stiffness_solver", "Linear and Nonlinear stiffness Solver Parameters.");
   serac::mfem_ext::EquationSolver::DefineInputFileSchema(stiffness_solver_table);
 
-  auto& dynamics_table = table.addTable("dynamics", "Parameters for mass matrix inversion");
+  auto& dynamics_table = table.addStruct("dynamics", "Parameters for mass matrix inversion");
   dynamics_table.addString("timestepper", "Timestepper (ODE) method to use");
   dynamics_table.addString("enforcement_method", "Time-varying constraint enforcement method to use");
 
   auto& bc_table = table.addGenericDictionary("boundary_conds", "Table of boundary conditions");
```
joshessman-llnl: While we're changing `addTable` -> `addStruct`, we might also want to change `addGenericDictionary` -> `addStructDictionary` and `addGenericArray` -> `addStructArray`.

joshessman-llnl (approved):
This should improve performance a bit since functions are overkill for constant vectors

src/serac/physics/nonlinear_solid.cpp
joshessman-llnl: With a trivial example `std::shared_ptr<T>` can't be constructed from a `T` - https://godbolt.org/z/98e59j

jamiebramwell: This works because `constructVector` and `constructScalar` now return unique pointers since we only know the base class.

joshessman-llnl: Ah, I missed that.  Thanks!

jamiebramwell:
LGTM

data/input_files/default.lua
```diff
@@ -41,41 +41,40 @@ nonlinear_solid = {
...
     },
 
     -- boundary condition parameters
     boundary_conds = {
         ['displacement'] = {
             attrs = {1},
-            vec_coef = function (v)
-                return Vector.new(0, 0, 0)
-            end
+            constant_vector = {
```
white238: I like that you removed the "coef" layer.  I assume that was an attempt to debug?

data/input_files/tests/nonlinear_solid/dyn_solve.lua
```diff
@@ -82,9 +84,11 @@ nonlinear_solid = {
         ['displacement'] = {
             -- boundary attribute 1 (index 0) is fixed (Dirichlet) in the x direction
             attrs = {1},
-            vec_coef = function (v)
-                return Vector.new(0, 0, 0)
-            end
+            constant_vector = {
```
white238: Is "constant_vector" descriptive enough?

data/input_files/default.lua
jamiebramwell: Yep. That was verifying that it wasn't getting saved in sidre.

data/input_files/tests/nonlinear_solid/dyn_solve.lua
jamiebramwell: I think so. This block describes coefficients, where `constant_vector` is a way to describe one. However, I'm open to other ideas if you have them!

src/serac/infrastructure/input.cpp
```diff
@@ -89,33 +92,100 @@ void BoundaryConditionInputOptions::defineInputFileSchema(axom::inlet::Table& ta
...
+        pw_constants(entry.first - 1) = entry.second[i];
+      }
+
+      // Set the spatial dimension coefficient to a newly constructed scalar piecewise coefficient
+      pw_vec_coeff->Set(i, new mfem::PWConstCoefficient(pw_constants));
+    }
+    return pw_vec_coeff;
+
+  } else {
+    SLIC_ERROR("Trying to build a vector coefficient when no appropriate input data exists.");
```
white238: Can we give more information in this error on how to fix it? or should that be higher up in an inlet verifier?

jamiebramwell: Hmmm. I think it could be in a verifier. I need to learn how to use them anyway. I can also make this error message better.

white238 (approved):

jamiebramwell:
LGTM

jamiebramwell:
LGTM

PR #393, opened 2021-02-17
Update solid.rst

RMFerencz:
A couple of small edits and a typo fix.


RMFerencz:
So should I be able to add an obvious label like "documentation" to this PR?  Or is that something that gets added during the review phase?

jamiebramwell (approved):

jamiebramwell:
@RMFerencz Hello! You can normally add tags to PRs at the right side of the page. I'm not sure if you can since you're username is not technically in the LLNL group yet.

jamiebramwell:
Corrections now live here: https://serac.readthedocs.io/en/doc-bramwell-theory_manual/sphinx/theory_reference/solid.html. Thanks!

PR #398, opened 2021-02-23
Refactor mesh utilities to reduce dependence on Inlet

joshessman-llnl:
To allow greater flexibility when using the C++ API instead of Inlet/Lua for simulation configuration, this PR intends to further separate Inlet deserialization and mesh instantiation logic.

Instead of having to pass an MPI communicator and refinement parameters to each mesh creation function, the common refinement/parallelizing logic is moved to a new `serac::mesh::finalize` function that accepts a serial mesh.  The individual generation functions are now only responsible for generating a serial mesh.

A new `serac::mesh::build` function will construct a complete ParMesh from a set of input options, meaning that some of the variant examining/accessing is kept out of user-facing code.

The work done in this PR exposed a bug in which refinement parameters for generated meshes are ignored, which was also fixed.
I'm not sure how superfluous the refinement logic is for the generated meshes, but because the `InputOptions` structure/corresponding input file schema still include it, I thought it made sense to include the matching logic.


jamiebramwell:
Is this ready to go or is it waiting on stuff?

joshessman-llnl:
I think this is probably a good starting point, though I'd like to make the other generated meshes (ball, disk, etc) available through Inlet and through this new approach in a future PR.

src/serac/numerics/mesh_utils.hpp
```diff
@@ -24,20 +24,15 @@
  */
 namespace serac {
 /**
- * @brief Constructs an MFEM parallel mesh from a file and refines it
+ * @brief Constructs an MFEM mesh from a file and refines it
```
jamiebramwell: This no longer refines it.

src/serac/numerics/mesh_utils.hpp
```diff
@@ -175,28 +170,30 @@ struct InputOptions {
...
+ * @param[in] refine_parallel The number of parallel refinements
+ * @param[in] MPI_Comm The MPI communicator
  *
- * @param[in] extra_options Rectangle Mesh Options
- * @param[in] MPI_Comm MPI Communicator
  * @return A shared_ptr containing the constructed mesh
  */
-std::shared_ptr<mfem::ParMesh> buildRectangleMesh(serac::mesh::GenerateInputOptions& options,
-                                                  const MPI_Comm                     comm = MPI_COMM_WORLD);
+std::shared_ptr<mfem::ParMesh> finalize(mfem::Mesh& serial_mesh, const int refine_serial = 0,
```
jamiebramwell: I'm not sure `finalize` is the best name. Maybe `buildParallel`?

jamiebramwell: `refineAndDistribute`?

src/serac/numerics/mesh_utils.cpp
```diff
@@ -16,22 +16,17 @@
...
+std::shared_ptr<mfem::Mesh> buildMeshFromFile(const std::string& mesh_file)
 {
-  // Get the MPI rank for logging purposes
-  int rank = 0;
-  MPI_Comm_rank(comm, &rank);
-
   // Open the mesh
   std::string msg = fmt::format("Opening mesh file: {0}", mesh_file);
-  SLIC_INFO_ROOT(rank, msg);
+  SLIC_INFO(msg);
```
jamiebramwell: I think these can go back to ROOT now.

src/serac/numerics/mesh_utils.hpp
```diff
@@ -175,28 +170,30 @@ struct InputOptions {
...
+ * @param[in] options The options used to construct the mesh
+ * @param[in] comm The MPI communicator to use with the parallel mesh
  *
- * @param[in] extra_options Cuboid Mesh Options
- * @param[in] MPI_Comm MPI Communicator
  * @return A shared_ptr containing the constructed mesh
  */
-std::shared_ptr<mfem::ParMesh> buildCuboidMesh(serac::mesh::GenerateInputOptions& options,
-                                               const MPI_Comm                     comm = MPI_COMM_WORLD);
+std::shared_ptr<mfem::ParMesh> build(const InputOptions& options, const MPI_Comm comm = MPI_COMM_WORLD);
```
jamiebramwell: `buildSerialMesh`?

src/serac/numerics/mesh_utils.hpp
```diff
@@ -175,28 +170,30 @@ struct InputOptions {
...
 
 /**
- * @brief Constructs a 2D MFEM mesh of a rectangle
+ * @brief Finalizes a serial mesh into a refined parallel mesh
+ *
+ * @param[in] serial_mesh The "base" serial mesh
+ * @param[in] refine_serial The number of serial refinements
+ * @param[in] refine_parallel The number of parallel refinements
+ * @param[in] MPI_Comm The MPI communicator
```
jamiebramwell: I would add this
```c++
* @note It is sometimes required to refine serially first if your number of processors is less than the original number of mesh elements
```

tests/serac_linelastic_solver.cpp
```diff
@@ -23,7 +23,7 @@ TEST(elastic_solver, static_solve)
   // Open the mesh
   std::string mesh_file = std::string(SERAC_REPO_DIR) + "/data/meshes/beam-quad.mesh";
 
-  auto pmesh = buildMeshFromFile(mesh_file, 1, 0);
+  auto pmesh = mesh::finalize(*buildMeshFromFile(mesh_file), 1, 0);
```
jamiebramwell: Dereferencing this looks a bit weird for something so user-facing. I think `finalize` should just take a pointer. Thoughts @white238 ?

joshessman-llnl: My preference here would be to have the `buildMeshFromFile` etc functions return an `mfem::Mesh` instead of a smart pointer (since they have no need to be on the heap), but for consistency that would require changing all the other signatures, which I was planning on doing when making all the other shapes available through Inlet.

src/serac/numerics/mesh_utils.hpp
joshessman-llnl: hmm, that might be confusing because it returns a `ParMesh` - maybe `buildParallelMesh`, though that might be confusing if we also have `refineAndDistribute`.

jamiebramwell: Oh you're right! Whoops!

jamiebramwell (approved):

jamiebramwell:
@jonwong12 @samuelpmishLLNL could you take a look at this?

samuelpmishLLNL (approved):

joshessman-llnl:
LGTM

tests/serac_mesh.cpp
```diff
@@ -20,7 +20,7 @@ TEST(mesh, load_exodus)
   MPI_Barrier(MPI_COMM_WORLD);
   std::string mesh_file = std::string(SERAC_REPO_DIR) + "/data/meshes/bortel_echem.e";
 
-  auto pmesh = buildMeshFromFile(mesh_file, 1, 1);
+  auto pmesh = mesh::refineAndDistribute(*buildMeshFromFile(mesh_file), 1, 1);
```
white238, line 23: 👍  on naming

tests/serac_solid.cpp
```diff
@@ -58,10 +58,12 @@ TEST(solid_solver, qs_custom_solve)
   test_utils::defineTestSchema<Solid>(inlet);
 
   // Build the mesh
-  auto mesh_options   = inlet["main_mesh"].get<serac::mesh::InputOptions>();
-  auto full_mesh_path = serac::input::findMeshFilePath(
-      std::get<serac::mesh::FileInputOptions>(mesh_options.extra_options).relative_mesh_file_name, input_file_path);
-  auto mesh = serac::buildMeshFromFile(full_mesh_path, mesh_options.ser_ref_levels, mesh_options.par_ref_levels);
+  auto mesh_options = inlet["main_mesh"].get<serac::mesh::InputOptions>();
```
white238, line 61: This seems to be in a lot of places.  Can it be in a function?

white238 (approved):

jamiebramwell:
LGTM

jonwong12 (approved):
Thanks. I think that cleans up things nicely.

jamiebramwell:
LGTM

PR #402, opened 2021-02-25
Adjust SLIC*_ROOT* macros to automatically retrieve rank

### LINKED ISSUE #400: Enable SLIC_ERROR_ROOT without a rank argument
We have a bunch of SLIC_ERROR and SLIC_ERROR_IF calls that should really be SLIC_ERROR_ROOT. This is because we often don't have clean access to the MPI rank. Would it be possible to  not require the rank as an argument? I know @white238 looked into this before and thought it would be messy, but I see this becoming a fairly large problem in terms of verbosity of output on large runs.

joshessman-llnl:
Resolves #400

### DIFF
--- a/tests/serac_input.cpp
+++ b/tests/serac_input.cpp
@@ -42,7 +42,7 @@ namespace serac {
 TEST_F(InputTest, vec_1d)
 {
   reader_->parseString("vec = { x = 4.75 }");
-  auto& vec_table = inlet_->addTable("vec");
+  auto& vec_table = inlet_->addStruct("vec");
   input::defineVectorInputFileSchema(vec_table);
   auto vec = vec_table.get<mfem::Vector>();
   EXPECT_EQ(vec.Size(), 1);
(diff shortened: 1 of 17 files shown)


tests/serac_input.cpp
```diff
@@ -42,7 +42,7 @@ namespace serac {
 TEST_F(InputTest, vec_1d)
 {
   reader_->parseString("vec = { x = 4.75 }");
-  auto& vec_table = inlet_->addTable("vec");
+  auto& vec_table = inlet_->addStruct("vec");
```
jamiebramwell, line 45: oh no! I think I ran my sed script on `src` only! Thanks for catching this.

jamiebramwell (approved):
This will clean things up nicely! Thanks!

tests/serac_input.cpp
joshessman-llnl: I'm using a new Axom version so I've had to fix it in all of my branches (e.g. #398) to even get it to compile 😆

white238: This does keep happening.. should we move unit tests under src next to their units when we move to our more strict integration/unit tests split?

white238 (approved):

white238:
LGTM

PR #404, opened 2021-02-26
Move the SLIC calls back to ROOT

jamiebramwell:
Most of our SLIC calls are in problem definition and set up and therefore should only appear on the root processor. This changes them to ROOT since we no longer need the MPI rank at the logging call site.


jamiebramwell:
LGTM

white238 (approved):

joshessman-llnl (approved):

jamiebramwell:
LGTM

jamiebramwell:
LGTM

PR #409, opened 2021-03-02
Enable reuse of a mesh with two solid modules

### LINKED ISSUE #408: serac::Solid GetNodes() reuse issue
A segmentation fault will occur when there is a mismatch between `pmesh->Nodes()->FESpace()` and `serac::Solid`'s finite element space. [line] (https://github.com/LLNL/serac/blob/9fad67ad4179c1e7959ea7faf2ffd857de24c1e1/src/serac/physics/solid.cpp#L38) 

Normally we have one `serac::solid` for a give mesh, so we won't run into this issue. However, the following will sequence of events will cause issues because of the way `mesh::NewNodes()` works. 
```
auto pmesh = buildMeshFromFile(....)
for (int i = 0; i < ....) {
auto solid = Solid(1, pmesh, ......);
}
```
when `i=1`, the initial `serac::Solid` will have gone out of scope. Because pmesh->Nodes() was set to a gridfunction that uses `solid(i=0)::FESpace()`, we're now in trouble, and the line above will segfault.

jamiebramwell:
This aims to enable mesh reuse by multiple solid modules.

Resolves https://github.com/LLNL/serac/issues/408.

### DIFF
--- a/src/serac/physics/solid.cpp
+++ b/src/serac/physics/solid.cpp
@@ -108,7 +108,7 @@ Solid::Solid(std::shared_ptr<mfem::ParMesh> mesh, const Solid::InputOptions& opt
       }
     } else if (name.find("traction") != std::string::npos) {
       std::shared_ptr<mfem::VectorCoefficient> trac_coef(bc.coef_opts.constructVector(dim));
-      if (geom_nonlin_ == GeometricOption::Linear) {
+      if (geom_nonlin_ == GeometricOption::Off) {
         setTractionBCs(bc.attrs, trac_coef, true);
       } else {
         setTractionBCs(bc.attrs, trac_coef, false);
@@ -118,7 +118,7 @@ Solid::Solid(std::shared_ptr<mfem::ParMesh> mesh, const Solid::InputOptions& opt
       setTractionBCs(bc.attrs, trac_coef, true);
     } else if (name.find("pressure") != std::string::npos) {
       std::shared_ptr<mfem::Coefficient> pres_coef(bc.coef_opts.constructScalar());
-      if (geom_nonlin_ == GeometricOption::Linear) {
+      if (geom_nonlin_ == GeometricOption::Off) {
         setPressureBCs(bc.attrs, pres_coef, true);
       } else {
         setPressureBCs(bc.attrs, pres_coef, false);
@@ -513,7 +513,7 @@ Solid::InputOptions FromInlet<Solid::InputOptions>::operator()(const axom::inlet
   if (input_geom_nonlin) {
     result.geom_nonlin = serac::GeometricOption::Nonlinear;
   } else {
-    result.geom_nonlin = serac::GeometricOption::Linear;
+    result.geom_nonlin = serac::GeometricOption::Off;
   }
 
   // Set the material nonlinearity flag
--- a/src/serac/physics/solid.cpp
+++ b/src/serac/physics/solid.cpp
@@ -221,6 +221,8 @@ void Solid::resetToReferenceConfiguration()
 
   velocity_.initializeTrueVec();
   displacement_.initializeTrueVec();
+
+  mesh_->NewNodes(*reference_nodes_);
 }
 
 void Solid::completeSetup()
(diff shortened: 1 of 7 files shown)


jamiebramwell:
I still need to run this through memcheck, but I think it's working.

jamiebramwell:
LGTM

jamiebramwell:
LGTM

jamiebramwell:
LGTM

jonwong12 (approved):
Thanks. I think it's a good solution.

jamiebramwell:
LGTM

src/serac/physics/solid.hpp
```diff
@@ -175,7 +175,8 @@ class Solid : public BasePhysics {
    * @param[in] options The options for the linear, nonlinear, and ODE solves
    * @param[in] geom_nonlin Flag to include geometric nonlinearities
    */
-  Solid(int order, std::shared_ptr<mfem::ParMesh> mesh, const SolverOptions& options, bool geom_nonlin = true);
+  Solid(int order, std::shared_ptr<mfem::ParMesh> mesh, const SolverOptions& options, bool geom_nonlin = true,
+        bool keep_deformation = false);
```
joshessman-llnl: `keep_deformation` needs a Doxygen annotation

src/serac/physics/solid.cpp
```diff
@@ -20,13 +20,15 @@ namespace serac {
  */
 constexpr int NUM_FIELDS = 2;
 
-Solid::Solid(int order, std::shared_ptr<mfem::ParMesh> mesh, const SolverOptions& options, bool geom_nonlin)
+Solid::Solid(int order, std::shared_ptr<mfem::ParMesh> mesh, const SolverOptions& options, bool geom_nonlin,
```
joshessman-llnl: Having the two `bool`s next to each other in an argument list worries me a little bit, one option would be to use an enum for one or both of these parameters (at least in the interface, for now I think `bool` member variables are fine).

That seems almost like overkill though.

Also, is this something we might want to use with a Lua-configured physics module?  If so, we should update the schema and the other constructor accordingly.

src/serac/physics/solid.cpp
```diff
@@ -122,6 +124,27 @@ Solid::Solid(std::shared_ptr<mfem::ParMesh> mesh, const Solid::InputOptions& opt
   }
 }
 
+Solid::~Solid()
+{
+  // Update the mesh with the new deformed nodes if requested
+  if (keep_deformation_after_destructor_) {
+    reference_nodes_->Add(1.0, displacement_.gridFunc());
```
joshessman-llnl: Minor: `*reference_nodes_ += displacement_.gridFunc();`?

src/serac/physics/solid.hpp
```diff
@@ -311,7 +317,7 @@ class Solid : public BasePhysics {
   /**
    * @brief Destroy the Nonlinear Solid Solver object
    */
-  virtual ~Solid() = default;
```
joshessman-llnl, line 314: I must have jinxed it with my comment yesterday 😆

joshessman-llnl (approved):

src/serac/physics/solid.cpp
jamiebramwell: This is a fairly niche case that I can only image someone using via the C++ interface as you would have to construct multiple `Solid` modules on the same mesh. I agree that the two bools isn't ideal, but I came down on the side of the enums being overkill. I'm happy to change if the consensus is the other method however,

src/serac/physics/solid.cpp
jamiebramwell: I keep forgetting about the expression templates!

src/serac/physics/solid.hpp
jamiebramwell: I blame @jonwong12 for finding this problem!

src/serac/physics/solid.cpp
jamiebramwell: ... I changed my mind. The two bools is bothering me. I will add enums.

jamiebramwell:
LGTM

jamiebramwell:
LGTM

PR #411, opened 2021-03-03
Return meshes by value, expose disk + ball meshes from Inlet

### LINKED ISSUE #406: Use value semantics for the mesh
My preference here would be to have the `buildMeshFromFile` etc functions return an `mfem::Mesh` instead of a smart pointer (since they have no need to be on the heap), but for consistency that would require changing all the other signatures, which I was planning on doing when making all the other shapes available through Inlet.

_Originally posted by @joshessman-llnl in https://github.com/LLNL/serac/pull/398#discussion_r584713987_

joshessman-llnl:
Resolves #406 - possibly this PR should wait until https://github.com/mfem/mfem/pull/2016 is merged, depending on how annoying all the extra copies are.  After that MFEM PR is merged these functions will qualify for copy/move elision.

This PR introduces "box" and "n-ball" to the implementation to refer to the supersets of cuboids/rectangles and balls/disks, respectively - not sure if there are better terms we can use here without sacrificing accuracy.

This PR also allows for the removal of the custom mesh logic from the test driver.


data/input_files/tests/thermal_conduction/static_amgx_solve.lua
```diff
@@ -6,12 +6,12 @@ epsilon = 0.00001
 dt      = 1.0
 
 main_mesh = {
-    type = "file",
-    -- mesh file
-    mesh = "../../../meshes/star_with_2_bdr_attributes.mesh",
+    type = "ball",
+    -- number of elements in the mesh
+    approx_elements = 10000,
```
jamiebramwell, line 11: 👍

src/serac/numerics/mesh_utils.cpp
```diff
@@ -39,7 +39,7 @@ std::shared_ptr<mfem::Mesh> buildMeshFromFile(const std::string& mesh_file)
     SLIC_ERROR_ROOT(err_msg);
   }
 
-  return std::make_shared<mfem::Mesh>(imesh, 1, 1, true);
+  return mfem::Mesh{imesh, 1, 1, true};
```
jamiebramwell, line 42: 🎉

src/serac/numerics/mesh_utils.cpp
```diff
@@ -101,10 +101,12 @@ std::shared_ptr<mfem::ParMesh> buildDiskMesh(int approx_number_of_elements, cons
 
   squish(mesh);
 
-  return std::make_shared<mfem::ParMesh>(comm, mesh);
+  // The copy ctor is marked explicit, but this should qualify for NRVO...
+  // It looks like this (and its analogues) don't have their copies elided...
+  return mfem::Mesh(mesh);
```
jamiebramwell, line 106: We should add an issue (with a link to the MFEM PR) so we don't forget to get rid of this extra copy.

src/serac/numerics/mesh_utils.cpp
```diff
@@ -389,31 +380,40 @@ void InputOptions::defineInputFileSchema(axom::inlet::Table& table)
...
   if (const auto file_opts = std::get_if<FileInputOptions>(&options.extra_options)) {
     SLIC_ERROR_ROOT_IF(file_opts->absolute_mesh_file_name.empty(),
                        "Absolute path to mesh file was not configured, did you forget to call findMeshFilePath?");
-    serial_mesh = buildMeshFromFile(file_opts->absolute_mesh_file_name);
-  } else if (const auto generate_opts = std::get_if<GenerateInputOptions>(&options.extra_options)) {
-    const auto& eles  = generate_opts->elements;
-    const auto& sizes = generate_opts->overall_size;
+    serial_mesh.emplace(buildMeshFromFile(file_opts->absolute_mesh_file_name));
+  } else if (const auto box_opts = std::get_if<BoxInputOptions>(&options.extra_options)) {
+    const auto& eles  = box_opts->elements;
```
jamiebramwell: `elements` or `elems`

src/serac/numerics/mesh_utils.cpp
```diff
@@ -461,7 +461,11 @@ serac::mesh::InputOptions FromInlet<serac::mesh::InputOptions>::operator()(const
       overall_size = std::vector<double>(overall_size.size(), 1.);
     }
 
-    return {serac::mesh::GenerateInputOptions{elements, overall_size}, ser_ref, par_ref};
+    return {serac::mesh::BoxInputOptions{elements, overall_size}, ser_ref, par_ref};
+  } else if (mesh_type == "disk" || mesh_type == "ball") {
+    int approx_elements = base["approx_elements"];
+    int dim             = (mesh_type == "disk") ? 2 : 3;
```
jamiebramwell: I generally don't like using ternary operators unless there's a good reason. I think an if statement is clearer. However, I don't feel extremely strongly about this.

tests/serac_thermal_solver.cpp
```diff
@@ -40,22 +30,17 @@ TEST_P(InputFileTest, thermal_conduction)
...
-TEST(thermal_solver, static_amgx_solve)
-{
-  MPI_Barrier(MPI_COMM_WORLD);
-  std::string input_file_path =
-      std::string(SERAC_REPO_DIR) + "/data/input_files/tests/thermal_conduction/static_amgx_solve.lua";
-  auto pmesh = buildBallMesh(10000);
-  test_utils::runModuleTest<ThermalConduction>(input_file_path, "static_amgx_solve", pmesh);
-  MPI_Barrier(MPI_COMM_WORLD);
-}
+                                   "static_amgx_solve",
```
jamiebramwell, line 35: much cleaner!

jamiebramwell (commented):
I think this is worth merging now even though we have the extra copy with a note that this will go away with the upcoming MFEM PR.

jamiebramwell:
@samuelpmishLLNL , @jonwong12  can you take a look at this?

jamiebramwell (approved):

src/serac/numerics/mesh_utils.hpp
```diff
@@ -205,7 +205,7 @@ struct InputOptions {
    * @brief The mesh input options (either file or generated)
    *
    */
-  std::variant<FileInputOptions, GenerateInputOptions> extra_options;
+  std::variant<FileInputOptions, BoxInputOptions, NBallInputOptions> extra_options;
```
jonwong12, line 208: So in the future, we might have `RingInputOptions` and etc?

jonwong12 (approved):

src/serac/numerics/mesh_utils.hpp
joshessman-llnl: Yep, the idea was that those could be added in as we need them (for tests etc)

joshessman-llnl:
LGTM

joshessman-llnl:
LGTM

PR #414, opened 2021-03-09
Switch to conduit for AMGX configuration

joshessman-llnl:
This PR revises #302 - I think when I put that one together I wasn't aware of the extent of JSON support in Conduit, which we're already building as a TPL.


src/serac/physics/utilities/equation_solver.cpp
```diff
@@ -40,130 +40,28 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverOptions& lin_opt
 
 namespace detail {
 #ifdef MFEM_USE_AMGX
-using JSONLiteral  = std::variant<std::string, double, int>;
```
jamiebramwell, line 43: and another JSON parser bites the dust! 🎉

jamiebramwell (approved):

src/serac/physics/utilities/equation_solver.cpp
```diff
@@ -40,130 +40,28 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverOptions& lin_opt
...
-    sep = ',';
-  }
-  out << "\n" << indent << "}";
-  return out;
-}
-
 std::unique_ptr<mfem::AmgXSolver> configureAMGX(const MPI_Comm comm, const AMGXPrec& options)
 {
   auto              amgx                 = std::make_unique<mfem::AmgXSolver>();
   static const auto default_prec_options = []() {
```
white238: Why are these 2 (was 3) lambdas here?

joshessman-llnl: Copy semantics are different for `conduit::Node`s so it made more sense to just assign the 3 values when in verbose mode

white238 (approved):

joshessman-llnl:
LGTM

Issue #401, opened 2021-02-25
Thermal-only input files don't work


jamiebramwell:
Due to issues with inlet verifiers, thermal-only input files currently do not work.

joshessman-llnl:
It looks like this will be fixed by https://github.com/LLNL/axom/pull/481

jamiebramwell:
Yeah, I looked into fixing this, and decided it was worth it to just wait for this to come down the pipe and do it right.

joshessman-llnl:
Tested + confirmed to work with:

- [x] https://github.com/LLNL/axom/pull/495

PR #415, opened 2021-03-09
Bump axom version to develop (Table -> Container)

joshessman-llnl:
It looks like there are still some slight tweaks that need to be made for #401

### DIFF
--- a/src/serac/infrastructure/input.cpp
+++ b/src/serac/infrastructure/input.cpp
@@ -69,7 +69,7 @@ std::string fullDirectoryFromPath(const std::string& path)
   return dir;
 }
 
-void defineVectorInputFileSchema(axom::inlet::Table& table)
+void defineVectorInputFileSchema(axom::inlet::Container& table)
 {
   // TODO: I had to remove the required tag on x as we now have an optional vector input in the coefficients. IT would
   // be nice to support "If this exists, this subcomponent is required."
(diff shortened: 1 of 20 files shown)


jamiebramwell (approved):

jamiebramwell:
I believe this still needs the LC installs, but otherwise it's good to go!

src/serac/infrastructure/input.cpp
```diff
@@ -168,33 +168,33 @@ std::unique_ptr<mfem::Coefficient> CoefficientInputOptions::constructScalar() co
   }
 }
 
-void CoefficientInputOptions::defineInputFileSchema(axom::inlet::Table& table)
+void CoefficientInputOptions::defineInputFileSchema(axom::inlet::Container& container)
```
white238, line 171: do we want to be consistent on variable names in these functions between container/base?

src/serac/numerics/mesh_utils.hpp
```diff
@@ -149,9 +149,9 @@ struct FileInputOptions {
   /**
    * @brief Input file parameters specific to this class
    *
-   * @param[in] table Inlet's SchemaCreator that input files will be added to
+   * @param[in] container Inlet's SchemaCreator that input files will be added to
```
white238: There is no longer a SchemaCreator class

white238:
I am okay with a follow up handling the continued clarification of the verbage.

white238 (approved):

src/serac/infrastructure/input.cpp
joshessman-llnl: I think the current convention is that `defineSchema` et al uses `container` while the `FromInlet` specializations use `base`, though I wouldn't have a problem with changing that.  I don't think either of them are as descriptive as they could be.

joshessman-llnl:
LGTM

PR #417, opened 2021-03-11
Upgrade to Spack's clingo concretizer

### LINKED ISSUE #327: Add Spack variants for optional packages
For example, Sundials could be optional but this requires the new clingo Spack concretizer to work around the bugs in the current Spack concretizer.

### LINKED ISSUE #428: Update conduit to 0.6.0
We are slightly behind and Axom has already updated to 0.6.0. We need to stay in sync.

joshessman-llnl:
Fixes #327
Fixes #428 

This PR switches to Spack's new concretizer which is based on the `clingo` answer set programming library.

Getting `clingo` on LC required binary packaging with the PyPA toolchain: https://github.com/joshessman-llnl/package-clingo-manylinux which is managed by testPyPI: https://test.pypi.org/project/clingo-cffi-wheel/
The next clingo release will be part of the "real" PyPI through clingo's maintainer, so this approach *should* be temporary.

~~Requires an Axom branch  (which itself requires a BLT branch) and a RAJA fork - presumably we'd want these changes to get merged in before merging this PR.~~

Upstream PRs:
 - [x] https://github.com/LLNL/axom/pull/494
 - [x] https://github.com/LLNL/blt/pull/449
 - [x] https://github.com/LLNL/RAJA/pull/999
 - [x] https://github.com/LLNL/uberenv/pull/62 
 - [ ] https://github.com/spack/spack/issues/23259


joshessman-llnl:
LGTM

scripts/llnl/common_build_functions.py
```diff
@@ -125,6 +125,40 @@ def log_failure(prefix, msg, timestamp=""):
...
+def setup_clingo():
+    """
+    Attempts to install the clingo answer set programming library
+    if it is not already available as a Python module
+    """
+    try:
+        import clingo
+    except ModuleNotFoundError:
+        import pip
+        pip_ver = pip.__version__
```
white238: Should this be moved into uberenv?

scripts/spack/configs/blueos_3_ppc64le_ib_p9/packages.yaml
```diff
@@ -134,7 +134,7 @@ packages:
     version: [10.0.0]
     buildable: false
     externals:
-    - spec: llvm+clang
+    - spec: llvm+clang@10.0.0
```
white238, line 137: The version above this is supposed to handle this.  Was this a problem?

scripts/spack/configs/toss_3_x86_64_ib/packages.yaml
```diff
@@ -26,9 +26,9 @@ packages:
     buildable: false
   mvapich2:
     externals:
-    - spec: mvapich2@2.3%clang@10.0.0 process_managers=slurm arch=linux-rhel7-ivybridge
+    - spec: mvapich2@2.3%clang@10.0.0 arch=linux-rhel7-ivybridge process_managers=slurm
```
white238: This ordering matters?

scripts/spack/packages/axom/package.py
```diff
@@ -251,7 +251,13 @@ def hostconfig(self, spec, prefix):
             flags = ""
             for _libpath in [libdir, libdir + "64"]:
                 if os.path.exists(_libpath):
-                    flags += " -Wl,-rpath,{0}".format(_libpath)
+                    # SERAC EDIT BEGIN - BLT_EXE_LINKER_FLAGS aren't filtered
+                    # for the Wl/Xlinker incompability
+                    if "+cuda" in spec:
+                        flags += " -Xlinker -rpath -Xlinker {0}".format(_libpath)
```
white238, line 171: Can you move these back to axom/spack proper?

scripts/spack/packages/serac/package.py
```diff
@@ -115,35 +118,44 @@ class Serac(CMakePackage, CudaPackage):
...
     depends_on("mfem+petsc", when="+petsc")
+    depends_on("mfem+sundials", when="+sundials")
+    depends_on("sundials~shared", when="+sundials")
+    depends_on("netcdf-c@4.7.4~shared", when="+netcdf")
 
     # Needs to be first due to a bug with the Spack concretizer
     depends_on("hdf5+hl@1.8.21~shared")
 
+    # Axom enables RAJA/Umpire by default
+    depends_on("axom~raja", when="~raja")
```
white238, line 109: Should we be toggling this based on the cuda variant instead of individually?

scripts/spack/configs/blueos_3_ppc64le_ib_p9/packages.yaml
joshessman-llnl: They serve different purposes - the first one tells Spack that it needs to be using LLVM 10, the second one tells Spack that the external LLVM installation is version 10.  The new concretizer requires that all externals have a version associated with it (hence all the changes).

scripts/spack/packages/serac/package.py
joshessman-llnl: I think it makes sense to keep them as a separate variant, if we start using e.g. RAJA in serac code directly we would probably want it to still be compiled when GPUs are not available, and just execute serially with RAJA in that case.

scripts/llnl/common_build_functions.py
joshessman-llnl: I can open an Uberenv PR, presumably we'd want the concretizer to be configurable via just the JSON file?

scripts/spack/configs/toss_3_x86_64_ib/packages.yaml
joshessman-llnl: I'll revert this, I had to remove and then re-add this as a spack bug was found and then fixed.

scripts/llnl/common_build_functions.py
white238: yes please

.uberenv_config.json
```diff
@@ -5,8 +5,9 @@
 "package_source_dir" : "../..",
 "force_commandline_prefix" : false,
 "spack_url": "https://github.com/spack/spack",
-"spack_commit": "e0e6a0199ef4db473ea0bd2e924f3e96fa10297a",
+"spack_commit": "1a080b9a00a7952dd636d9eba7137e61ae4af4b6",
 "spack_configs_path": "scripts/spack/configs",
 "spack_packages_path": "scripts/spack/packages",
-"spack_activate" : {}
+"spack_activate" : {},
+"use_clingo": true
```
white238: Please prefix this with "spack_" like all the other spack options.  Also is "use_clingo" descriptive enough without having to understand the internals of spack? spack_use_new_concretizer?

scripts/spack/configs/blueos_3_ppc64le_ib_p9/packages.yaml
```diff
@@ -72,37 +72,37 @@ packages:
...
   libtool:
     externals:
-      - spec: libtool
+      - spec: libtool@2.4.2
         prefix: /usr
     buildable: false
   m4:
     externals:
-      - spec: m4
+      - spec: m4@1.4.16
```
white238, line 100: @becker33 is there any way around this? Bumping these versions is going to be tedious.  Not to mention this probably isn't accurate across multiple machines...

becker33: @white238 I think you should be able to remove that entirely and run `spack external find --scope site --not-buildable m4` to generate the config for m4. That won't be true for every package you have in these files, but it is true for many of them (basically all of the build tools, for example).

becker33: You might not even need the scope argument, I'm not remembering exactly how you munge the configs inside uberenv, so I was trying to be careful with it.

white238: I guess I view these kind of differently.  More like "these are system libraries and I don't care what version they are".  We will see in the future how much this causes a maintenance nightmare.

becker33: The problem is there are occasionally packages that actually care what version it is. I don't think I've seen that for m4, but I have for automake, autoconf, and bison.

What the old concretizer did was just assume it was the newest version Spack was aware of if it didn't have a version -- if that was a problem, you just got heinous errors, but that was relatively uncommon. I can look into how hard that would be to implement in the new concretizer.

Actually, maybe a better solution would be to allow externals not to have a version -- then you get an error if any package cares what version it was, but otherwise it can be fine without. That would probably be pretty complicated for us to implement, though, so that's a downside.

.uberenv_config.json
```diff
@@ -2,11 +2,12 @@
 "package_name" : "serac",
 "package_version" : "develop",
 "package_final_phase" : "initconfig",
-"package_source_dir" : "../..",
+"package_source_dir" : ".",
```
white238: I'm pretty sure this was an unintentional change uberenv.  Do you know the line that caused this?

scripts/llnl/common_build_functions.py
```diff
@@ -160,7 +159,7 @@ def uberenv_build(prefix, spec, project_file, mirror_path):
     Calls uberenv to install tpls for a given spec to given prefix.
     """
     assertUberenvExists()
-    cmd  = "python {0} -k ".format(get_uberenv_path())
+    cmd  = "{0} {1} -k ".format(sys.executable, get_uberenv_path())
```
white238, line 162: You are just running into all the problems this PR arent you....

scripts/spack/packages/axom/package.py
```diff
@@ -212,6 +227,13 @@ def initconfig_hardware_entries(self):
                     cudaflags += " -std=c++14"
                 else:
                     cudaflags += " -std=c++11"
+
+                # SERAC EDIT BEGIN - debug flag causes the NVIDIA assembler to fail
```
white238: is this debug comment still relavent?

scripts/spack/packages/axom/package.py
```diff
@@ -212,6 +227,13 @@ def initconfig_hardware_entries(self):
...
+                # NVCC ignores the host compiler when linking??
+                # and some MPI -Wl,-rpath, flags are added from somewhere
+                cudaflags += " -ccbin ${CMAKE_CXX_COMPILER} -forward-unknown-to-host-compiler "
```
white238, line 234: I'm concerned why this is now showing up when it never has before.  What was led to this fix?

.uberenv_config.json
joshessman-llnl: It looks like this was a result of a change related to https://github.com/LLNL/uberenv/issues/44

With the current (old) version of uberenv we use, hostconfigs end up in the directory from which we run uberenv/build_tpls (the repo root) and in the TPL directory (/usr/WS2/smithdev/libs/serac/blueos_3_ppc64le_ib_p9/2021_04_21_14_18_07 for example).  It seems that placing hostconfigs in the repo tree is undesirable (per the above issue) and was removed by [this line](https://github.com/LLNL/uberenv/pull/46/files#diff-815c7c31845f6d60a7744f452038d9aacb5b2614e3005e5401ea3ecb0c395951R640), which uses `package_source_dir` relative to the `--prefix` and not the directory from which uberenv was run.

scripts/llnl/common_build_functions.py
joshessman-llnl: To be fair this one was (mostly) my fault since there's some Python2-incompatible logic I added to uberenv.

scripts/spack/packages/axom/package.py
joshessman-llnl: So this was not showing up for us before because I had punted (in #254) on separable compilation, which axom+raja+cuda requires (not req'd by axom+cuda~raja).  I think it didn't show up for other Axom users because we're the only ones who use axom+cuda but don't build our project with NVCC/as CUDA C++ (which necessitated LLNL/axom#494).

white238: I thought cmake added the ccbin option automatically due to setting `CMAKE_CUDA_HOST_COMPILER` which we do through BLT here:

https://github.com/LLNL/blt/blob/68c98b7306e0b1f19cb383ac6af210d6f75627fc/cmake/thirdparty/SetupCUDA.cmake#L33

scripts/spack/packages/serac/package.py
white238: should we add a conflict for now for +cuda and ~raja~umpire?

scripts/spack/packages/axom/package.py
joshessman-llnl: I would have thought so as well, but my guess is that that doesn't apply to the link stage somehow (hence the comment)

joshessman-llnl:
This is ready to go, but should we hold out on MFEM 4.3 and/or Axom 0.5.0?

scripts/spack/devtools.json
```diff
@@ -4,7 +4,7 @@
 "package_source_dir" : "../..",
 "package_final_phase" : "",
 "spack_url": "https://github.com/spack/spack",
-"spack_commit": "e0e6a0199ef4db473ea0bd2e924f3e96fa10297a",
+"spack_commit": "e75b76f433016a720027b41c0d36cdb9ef24a827",
```
white238, line 7: Can you bump this to the sha that went in with the OSX branch?  I am okay with not rebuilding everything.

scripts/spack/packages/mfem/package.py
```diff
@@ -99,8 +99,10 @@ class Mfem(Package):
     variant('openmp', default=False,
             description='Enable OpenMP parallelism')
     variant('cuda', default=False, description='Enable CUDA support')
+    # SERAC EDIT BEGIN - need to explicitly enumerate values
     variant('cuda_arch', default='sm_60',
-            description='CUDA architecture to compile for')
+            description='CUDA architecture to compile for', values=['sm_' + val for val in CudaPackage.cuda_arch_values])
```
white238, line 104: Can you push these upstream?

scripts/spack/packages/serac/package.py
```diff
@@ -57,12 +57,14 @@ class Serac(CachedCMakePackage, CudaPackage):
             description='Build the glvis visualization executable')
     variant('petsc', default=False,
             description='Enable PETSC')
-    # netcdf and sundials variants commented out until a bug in the spack
-    # concretizer is fixed
-    #variant('netcdf', default=True,
-    #        description='Enable Cubit/Genesis reader')
-    #variant('sundials', default=True,
-    #        description='Build MFEM with SUNDIALS nonlinear/ODE solver')
+    variant('netcdf', default=True,
```
white238, line 60: What is this brave new world where variants actually work?!?!

white238 (approved):

jonwong12:
Looks good. I'll run it on my WSL machines tonight just to verify.

jonwong12:
So the the only thing I needed was `python3-pip` in order to grab clingo. Can we add `python3-pip` or `python-pip` to the WSL sphinx documentation? Otherwise using the clingo concretizer seems to work better than `develop` on my wsl2-build.

jonwong12 (changes requested):
Just add `python/3-pip` package to the required dependencies in sphinx and it seems to work on WSL 2 UBuntu 20.04 for me.

jamiebramwell (approved):
Works on my system! Thanks for taking the time to figure out all of this!

scripts/spack/configs/config.yaml
```diff
@@ -33,48 +42,163 @@ config:
...
+  # Cache directory for already downloaded source tarballs and archived
+  # repositories. This can be purged with `spack clean --downloads`.
   source_cache: $spack/var/spack/cache
 
 
   # Cache directory for miscellaneous files, like the package index.
-  # This can be purged with `spack purge --misc-cache`
-  misc_cache: .spack/misccache
+  # This can be purged with `spack clean --misc-cache`
+  misc_cache: ~/.spack/cache
```
adrienbernede: I recently ran into issues with environments where the misc_cache needed to be cleaned because I was switching to an older spack.
In order to keep Uberenv "bullet-proof" I think the misc_cache should remain uberenv-local, e.g.
```suggestion
  misc_cache: $spack/.misc_cache
```

adrienbernede (changes requested):
I found a minor thing to complain about...

jonwong12 (approved):

scripts/spack/configs/config.yaml
joshessman-llnl: Good catch, I think I may have missed this when updating the `config.yaml`

scripts/spack/devtools.json
joshessman-llnl: It's now at the hdf5 typo commit - https://github.com/spack/spack/commit/6394604d0adb57d04b715a6d293ef4b18c9c26ef

adrienbernede (approved):

.gitlab/build_lassen.yml
```diff
@@ -15,6 +15,10 @@
 .with_cuda:
   before_script:
     - module load cuda/11.1.1
+    # python3.8 is needed on lassen to avoid trampling on the x86 clingo wheel
+    # This is not the ideal place for it but the pipeline syntax doesn't allow
+    # for multiple before_scripts - see https://gitlab.com/gitlab-org/gitlab-runner/-/issues/2301
```
adrienbernede: That’s true:
Yaml lists are not merged when used in the `extends` mechanism. True for script, before/after_script, rules, even tags.

An alternate solution would be to have a top level before script (one per machine for example) and adapt the behavior depending on the job name (requires a convention) or maybe better, on variables:
```yaml
.on_lassen:
  before_script:
    - module load python/3.8
    - if [[ $CUDA == "ON" ]]; then module load cuda/11.1.1; fi

.with_cuda:
  variables:
    CUDA: "ON"
```
That’s because variables can be merged through `extends`.

PR #419, opened 2021-03-12
Discuss Entity sets

jonwong12:
This is a first pass at trying to provide some `Set` functionality to help handle the complexity of organizing different sets of material properties and boundary conditions.

- Former `mfem_ext` helper functions were templated to output both `mfem::Array` (which is currently used in serac) and `std::vector` to be used with the `Set` abstraction. (This relies on #437.)
- Set operations are computed and converted back to a dense representation.

Two simple tests were implemented. One just evaluates different types of set operations. The second test creates a flag by using Sets to facilitate creation of a flag. 

Internally a list of attributes is converted to an unordered_map<attribute, index list>. So `Set<int>({1, 2, 4, 3})` gets converted to ```
{ {1, {0}}, 
  {2, {1}},
  {4, {2}}, 
  {3, {3}}}```. 

This allows us to grab a list of elements that have a particular attribute and apply set operations to those lists. Alternatively we can selectively assemble or perform other operations using this list as input.

Note: It might be worthwhile in the future to convert this implementation to better take advantage of c++20 ref_views.

TODO:
- [ ] Psuedocode example for a physics problem


src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,151 @@
...
+}
+
+template <>
+typename eval_t<mfem::Coefficient>::type eval<mfem::Coefficient>(mfem::Coefficient& c, mfem::ElementTransformation& Tr,
+                                                                 const mfem::IntegrationPoint& ip);
+
+template <>
+typename eval_t<mfem::VectorCoefficient>::type eval<mfem::VectorCoefficient>(mfem::VectorCoefficient&      v,
+                                                                             mfem::ElementTransformation&  Tr,
+                                                                             const mfem::IntegrationPoint& ip);
```
joshessman-llnl: I don't think these need to be template specializations - they can just be overloads (which will take higher priority than the templated fallback above)

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,151 @@
...
+template <typename T>
+struct index_t<mfem::Array<T>> {
+  using type = int;
+};
+
+// methods for determining the size of a container
+template <typename T>
+int size(T container)
+{
+  return static_cast<int>(container.size());
```
white238: You are probably losing precision here.  Most stl `size()` calls return `size_t` which is normally `unsigned int`

jonwong12: @white238 You're absolutely right. I should probably add a comment, but the original reason why I moved to int is because mfem's size container returns `int`. I can try moving to `auto` for the std collections though.

white238: you can promote int to unsigned in safely so they are both the same..

jonwong12: That's true, but if I use implicit conversions I get the following error:
`error: conversion from 'std::vector<int>::size_type' {aka 'long unsigned int'} to 'int' may change value [-Werror=conversion]`

scripts/spack/packages/mfem/package.py
```diff
@@ -525,11 +525,11 @@ def find_optional_library(name, prefix):
...
-            hdf5 = spec['hdf5:hl']
-            if hdf5.satisfies('~shared'):
-                hdf5_libs = hdf5.libs
-                hdf5_libs += LibraryList(find_system_libraries('libdl'))
-                lib_flags += " " + ld_flags_from_library_list(hdf5_libs)
+#            hdf5 = spec['hdf5:hl']
+#            if hdf5.satisfies('~shared'):
+#                hdf5_libs = hdf5.libs
+#                hdf5_libs += LibraryList(find_system_libraries('libdl'))
+#                lib_flags += " " + ld_flags_from_library_list(hdf5_libs)
```
joshessman-llnl: This worries me a little - Spack will still be building an HDF5 (for conduit etc), but the system package will be using a system-provided HDF5, and I'm a little hesitant about mixing two HDF5s.  What was the motivation for switching to the system-provided netcdf package?

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,151 @@
...
+
+template <typename T>
+struct eval_t<T, std::enable_if_t<std::is_base_of_v<mfem::Coefficient, T>>> {
+  using type = double;
+};
+
+template <typename T>
+struct eval_t<T, std::enable_if_t<std::is_base_of_v<mfem::VectorCoefficient, T>>> {
+  using type = mfem::Vector;
+};
```
joshessman-llnl: I think if `eval` switches to overloads, `eval_t` is not needed - there can just be Coef/VecCoef overloads with a fallback `T eval(T&...)`

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -65,8 +175,42 @@ mfem::Array<int> MakeEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoef
...
  */
-mfem::Array<int> MakeAttributeList(
-    mfem::Mesh& m, mfem::Coefficient& c, std::function<int(double)> digitize = [](double v) { return v > 0. ? 2 : 1; });
+
+template <typename T>
+T MakeAttributeList(mfem::Mesh& m, mfem::Coefficient& c, std::function<int(double)> digitize = digitize::threshold)
+{
+  mfem::L2_FECollection    l2_fec(0, m.SpaceDimension());
+  mfem::FiniteElementSpace fes(&m, &l2_fec);
+  T                        attr_list(static_cast<typename detail::index_t<T>::type>(fes.GetNE()));
```
joshessman-llnl: Probably want to document the implicit precondition on `T` here - presumably this requires that `T` be some sort of container with a certain constructor and begin/end functionality?

These could also be enforced with a `static_assert` - see e.g. `std::is_constructible`

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -65,8 +175,42 @@ mfem::Array<int> MakeEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoef
...
+ *
+ * @param[in] m the mesh
+ * @param[in] attr_list a list of attributes to assign to the given mesh
+ */
+
+template <typename T>
+void AssignMeshElementAttributes(mfem::Mesh& m, T&& list)
+{
+  // check to make sure the lists are match the number of elements
+  SLIC_ERROR_IF(detail::size(list) != static_cast<std::size_t>(m.GetNE()), "list size does not match the number of mesh elements");
```
joshessman-llnl: Also want to document the preconditions on `T` here

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -85,16 +229,52 @@ mfem::Array<int> MakeAttributeList(
  * attribute 2.
  * @return An array holding the attributes that correspond to each element
  */
-mfem::Array<int> MakeBdrAttributeList(
-    mfem::Mesh& m, mfem::Coefficient& c,
-    std::function<int(double)> digitize = [](double v) { return v == 1. ? 2 : 1; });
+template <typename T>
+T MakeBdrAttributeList(mfem::Mesh& m, mfem::Coefficient& c, std::function<int(double)> digitize = digitize::equals1)
```
joshessman-llnl: My recollection of the discussion on templates here was that it would be desirable to write these functions in a generic way such that they could be used outside the MFEM ecosystem at some point (i.e., independent of `mfem::Vector`).  It looks like most of these functions are highly dependent on other MFEM domain objects (FESpace etc), though - possibly I'm missing something?

src/serac/infrastructure/set.hpp
```diff
@@ -0,0 +1,324 @@
...
+  friend bool operator==(const Set<T>& s1, const Set<T>& s2)
+  {
+    bool isSame = (s1.keys == s2.keys);
+    if (isSame) {
+      for (auto k : s1.keys) {
+        isSame = isSame && (s1.values_index_list.at(k) == s2.values_index_list.at(k));
+        if (!isSame) break;
+      }
+    }
+    return isSame;
```
joshessman-llnl: `isSame` -> `is_same`

Minor: the control flow can probably be cleaned up a little bit since I don't think we're worried about multiple returns

scripts/spack/packages/mfem/package.py
jonwong12: Netcdf/hdf5 failed to compile on WSL 2. Somehow it failed to find the C compiler 20 packages down the concretizer. 

It would be nice to have a method of better overriding netcdf when things just aren't working. I should probably move this packaging stuff into another PR.

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: Previously this method returned a `mfem::Array`. In this case I'm just allowing `std::vector` to be returned instead so it can be used as input to `Set` which lives outside of the mfem ecosystem.

I think the discussion to make this work better outside of the MFEM ecosystem is a slightly separate enhancement? This is more of an incremental move using the MFEM ecosystem as reference.

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: I think `eval_t` is needed for the `Transformed*Coefficient` modification which takes in a std::function and corresponding args instead of several fixed constructors.

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: Changed in #437

jamiebramwell:
@jonwong12 , should we start looking at this now that the templated MFEM extensions branch is merged?

src/serac/coefficients/coefficient_extensions.cpp
```diff
@@ -11,141 +11,8 @@
 
 namespace serac::mfem_ext {
 
-mfem::Array<int> MakeTrueEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoefficient& c)
-{
-  mfem::Array<int> ess_tdof_list(0);
+namespace detail {
```
white238: was this, now empty, file left for a reason?  I can think of a few reasons for an empty file but not in this case.

jonwong12: Ahh yes. I need to update this branch with what's recently merged into develop. It should disappear.

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,170 @@
...
+}  // namespace detail
+
+namespace digitize {
 /**
- * @brief MakeTrueEssList takes in a FESpace, a vector coefficient, and produces a list
- *  of essential boundary conditions
+ * @brief takes floating point value and rounds down to the nearest integer
+ * @param[in] v floating point value
+ */
+
```
white238: Remove this empty line. Also capitalize Takes

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,170 @@
...
+template <typename T, typename = void>
+struct has_brackets : std::false_type {};
+  
+template <typename T>
+struct has_brackets<T, std::void_t<decltype(std::declval<T>()[0])>>
+    : std::true_type {};
+  
+}  // namespace detail
+
+namespace digitize {
```
white238, line 131: I am not super sold on this name for the namespace.  Also these seem more like general utility functions that should go in a common place like `infrastructure` or if they are specific enough for this file, `detail` is fine.

tests/CMakeLists.txt
```diff
@@ -29,7 +29,8 @@ if (ENABLE_GTEST)
         serac_boundary_cond.cpp
         serac_mesh.cpp
         mfem_ex9p_blockilu.cpp
-        serac_newmark_test.cpp)
+	serac_newmark_test.cpp
```
white238: indenting.

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: Good point. Currently it's only used in conjunction with coefficient_extensions, but they have mostly been provided for convenience. All of the functions take in a `double` and return an `int`.  Let me know if there's a better namespace.

tests/serac_sets.cpp
```diff
@@ -0,0 +1,296 @@
...
+  void SetUp() override {}
+};
+
+namespace serac {
+
+TEST_F(SetTest, set)
+{
+  Set<int> a1(std::vector<int>{1, 2, 4, 2, 2, 3, 4});
+  std::cout << "Set a1:" << std::endl << a1 << std::endl;
+  EXPECT_EQ(a1, Set<int>({1, 2, 4, 2, 2, 3, 4}));
```
white238, line 41: Is this tested somewhere else to make sure you aren't getting duplicates in this case?

```
EXPECT_EQ(a1, Set<int>({1, 2, 4, 3}));
```

tests/CMakeLists.txt
jonwong12: Sorry. I really need to change my default emacs settings.

tests/serac_sets.cpp
jonwong12: I'll try to provide a better explanation of what `Set` is trying to do in the PR summary.
But essentially `Set<int>({1, 2, 4, 3})` gets converted into a std::unordered_map<attribute, index list> which would be `{ {1, {0}}, {2, {1}}, {4, {2}}, {3, {3}}`. This allows us to grab a list of elements that have a particular attribute and apply set operations to those lists.

jonwong12: The `a3` example a few lines down does test for a sort of duplicate key case.

`Union` is currently defined such that in the case where there are two different sets that have an overlapping offset, the attribute the overlapping offset will be assigned to will be the attribute key that is greater. 

We could throw an error in that case, but this would require users to check using intersection/difference/complement before executing particular union calls. `Set` also tries to make it easy to create new sets from extracted subsets.

white238: Should this class be called `EntitySet` then to not clash/cause confusion with `std::set`?

jamiebramwell:
I'm still having some issues understanding how this would be used in a real simulation. Could you describe a realistic use case (materials, boundary conditions, etc)?

jonwong12:
@jamiebramwell Great question. I think being able to go to and from attribute lists to subsets is generally useful. I think it would be useful for dispatching GPU kernels. 

I'm not sure about realistic scenarios, but if you had an example of a multi-material problem that would be difficult in MFEM, perhaps I could sketch out how it could be used.

For multiphysics problems you can run into the combinatorial issues. A hypothetical multiphysics thermal mechanical problem might have thermal/non-thermal attributes and 4 different material properties (anisotropic elastic, isotropic elastic, orthotropic, meta-material). You could use 8 different attributes (combinatorial) or you could have 6 different base attributes. You just have more control over what you want to compute when you can perform sets. You can always form combinatorial sets manually of course, but I think being able to perform set operations on attribute lists helps you pin point exactly which elements/surfaces/nodes/vertices you want to compute on.

tests/serac_sets.cpp
```diff
@@ -0,0 +1,296 @@
...
+      *pmesh, FiniteElementState::Options{.order      = 0,
+                                          .vector_dim = 1,
+                                          .coll = {std::make_unique<mfem::L2_FECollection>(0, pmesh->SpaceDimension())},
+                                          .ordering = mfem::Ordering::byVDIM,
+                                          .name     = "stripes",
+                                          .alloc_gf = true});
+  stripes.project(stripes_coef);
+
+  auto stripes_attr_set =
+      Set(mfem_ext::MakeAttributeList<std::vector<int>>(*pmesh, stripes_coef, mfem_ext::digitize::floor));
```
jonwong12, line 140: The overall idea behind `Set` is to create a pair between an index of _something_ and an attribute (currently an integer), and then provide set operations on these pairs based on their attributes. Conversion to and from a vector of attributes or a dictionary from attributes -> index lists with the same attribute are also provided.

Since we've moved away from using coefficients, one might want to update `MakeAttributeList` which can return a container containing the attributes associated from index 0 to `N`. So if N = 5, the resulting `std::vector<int>` might return `{0, 1, 2, 2,1}`, which represents the following pairs of indices and attributes. `{0, 0}, {1,1}, {2,2}, {3,2}, {4,1}`. Internally `Set` converts the pairwise index -> attribute representation to a mapping from attributes to indices (e.g. `{0,{0}}, {1, {1,4}}, {2, {2,3}}` which we'll call a sparse representation).

`Set` doesn't require using coefficients, either the dense `std::vector<int>` is sufficient or the sparse _dictionary_ should be sufficient. One modification is to replace `mfem::Coefficient` with a lambda that returns the attribute given a particular rank-local element/_thing_ index.

tests/serac_sets.cpp
```diff
@@ -0,0 +1,296 @@
...
+    if (fabs(x[1] - 0.) < 1.e-5) {
+      return 2.;
+    } else if (fabs(x[0] - flag_width) < 1.e-5) {
+      return 3.;
+    }
+    return 1.;
+  });
+
+  mfem_ext::AssignMeshBdrAttributes(
+      *pmesh, mfem_ext::MakeBdrAttributeList<std::vector<int>>(*pmesh, bdr_assign_coef, mfem_ext::digitize::floor));
```
jonwong12, line 165: `AssignMeshBdrAttributes` is a helper method that modifies the boundary element attributes in the mesh to match that of the provided dense vector on a `mfem::ParMesh`.

tests/serac_sets.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+  // we'll get the top blue corner
+  auto blue = stars_attr_set.getComplement({1});
```
jonwong12, line 168: This method finds all the index - attribute pairs that are the complement of `1` to get the "blue/stars" component of domain defined by `stars_coef` (only element indices with attributes 2,3). 

The `stripes_attr_set` contains the entire domain where element index attributes of 1 ("red") and 2 ("white") are assigned. `getDifference` provides the set difference of the entire domain and the blue portion of the flag and yields only the ("red/white") element indices that do not overlap with the top left blue portion.

Lastly we combine the stars with the stripes to get a domain which has elements that look like a flag (red = 1, white = 2, blue = 3). Both coefficients used defined white = 2, so when we `getUnion` on the attribute/index pairs we get the result we want.

The `set` infrastructure might be better extended to better allow for "remapping" of different attribute subsets. Currently I'm just copying dictionaries and list collections, but view semantics (or pairs of attributes - references) might be more efficient.

tests/serac_sets.cpp
```diff
@@ -0,0 +1,296 @@
...
+  // integrate on the boundary of the bottom
+  mfem::ParFiniteElementSpace& fes = *density.gridFunc().ParFESpace();
+  mfem::ParGridFunction        one_gf(&fes);
+  one_gf = 1.;
+
+  mfem::Array<int> bottom_marker(3);
+  bottom_marker    = 0;
+  bottom_marker[1] = 1;
+
+  auto surface_density_coef = mfem_ext::SurfaceElementAttrCoefficient(*pmesh, density_coef);
```
jonwong12, line 209: Sometimes it's useful to be able to get "domain" elements attributes from their associated surface (bdr elements). This coefficient does that, and if we move away from using coefficients, it would be useful to have this information available in some form.

Some examples of how this is used is if you specifically only wanted to integrate on the surface of bdr elements associated with volume elements with a specific attribute.

tests/serac_sets.cpp
```diff
@@ -0,0 +1,296 @@
...
+  right_marker[2] = 1;
+
+  mfem::ParLinearForm right(&fes);
+  right.AddBoundaryIntegrator(new mfem::BoundaryLFIntegrator(surface_density_coef), right_marker);
+  right.Assemble();
+  double integrate_right = right(one_gf);
+  EXPECT_NEAR(6. / 11., integrate_right, 5.e-2);
+
+  // Reverse attributes: white = 1, red = 2, blue = 3 and integrate on right side
+  std::vector<int> orig_elem_attr = flag.toList();
```
jonwong12, line 246: Here `orig_elem_attr` is a dense vector list of attributes from indices 0 to `N` element no. We can use `std::transform` to apply a lambda on the dense list to swap _red_ with _white_. The example then shows how we can "hot swap" the reversed red/white coefficient using the `mfem::Integrator` infrastructure and perform an integration on the boundary.

In the move away from coefficients, I think it would be helpful to have an abstraction like `Set` that handles `indices <-> attribute` mappings. `Functional` could then take in a `Set` as an argument, and you would integration over certain attribute/index pairs directly instead of masking. It also might be useful to distinguish between `domain` attributes, `boundary` attributes, and maybe `quadrature`-point attributes using something like types. Based on boundary <-> domain mappings you might be able to make it easier to integrate boundary indices associated with domain indices behind the scenes and avoid having to write something like `SurfaceElementAttrCoefficient`.

jonwong12: The abstract `Set` type could be useful when launching material kernels where the material properties are batched (GPU or ROM-based elements that benefit from caching). Since `Set` really is just an abstraction layer for `(index <-> attribute)` pairwise mappings and operations on those, the information can be reused in a lot of different ways.

Some of these operations could be used in determining the actual _consistent_ indexing of domain and boundary d.o.f.s for example and is especially useful if you want to assemble on the GPU.

jonwong12 (commented):
Some thoughts on extending entity sets to be used with functional.

PR #426, opened 2021-03-17
FIX: CI quartz skip and resource release

adrienbernede:
Two small fixes for the CI:

- When `needs` is not present (default behavior), a job expects all the pervious stages to pass. In order to allow quartz jobs to run independently of lassen status (and in parallel) I added a `needs` relationship with the allocation job.
- I also fixed the `release_resources` job conditional execution so that it will always free resources even on failure.


jamiebramwell (approved):
Thanks for fixing this!

joshessman-llnl (approved):
Thanks!

PR #432, opened 2021-03-29
Change default output and reorient tet mesh

jamiebramwell:
This changes the default output type to GLVis which does not require initialization. It also reorients the tet-based ball mesh for use with high order H curl elements.


jamiebramwell:
LGTM

jonwong12 (approved):
It looks fine to me. It would be nice in serac if we could return a warning if the "root_name" already is being used. I'm not sure if there is some mechanism to check this already in mfem.

joshessman-llnl (approved):

jamiebramwell:
LGTM

jamiebramwell:
LGTM

PR #434, opened 2021-04-13
Add option to select language/sidre path for Inlet

joshessman-llnl:
This change generalizes Serac's wrapper over Inlet to support input files other than the "primary" lua file.


jamiebramwell (approved):

white238:
Can we turn this on for the serac driver?

white238 (approved):

jamiebramwell:
> 
> 
> Can we turn this on for the serac driver?

I believe it is since we don't pass the language argument in the driver and it defaults to Lua.

jamiebramwell:
LGTM

jamiebramwell:
> 
> 
> > Can we turn this on for the serac driver?
> 
> I believe it is since we don't pass the language argument in the driver and it defaults to Lua.

Wait, you were asking to turn this on instead of off. I need more ☕

PR #437, opened 2021-04-15
Templated mfem_ext functions

jonwong12:
This PR templates methods the formerly just worked for `mfem::Array` so that they could be used with `std::vector` and possibly other std collections.

Some notable changes:
- Added `AssignMeshAttributes` which sets element and bdr attributes according to an attribute list (`mfem::Array` or e.g. `std::vector`)
- Transformed*Coefficient have been changed so that they take in a std::function and a list of corresponding args (`double, mfem::Coefficient, mfem::VectorCoefficient`). This allows us to do specialized one-off dot-products and things not defined in MFEM without needing a combinatorial number of constructors.
- Changed `AttributeModifier*Coefficient` so can temporarily modify mfem's view of the attribute list (`mfem::Array` or `std::vector`).
- Added `SurfaceVectorElementAttrCoefficient` which can evaluate an element-attribute-based coefficient on the surface (e.g. using a BdrIntegrator)


src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,170 @@
...
 #include <functional>
 #include <memory>
-
+#include <variant>
+#include <vector>
 #include "mfem.hpp"
 
 #include "serac/numerics/expr_template_ops.hpp"
-
+#include "serac/physics/utilities/finite_element_state.hpp"
```
joshessman-llnl: What is this included for? If we want to have the coefficients depend on the physics_utilities, that dependency should probably be propagated to the CMakeLists.

src/serac/coefficients/coefficient_extensions.cpp
```diff
@@ -11,141 +11,8 @@
 
 namespace serac::mfem_ext {
 
-mfem::Array<int> MakeTrueEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoefficient& c)
-{
-  mfem::Array<int> ess_tdof_list(0);
+namespace detail {
```
joshessman-llnl: It looks like this file can be deleted entirely?

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,170 @@
...
+};
+
+template <typename T>
+struct index_t<mfem::Array<T>> {
+  using type = int;
+};
+
+// methods for determining the size of a container
+template <typename T>
+auto size(T container)
```
joshessman-llnl: To avoid making copies this should be changed to a [universal reference](https://isocpp.org/blog/2012/11/universal-references-in-c11-scott-meyers), i.e., `T&& container`.

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,170 @@
...
+
+template <typename T>
+std::size_t size(mfem::Array<T> container)
+{
+  return container.Size();
+}
+
+// methods for determining type of coefficient evaluations
+template <typename T, typename = void>
+struct eval_t {
```
joshessman-llnl: `eval_result`?

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,170 @@
...
- * @return The list of true dofs that should be part of the essential boundary conditions
+ * @return The list of vector dofs that should be
+ * part of the essential boundary conditions
  */
-mfem::Array<int> MakeTrueEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoefficient& c);
+  template <typename T, typename SFINAE = std::enable_if_t<std::is_base_of_v<mfem::Coefficient, T> ||
+							   std::is_base_of_v<mfem::VectorCoefficient, T>>>
+mfem::Array<int> MakeEssList(mfem::ParFiniteElementSpace& pfes, T& c)
+{
+  mfem::Array<int> ess_vdof_list(0);
```
joshessman-llnl: Minor: Would this be more readable if it were just default-constructed?

Also, unlike most `std::vector` implementations, MFEM resizing here is naive (resizes to size + 1 instead of e.g. size * 2) and will reallocate + copy after every `Append`. Would it be worth considering making a single conservative allocation of `pfes.GetVSize()` here and then shrinking at the end?

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -65,8 +197,49 @@ mfem::Array<int> MakeEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoef
...
  * @return An array holding the attributes that correspond to each element
  */
-mfem::Array<int> MakeAttributeList(
-    mfem::Mesh& m, mfem::Coefficient& c, std::function<int(double)> digitize = [](double v) { return v > 0. ? 2 : 1; });
+
+template <typename T>
+T MakeAttributeList(mfem::Mesh& m, mfem::Coefficient& c, std::function<int(double)> digitize = digitize::threshold)
+{
+
+  static_assert( std::is_constructible<T, typename detail::index_t<T>::type>::value, "T does not have a constructor of type detail::index_t<T>");
```
joshessman-llnl: Suggestion: "T must be constructible with an instance of detail::index_t<T>"

Also, the `@pre` and `@tparam` Doxygen commands might be useful here for documenting template preconditions.

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -14,41 +14,170 @@
...
+ * @brief Thresholds a real function so that it is 2 if greater than 0.
+ * @param[in] v floating point value
+ */
+[[maybe_unused]] static int threshold(double v) { return v > 0. ? 2 : 1; }
+
+/**
+ * @brief Checks if floating point value is equal to 1, if return 2 otherwise return 1.
+ */
+[[maybe_unused]] static int equals1(double v) { return v == 1. ? 2 : 1; }
+}  // namespace digitize
```
samuelpmishLLNL: It feels like these functions should return `bool` rather than `int`. The convention of "2" for true and "1" for false seems dangerous, as all other C/C++ code assumes "1" for true and "0" for false.

The function name "threshold" is ambiguous too, maybe we could change it to be self-documenting, e.g. `is_positive`, or `greater_than_zero`

src/serac/coefficients/coefficient_extensions.hpp
samuelpmishLLNL: it seems like it seems like `mfem::Array` resizing is not as naive as that (thankfully):

```cpp
template <class T>
inline void Array<T>::GrowSize(int minsize) {
  const int nsize = std::max(minsize, 2 * data.Capacity());
  Memory<T> p(nsize, data.GetMemoryType());
  p.CopyFrom(data, size);
  p.UseDevice(data.UseDevice());
  data.Delete();
  data = p;
}
```

so it shouldn't be too bad to keep as is. Alternatively, maybe just reserve something like 1/8 of the `pfes.GetVSize()` before the loop.

joshessman-llnl: Thanks for the correction - I think maybe I was thinking of the `mfem::Vector` resize logic?

jonwong12: What do you think about using `std::vector` and then just converting to `mfem::Array` at the end?

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: It just so happens that `threshold` and `equals1` return 1 or 2, but `floor` indicates that you can return a discrete set of integer attributes. It just so happens in these functions the assumption is to have two attributes. Mfem attributes must be `0` because they do something like `array[T.attribute-1]`.

I can add some more comments. These are mostly convenience functions for common "digitization" functions.

I can rename to `greater_than_zero`.

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -25,30 +26,172 @@
...
+// Methods for determining the size of a container
+template <typename T>
+auto size(T&& container)
+{
+  return container.size();
+}
+
+/// mfem:Array size call
+template <typename T>
+std::size_t size(mfem::Array<T> container)
```
joshessman-llnl:
```suggestion
std::size_t size(const mfem::Array<T>& container)
```

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -25,30 +26,172 @@
...
- * @brief MakeTrueEssList takes in a FESpace, a vector coefficient, and produces a list
- *  of essential boundary conditions
+ * @brief This namespace provides a set of commonly used "digitizing" functions for convenience.
  *
- * @param[in] pfes A finite element space for the constrained grid function
- * @param[in] c A VectorCoefficient that is projected on to the mesh. All
- * d.o.f's are examined and those that are the condition (> 0.) are appended to
- * the vdof list.
- * @return The list of true dofs that should be part of the essential boundary conditions
+ * The goal of a "digizitizing" function is to convert a floating-point number to an integer to be used as an mfem
```
joshessman-llnl:
```suggestion
 * The goal of a "digitizing" function is to convert a floating-point number to an integer to be used as an mfem
```

Though "digizitizing" sounds much more fun...

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -25,30 +26,172 @@
...
+ * attribute. Mfem attributes must be integers > 0 (1,2,... etc).
+ */
+namespace digitize {
+/**
+ * @brief takes floating point value and rounds down to the nearest integer
+ * @param[in] v floating point value
  */
-mfem::Array<int> MakeTrueEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoefficient& c);
+
+[[maybe_unused]] static int floor(double v) { return static_cast<int>(v); }
```
joshessman-llnl: Since this is in a header I think the `static` keyword on this function (and on the others) doesn't do anything.

Also, is this function any different from `std::floor`?

src/serac/coefficients/coefficient_extensions.hpp
joshessman-llnl: I think that's reasonable, and definitely in the spirit of using MFEM containers only when necessary

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -65,8 +214,51 @@ mfem::Array<int> MakeEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoef
...
  * @return An array holding the attributes that correspond to each element
  */
-mfem::Array<int> MakeAttributeList(
-    mfem::Mesh& m, mfem::Coefficient& c, std::function<int(double)> digitize = [](double v) { return v > 0. ? 2 : 1; });
+
+template <typename T>
+T MakeAttributeList(mfem::Mesh& m, mfem::Coefficient& c,
+                    std::function<int(double)> digitize = digitize::greater_than_zero)
+{
+  static_assert(std::is_constructible<T, typename detail::index_t<T>::type>::value,
```
joshessman-llnl, line 221: 👍

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -57,6 +200,12 @@ mfem::Array<int> MakeEssList(mfem::ParFiniteElementSpace& pfes, mfem::VectorCoef
...
  *
+ * @pre The template type, T, should have a constructor that takes in the size, and
+ * should have begin() and end() methods.
+ *
+ * @pre T must be constructible with an instance of detail::index_t
+ *
+ * @tparam T Return type is either a suitable std collection or mfem::Array
  * @param[in] m The mesh
  * @param[in] c The coefficient provided that will be evaluated on the mesh
  * @param[in] digitize An optional function that can be
```
joshessman-llnl: Maybe "A function mapping coefficient values onto integer attribute values" or something like that?

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: Not really, except that it returns an int which is needed for mfem element attributes.

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: Should I just move to "default constructed" then? It seems like if I use `std::vector` we'll have an extra copy. At the end of the day, this information probably needs to be `mfem::Array<int>`.

jamiebramwell:
@joshessman-llnl @samuelpmishLLNL , can you take a look to see if all of your comments were addressed? Thanks!

src/serac/coefficients/coefficient_extensions.hpp
```diff
@@ -133,88 +391,141 @@ class AttributeModifierCoefficient : public mfem::Coefficient {
...
+   * @param[in] dim d.o.f of this mfem::Vectorcoefficient
+   * @param[in] func A function to apply to the evaluations of all args and return mfem::Vector
+   * @param[in[ args A list of mfem::Coefficients, mfem::VectorCoefficients, or numbers
    */
-  std::function<void(const mfem::Vector&, mfem::Vector&)> mono_function_;
+  TransformedVectorCoefficient(int                                                                          dim,
+                               std::function<mfem::Vector(typename detail::eval_result_t<Types>::type&...)> func,
+                               Types&... args)
+      : mfem::VectorCoefficient(dim), references_(std::make_tuple(std::ref(args)...)), function_(func)
+  {
```
joshessman-llnl, line 476: Would it be possible to use a `static_assert(std::is_invocable<...` here to intercept the nasty compiler errors a user might see if they attempt to instantiate this incorrectly?

joshessman-llnl (approved):
Looks good, thanks for putting this together!

src/serac/coefficients/coefficient_extensions.hpp
jonwong12: Adjusted and now uses `std::floor` and static_cast to return int.

samuelpmishLLNL (approved):

jamiebramwell (approved):
This is really awesome! Thanks for putting it together!

PR #441, opened 2021-04-15
fix: FindHDF5 try_compile requires enabling C lang

goxberry:
This commit adds `C` to the Serac project languages in the project's
root `CMakeLists.txt` file to resolve an unknown file extension error
emitted by CMake when invoked by `spack install serac` within
`uberenv`. During this invocation, if `C` is not a member of the
`LANGUAGES` argument in the `project(serac ...)` macro invocation, the
built-in `FindHDF5` CMake module will throw an error. Adding `C to the
`LANGUAGES` argument of this macro invocation resolves this error.

PR #439, opened 2021-04-15
fix: serac spack pkg: ' install' -> 'install'

goxberry:
This commit fixes a bug in the `phases` list in the Serac Spack
package. In this list, the last element was ' install' and has been
changed in this commit to 'install'. The leading space causes an error
when Serac clients attempt to install Serac using this Spack package,
and deleting this leading space resolves this error.

PR #444, opened 2021-04-15
merged PR from forks from Geoffrey

white238:
#439 
#440 
#441

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@
 
 cmake_minimum_required(VERSION 3.8)
 
-project(serac LANGUAGES CXX)
+project(serac LANGUAGES CXX C)
 
 # MPI is required in Serac.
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -10,7 +10,7 @@
 
 cmake_minimum_required(VERSION 3.8)
 
-project(serac LANGUAGES CXX)
```
joshessman-llnl, line 13: Do we know where the `find_package(HDF5)` is?  It might be preferable to add an `enable_language(C)` near where that is.

joshessman-llnl (approved):

CMakeLists.txt
white238: Let's merge this and investigate afterwards so @goxberry is not held up.

white238: #445

jonwong12 (approved):

PR #448, opened 2021-04-20
WIP: Doxygen documentation/misc cleanup of variational form

joshessman-llnl:
https://serac.readthedocs.io/en/docs-essman-weak_form/doxygen/html/classserac_1_1WeakForm_3_01test_07trial_08_4.html


src/serac/physics/utilities/variational_form/array.hpp
```diff
@@ -80,10 +135,16 @@ constexpr auto rotate(const array<T, n>& original, int rotation)
...
 constexpr int orientation(const array<int, 2> a, const array<int, 2> b) { return (a[0] == b[0]) ? 1 : -1; }
+/// @endcond
 
+// NOTE: Currently not used anywhere
+/**
+ * @brief
+ */
 template <int n>
-constexpr auto orientation(const array<int, n> a, const array<int, n> b)
+constexpr auto orientation(const array<int, n>& a, const array<int, n>& b)
```
samuelpmishLLNL: I believe this function was supposed to help "undo" a rotation (defined a few lines above). So, if it was the case that `auto b = rotate(a, 3)`, then `orientation(a, b)` would return either 3 or -3 (or something like that). We can also just delete this here, as it's unused.

Originally, this was used for detecting orientations of shared faces between Nedelec elements, to figure out the appropriate local rotation to apply to them.

src/serac/physics/utilities/variational_form/array.hpp
```diff
@@ -93,8 +154,12 @@ constexpr auto orientation(const array<int, n> a, const array<int, n> b)
   return std::tuple{0, 0};
 }
 
+/**
+ * @brief Computes the successive sum of a one-dimensional array
```
samuelpmishLLNL:
```suggestion
 * @brief Computes the sum of all entries in the array
```

src/serac/physics/utilities/variational_form/array.hpp
```diff
@@ -103,8 +168,12 @@ constexpr auto total(const array<T, n> a)
   return p;
 }
 
+/**
+ * @brief Computes the successive product of a one-dimensional array
```
samuelpmishLLNL:
```suggestion
 * @brief Computes the product of all entries in the array
```

src/serac/physics/utilities/variational_form/array.hpp
```diff
@@ -240,21 +342,45 @@ constexpr auto join(array<T, n1> v1, array<T, n2> v2)
...
 template <int... n>
 struct IndexSpace {
   static constexpr int size = sizeof...(n);
-  array<char, size>    i;
-  constexpr auto       begin() { return IndexSpace<n...>{}; }
-  constexpr auto       end()
+  /**
+   * @brief The current indices within the array/tensor
+   */
+  array<char, size> i;  // FIXME: Is char too small here?
```
samuelpmishLLNL, line 369: IndexSpace is supposed to be for iterating over a small multidimensional array, like for indices in a tensor/array. If those dimensions are getting bigger than 128/256, then tensor and array are likely not the right containers anyway.

Also, `for_constexpr` fills a similar role to IndexSpace, and tends to incur less runtime overhead than this IndexSpace class (but increases compile times, as it forces inlining)

src/serac/physics/utilities/variational_form/detail/hexahedron_hcurl.inl
```diff
@@ -1,7 +1,7 @@
 template <int p>
-struct finite_element<::Geometry::Hexahedron, Hcurl<p>> {
-  static constexpr auto geometry   = ::Geometry::Hexahedron;
-  static constexpr auto family     = ::Family::HCURL;
+struct finite_element<Geometry::Hexahedron, Hcurl<p>> {
```
samuelpmishLLNL, line 2: I think I ran into issues using `Geometry` rather than `::Geometry` when a sourcefile had `using namespace mfem`, since they also define `Geometry`. I'm not sure if that issue will still exist after putting these definitions in the serac namespace. e.g. if I have a .cpp file with

```cpp
using namespace mfem
using namespace serac

auto g = Geometry::Hexahedron;
```
will it still work?

src/serac/physics/utilities/variational_form/integral.hpp
```diff
@@ -8,44 +19,64 @@
...
+ * @tparam space A test or trial space
+ * @param[in] u The raw data to reshape into an mfem::DeviceTensor
+ * @param[in] n1 The first dimension to reshape into
+ * @param[in] n2 The second dimension to reshape into
+ * @see mfem::Reshape
+ */
 template <typename space>
 auto Reshape(double* u, int n1, int n2)
 {
+  // TODO: Helpful static_assert
```
samuelpmishLLNL, line 36: what should test here with a `static_assert`?

src/serac/physics/utilities/variational_form/integral.hpp
```diff
@@ -74,19 +114,26 @@ void Add(const mfem::DeviceTensor<3, double>& r_global, tensor<double, ndof, com
...
+ *
+ * @param[in] u The DOF values for the element
+ * @param[in] xi The position of the quadrature point in reference space
+ * @param[in] J The Jacobian of the element transformation at the quadrature point
+ * @tparam element_type The type of the element (used to determine the family)
+ */
 template <typename element_type, typename T, int dim>
 auto Preprocess(T u, const tensor<double, dim> xi, const tensor<double, dim, dim> J)
 {
+  // TODO: Static assert that element_type is an element??
```
samuelpmishLLNL, line 135: maybe something like this
```suggestion
static_assert(is_finite_element<element_type>::value, "impl::Preprocess error: template argument must be of type `finite_element<>`");
```

src/serac/physics/utilities/variational_form/detail/hexahedron_hcurl.inl
joshessman-llnl: So we still wouldn't be able to have that snippet work unless we disambiguated the two somehow (renamed the enum in serac, or specified the namespace explicitly, etc)

src/serac/physics/utilities/variational_form/integral.hpp
```diff
@@ -121,30 +170,33 @@ auto Preprocess(T u, const tensor<double, geometry_dim> xi,
...
+ * @tparam T The type of the output from the user-provided q-function
+ * @pre T must be a pair type for H1, H(curl) and H(div) family elements
+ * @param[in] f The value component output of the user's quadrature function (as opposed to the value/derivative pair)
+ * @param[in] xi The position of the quadrature point in reference space
+ * @param[in] J The Jacobian of the element transformation at the quadrature point
+ */
 template <typename element_type, typename T, int dim>
 auto Postprocess(T f, const tensor<double, dim> xi, const tensor<double, dim, dim> J)
 {
+  // TODO: Helpful static_assert about f being tuple or tuple-like for H1, hcurl, hdiv
```
samuelpmishLLNL, line 198: Later on, the type `T` will not always be a tuple here (e.g. once we support gradient-only annotations). But in the near future, we can have
```suggestion
static_assert(is_tuple<T>::value, "impl::Postprocess error: first argument must be a std::tuple of two entries (see equation (1) here https://libceed.readthedocs.io/en/latest/libCEEDapi/)");
```

src/serac/physics/utilities/variational_form/integral.hpp
joshessman-llnl: Was thinking that we'd implement a `is_finite_element_space` trait and then `static_assert(detail::is_finite_element_space<space>::value);` or something - there's a fair number of places where I thought we could go back later and add these little helpers that get us slightly less incomprehensible template error messages from the compiler, but I don't think they're critical for this big initial merge

src/serac/physics/utilities/variational_form/integral.hpp
```diff
@@ -411,9 +521,10 @@ struct Integral {
     using derivative_type = decltype(get_gradient(qf(x_t{}, make_dual(u_du_t{}))));
 
     auto num_quadrature_points = static_cast<uint32_t>(X.Size() / spatial_dim);
-    qf_derivatives.resize(sizeof(derivative_type) * num_quadrature_points);
+    qf_derivatives_.resize(sizeof(derivative_type) * num_quadrature_points);
 
-    auto qf_derivatives_ptr = reinterpret_cast<derivative_type*>(qf_derivatives.data());
+    // FIXME: Strict aliasing rule?
```
samuelpmishLLNL, line 525: `qf_derivatives` may not need to be owned by the `Integral` object, we might be able to just make it a `std::shared_ptr` in the scope of constructor (and change the captures on the lambdas that need it). That way, it could just store the actual derivative type instead of going aliasing a `char *`

src/serac/physics/utilities/variational_form/array.hpp
joshessman-llnl: I think we should be able to verify this statically (that all the sizes are less than char_max)

src/serac/physics/utilities/variational_form/integral.hpp
samuelpmishLLNL: I 100% agree with the spirit of using `static_assert` to spare the user nasty template compiler errors, but I think `impl::Reshape` might be the wrong place to do it, since it's so deep into the implementation. Instead, we could catch the same issue higher up by `static_assert`ing on `is_finite_element_space<test_space>::value` at the WeakForm or Integral level, and it would output a smaller template error "backtrace".

samuelpmishLLNL:
Thank you again for making all of these much-needed changes!

src/serac/physics/utilities/variational_form/integral.hpp
joshessman-llnl: That definitely makes more sense, it's also possible that I've already left a similar FIXME/TODO comment at a higher level that would obviate the need for this one

src/serac/physics/utilities/variational_form/array.hpp
```diff
@@ -1,35 +1,87 @@
...
+namespace serac {
+
+/// @cond
 template <typename T, int... n>
 struct array;
+/// @endcond
 
+/**
+ * @brief Multidimensional array class
+ * @tparam T The scalar type of the array element
```
jamiebramwell, line 27: `first` and `rest` also need docs.

src/serac/physics/utilities/variational_form/finite_element.hpp
```diff
@@ -69,38 +118,47 @@ enum class Evaluation
...
+template <Geometry g, int p>
 struct is_finite_element<finite_element<g, Hcurl<p> > > {
   static constexpr bool value = true;
 };
 
-template <typename T>
-struct quadrature_data {
-  using type = T;
-};
+// TODO: Implemented in separate branch
```
jamiebramwell: I think this should just be deleted.

jamiebramwell (approved):
This is really great! Thanks for putting it together. I'll go through this a bit more carefully when it's merged into the weak form branch and we do a full review of that.

samuelpmishLLNL (approved):

PR #457, opened 2021-05-06
Discuss: Caliper region/expr implementation

jonwong12:
This PR seeks to make it easier to profile both single line expressions and regions using caliper. There's one implementation for profiling a single line expression.

Discussion Topic:
There are currently two different region implementations (one that uses a `Region` class to handle scoping) and the other uses a C macro which is syntactically nice. 

TODO: 

- [x] Extended serac_profile.cpp test to use new region and expr markers
- [x] Use new cali_*scope* marker
- [x] Get rid of possibly harder to debug region2 macro
- [x] Write test to verify that caliper closes end tags when exceptions occur

For single line expression profiling:
```
mfem::Vector r1 =
    SERAC_PROFILE_EXPR(concat("mfem_Apply", postfix),
		       A * x - f);
```
However, for single line expressions that require the "scope" to extend to the rest of the code, one should use `SERAC_MARK_START/END` region markers which are a little more cumbersome.

```
  SERAC_MARK_START(concat("mfem_parallelAssemble", postfix));
  std::unique_ptr<mfem::HypreParMatrix> J(A.ParallelAssemble());
  SERAC_MARK_END(concat("mfem_parallelAssemble", postfix));
```


jamiebramwell:
I like the C macro since it scope is obvious from the brackets, but I defer to the CS experts for their opinion.

white238:
Can you add some documentation and examples here: src/docs/sphinx/dev_guide/profiling.rst

Also while you are in that file. Please fix the weird formatting that is going on.

thanks @jonwong12 !

white238 (approved):

white238:
https://serac.readthedocs.io/en/feature-jonwong12-caliper_regions/sphinx/dev_guide/profiling.html

jamiebramwell (approved):
Looks good!

src/serac/infrastructure/profiling.cpp
```diff
@@ -78,5 +78,34 @@ void setCaliperMetadata([[maybe_unused]] const std::string& name, [[maybe_unused
...
+  CALI_MARK_BEGIN(name);
+#endif
+}
+
+void startCaliperRegion([[maybe_unused]] const std::string& name)
+{
+#ifdef SERAC_USE_CALIPER
+  CALI_MARK_BEGIN(name.c_str());
+#endif
+}
```
joshessman-llnl: I think that the `const char* -> std::string` conversion would make the overloading here unnecessary, is the intent to just avoid the extra work where possible?

src/serac/infrastructure/profiling.hpp
```diff
@@ -56,16 +56,37 @@
...
+#define SERAC_CONCAT_(a, b) a##b
+#define SERAC_CONCAT(a, b) SERAC_CONCAT_(a, b)
+
+#define SERAC_PROFILE_SCOPE(name) cali::ScopeAnnotation SERAC_CONCAT(region, __LINE__)(name)
+
+#define SERAC_PROFILE_EXPR(name, expr)                                            \
+  [&]() -> typename serac::profiling::detail::expr_t<decltype(expr)>::expr_type { \
+    const cali::ScopeAnnotation SERAC_CONCAT(region, __LINE__)(name);             \
+    return expr;                                                                  \
+  }()
```
joshessman-llnl: Does this work with noncopyable types/references?

src/serac/infrastructure/profiling.hpp
```diff
@@ -118,6 +141,36 @@ void setCaliperMetadata(const std::string& name, double data);
...
+/*!
+  @overload
+*/
+void endCaliperRegion(const std::string& name);
+
+template <typename T>
+struct expr_t {
+  using expr_type = T;
+};
+
```
joshessman-llnl: Minor: Could we just use the implementation of [std::type_identity](https://en.cppreference.com/w/cpp/types/type_identity) here so it's a drop-in replacement when C++20 is available?

joshessman-llnl (approved):

src/serac/infrastructure/profiling.cpp
jonwong12: When I had tried passing in `std::string` earlier I had an issue, but I can test it again.

jonwong12:
Thanks everyone! Special thanks to @joshessman-llnl  for making SERAC_PROFILE_EXPR() much better!

PR #459, opened 2021-05-07
Clean up the weak form tests

jamiebramwell:
This adds some comments to the weak form tests and moves the unused performance test to a separate directory for now.


src/serac/physics/utilities/variational_form/tests/l2_comparison.cpp
```diff
@@ -1,149 +0,0 @@
-#include "mfem.hpp"
```
samuelpmish: Jonathan and I were recently talking about trying to add / debug L2 support, so deleting this might be a little premature. I suppose we can just grab it from an older commit though.

src/serac/physics/utilities/variational_form/tests/weak_form_comparisons.cpp
```diff
@@ -237,9 +260,9 @@ void weak_form_test(mfem::ParMesh& mesh, Hcurl<p> test, Hcurl<p> trial, Dimensio
       Dimension<dim>{},
       [&](auto x, auto vector_potential) {
         auto [A, curl_A] = vector_potential;
-        auto J = a * A - tensor<double, dim>{10 * x[0] * x[1], -5 * (x[0] - x[1]) * x[1]};
-        auto H = b * curl_A;
-        return std::tuple{J, H};
+        auto J_term      = a * A - tensor<double, dim>{10 * x[0] * x[1], -5 * (x[0] - x[1]) * x[1]};
+        auto H_term      = b * curl_A;
+        return std::tuple{J_term, H_term};
```
samuelpmish, line 265: what is motivating these name changes?

samuelpmishLLNL (approved):

src/serac/physics/utilities/variational_form/tests/l2_comparison.cpp
jamiebramwell: That's fair! I think I was being overly ambitious here.

src/serac/physics/utilities/variational_form/tests/weak_form_comparisons.cpp
jamiebramwell: Since `J` was declared in the containing scope as well, this was issuing a shadowing warning.

PR #471, opened 2021-05-25
Harden BLT tools options

### LINKED ISSUE #442: Guard Code Checks
This should not be run by our customers so we should be guarding this against whether the individual executables are found:

https://github.com/LLNL/serac/blob/b5191c08fbf2cb106491489bce40ecfde0d41b6f/CMakeLists.txt#L87-L89

@goxberry ran into an issue when Serac is installed via Spack.

white238:
- Guards against auto-detecting BLT developer tools when users are running the code
- Lock down `clangformat` to version 10

Closes #442

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,9 +48,35 @@ endif()
 if (NOT BLT_CXX_STD)
     set(BLT_CXX_STD "c++17" CACHE STRING "")
 endif()
-set(ENABLE_ASTYLE      OFF CACHE BOOL "")
-set(ENABLE_CLANGQUERY  OFF CACHE BOOL "")
-set(ENABLE_UNCRUSTIFY  OFF CACHE BOOL "")
+
+# These BLT tools are not used in Serac, turn them off
+set(_unused_blt_tools
+    CLANGQUERY
+    VALGRIND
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -48,9 +48,37 @@ endif()
...
+    CLANGFORMAT
+    CLANGTIDY
+    CPPCHECK
+    DOXYGEN
+    SPHINX)
+foreach(_tool ${_used_blt_tools})
+    if(NOT ${_tool}_EXECUTABLE)
+        set(ENABLE_${_tool} OFF CACHE BOOL "")
+    else()
+        set(ENABLE_${_tool} ON CACHE BOOL "")
```
joshessman-llnl, line 76: My understanding is that these are ON by default, so is this just for clarity/explicitness?

joshessman-llnl (approved):

jamiebramwell (approved):
Thanks!

CMakeLists.txt
white238: Explicitness and I'm thinking of turning them off by default due to the large set we support in BLT at this point.  Easier to turn on the ones you want then turn off all the ones you don't.

PR #472, opened 2021-05-25
Remove BLT_EXPORT_THIRDPARTY workaround

### LINKED ISSUE #470: Investigate why BLT exported targets are working
We are not exporting our BLT targets we depend on (e.g. `mpi`).  I'm not sure why our `using-cmake` examples are working at this point.  This should be investigated.

joshessman-llnl:
This resolves the following FIXME by incorporating the functionality added to BLT in https://github.com/LLNL/blt/pull/436:

https://github.com/LLNL/serac/blob/4dbb20800e4914ccfb5b42903f65f589cb03b610/examples/using-with-cmake/CMakeLists.txt#L7-L10

Fixes #470 

Possibly this would have been noticed earlier if we had opened a GitHub issue instead of just leaving it as a TODO?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,9 @@ set(ENABLE_BENCHMARKS  OFF CACHE BOOL "")
 set(ENABLE_ALL_WARNINGS ON CACHE BOOL "")
 set(ENABLE_WARNINGS_AS_ERRORS ON CACHE BOOL "")
 
+# Allows MPI/CUDA/etc targets to be exported in the serac namespace
+set(BLT_EXPORT_THIRDPARTY ON CACHE BOOL "")
+
 include(${BLT_SOURCE_DIR}/SetupBLT.cmake)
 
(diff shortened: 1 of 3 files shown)


white238 (approved):

jamiebramwell (approved):

jamiebramwell:
> 
> 
> This resolves the following FIXME by incorporating the functionality added to BLT in [LLNL/blt#436](https://github.com/LLNL/blt/pull/436):
> 
> https://github.com/LLNL/serac/blob/4dbb20800e4914ccfb5b42903f65f589cb03b610/examples/using-with-cmake/CMakeLists.txt#L7-L10
> 
> Fixes #470
> 
> Possibly this would have been noticed earlier if we had opened a GitHub issue instead of just leaving it as a TODO?

Yeah, I try to catch this stuff and add issues as I see them, but I know we have some hanging out in the code base. We should probably do a search to grab all the TODO items.

PR #474, opened 2021-05-25
Fix WSL/Ubuntu 20 TPL build

jamiebramwell:
I needed to do this to resolve missing pthread libraries at link time in the HDF5 build.


joshessman-llnl (approved):
Does this mean we no longer need the `lpthread` linker flag? I think in general `pthread` should be preferred to `lpthread` where possible.

white238 (approved):

jamiebramwell:
> 
> 
> Does this mean we no longer need the `lpthread` linker flag? I think in general `pthread` should be preferred to `lpthread` where possible.

I'll try that.

PR #418, opened 2021-03-11
Feature/variational form

samuelpmishLLNL:
This PR aims to add additional tools for writing differentiable nonlinear material models and objective functions that run performantly on CPUs and GPUs.

Built docs [here](https://serac.readthedocs.io/en/feature-variational_form/sphinx/dev_guide/weak_form.html).

The main file of interest is `serac/physics/utilities/variational_form/weak_form.hpp`. Its design is based on the idea of 
writing a finite element calculation kernel that takes a handful of template parameters, including:
- test space
- trial space
- element geometry (quadrilateral, triangle, hexahedron, ... )
- quadrature rule
- "q-function" (see [here](https://libceed.readthedocs.io/en/latest/libCEEDapi/#theoretical-framework)) to characterize material response

This approach allows the user to define custom kernels with interesting material models succinctly, addressing issues like
https://github.com/mfem/mfem/issues/1870
https://github.com/mfem/mfem/issues/1913
https://github.com/mfem/mfem/issues/1982

In addition, it is an opportunity to potentially deduplicate many copy-and-pasted kernels in mfem:
[bilininteg_gradient: 823 lines](https://github.com/mfem/mfem/blob/master/fem/bilininteg_gradient.cpp)
[bilininteg_divergence: 1069 lines](https://github.com/mfem/mfem/blob/master/fem/bilininteg_divergence.cpp)
[bilininteg_hcurl: 6697 lines](https://github.com/mfem/mfem/blob/master/fem/bilininteg_hcurl.cpp)
[bilininteg_hdiv: 2026 lines](https://github.com/mfem/mfem/blob/master/fem/bilininteg_hdiv.cpp)

Some other important aspects of this PR include:
- a statically allocated `tensor` class for arithmetic on small, dense tensors. Template arguments for the tensor extents allows for better compiler optimization, enforcing compatible sizes (e.g. addition of tensors of different shapes or dot products between incompatible tensors will not compile) and GPU compatibility.
```cpp
  tensor < double, 3 > u = {1, 2, 3};
  tensor < double, 3 > v = {4, 5, 6};
  tensor < double, 3, 3 > A = make_tensor<3,3>([](int i, int j){ return i + 2.0 * j; })
  tensor < double, 3, 3 > I = Identity<3>();

  tensor result = 3.0 * tr(A) * I + outer(u, v);
```
- forward-mode automatic differentiation via dual numbers
```cpp
 // AD works with tensors of all shapes
  static constexpr auto I = Identity<3>();
[...]

PR #477, opened 2021-05-27
add support for q-function interface with discontinuous Galerkin fields

samuelpmishLLNL:
- adds a new function space `L2` (the name mfem uses, should we keep this, or call it `DG` or something else)?
- implement `finite_element` specializations for `L2` with `segment`, `quadrilateral` and `hexahedron` geometries
  - note: the shape functions for these elements interpolate at the Gauss-Legendre nodes
- adds a new test `functional_comparison_L2` that compares against mfem's implementation in 2D/3D (linear, quadratic and cubic)
- removed / renamed some functions in `polynomials.hpp`


src/serac/physics/utilities/functional/detail/hexahedron_L2.inl
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
```
joshessman-llnl, line 1: Extremely minor: the H1 file is called `hexahedron_h1.inl`, so should the L2 file be called `hexahedron_l2.inl` for consistency in terms of capitalization? I realize the lowercase L is sort of ambiguous so maybe h1 -> H1 instead?

joshessman-llnl (approved):

jandrej:
The changes look good so far but I have a few questions. There is no interface addition to support jumps or averages of face contributions between two elements right? This is important for any DG discretization.

samuelpmishLLNL:
> The changes look good so far but I have a few questions. There is no interface addition to support jumps or averages of face contributions between two elements right? This is important for any DG discretization.

Correct, for now. My understanding was that LIDO has been using L2 fields to express material parameters that vary discontinuously across element boundaries, but they didn't (immediately) need to perform any calculations of the type you're describing. I do believe that a similar q-function interface could be extended for that case as well, maybe something like the second part of the code below:

```cpp
Functional < test_space(trial_space) > residual(...);

// this capability already exists
// evaluate this q-function at each quadrature point of each volume element
residual.AddVolumeIntegral([](auto displacement) {
  auto [u, du_dx] = displacement;
  auto source = ...;
  auto flux = ...;
  return std::tuple{source, flux};
}, mesh);

// this capability does not currently exist
// evaluate this q-function at each quadrature point on the boundaries between volume elements
residual.Add????Integral([](auto displacement) {
  auto [u_avg, u_jump /* ??? */] = displacement;
  auto source = ...;
  auto flux = ...;
  return std::tuple{source, flux};
}, mesh);

```

i.e. provide the q-function the average and jump values for the trial space, and return the same {source, flux} tuple (do we need to return a source term?)

src/serac/physics/utilities/functional/integral.hpp
```diff
@@ -557,8 +562,11 @@ class Integral {
            Dimension<spatial_dim>, lambda_type&& qf)
       : J_(J), X_(X)
   {
-    constexpr auto geometry = supported_geometries[geometry_dim];
-    constexpr auto Q        = std::max(test_space::order, trial_space::order) + 1;
+    constexpr auto geometry                      = supported_geometries[geometry_dim];
+    constexpr auto Q                             = std::max(test_space::order, trial_space::order) + 1;
+    constexpr auto quadrature_points_per_element = (spatial_dim == 2) ? Q * Q : Q * Q * Q;
+
+    uint32_t num_quadrature_points = quadrature_points_per_element * uint32_t(num_elements);
```
jamiebramwell, line 569: Why are you using a fixed width type?

src/serac/physics/utilities/functional/integral.hpp
```diff
@@ -569,11 +577,7 @@ class Integral {
     using u_du_t          = typename detail::lambda_argument<trial_space, geometry_dim, spatial_dim>::type;
     using derivative_type = decltype(get_gradient(qf(x_t{}, make_dual(u_du_t{}))));
 
-    auto num_quadrature_points = static_cast<uint32_t>(X.Size() / spatial_dim);
-    qf_derivatives_.resize(sizeof(derivative_type) * num_quadrature_points);
-
-    // FIXME: Strict aliasing rule?
-    auto qf_derivatives_ptr = reinterpret_cast<derivative_type*>(qf_derivatives_.data());
+    std::shared_ptr<derivative_type[]> qf_derivatives(new derivative_type[num_quadrature_points]);
```
jamiebramwell, line 591: Why does this need to be a `shared_ptr`?

joshessman-llnl: I think that's what this comment refers to:

https://github.com/LLNL/serac/pull/418#discussion_r628635383

Instead of performing a (dubious) reinterpret_cast, this approach avoids type punning entirely.

joshessman-llnl: Though I'm concerned the lifetime of `qf_derivatives` ends at the end of the constructor - perhaps a `shared_ptr` should be the type of the argument to the kernels so the reference count is incremented appropriately?

samuelpmishLLNL: `qf_derivatives` can't be a member variable (`derivative_type` is specific to the lambda function, so it can't be known a priori), so there's no way to `delete` it in the destructor.

By making it a shared pointer, its lifetime should roughly match the lifetime of the `Integral` that created it.
e.g. 
1. Integral constructor: creates `qf_derivatives` and the two lambda functions capture it, incrementing the reference count.
2. Integral destructor: `std::function`s that capture `qf_derivatives` get deallocated, and the reference count of `qf_derivatives` goes to zero again, triggering its cleanup.

The previous implementation involved essentially making a member variable `void * qf_derivatives` and aliasing that pointer in a way that isn't really type safe. 

If there's a better way to do it, I'm interested! This was just one implementation that came to mind.

src/serac/physics/utilities/functional/integral.hpp
samuelpmishLLNL: I'm not sure what the appropriate type should be. Part of the weirdness is `mfem::Mesh::GetNE()` returning a 32-bit signed integer. (?!)

Would you prefer: `size_t`, `uint64_t`, or something else?

jamiebramwell: I typically like to use `size_t` to avoid portability issues with fixed with stuff, however I don't have a super strong opinion. Any thoughts @white238 @joshessman-llnl ?

src/serac/physics/utilities/functional/integral.hpp
jamiebramwell: I'd add a comment reflecting this. I can't think of a better way to do this at the moment. Thank you for explaining!

jamiebramwell (approved):

src/serac/physics/utilities/functional/detail/hexahedron_L2.inl
```diff
@@ -0,0 +1,76 @@
...
+
+// interpolate at Gauss-Lobatto nodes for the appropriate polynomial order
+//
+// note: mfem assumes the parent element domain is [0,1]x[0,1]x[0,1]
+// for additional information on the finite_element concept requirements, see finite_element.hpp
+/// @cond
+template <int p, int c>
+struct finite_element<Geometry::Hexahedron, L2<p, c> > {
+  static constexpr auto geometry   = Geometry::Hexahedron;
+  static constexpr auto family     = Family::H1;
```
white238: Should this be L2?

src/serac/physics/utilities/functional/detail/hexahedron_L2.inl
```diff
@@ -0,0 +1,76 @@
...
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file hexahedron_L2.inl
+ *
+ * @brief Specialization of finite_element for L2 on hexahedron geometry
+ */
+
+// interpolate at Gauss-Lobatto nodes for the appropriate polynomial order
```
white238: Also i think this should be Gauss-Legendre...

src/serac/physics/utilities/functional/detail/quadrilateral_L2.inl
```diff
@@ -12,24 +12,27 @@
 
 // specialization of finite_element for L2 on quadrilateral geometry
 //
-// this specialization defines shape functions that interpolate
-// at Gauss-Lobatto nodes for the appropriate polynomial order
+// this specialization defines shape functions (and their gradients) that
+// interpolate at Gauss-Lobatto nodes for the appropriate polynomial order
```
white238: I think this is Gauss-Legendre as well

src/serac/physics/utilities/functional/detail/hexahedron_L2.inl
samuelpmishLLNL:
```suggestion
// interpolate at Gauss-Legendre nodes for the appropriate polynomial order
```

src/serac/physics/utilities/functional/detail/quadrilateral_L2.inl
samuelpmishLLNL:
```suggestion
// this specialization defines shape functions (and their gradients) that
// interpolate at Gauss-Legendre nodes for the appropriate polynomial order
```

src/serac/physics/utilities/functional/detail/segment_L2.inl
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @file segment_L2.inl
+ *
+ * @brief Specialization of finite_element for L2 on segment geometry
+ */
+
+// specialization of finite_element for L2 on segment geometry
+//
+// this specialization defines shape functions (and their gradients) that
+// interpolate at Gauss-Lobatto nodes for the appropriate polynomial order
```
samuelpmishLLNL:
```suggestion
// interpolate at Gauss-Legendre nodes for the appropriate polynomial order
```

src/serac/physics/utilities/functional/tests/functional_comparison_L2.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include "serac/serac_config.hpp"
+#include "serac/numerics/mesh_utils.hpp"
+#include "serac/numerics/expr_template_ops.hpp"
+#include "serac/physics/operators/stdfunction_operator.hpp"
+#include "serac/physics/utilities/functional/functional.hpp"
+#include "serac/physics/utilities/functional/tensor.hpp"
+
+#include <gtest/gtest.h>
+
+using namespace std;
```
white238: I understand this is a test.  But having this many using's makes it very hard to read and understand where everything is coming from.  Can you remove the std and mfem one at least? Especially since you have std and mfem already on most places...

src/serac/physics/utilities/functional/polynomials.hpp
```diff
@@ -247,94 +249,72 @@ constexpr tensor<T, n> GaussLobattoInterpolationDerivative01([[maybe_unused]] T
...
-  if constexpr (n == 2) {
-    return {-1.732050807568877, 1.732050807568877};
-  }
-
-  if constexpr (n == 3) {
-    return {-4.6243277820691 + 6.6666666666667 * x, 6.6666666666667 - 13.3333333333333 * x,
-            -2.04233888459753 + 6.6666666666667 * x};
-  }
-
+  if constexpr (n == 1) return {0};
```
white238, line 307: My inner CS person who hates magic numbers dies a little here... no fix is needed.

src/serac/physics/utilities/functional/integral.hpp
white238: Is there a guarantee this line is not gonna go past 32 bits after you multiply them together? `size_t` is generally 64 so it does give you quite a bit more buffer but I'm not familiar enough with how large these two things end up being

src/serac/physics/utilities/functional/detail/segment_Hcurl.inl
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file segment_hcurl.inl
```
white238:
```suggestion
 * @file segment_Hcurl.inl
```

white238 (approved):

src/serac/physics/utilities/functional/polynomials.hpp
samuelpmishLLNL: At the very least, the mathematica code that generates this C++ is available, so the source of those values is known.

It could be a cute C++ exercise to write a `constexpr` routine that finds the roots of Legendre polynomials to determine this values without code generation though!

jandrej (approved):

src/serac/physics/utilities/functional/integral.hpp
samuelpmishLLNL: There's definitely a chance of overflow on the multiplication, so it should be greater than 32 bits. It would be good to update the mfem routines to use a larger type too.

goxberry:
> Correct, for now. My understanding was that LIDO has been using L2 fields to express material parameters that vary discontinuously across element boundaries, but they didn't (immediately) need to perform any calculations of the type you're describing.

Yes, that statement is accurate. Without looking at the details of the pull request, Serac-native support for L2 fields would likely be helpful for LiDO regardless of whether jump or average operations across faces are implemented because LiDO currently doesn't use jump or average operations with L2 fields. LiDO currently only uses L2 fields as "containers with inner product information" to parameterize material properties (e.g., as a function of an L2 density field), and LiDO currently only needs to compute derivatives of quantities like material properties with respect to those parameterizations, not derivatives (or derivative-like quantities) with respect to space coordinates. If LiDO uses DG schemes in the future, it would likely do so via Serac-implemented physics.

PR #478, opened 2021-06-02
Add profiling tags to functional_comparisons test

jonwong12:
Adds caliper profiling calls to `functional_comparisons` test. 

Also has workarounds to such that CUDA related files can be successfully compiled with nvcc on lassen. (Starting point).

Noted Workaround issues:
- nvcc on LC seems to have issues with std::variant (@joshessman-llnl has reported this already). Avoid having nvcc compile files with `std::variant` seems to get around the issue, but results in some duplicated code.
- nvcc on LC seems to have some problems deducing templates (specifying the arguments seems to mitigate these issues)


src/serac/physics/utilities/functional/tensor.hpp
```diff
@@ -82,9 +82,8 @@ struct tensor;
 
 template <typename T>
 struct tensor<T> {
-  using type                    = T;
-  static constexpr int ndim     = 0;
-  static constexpr int shape[1] = {0};
+  using type                = T;
+  static constexpr int ndim = 0;
```
samuelpmishLLNL: I think this specialization should have a `first_dim` member too, to be consistent. In the same way, it should also probably have an `operator[]` as well!
```suggestion
  static constexpr int ndim = 0;
  static constexpr int first_dim = 1;
  SERAC_HOST_DEVICE constexpr auto& operator[](int) { return value; }
  SERAC_HOST_DEVICE constexpr auto operator[](int) const { return value; }
```

src/serac/physics/utilities/functional/tensor.hpp
```diff
@@ -148,7 +147,7 @@ struct tensor<T, first, rest...> {
    * @brief The array of dimensions containing the shape (not the data itself)
    * Similar to numpy.ndarray.shape
    */
-  static constexpr int shape[ndim] = {first, rest...};
+  static constexpr int first_dim = first;
```
samuelpmishLLNL, line 153: for context, this change was made because only `shape[0]` was being used, and nvcc was having trouble compiling `static constexpr` C-array members

src/serac/physics/utilities/functional/tensor.hpp
```diff
@@ -376,8 +375,9 @@ SERAC_HOST_DEVICE constexpr auto operator-(const tensor<T, n...>& A)
 template <typename S, typename T, int... n>
 SERAC_HOST_DEVICE constexpr auto operator-(const tensor<S, n...>& A, const tensor<T, n...>& B)
 {
+  // FIX: hcurl_unit_tests
```
samuelpmishLLNL: what does this refer to?

src/serac/physics/utilities/functional/tests/functional_comparison_L2.cpp
```diff
@@ -30,8 +30,11 @@ std::unique_ptr<mfem::ParMesh> mesh3D;
 template <int p, int dim>
 void functional_test(mfem::ParMesh& mesh, L2<p> test, L2<p> trial, Dimension<dim>)
 {
-  [[maybe_unused]] static constexpr double a = 1.7;
-  [[maybe_unused]] static constexpr double b = 0.0;
+  [[maybe_unused]] static constexpr double a       = 1.7;
+  [[maybe_unused]] static constexpr double b       = 0.0;
+  std::string                              postfix = serac::profiling::concat("_L2<", p, ">");
```
samuelpmishLLNL: This could be more consistent about keeping or omitting the `serac::profiling::` prefix, given the `using namespace serac::profiling;`

Later on, `concat` appears without explicitly specifying its namespace.

src/serac/physics/utilities/functional/tests/functional_comparison_L2.cpp
```diff
@@ -44,7 +47,7 @@ void functional_test(mfem::ParMesh& mesh, L2<p> test, L2<p> trial, Dimension<dim
   A.AddDomainIntegrator(new mfem::MassIntegrator(a_coef));
 
   // Assemble the bilinear form into a matrix
-  A.Assemble(0);
+  SERAC_PROFILE_VOID_EXPR(concat("mfem_localAssemble").c_str(), A.Assemble(0));
```
samuelpmishLLNL: does this `concat` do anything?

jonwong12: Good catch. I forgot to add the `postfix` here.

src/serac/physics/utilities/functional/tests/functional_comparison_L2.cpp
```diff
@@ -127,6 +134,60 @@ TEST(L2, 3D_linear) { functional_test(*mesh3D, L2<1>{}, L2<1>{}, Dimension<3>{})
...
 TEST(L2, 3D_cubic) { functional_test(*mesh3D, L2<3>{}, L2<3>{}, Dimension<3>{}); }
 
+/** CUDA workaround
+Issue with std::variant for InputOptions in mesh_utils.hpp
+
+// this file has a lot of warnings
+serac/src/serac/infrastructure/../../serac/physics/utilities/functional/tensor.hpp(347): warning: calling a __host__
+function("std::tuple< ::serac::tensor<double, (int)3 > ,  ::serac::zero > ::operator =") from a __host__ __device__
+function("serac::operator +< ::serac::dual<    ::std::tuple< ::serac::tensor<double, (int)3 > ,  ::serac::zero > > ,
+double, (int)3 > ") is not allowed
```
samuelpmishLLNL: a more human friendly translation of this warning:

`std::tuple`'s operator= is not marked with `__device__`, so it can't be called from a `__device__` context. 

This is confusing, because nvcc is supposed to support `std::tuple` since 9.2:
https://docs.nvidia.com/cuda/archive/9.2/cuda-toolkit-release-notes/index.html#cuda-compiler-new-features

Screen Shot 2021-06-07 at 4 03 14 PM

My guess is that nvcc only supports the `constexpr` parts of `std::tuple` in device code, and since this copy constructor isn't `constexpr` until C++20, it won't work in device code. That being said, if we want to pursue implementing a small `serac::tuple` object, nothing would prevent us from making those constructors `constexpr` and working on GPUs.

samuelpmishLLNL:
it looks like almost all of the uses of `serac::profiling::concat` are immediately calling `std::string::c_str()` on the result to pass in to the first argument of `SERAC_PROFILE_SOMETHING_EXPR`. Is it worth having something like `serac::profiling::concat_cstr` that does that, or just making the macros accept a `std::string` instead of (or in addition to) `const char *`?

src/serac/physics/utilities/functional/tensor.hpp
jonwong12: Sorry. I think it was related to some of the type deduction nvcc issues.

src/serac/physics/utilities/functional/tests/functional_comparison_L2.cpp
jonwong12: Thanks. I'll permalink this within the comments itself.

jonwong12:
> 
> 
> it looks like almost all of the uses of `serac::profiling::concat` are immediately calling `std::string::c_str()` on the result to pass in to the first argument of `SERAC_PROFILE_SOMETHING_EXPR`. Is it worth having something like `serac::profiling::concat_cstr` that does that, or just making the macros accept a `std::string` instead of (or in addition to) `const char *`?

Great question. Personally I like `std::string` and it works well with `serac::profiling::concat`; at the same time CALIPER seems to want `const char *`. I agree it's a bit cumbersome to add in c_str() everywhere.

I added a templated function to the macros to enforce conversions to `const char *` from either `std::string` or c strings so caliper is happy.

src/serac/infrastructure/profiling.hpp
```diff
@@ -111,21 +120,53 @@ auto&& forwarder(T&& thing)
...
+    return str;
+}
+
+  /**
+   * @brief Converts a std::string into a c string
+   */  
+template <typename T>
+std::enable_if_t<std::is_same_v<std::string, T>, const char * > make_cstr(const T & str) {
+    return str.c_str();
+}
```
joshessman-llnl: Would just an overload be sufficient here?
```cpp
const char* make_cstr(const char*);
const char* make_cstr(const std::string&);
```

jonwong12: Funny. I swore I tried this in godbolt yesterday and it claimed it was ambiguous.

src/serac/infrastructure/profiling.hpp
```diff
@@ -111,21 +119,48 @@ auto&& forwarder(T&& thing)
...
-}  // namespace detail
+/**
+ * @brief Guarantees str is a c string
+ */
+const char* make_cstr(const char* str) { return str; }
+
+/**
+ * @brief Converts a std::string into a c string
+ */
+const char* make_cstr(const std::string& str) { return str.c_str(); }
```
joshessman-llnl:
```suggestion
inline const char* make_cstr(const char* str) { return str; }

/**
 * @brief Converts a std::string into a c string
 */
inline const char* make_cstr(const std::string& str) { return str.c_str(); }
```

src/serac/physics/utilities/functional/CMakeLists.txt
```diff
@@ -5,10 +5,11 @@
 # SPDX-License-Identifier: (BSD-3-Clause)
 
 set(functional_depends serac_physics mfem mpi)
+blt_list_append( TO functional_depends ELEMENTS caliper IF ${SERAC_USE_CALIPER} )
 
-#if(ENABLE_CUDA)
-#    list(APPEND functional_depends cuda)
-#endif()
+if(ENABLE_CUDA)
```
white238: Please use blt_list_append here.  

Small side note:  you dont have to use the value of the IF clause, you can send in the variable:

```
blt_list_append( TO functional_depends ELEMENTS caliper IF SERAC_USE_CALIPER )
```

src/serac/infrastructure/profiling.hpp
```diff
@@ -112,22 +119,49 @@ auto&& forwarder(T&& thing)
...
+  serac::profiling::detail::remove_rvalue_reference<decltype(serac::profiling::detail::forwarder(expr))>::type
 
-#define SERAC_PROFILE_EXPR(name, expr)                                \
-  [&]() -> SERAC_PROFILE_EXPR_RETURN_TYPE(expr) {                     \
-    const cali::ScopeAnnotation SERAC_CONCAT(region, __LINE__)(name); \
-    return static_cast<SERAC_PROFILE_EXPR_RETURN_TYPE(expr)>(expr);   \
+#define SERAC_PROFILE_EXPR(name, expr)                                                                     \
+  [&]() -> typename SERAC_PROFILE_EXPR_RETURN_TYPE(expr) {                                                 \
+    const cali::ScopeAnnotation SERAC_CONCAT(region, __LINE__)(serac::profiling::detail::make_cstr(name)); \
+    return static_cast<typename SERAC_PROFILE_EXPR_RETURN_TYPE(expr)>(expr);                               \
```
samuelpmishLLNL, line 149: do we need the trailing return type here, or does automatic return type deduction do the right thing with the `static_cast`?

joshessman-llnl: I'd have to look up what the exact rules are but if references are involved (as they are here) typically you have to specify the return type explicitly like with `auto` vs `auto&`.

white238 (approved):

jamiebramwell:
@joshessman-llnl @samuelpmishLLNL , are there any more outstanding issues on this PR?

src/serac/infrastructure/profiling.hpp
```diff
@@ -112,22 +119,49 @@ auto&& forwarder(T&& thing)
...
   }()
 
+/**
+ * @brief Profiles an expression several times; Return the last evaluation
+ */
+#define SERAC_PROFILE_EXPR_LOOP(name, expr, ntest)                                                                  \
```
joshessman-llnl, line 155: In this case I think it could be pretty easy for the compiler to recognize that the loop has no side effects, in which case it could optimize it out.

Google Benchmark implements a [DoNotOptimize](https://github.com/google/benchmark/blob/main/include/benchmark/benchmark.h#L330-L366) function that avoids this - possibly we should look into this.

joshessman-llnl (approved):

PR #482, opened 2021-06-04
Put bank name in a variable so that we override it

adrienbernede:
Let's say that struggling to get access to `eng` bank on `lassen` forced me to think of a more generic solution.
Actually I had this idea in the back of my head, and should have gone for it a long time ago:

This simple PR makes the bank configurable, so that anyone can run the pipeline with a bank he/she has access to.
Of course, one still needs to belong to the `smithdev` group in LC to access the shared files (dependencies installations), which makes this not a complete game-changer.

Note: I need this to finally fix Uberenv CI (https://github.com/LLNL/uberenv/pull/65). Uberenv CI will use the radiuss bank. Which is also for the best, I guess.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -8,6 +8,7 @@ variables:
   PROJECT_ALLOC_NAME: ${CI_PROJECT_NAME}_ci_${CI_PIPELINE_ID}
   BUILD_ROOT: ${CI_PROJECT_DIR}
   FULL_BUILD_ROOT: ${CI_BUILDS_DIR}/serac/${CI_JOB_NAME}
+  ALLOC_BANK: eng
 
 stages:
   - allocate
--- a/.gitlab/build_lassen.yml
+++ b/.gitlab/build_lassen.yml
@@ -26,15 +26,15 @@
 # Template
 .src_build_on_lassen:
   stage: build
-  variables: 
-    ALLOC_COMMAND: lalloc 1 -W 10 -q pdebug -G eng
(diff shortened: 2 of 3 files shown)


white238:
eng is not what controls our shared files.  it is smithdev

.gitlab-ci.yml
```diff
@@ -8,6 +8,7 @@ variables:
   PROJECT_ALLOC_NAME: ${CI_PROJECT_NAME}_ci_${CI_PIPELINE_ID}
   BUILD_ROOT: ${CI_PROJECT_DIR}
   FULL_BUILD_ROOT: ${CI_BUILDS_DIR}/serac/${CI_JOB_NAME}
+  ALLOC_BANK: eng
```
white238, line 11: How does putting this in a variable help you to run our CI if you don't have `eng`?  Is it overrideable through the environment or does the calling pipeline take precedence?

adrienbernede: Because then, either in the UI or in a pipeline trigger, I can override it.

adrienbernede: Like here: https://github.com/LLNL/uberenv/blob/74400647c8b273c7e74d3f8d5947c4dfec929a99/.gitlab-ci.yml#L63

adrienbernede:
> eng is not what controls our shared files. it is smithdev

Someone is reading my verbiage!

white238 (approved):

jamiebramwell (approved):

PR #484, opened 2021-06-08
Functional feature: Computes matrix entries for H1/Hcurl

jonwong12:
This PR adds a `gradient_matrix_kernel` to functional which will compute the element tangent entries of the residual in the form of a continuous vector `K_e` of size (test_ndof * test_dim * trial_ndof * trial_dim * num_elem). It should work for H1, Hcurl, and L2 spaces. However L2 is not tested specifically in this PR.

Afterwards an `serac::mfem_ext::AssembledSparseMatrix` is constructed. This holds the mapping from `K_e` to a rank-local assembled CSR-matrix. Calling `AssembledSparseMatrix.FillData(K_e)` will maintain the same sparsity pattern but change matrix entries. `AssembledSparseMatrix` is derived from `mfem::SparseMatrix`; however `FillData()` and `ParallelAssemble()` methods were added.

Note:
Currently `AssembledSparseMatrix` uses some "forbidden C++ calls" to get at protected information in `mfem::ElementRestriction`. Ideally this information would be `public` in mfem. Additionally there is an mfem issue with L2 when using LEXICOGRAPHIC ordering which a follow-up PR will try to address. For `NATIVE` ordering, see this [branch](https://github.com/mfem/mfem/blob/f1366b463d61437d330f886729ccd3713b052cae/examples/ex1.cpp#L239) for a working H1-L2 matrix assembly.

TODO:
- [x] Add validation checks for hcurl and elasticity
- [x] Add second method in functional.hpp to update an AssembledSparseMatrix directly

### DIFF
--- a/src/serac/numerics/assembled_sparse_matrix.cpp
+++ b/src/serac/numerics/assembled_sparse_matrix.cpp
@@ -1,0 +1,273 @@
+
+#include "mfem/linalg/dtensor.hpp"
+#include "assembled_sparse_matrix.hpp"
+
+template <auto offsetV, auto indicesV, auto gatherV>
+struct forbidden_restriction {
+  //  friend result_fes _get_fes(From & from) { return & from.*fesV; }
+  friend mfem::Array<int>& __get_offsets(mfem::ElementRestriction& from) { return from.*offsetV; }
+  friend mfem::Array<int>& __get_indices(mfem::ElementRestriction& from) { return from.*indicesV; }
+  friend mfem::Array<int>& __get_gatherMap(mfem::ElementRestriction& from) { return from.*gatherV; }
+};
+
+// const FiniteElementSpace * __get_fes(mfem::ElementRestriction &);
+mfem::Array<int>& __get_offsets(mfem::ElementRestriction&);
+mfem::Array<int>& __get_indices(mfem::ElementRestriction&);
+mfem::Array<int>& __get_gatherMap(mfem::ElementRestriction&);
+
+template struct forbidden_restriction<&mfem::ElementRestriction::offsets, &mfem::ElementRestriction::indices,
+                                      &mfem::ElementRestriction::gatherMap>;
+
+namespace serac {
+  namespace mfem_ext {
+
+    AssembledSparseMatrix::AssembledSparseMatrix(const mfem::FiniteElementSpace& test,   // test_elem_dofs * ne * vdim x vdim * test_ndofs
+						 const mfem::FiniteElementSpace& trial,  // trial_elem_dofs * ne * vdim x vdim * trial_ndofs
+						 mfem::ElementDofOrdering        elem_order)
+      : mfem::SparseMatrix(test.GetNDofs() * test.GetVDim(), trial.GetNDofs() * trial.GetVDim()),
+      test_fes(test),
+      trial_fes(trial),
+      test_restriction(test, elem_order),
+      trial_restriction(trial, elem_order),
+      elem_ordering(elem_order)
+    {
+      GetMemoryI().New(Height() + 1, GetMemoryI().GetMemoryType());
+
+      const int nnz = FillI();
+      GetMemoryJ().New(nnz, GetMemoryJ().GetMemoryType());
+      GetMemoryData().New(nnz, GetMemoryData().GetMemoryType());
+      FillJ();
+
+      // zero initialize the data
+      for (int i = 0; i < nnz; i++) {
+	A[i] = 0.;
+      }
+    }
+
+    int AssembledSparseMatrix::FillI()
+    {
+      // ElementRestriction creates a CSR matrix that maps vdof -> (dof, ne).
+      // offsets are the row offsets corresponding to a vdof
+      // indices maps a given vdof to the the assembled element matrix vector (dof * ne + d).
+      // gatherMap takes an element matrix vector offset (dof, ne) and returns the partition-local vdof (d.o.f. id).
+      [[maybe_unused]] auto& test_offsets = __get_offsets(test_restriction);
+      [[maybe_unused]] auto& test_indices = __get_indices(test_restriction);    
+      [[maybe_unused]] auto& test_gatherMap  = __get_gatherMap(test_restriction);
+      [[maybe_unused]] auto& trial_offsets   = __get_offsets(trial_restriction);
+      [[maybe_unused]] auto& trial_indices   = __get_indices(trial_restriction);
+      [[maybe_unused]] auto& trial_gatherMap = __get_gatherMap(trial_restriction);
+
+      /**
+	 We expect mat_ea to be of size (test_elem_dof * test_vdim, trial_elem_dof * trial_vdim, ne)
+	 We assume a consistent striding from (elem_dof, vd) within each element
+      */
+      const int test_elem_dof  = test_fes.GetFE(0)->GetDof();
+      const int trial_elem_dof = trial_fes.GetFE(0)->GetDof();
+      const int test_vdim      = test_fes.GetVDim();
+      const int trial_vdim     = trial_fes.GetVDim();
+      const int test_ndofs     = test_fes.GetNDofs();
+
+      for (int i = 0; i < test_vdim * test_ndofs; i++) {
+	I[i] = 0;
+      }
+
+      for (int test_vdof = 0; test_vdof < test_fes.GetNDofs(); test_vdof++) {
+	// Look through each element corresponding to a test_vdof
+	const int test_row_offset = test_offsets[test_vdof];
+	const int nrow_elems      = test_offsets[test_vdof + 1] - test_row_offset;
+
+	// Build temporary array to get rid of duplicates
+	mfem::Array<int> trial_vdofs(nrow_elems * trial_elem_dof);
+	trial_vdofs = -1;
+	int nnz_row = 0;
+	for (int e_index = 0; e_index < nrow_elems; e_index++) {
+	  // test_indices can be negative in the case of Hcurl
+	  const int                  test_index_v = test_indices[test_row_offset + e_index];
+	  const int                  test_index = test_index_v >= 0 ? test_index_v : -test_index_v - 1;
+	  const int                  e           = test_index / test_elem_dof;
+	  [[maybe_unused]] const int test_i_elem = test_index % test_elem_dof;
+
+	  // find corresponding trial_vdofs
+	  mfem::Array<int> trial_elem_vdofs(trial_elem_dof);
+	  for (int j = 0; j < trial_elem_dof; j++) {
+	    // this might be negative
+	    const auto trial_j_vdof_v = trial_gatherMap[trial_elem_dof * e + j];
+	    const auto trial_j_vdof = trial_j_vdof_v >= 0 ? trial_j_vdof_v : -1 -trial_j_vdof_v;
+	    trial_elem_vdofs[j]     = trial_j_vdof;
+	    if (trial_vdofs.Find(trial_j_vdof) == -1) {
+	      // we haven't seen this before
+	      trial_vdofs[nnz_row] = trial_j_vdof;
+	      nnz_row++;
+	    }
+	  }
+	}
+
+	// add entries to I
+	for (int vi = 0; vi < test_vdim; vi++) {
+	  const auto nnz_index_v = test_fes.DofToVDof(test_vdof, vi);
+	  const auto nnz_index = nnz_index_v >= 0 ? nnz_index_v : -1 -nnz_index_v;
+	  I[nnz_index] = nnz_row * trial_vdim;
+	}
+      }
+
+      // Perform inclusive scan on all entries
+      int nnz = 0;
+      for (int i = 0; i < test_ndofs * trial_vdim; i++) {
+	int temp = I[i];
+	I[i]     = nnz;
+	nnz += temp;
+      }
+      I[test_ndofs * trial_vdim] = nnz;
+
+      return nnz;
+    }
+
+    void AssembledSparseMatrix::FillJ()
+    {
+
+      [[maybe_unused]] auto& test_offsets = __get_offsets(test_restriction);
+      [[maybe_unused]] auto& test_indices = __get_indices(test_restriction);    
+      [[maybe_unused]] auto& test_gatherMap  = __get_gatherMap(test_restriction);
+      [[maybe_unused]] auto& trial_offsets   = __get_offsets(trial_restriction);
+      [[maybe_unused]] auto& trial_indices   = __get_indices(trial_restriction);
+      [[maybe_unused]] auto& trial_gatherMap = __get_gatherMap(trial_restriction);
+  
+      const int                  test_elem_dof  = test_fes.GetFE(0)->GetDof();
+      const int                  trial_elem_dof = trial_fes.GetFE(0)->GetDof();
+      const int                  test_vdim      = test_fes.GetVDim();
+      const int                  trial_vdim     = trial_fes.GetVDim();
+      [[maybe_unused]] const int test_ndofs     = test_fes.GetNDofs();
+
+      const int ne = trial_fes.GetNE();
+      ea_map.SetSize(test_elem_dof * test_vdim * trial_elem_dof * trial_vdim * ne);
+      auto map_ea = mfem::Reshape(ea_map.ReadWrite(), test_elem_dof * test_vdim, trial_elem_dof * trial_vdim, ne);
+
+      // initialize J
+      for (int j = 0; j < this->J.Capacity(); j++) {
+	this->J[j] = -1;
+      }
+
+      for (int test_vdof = 0; test_vdof < test_fes.GetNDofs(); test_vdof++) {
+	// Look through each element corresponding to a test_vdof
+	const int test_row_offset = test_offsets[test_vdof];
+	const int nrow_elems      = test_offsets[test_vdof + 1] - test_row_offset;
+
+	// here we assume all the components have the same number of columns
+	const int        nnz_row = I[test_fes.DofToVDof(test_vdof, 0) + 1] - I[test_fes.DofToVDof(test_vdof, 0)];
+	mfem::Array<int> trial_vdofs(nnz_row);
+	trial_vdofs      = -1;
+	int j_vdof_index = 0;
+
+	// Build temporary array for assembled J
+	for (int e_index = 0; e_index < nrow_elems; e_index++) {
+	  // test_indices can be negative in the case of Hcurl
+	  const int                  test_index_v = test_indices[test_row_offset + e_index];
+	  const int                  test_index = test_index_v >= 0 ? test_index_v : -test_index_v - 1;
+	  const int                  e           = test_index / test_elem_dof;
+	  [[maybe_unused]] const int test_i_elem = test_index % test_elem_dof;
+
+	  // find corresponding trial_vdofs
+	  mfem::Array<int> trial_elem_vdofs(trial_elem_dof);
+	  for (int j_elem = 0; j_elem < trial_elem_dof; j_elem++) {
+	    // could be negative.. but trial_elem_vdofs is a temporary array
+	    const auto trial_j_vdof_v  = trial_gatherMap[trial_elem_dof * e + j_elem];
+	    const auto trial_j_vdof = trial_j_vdof_v >= 0 ? trial_j_vdof_v : -1 -trial_j_vdof_v;
+	    trial_elem_vdofs[j_elem] = trial_j_vdof;
+
+	    // since trial_j_vdof could be negative but there are now two indices that point to the same dof (just oriented differently).. we only want to search for positive oriented indices
+	    auto find_index = trial_vdofs.Find(trial_j_vdof);
+	    if (find_index == -1) {
+	      // we haven't seen this before
+	      trial_vdofs[j_vdof_index] = trial_j_vdof;
+
+	      // we can add this entry to J
+	      for (int vi = 0; vi < test_vdim; vi++) {
+		const auto i_dof_offset = I[test_fes.DofToVDof(test_vdof, vi)];
+
+		// this access pattern corresnponds to j_vdof_index + vj * nnz_row
+		for (int vj = 0; vj < trial_vdim; vj++) {
+		  const auto column_index = j_vdof_index + vj * nnz_row / trial_vdim;
+		  const auto j_nnz_index  = i_dof_offset + column_index;
+		  // this index may be negative, but J needs to be positive
+		  const auto j_value = trial_fes.DofToVDof(trial_j_vdof_v, vj);
+		  J[j_nnz_index]          = j_value >= 0 ? j_value : -1-j_value;
+		}
+	      }
+
+	      // write mapping from ea to csr_nnz_index (can probably optimize this)
+	      for (int vi = 0; vi < test_vdim; vi++) {
+		const auto i_dof_offset = I[test_fes.DofToVDof(test_vdof, vi)];
+		for (int vj = 0; vj < trial_vdim; vj++) {
+		  const auto column_index = j_vdof_index + vj * nnz_row / trial_vdim;
+		  const int index_val = i_dof_offset + column_index;
+		  const int trial_index = trial_fes.DofToVDof(trial_j_vdof_v, vj);
+		  const int orientation_factor = (test_index_v >= 0 ? 1 : -1) * (trial_index >=0 ? 1 : -1);
+		  map_ea(test_i_elem + test_elem_dof * vi, j_elem + trial_elem_dof * vj, e) =
+		    orientation_factor > 0 ? index_val : -1-index_val;
+		}
+	      }
+
+	      j_vdof_index++;
+	    } else {
+	      // this is a duplicate entry
+	      // write mapping from ea to csr_nnz_index (can probably optimize this)
+	      for (int vi = 0; vi < test_vdim; vi++) {
+		const auto i_dof_offset = I[test_fes.DofToVDof(test_vdof, vi)];
+		for (int vj = 0; vj < trial_vdim; vj++) {
+		  const auto column_index = find_index + vj * nnz_row / trial_vdim;
+		  const int index_val = i_dof_offset + column_index;
+		  const int trial_index = trial_fes.DofToVDof(trial_j_vdof_v, vj);
+		  const int orientation_factor = (test_index_v >= 0 ? 1 : -1) * (trial_index >=0 ? 1 : -1);      
+		  map_ea(test_i_elem + test_elem_dof * vi, j_elem + trial_elem_dof * vj, e) =
+		    orientation_factor > 0 ? index_val : -1-index_val;
+		}
+	      }
+	    }
+	  }
+	}
+      }
+    }
+    void AssembledSparseMatrix::FillData(const mfem::Vector& ea_data)
+    {
+      auto Data = WriteData();
+
+      [[maybe_unused]] auto& test_offsets = __get_offsets(test_restriction);
+      [[maybe_unused]] auto& test_indices = __get_indices(test_restriction);    
+      [[maybe_unused]] auto& test_gatherMap  = __get_gatherMap(test_restriction);
+      [[maybe_unused]] auto& trial_offsets   = __get_offsets(trial_restriction);
+      [[maybe_unused]] auto& trial_indices   = __get_indices(trial_restriction);
+      [[maybe_unused]] auto& trial_gatherMap = __get_gatherMap(trial_restriction);
+  
+      const int                  test_elem_dof  = test_fes.GetFE(0)->GetDof();
+      const int                  trial_elem_dof = trial_fes.GetFE(0)->GetDof();
+      const int                  test_vdim      = test_fes.GetVDim();
+      const int                  trial_vdim     = trial_fes.GetVDim();
+      [[maybe_unused]] const int test_ndofs     = test_fes.GetNDofs();
+
+      const int ne = trial_fes.GetNE();
+
+      //      auto map_ea = mfem::Reshape(ea_map.Read(), test_elem_dof * test_vdim, trial_elem_dof * trial_vdim, ne);
+      auto map_ea = mfem::Reshape(ea_map.Read(), test_elem_dof * test_vdim, trial_elem_dof * trial_vdim, ne);
+
+      auto mat_ea = mfem::Reshape(ea_data.Read(), test_elem_dof * test_vdim, trial_elem_dof * trial_vdim, ne);
+
+      // Use map_ea to take ea_data directly to CSR entry
+      for (int e = 0; e < ne; e++) {
+	for (int i_elem = 0; i_elem < test_elem_dof; i_elem++) {
+	  for (int vi = 0; vi < test_vdim; vi++) {
+	    for (int j_elem = 0; j_elem < trial_elem_dof; j_elem++) {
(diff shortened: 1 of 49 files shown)


src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,274 @@
...
+  //  friend result_fes _get_fes(From & from) { return & from.*fesV; }
+  friend mfem::Array<int>& __get_offsets(mfem::ElementRestriction& from) { return from.*offsetV; }
+  friend mfem::Array<int>& __get_indices(mfem::ElementRestriction& from) { return from.*indicesV; }
+  friend mfem::Array<int>& __get_gatherMap(mfem::ElementRestriction& from) { return from.*gatherV; }
+};
+
+// const FiniteElementSpace * __get_fes(mfem::ElementRestriction &);
+mfem::Array<int>& __get_offsets(mfem::ElementRestriction&);
+mfem::Array<int>& __get_indices(mfem::ElementRestriction&);
+mfem::Array<int>& __get_gatherMap(mfem::ElementRestriction&);
```
samuelpmishLLNL: It's unlikely to actually be an issue, but names starting with double underscore are technically reserved for the language implementation / compiler, so I think these should be renamed.

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,274 @@
...
+#include "mfem/linalg/dtensor.hpp"
+#include "assembled_sparse_matrix.hpp"
+
+template <auto offsetV, auto indicesV, auto gatherV>
+struct forbidden_restriction {
+  //  friend result_fes _get_fes(From & from) { return & from.*fesV; }
+  friend mfem::Array<int>& __get_offsets(mfem::ElementRestriction& from) { return from.*offsetV; }
+  friend mfem::Array<int>& __get_indices(mfem::ElementRestriction& from) { return from.*indicesV; }
+  friend mfem::Array<int>& __get_gatherMap(mfem::ElementRestriction& from) { return from.*gatherV; }
+};
```
samuelpmishLLNL: I presume this is related to getting at protected members of `mfem::ElementRestriction`, but this needs some explanation.

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,274 @@
...
+  // ElementRestriction creates a CSR matrix that maps vdof -> (dof, ne).
+  // offsets are the row offsets corresponding to a vdof
+  // indices maps a given vdof to the the assembled element matrix vector (dof * ne + d).
+  // gatherMap takes an element matrix vector offset (dof, ne) and returns the partition-local vdof (d.o.f. id).
+  [[maybe_unused]] auto& test_offsets    = __get_offsets(test_restriction);
+  [[maybe_unused]] auto& test_indices    = __get_indices(test_restriction);
+  [[maybe_unused]] auto& test_gatherMap  = __get_gatherMap(test_restriction);
+  [[maybe_unused]] auto& trial_offsets   = __get_offsets(trial_restriction);
+  [[maybe_unused]] auto& trial_indices   = __get_indices(trial_restriction);
+  [[maybe_unused]] auto& trial_gatherMap = __get_gatherMap(trial_restriction);
```
samuelpmishLLNL: these appear to be used, do we need the `[[maybe_unused]]` attribute?

src/serac/numerics/assembled_sparse_matrix.hpp
```diff
@@ -0,0 +1,41 @@
...
+  const mfem::FiniteElementSpace& trial_fes;
+  // class local ElementRestriction objects
+  mfem::ElementRestriction test_restriction;
+  mfem::ElementRestriction trial_restriction;
+  mfem::ElementDofOrdering elem_ordering;
+  mfem::Array<int>         ea_map;
+
+private:
+  int  FillI();
+  void FillJ();
```
samuelpmishLLNL: what do these functions do?

src/serac/physics/utilities/functional/detail/metaprogramming.hpp
```diff
@@ -118,3 +118,82 @@ inline constexpr void for_constexpr(lambda&& f)
 {
   detail::for_constexpr(f, std::make_integer_sequence<int, n>{}...);
 }
+
+// sam's templates for undoing for_constexpr for faster compile times
+template <int n1, typename lambda>
+void for_loop(lambda f)
```
samuelpmishLLNL, line 124: we may want to consider passing the lambda by rvalue reference, but I'm not 100% sure when that is preferable

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -292,6 +292,16 @@ class Functional<test(trial)> : public mfem::Operator {
...
 
+  /**
+   * @brief Obtains the element stiffness matrix reshaped a mfem::Vector
+   * @param[inout] K_e A mfem::Vector containing the assembled element stiffness matrix (test_dim * test_ndof, trial_dim
+   * * trial_ndof, nelem)
+   */
+  virtual void GradientMatrix(mfem::Vector& K_e) const
+  {
+    for (auto domain : domain_integrals_) domain.GradientMatrix(K_e);
+  }
```
samuelpmishLLNL: Does this need to be `virtual`? 

The function is also a little bit misleading, since it is returning a packed array of element matrices, not the actual gradient matrix itself. Maybe `ComputeElementMatrices()` or something similar would be clearer (or `GradientMatrix` could return the `AssembledSparseMatrix` object?)

I prefer return values to out parameters when possible, but I'm not sure if we have anything about that choice in our style guide.

src/serac/physics/utilities/functional/tensor.hpp
```diff
@@ -1471,4 +1490,35 @@ auto chain_rule(const tensor<double, m, n, p...>& df_dx, const tensor<double, p.
   return total;
 }
 
+/**
+ * @brief tensor conversions since double and tensor are different types
+ */
```
samuelpmishLLNL: maybe call this `convert_to_tensor` or `convert_to_tensor_with_shape` (or even overload `make_tensor`), since `convert` doesn't really convey what is happening

src/serac/physics/utilities/functional/tests/functional_comparisons.cpp
```diff
@@ -113,18 +122,45 @@ void functional_test(mfem::ParMesh& mesh, H1<p> test, H1<p> trial, Dimension<dim
...
+  serac::mfem_ext::AssembledSparseMatrix A_serac_mat(fespace, fespace, mfem::ElementDofOrdering::LEXICOGRAPHIC);
+  {
+    SERAC_PROFILE_SCOPE(concat("AssembledSparseMatrix_FillData", postfix).c_str());
+    A_serac_mat.FillData(K_e);
+  }
+
+  A_serac_mat.Finalize();
+
+  std::unique_ptr<mfem::HypreParMatrix> J2(
+      SERAC_PROFILE_EXPR(concat("functional_gradParAssemble", postfix).c_str(), A.ParallelAssemble(&A_serac_mat)));
```
samuelpmishLLNL: will these intermediate quantities (K_e, A_serac_mat) be generally useful, or should `Functional::GradientMatrix` just return the HypreParMatrix itself?

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,274 @@
+
+#include "mfem/linalg/dtensor.hpp"
+#include "assembled_sparse_matrix.hpp"
+
+template <auto offsetV, auto indicesV, auto gatherV>
+struct forbidden_restriction {
+  //  friend result_fes _get_fes(From & from) { return & from.*fesV; }
```
joshessman-llnl: Dead code?

Also, can these "tricks" be moved into a `detail` namespace?

src/serac/numerics/assembled_sparse_matrix.cpp
joshessman-llnl: Here's an article that explains the trick being used: https://accu.org/journals/overload/28/156/harrison_2776/

In addition to linking this we could also document why it's needed for this particular case and maybe raise an MFEM issue since presumably this isn't the long-term solution we want

src/serac/physics/utilities/functional/functional.hpp
jonwong12: Great suggestions and questions. As is, I don't think it needs to be `virtual`. I think it depends on how we want it to interface with CUSPARSE?

I intended the sparsity pattern to be encoded when constructed and updated as we go along. Since every new state will involve new values, but not necessarily new sparsity, I didn't want to return another `AssembledSparseMatrix`? I also wanted to be able to accumulate various domain_integrals as we go along. I think ideally on the GPU we want a block of memory that's pre-allocated that we just change the values directly on the GPU itself.

src/serac/physics/utilities/functional/tests/functional_comparisons.cpp
jonwong12: I think it depends on how many times you think you're going to re-assemble the matrix. Here I've assumed we're going to possibly assemble the same sparsity pattern /connectivity multiple times, so I would claim that `A_serac_mat` is more of a permanent quantity that is updated with `K_e`. I'm not sure, but I think HypreParMatrix needs to be possibly be "recomputed" each time the rank-local matrix is assembled. So I think that is the "temporary" quantity that we actually use to compute with.

src/serac/physics/utilities/functional/functional.hpp
jonwong12: Maybe one possibility is to pass in an `AssembledSparseMatrix`. We can then do some check on that and encode it directly into the final storage pattern. While `K_e` is a temporary, it's matrix-format agnostic; I know in the past we had talked about using other CUDA-supported matrix formats.

jonwong12:
@joshessman-llnl I have a issue/branch in MFEM to do this matrix assembly, but I don't know what the timeline for that is. I think they are interested in seeing some benchmarks. In that branch, I just `friend` `AssembledSparseMatrix` in `mfem::ElementRestriction`, so I agree this is a temporary solution. 

One point of contention is that MFEM seems more concerned with higher order elements where assembling a matrix may have declining payoffs, whereas for the problems we solve, matrices are useful.

Alternatively this PR doesn't support L2 (LEXICOGRAPHIC ordering, but NATIVE ordering is fine?), because `mfem::ElementRestriction` and `mfem::L2ElementRestriction` are not derived from each other. We could re-implement or use variants (and get at protected arrays). @jamiebramwell  What are your thoughts on this?

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: Another option is to make the `Functional` object maintain a `AssembledSparseMatrix` as a member variable. If the user calls `GradientMatrix`, it could form that matrix (and only figure out sparsity on the first call to `GradientMatrix`, caching it on subsequent calls), and return a reference to it.

jonwong12: @samuelpmishLLNL I think that might work nicely. I'd like to keep support for `K_e` though, as it would make it easier to compare new matrix formats and so forth. Once you "assemble", it will be difficult to retrieve `K_e`.

samuelpmishLLNL: There are definitely cases where you want just the element matrices, but I would like to also provide an interface that doesn't require the user jumping through too many hoops to just get the gradient matrix. Something like this would allow for both use cases:

```cpp
class Functional{
  AssembledSparseMatrix K; // gradient matrix
  mfem::Vector K_e; // element stiffness matrices packed into 1D array

  ...

  const auto & ElementMatrices(){
    calculateElementMatrices(K_e);
    return K_e; 
  }

  const auto & GradientMatrix(){
    calculateElementMatrices(K_e);
    return K = assemble(K_e);
  }
};

File: src/serac/physics/utilities/functional/detail/metaprogramming.hpp [hunk 10]
```diff
@@ -118,3 +118,82 @@ inline constexpr void for_constexpr(lambda&& f)
 {
   detail::for_constexpr(f, std::make_integer_sequence<int, n>{}...);
 }
+
+// sam's templates for undoing for_constexpr for faster compile times
+template <int n1, typename lambda>
+void for_loop(const lambda& f)
```
joshessman-llnl: It's usually fine to pass these "by value" but if they're references they need to be forwarding/universal references, i.e., `lambda&& f`. I think `mutable` lambdas are not `const` and couldn't be used with a `const lambda& f`

src/serac/physics/utilities/functional/functional.hpp
jonwong12: I changed it now to build an AssembledSparseMatrix on the first call to "GradientMatrix()". It then always returns the internal K.

src/serac/physics/utilities/functional/detail/metaprogramming.hpp
jonwong12: Alright I'll change it back to "pass by value".

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,263 @@
...
+#include "assembled_sparse_matrix.hpp"
+
+namespace detail {
+/**
+ * We'd like to access the protected offset, indices, and gatherMap variables from mfem::ElementRestriction
+ * to compute our sparsity patterns in a manner consistent with MFEM.
+ * In order to do that, we use the following template tricks.
+ * Here's an article that explains the trick being used: https://accu.org/journals/overload/28/156/harrison_2776/
+ */
+template <auto offsetV, auto indicesV, auto gatherV>
```
samuelpmishLLNL, line 12: Interesting, I missed the `auto` templates when I skimmed this earlier. I thought this was encoding type information into `forbidden_restriction` initially, but it's actually tracking member variable pointers. What a weird trick!

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,263 @@
...
+struct forbidden_restriction {
+  friend mfem::Array<int>& ElementRestriction_offsets(mfem::ElementRestriction& from) { return from.*offsetV; }
+  friend mfem::Array<int>& ElementRestriction_indices(mfem::ElementRestriction& from) { return from.*indicesV; }
+  friend mfem::Array<int>& ElementRestriction_gatherMap(mfem::ElementRestriction& from) { return from.*gatherV; }
```
samuelpmishLLNL: the `.*` makes me think of https://www.mathworks.com/help/matlab/ref/times.html

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,263 @@
...
+  // indices maps a given vdof to the the assembled element matrix vector (dof * ne + d).
+  // gatherMap takes an element matrix vector offset (dof, ne) and returns the partition-local vdof (d.o.f. id).
+  auto& test_offsets    = detail::ElementRestriction_offsets(test_restriction_);
+  auto& test_indices    = detail::ElementRestriction_indices(test_restriction_);
+  auto& trial_gatherMap = detail::ElementRestriction_gatherMap(trial_restriction_);
+
+  /**
+     We expect mat_ea to be of size (test_elem_dof * test_vdim, trial_elem_dof * trial_vdim, ne)
+     We assume a consistent striding from (elem_dof, vd) within each element
+  */
```
samuelpmishLLNL, line 106: 👍

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  // Perform inclusive scan on all entries
+  int nnz = 0;
+  for (int i = 0; i < test_ndofs * trial_vdim; i++) {
+    int temp = I[i];
+    I[i]     = nnz;
+    nnz += temp;
+  }
+  I[test_ndofs * trial_vdim] = nnz;
```
samuelpmishLLNL: If `I[0]` is always zero, I think it's an exclusive scan. I have not tried it yet, but apparently C++17 introduced
https://en.cppreference.com/w/cpp/algorithm/exclusive_scan as well. If that fits our needs, I think it would be good to use the standard library algorithms when possible.

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,263 @@
...
+      const int test_index   = test_index_v >= 0 ? test_index_v : -test_index_v - 1;
+      const int e            = test_index / test_elem_dof;
+      const int test_i_elem  = test_index % test_elem_dof;
+
+      // find corresponding trial_vdofs
+      mfem::Array<int> trial_elem_vdofs(trial_elem_dof);
+      for (int j_elem = 0; j_elem < trial_elem_dof; j_elem++) {
+        // could be negative.. but trial_elem_vdofs is a temporary array
+        const auto trial_j_vdof_v = trial_gatherMap[trial_elem_dof * e + j_elem];
+        const auto trial_j_vdof   = trial_j_vdof_v >= 0 ? trial_j_vdof_v : -1 - trial_j_vdof_v;
```
samuelpmishLLNL: This same ternary construction shows up all over mfem too. I wish this weird sign encoding was encapsulated in a type rather than just being copy and pasted everywhere (e.g. https://godbolt.org/z/xGhnbMefo) 

That is probably out of our control though, since we're using mfem's data types.

src/serac/numerics/assembled_sparse_matrix.hpp
```diff
@@ -0,0 +1,76 @@
+#pragma once
+
+#include <mfem.hpp>
+
+namespace serac {
+namespace mfem_ext {
+
+/**
+ @brief Creates a CSR sparse matrix from element matrices assembled usign a mfem::ElementDofOrdering
```
samuelpmishLLNL:
```suggestion
 @brief Creates a CSR sparse matrix from element matrices assembled using a mfem::ElementDofOrdering
```

src/serac/numerics/assembled_sparse_matrix.hpp
```diff
@@ -0,0 +1,76 @@
...
+  // Consistent element ordering
+  mfem::ElementDofOrdering elem_ordering_;
+
+  // Maps individual element matrix entries in the K_e vector to the final CSR data offset
+  mfem::Array<int> ea_map_;
+
+private:
+  // Computes the row offsets for the CSR sparsity pattern given by the spaces test(trial)
+  int FillI();
+  // Computes the column indicies per row for a CSR sparsity pattern given by the spaces test(trial)
```
samuelpmishLLNL:
```suggestion
  // Computes the column indices per row for a CSR sparsity pattern given by the spaces test(trial)
```

src/serac/numerics/assembled_sparse_matrix.cpp
jonwong12: It is always zero for hypre CSR I think. I didn't know there was an inclusive_scan/exclusive_scan available. The CSR format requires nrows+1 offsets, so I guess you can generate it a few different ways.

I guess in the GPU or RAJA case we can use a parallel exclusive_scan operation.

src/serac/numerics/assembled_sparse_matrix.cpp
jonwong12: I think we can add it in our side. That's a good idea to tidy up the code.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -522,6 +561,20 @@ class Functional<test(trial)> : public mfem::Operator {
...
+
+  /**
+   * @brief The element tangent matrix vector for ComputeElementMatrices() and UpdateAssembledSparseMatrix() calls
+   */
+  mutable mfem::Vector K_e_;
+
+  /**
+   * @brief Local AssembledSparseMatrix storage for ComputeElementMatrices
+   *
+   * If unique_ptr is empty, construct AssembledSparsMatrix.
```
```suggestion
   * If unique_ptr is empty, construct AssembledSparseMatrix.
```

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -522,6 +561,20 @@ class Functional<test(trial)> : public mfem::Operator {
    * @brief The gradient object used to implement @p GetGradient
    */
   mutable Gradient grad_;
+
+  /**
+   * @brief The element tangent matrix vector for ComputeElementMatrices() and UpdateAssembledSparseMatrix() calls
```
samuelpmishLLNL: "matrix vector" sounds funny to me, maybe something like

```suggestion
   * @brief storage buffer for element stiffness matrices, used in ComputeElementMatrices() and UpdateAssembledSparseMatrix()
```

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -522,6 +561,20 @@ class Functional<test(trial)> : public mfem::Operator {
...
+   */
+  mutable mfem::Vector K_e_;
+
+  /**
+   * @brief Local AssembledSparseMatrix storage for ComputeElementMatrices
+   *
+   * If unique_ptr is empty, construct AssembledSparsMatrix.
+   *
+   * @return reference to internal AssembledSparseMatrix
```
samuelpmishLLNL: it looks like `assembled_spmat_` is a member variable, but the documentation makes it sound like a function (having a return value)

src/serac/physics/utilities/functional/integral.hpp
```diff
@@ -439,6 +439,243 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
+      auto                    dxi_q = rule.weights[q];
+      auto                    J_q = make_tensor<spatial_dim, geometry_dim>([&](int i, int j) { return J(q, i, j, e); });
+      auto                    detJ_q = detail::Measure(J_q);
+      [[maybe_unused]] double dx     = detJ_q * dxi_q;
+
+      // recall the derivative of the q-function w.r.t. its arguments at this quadrature point
+      auto dq_darg = derivatives_ptr[e * int(rule.size()) + q];
+
+      // use the chain rule to compute the first-order change in the q-function output
+      // auto dq = chain_rule(dq_darg, darg);
```
samuelpmishLLNL: I think these comments don't apply here

src/serac/physics/utilities/functional/integral.hpp
```diff
@@ -439,6 +439,243 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
+                    M[i] * df0_dgradu[id][jd][dummy_i] * dN_dx[j][dummy_i] * dx;
+              });
+        }
+
+        // // df1_du stiffness contribution
+        // size(dM_dx) = test_ndof x spatial_dim
+        // size(N) = trial_ndof
+        // size(df1_du) = test_dim x spatial_dim x trial_dim
+        if constexpr (!is_zero<decltype(f10)>::value) {
+          auto& df1_du = f10;
```
samuelpmishLLNL, line 553: some of these are just references (i.e. f10, f11), others convert to tensors with some assumed shape (i.e. f00, f01).

This part of the code will likely be refactored in the near future, but I was wondering if there was a reason for the separate treatment of those different derivative terms.

src/serac/physics/utilities/functional/integral.hpp
```diff
@@ -624,6 +866,13 @@ class Integral {
    */
   void GradientMult(const mfem::Vector& input_E, mfem::Vector& output_E) const { gradient_(input_E, output_E); }
 
+  /**
+   * @brief Computes the element stiffness matrices in as a reshaped vector
```
samuelpmishLLNL:
```suggestion
   * @brief Computes the element stiffness matrices, storing them in an `mfem::Vector` that has been reshaped into a multidimensional array
```

src/serac/physics/utilities/functional/integral.hpp
```diff
@@ -624,6 +866,13 @@ class Integral {
...
+   * @param[inout] K_e The reshaped vector as a mfem::DeviceTensor of size (test_dim * test_dof, trial_dim * trial_dof,
```
samuelpmishLLNL, line 871: 👍

src/serac/numerics/assembled_sparse_matrix.cpp
samuelpmishLLNL: it looks like these scan operations support execution space template arguments, so they may already support parallel execution. 

inclusive and exclusive scans were added as part of the Parallelism_TS to C++17 by Jared Hoberock, NVIDIA's main developer of thrust. So it sounds like he likely just added many of the parallel algorithms from thrust directly into the C++ standard!

samuelpmishLLNL:
This all looks very promising to me! 

I want to apologize again for not providing a rich enough set of tools to make the implementation of `gradient_matrix_kernel` straightforward. I am hopeful that we can significantly simplify `gradient_matrix_kernel` by adding some additional features to the tensor class. In the meantime, having this working example will be incredibly useful when we eventually pursue that refactor.

src/serac/numerics/assembled_sparse_matrix.cpp
jonwong12: I've added some temporary sign-related templated methods just to help document the code a little more clearly.

src/serac/numerics/assembled_sparse_matrix.cpp
jonwong12: Interesting! Anyways I added a note in the code. since it seems that gcc 8.3.1 and clang 10.0.0 on LC don't support the scan operations unfortunately.

src/serac/physics/utilities/functional/integral.hpp
jonwong12: No. The conversion routines are really workarounds to handle the derivatives being `serac::zero` or `double` types. I tried to keep all my matrix code such that it uses tensor access patterns.

The cases that are not "converted" may also need to undergo "conversion". I don't think they have been well vetted by the functional_comparison tests.

jamiebramwell:
@samuelpmishLLNL @samuelpmishLLNL , any open issues in this PR?

samuelpmishLLNL:
I think it looks good, approving now

samuelpmishLLNL (approved):

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,297 @@
...
+namespace detail {
+/**
+ * We'd like to access the protected offset, indices, and gatherMap variables from mfem::ElementRestriction
+ * to compute our sparsity patterns in a manner consistent with MFEM.
+ * In order to do that, we use the following template tricks.
+ * Here's an article that explains the trick being used: https://accu.org/journals/overload/28/156/harrison_2776/
+ */
+template <auto offsetV, auto indicesV, auto gatherV>
+struct forbidden_restriction {
+  friend mfem::Array<int>& ElementRestriction_offsets(mfem::ElementRestriction& from) { return from.*offsetV; }
```
joshessman-llnl: Minor: we normally use PascalCase for struct names (with the exception being type traits) and camelCase for function names - for the methods here they're referring to the MFEM type so maybe `ElementRestrictionOffsets` etc?

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,297 @@
...
+template struct forbidden_restriction<&mfem::ElementRestriction::offsets, &mfem::ElementRestriction::indices,
+                                      &mfem::ElementRestriction::gatherMap>;
+/**
+ * @brief Account for HCURL-oriented index
+ * @param[in] index Index-value. Can be negative for HCURL
+ * @return corresponding valid index value in array (must be positive)
+ */
+
+template <typename T>
+T oriented_index(T index)
```
joshessman-llnl: Minor: naming, `orientedIndex`

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,297 @@
...
+template <typename T>
+T sign(T index)
+{
+  return index >= 0 ? 1 : -1;
+}
+
+/**
+ * @brief Adds HCURL orientation encoding to index
+ * @param[in] index positive index value
+ * @param[in] orientation Positive orientation or negative
```
joshessman-llnl: Looks like the parameter needs to be renamed here?

src/serac/numerics/assembled_sparse_matrix.cpp
```diff
@@ -0,0 +1,297 @@
...
+ */
+template <typename T>
+T orient_index(T index, bool pos)
```
joshessman-llnl: Naming, also, it isn't immediately obvious to me what the difference between `orientedIndex` and `orientIndex` would be, is there a different set of names or further clarification that could be provided?

src/serac/numerics/assembled_sparse_matrix.hpp
```diff
@@ -0,0 +1,76 @@
...
+   * @return Size of ea_map
+   */
+  auto GetElementDataSize() { return ea_map_.Size(); }
+
+  /**
+   * @brief Assembles a new HypreParMatrix
+   * @returns a new HypreParMatrix
+   */
+
+  auto ParallelAssemble()
```
joshessman-llnl:
```suggestion
   */
  auto ParallelAssemble()
```

src/serac/numerics/assembled_sparse_matrix.hpp
```diff
@@ -0,0 +1,76 @@
...
+  {
+    auto hypre_A =
+        std::make_unique<mfem::HypreParMatrix>(trial_fes_.GetComm(), test_fes_.GlobalVSize(), trial_fes_.GlobalVSize(),
+                                               test_fes_.GetDofOffsets(), trial_fes_.GetDofOffsets(), this);
+
+    return RAP(test_fes_.Dof_TrueDof_Matrix(), hypre_A.release(), trial_fes_.Dof_TrueDof_Matrix());
+  }
+
+protected:
+  // Test space describing the sparsity pattern
```
joshessman-llnl: Should these member variables and private functions have Doxygen-style annotations instead of just regular comments?

src/serac/physics/utilities/functional/tensor.hpp
```diff
@@ -1471,4 +1490,44 @@ auto chain_rule(const tensor<double, m, n, p...>& df_dx, const tensor<double, p.
   return total;
 }
 
+/**
+ * @brief Recast the shape of a tensor <m,n>
+ */
+template <int m, int n, typename T>
+auto convert_to_tensor_with_shape(tensor<T, m, n> A)
```
joshessman-llnl:
```suggestion
auto convert_to_tensor_with_shape(const tensor<T, m, n>& A)
```

src/serac/physics/utilities/functional/tensor.hpp
```diff
@@ -1471,4 +1490,44 @@ auto chain_rule(const tensor<double, m, n, p...>& df_dx, const tensor<double, p.
...
+  tensor<T, m, n> A;
+  A[0][0] = value;
+  return A;
+}
+
+/**
+ * @brief Recast the shape of a tensor<m, n, o>
+ */
+template <int m, int n, int o, typename T>
+auto convert_to_tensor_with_shape(tensor<T, m, n, o> A)
```
joshessman-llnl:
```suggestion
auto convert_to_tensor_with_shape(const tensor<T, m, n, o>& A)
```

joshessman-llnl (approved):
Just some minor style comments; otherwise, looks good to me!

PR #492, opened 2021-06-15
`Functional`: add support for surface integrals

samuelpmishLLNL:
This PR adds a new member function to the `Functional` class template called `AddBoundaryIntegral()`. Its interface is similar to `AddDomainIntegral` (i.e. the user provides a "q-function" to be evaluated at each quadrature point, and a domain over which to apply it), except it targets the boundary elements of a mesh. One example of using this function is:

```cpp
  Functional<test_space(trial_space)> residual(&fespace, &fespace);

  residual.AddBoundaryIntegral(
      Dimension<dim - 1>{},
      [&](auto x, auto n, auto u) {
        tensor<double, dim> b{sin(x[0]), x[0] * x[1]};

        // return a boundary source term that can depend arbitrarily on
        // the spatial position, surface normal, and 
        return x[0] * x[1] + dot(b, n) + rho * u;
      },
      mesh);
```

Note:
1. The first two arguments of the q-function are `x` (the spatial position) and `n` (the unit normal) evaluated at the boundary element quadrature point.
2. The 3rd argument is the interpolated value of the finite element field.
3. The output of the q-function should be a "source" or "traction" term, which will be integrated against the basis functions of the test space.
4. Only H1 and L2 spaces are supported.
5. Boundary integrals are applied to the entire boundary.

Notes 2 - 5 are short term compromises to the originally proposed interface, in an effort to work around current limitations in mfem. In the future, we hope to also provide tangential derivative values of the finite element fields, flux terms in the return value, support for Hcurl/Hdiv, and ways to describe which part of the boundary to integrate over.


src/serac/physics/utilities/functional/detail/segment_H1.inl
```diff
@@ -24,6 +24,9 @@ struct finite_element<Geometry::Segment, H1<p, c> > {
   static constexpr int  dim        = 1;
   static constexpr int  ndof       = (p + 1);
 
+  using residual_type =
```
jamiebramwell, line 27: Where is this used?

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -38,6 +49,17 @@ auto get_boundary_dof_ordering(mfem::ParFiniteElementSpace* pfes)
...
 #endif
 
+template <typename space>
+const mfem::Operator* GetFaceRestriction(mfem::ParFiniteElementSpace* pfes)
+{
+  if (!is_hcurl<space>::value) {
+    return pfes->GetFaceRestriction(mfem::ElementDofOrdering::LEXICOGRAPHIC, mfem::FaceType::Boundary,
+                                    mfem::L2FaceValues::SingleValued);
+  } else {
+    return nullptr;
```
jamiebramwell, line 52: A comment about why this isn't implemented yet is needed.

src/serac/physics/utilities/functional/integral_utilities.hpp
```diff
@@ -0,0 +1,233 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
```
jamiebramwell, line 1: 👍

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -135,65 +154,47 @@ class Functional<test(trial)> : public mfem::Operator {
...
 
-      const mfem::FiniteElement&   el = *test_space_->GetBE(0);
-      const mfem::IntegrationRule& ir = mfem::IntRules.Get(el.GetGeomType(), el.GetOrder() * 2);
-      constexpr auto flags            = mfem::FaceGeometricFactors::COORDINATES | mfem::FaceGeometricFactors::JACOBIANS;
+  template <int dim, typename lambda>
+  void AddBoundaryIntegral(Dimension<dim>, lambda&& integrand, mfem::Mesh& domain)
+  {
+    // TODO: fix mfem::FaceGeometricFactors
+    auto num_boundary_elements = domain.GetNBE();
+    SLIC_ERROR_IF(num_boundary_elements == 0, "Mesh has no boundary elements");
```
jamiebramwell: These should be `ROOT` errors.

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -5,215 +5,40 @@
...
-    // from parent element to physical element
-    auto value = dot(u, dot(element_type::shape_functions(xi), inv(J)));
-    auto curl  = dot(u, element_type::shape_function_curl(xi) / det(J));
-    if constexpr (dim == 3) {
-      curl = dot(curl, transpose(J));
-    }
-    return std::tuple{value, curl};
-  }
-}
+namespace boundary_integral {
```
jamiebramwell, line 27: Is there a reason to put this here instead of `detail`?

src/serac/physics/utilities/functional/detail/segment_H1.inl
samuelpmishLLNL: `residual_type` is part of the `finite_element` concept. It tells the finite element kernel templates what the type of the element residual vector should be (dictated by the elements from the test space). Using an abridged version of one of the kernels in boundary_integral.hpp as an example


```cpp
  using element_residual_type = typename trial_element::residual_type;

  // for each element in the domain
  for (int e = 0; e < num_elements; e++) {
   
    // this is where we will accumulate the element residual tensor
    element_residual_type r_elem{};

    // for each quadrature point in the element
    for (int q = 0; q < static_cast<int>(rule.size()); q++) {
      ...
      r_elem += residual_contribution_from_this_quadrature_point;
      ...
    }
    detail::Add(r, r_elem, e);
  }
```

It only got added in this PR, because the segment elements were previously unused, so the error went unnoticed.

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: My intention was to put the things shared by `integral.hpp` (which should probably be renamed to `domain_integral.hpp`) and `boundary_integral.hpp` in `integral_utilities.hpp`, and put the things that are different in their individual files, behind a unique namespace. If they're all in the `detail` namespace, some of the `domain_integral`-specific definitions might conflict with the `boundary_integral`-specific ones. Right now, there should not a problem because those functions (Preprocess, Postprocess, etc) actually have different function signatures (so they could just be an overload set), as a result of our current inability to get tangential derivative info from mfem.

jamiebramwell: Yeah, I'm ok with either the overload set or namespace solution as long as the domain and boundary integral functions have the same layout.

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: what does that mean?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -295,22 +127,22 @@ auto Measure(const tensor<double, m, n>& A)
...
-template <Geometry g, typename test, typename trial, int geometry_dim, int spatial_dim, int Q,
-          typename derivatives_type, typename lambda>
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type, typename lambda>
 void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type* derivatives_ptr,
                        const mfem::Vector& J_, const mfem::Vector& X_, int num_elements, lambda qf)
 {
   using test_element               = finite_element<g, test>;
   using trial_element              = finite_element<g, trial>;
   using element_residual_type      = typename trial_element::residual_type;
+  static constexpr int  dim        = dimension_of(g);
```
jonwong12, line 147: Nice! `dimension_of` seems useful.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -18,25 +18,39 @@
...
-#ifdef ENABLE_BOUNDARY_INTEGRALS
-// for some reason, mfem doesn't support lexicographic ordering for Nedelec segment elements,
-// so we have to specifically detect this case, and use a different ordering (?)
-auto get_boundary_dof_ordering(mfem::ParFiniteElementSpace* pfes)
+
+/**
+ * @brief a type trait used to identify the Hcurl family
+ */
+template <typename T>
+struct is_hcurl {
```
jonwong12, line 32: This should be quite handy too.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -135,65 +146,47 @@ class Functional<test(trial)> : public mfem::Operator {
...
 
-      const mfem::FiniteElement&   el = *test_space_->GetBE(0);
-      const mfem::IntegrationRule& ir = mfem::IntRules.Get(el.GetGeomType(), el.GetOrder() * 2);
-      constexpr auto flags            = mfem::FaceGeometricFactors::COORDINATES | mfem::FaceGeometricFactors::JACOBIANS;
+  template <int dim, typename lambda>
+  void AddBoundaryIntegral(Dimension<dim>, lambda&& integrand, mfem::Mesh& domain)
+  {
+    // TODO: fix mfem::FaceGeometricFactors
+    auto num_boundary_elements = domain.GetNBE();
+    SLIC_ERROR_IF(num_boundary_elements == 0, "Mesh has no boundary elements");
```
jonwong12: Is it possible for a rank-local mesh to have no boundary elements (an interior partition) and have this "error" trigger even though it's valid?

src/serac/physics/utilities/functional/functional.hpp
jamiebramwell: It should be `SLIC_ERROR_ROOT_IF` instead of `SLIC_ERROR_IF`. The `ROOT` means it will only display on processor 0. This can really help clean up logs for large scale runs, It's described in the docs [here](https://serac.readthedocs.io/en/latest/sphinx/dev_guide/logging.html).

samuelpmishLLNL:
```suggestion
    SLIC_ERROR_ROOT_IF(num_boundary_elements == 0, "Mesh has no boundary elements");
```

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: Good question, I have no idea! It's not clear to me if `mfem::ParMesh::GetNBE()` returns the number of boundary elements on the local mesh, or the entire mesh (in which case it wouldn't be a problem).

joshessman-llnl: My understanding is that it will return the information in the local mesh - I don't think any of the `Get*` methods implemented in `mfem::Mesh` are reimplemented in `ParMesh` so they'll all just return info on the local mesh.

jonwong12: Some of the parallel "Get" methods in the API have the word `Global` in them.

samuelpmishLLNL: Oops, I just asked Jamie the same question, and got the same answer! Thanks, I'll reimplement this to just be an early exit condition if there are no boundary elements.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -420,12 +395,13 @@ class Functional<test(trial)> : public mfem::Operator {
         }
       }
 
+      output_L_boundary_ = 0.0;
+
       // scatter-add to compute residuals on the local processor
       G_test_boundary_->MultTranspose(output_E_boundary_, output_L_boundary_);
```
jamiebramwell, line 406: After poking around, I believe this does not work in parallel due to this operator application. It looks like MFEM is not treating shared boundaries in the DOF numbering appropriately, causing the boundary contributions to be added at the wrong index.

jamiebramwell: As this could possibly be an MFEM bug-hunt rabbit hole, I propose we mark this as a TODO for now and merge it in as-is so we can continue functional development in parallel.

jamiebramwell (approved):

jamiebramwell:
After way too much MFEM excavation, I finally figured out the bug. The face restriction operators were relying on mesh data that was not initialized by default. I will push a fix to mfem to ensure this doesn't happen again.

jamiebramwell:
Resolves https://github.com/LLNL/serac/issues/508

jamiebramwell:
@joshessman-llnl @jonwong12 This is ready for a final review.

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
```diff
@@ -12,91 +12,201 @@
 #include "axom/slic/core/SimpleLogger.hpp"
 
 #include "serac/serac_config.hpp"
-#include "serac/numerics/mesh_utils.hpp"
+//#include "serac/numerics/mesh_utils.hpp"
```
jamiebramwell: dead code.

jamiebramwell:
@jonwong12 @joshessman-llnl , any other issues with this PR?

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -322,24 +125,25 @@ void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type*
...
       // and calculate the measure of that point in physical space.
-      auto   xi  = rule.points[q];
-      auto   dxi = rule.weights[q];
-      auto   x_q = make_tensor<spatial_dim>([&](int i) { return X(q, i, e); });  // Physical coords of qpt
-      auto   J_q = make_tensor<spatial_dim, geometry_dim>([&](int i, int j) { return J(q, i, j, e); });
-      double dx  = detail::Measure(J_q) * dxi;
+      auto                  xi  = rule.points[q];
+      auto                  dxi = rule.weights[q];
+      auto                  x_q = make_tensor<dim + 1>([&](int i) { return X(q, i, e); });  // Physical coords of qpt
+      [[maybe_unused]] auto n_q =
```
joshessman-llnl: It looks like `n_q` is used unconditionally on L142 - can the `maybe_unused` be removed here?

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -536,37 +251,37 @@ class BoundaryIntegral {
   using trial_space = trial_space_t<spaces>;  ///< the trial function space
 
   /**
-   * @brief Constructs an @p Integral from a user-provided quadrature function
+   * @brief Constructs an @p BoundaryIntegral from a user-provided quadrature function
    * @tparam dim The dimension of the element (2 for quad, 3 for hex, etc)
    * @param[in] num_elements The number of elements in the mesh
    * @param[in] J The Jacobians of the element transformations at all quadrature points
    * @param[in] X The actual (not reference) coordinates of all quadrature points
+   * @param[in] N The unit normals of all quadrature points
```
joshessman-llnl: Name mismatch - `N` vs `normals`

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -5,120 +5,21 @@
...
-template <int ndof, int components>
-void Add(const mfem::DeviceTensor<3, double>& r_global, tensor<double, ndof, components> r_local, int e)
-{
-  for (int i = 0; i < ndof; i++) {
-    for (int j = 0; j < components; j++) {
-      r_global(i, j, e) += r_local[i][j];
-    }
-  }
-}
+namespace domain_integral {
```
joshessman-llnl, line 28: I kind of see `struct`s with static members as alternatives to `namespace`s - switching to the former might allow us to verify that these integral implementations are complete - i.e., `static_assert(is_integral_impl_v<serac::domain_integral>);`

Not sure if this would be useful at all (or even possible), so feel free to ignore - just a thought

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
joshessman-llnl: This can actually be replaced with `mesh_utils_base.hpp` so we don't have to copy/paste the `buildMesh` functions here

jamiebramwell:
@joshessman-llnl , could you give this a once-over to make sure I did the merge with the quadrature data stuff correctly?

jamiebramwell:
@joshessman-llnl , I think I got all of your comments (minus the static struct one). Any other issues with this PR?

joshessman-llnl (approved):

PR #494, opened 2021-06-17
Add QuadratureData class for encapsulating quadrature point data

joshessman-llnl:
For `QuadratureFunction` data I attempted to use a structure very similar to what the StateManager already uses for `GridFunction` data, with the exception being that StateManager always retains ownership of the QuadratureData objects.  This is because we use a double-buffer approach (discussed with @samuelpmishLLNL) where the memory allocated for the `QuadratureFunction` (a `double*`) is only synced to immediately before saving - otherwise, we use a `std::vector<T>` for generality - though of course we could use any generic container.


src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -241,23 +247,175 @@ class FiniteElementState {
...
-  std::string                               name_ = "";
+  detail::MaybeOwningPointer<mfem::ParGridFunction> gf_;
+  mfem::HypreParVector                              true_vec_;
+  std::string                                       name_ = "";
 };
 
+/**
+ * @brief A shim class for describing the interface of something that can be synced
+ */
+class SyncableData {
```
jamiebramwell: Am I correct in thinking this is an interface for "history" data with various views, e.g. friction data on a surface?

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -6,39 +6,41 @@
...
       // Use a raw pointer here when unallocated, lifetime will be managed by the DataCollection
-      gf_(options.alloc_gf
-              ? MaybeOwningPointer<mfem::ParGridFunction>{std::make_unique<mfem::ParGridFunction>(&retrieve(space_))}
-              : MaybeOwningPointer<mfem::ParGridFunction>{new mfem::ParGridFunction(&retrieve(space_),
-                                                                                    static_cast<double*>(nullptr))}),
-      true_vec_(&retrieve(space_)),
+      gf_(options.alloc_gf ? detail::MaybeOwningPointer<mfem::ParGridFunction>{std::make_unique<mfem::ParGridFunction>(
+                                 &detail::retrieve(space_))}
+                           : detail::MaybeOwningPointer<mfem::ParGridFunction>{new mfem::ParGridFunction(
+                                 &detail::retrieve(space_), static_cast<double*>(nullptr))}),
```
samuelpmishLLNL: These long single-line expressions are hard for me to read, is it practical to put them in a separate function?

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -241,23 +247,182 @@ class FiniteElementState {
...
+  /**
+   * @brief Per-quadrature point data, stored as array of doubles for compatibility with Sidre
+   */
+  detail::MaybeOwningPointer<mfem::QuadratureFunction> qfunc_;
+
+  std::vector<T> data_;
+  /**
+   * @brief The stride of the array
+   */
+  static constexpr int stride_ = sizeof(T) / sizeof(double);
```
samuelpmishLLNL: if `T` is `int`, wouldn't `stride_` be 0? More generally, if `sizeof(T)` isn't divisible by 8, this seems like it gives the wrong value (or I am misinterpreting the meaning of `stride`).

```suggestion
  static constexpr int stride_ = (sizeof(T) + sizeof(double) - 1) / sizeof(double);
```
https://godbolt.org/z/34KvKoao9

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -241,23 +247,182 @@ class FiniteElementState {
...
+public:
+  virtual ~SyncableData() = default;
+  virtual void sync()     = 0;
+};
+
+/**
+ * @brief Stores instances of user-defined type for each quadrature point in a mesh
+ * @tparam T The type of the per-qpt data
+ * @pre T must be default-constructible (TODO: Do we want to allow non-default constructible types?)
+ * @pre T must be trivially copyable (due to the use of memcpy for type punning)
```
samuelpmishLLNL: Thank you for precisely describing which types `T` are admissible!

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -501,7 +511,26 @@ struct lambda_argument<Hcurl<p>, 2, 2> {
...
+template <typename lambda_type, typename x_t, typename u_du_t, typename PointData, typename SFINAE = void>
+struct qf_result {
+  using type = std::invoke_result_t<lambda_type, x_t, decltype(make_dual(std::declval<u_du_t>()))>;
+};
+
+template <typename lambda_type, typename x_t, typename u_du_t, typename PointData>
+struct qf_result<lambda_type, x_t, u_du_t, PointData, std::enable_if_t<!std::is_same_v<PointData, void>>> {
+  // Expecting that qf lambdas take an lvalue reference to a state
+  using type = std::invoke_result_t<lambda_type, x_t, decltype(make_dual(std::declval<u_du_t>())),
+                                    std::add_lvalue_reference_t<PointData>>;
```
samuelpmishLLNL: 👍

src/serac/physics/utilities/finite_element_state.hpp
joshessman-llnl: You are completely correct - I think semantically what we want to do is have `stride_` be of floating-point type but that opens another can of worms - with sizes (for example) we'd have to make sure to always round up, and there's also some pointer arithmetic logic that would have to be refactored somehow

samuelpmishLLNL: another option is to use bresenham's algorithm to handle the case of when the sizes don't divide nicely. In theory, there are some optimizations to be made for when `sizeof(T) < 8`, but I'm not sure it's worth worrying about too much (in practice the user data will almost always be greater than 8 bytes).

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -241,23 +247,182 @@ class FiniteElementState {
...
+
+  /**
+   * @brief Synchronizes data from the stored vector<T> to the raw double*
+   * array used by the underlying mfem::QuadratureFunction
+   *
+   * Used for saving to a file - MFEMSidreDataCollection
+   * (and by extension mfem::DataCollection's interface) only allow for
+   * quadrature-point-specific data via mfem::QuadratureFunction, so this logic
+   * is needed to glue together a generic array of data with that class
+   */
```
samuelpmishLLNL: This is a good comment, since the name "sync()" alone doesn't convey which way the data is being copied (i.e. `T* -> double*` vs `double* -> T*`), but it might be better placed on the base class, to ensure that classes inheriting from `SyncableData` follow the same convention.

joshessman-llnl: Good suggestion, I've added a comment (containing similar info) to the `SyncableData` Doxygen annotation.

src/serac/physics/utilities/finite_element_state.hpp
joshessman-llnl: I've expanded on the description/Doxygen documentation for this class to hopefully better capture its raison d'etre.

src/serac/physics/utilities/functional/integral.hpp
```diff
@@ -339,7 +344,17 @@ void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type*
...
+        if constexpr (std::is_same_v<PointData, void>) {
+          // [[maybe_unused]] not supported in captures
+          (void)data;
+          (void)e;
+          (void)q;
+          return qf(x_q, make_dual(arg));
+        } else {
+          return qf(x_q, make_dual(arg), data(e, q));
+        }
+      }();
```
samuelpmishLLNL: let's mark this with a `TODO: refactor`, so we might add it to the list of things to revisit in an effort to clean up these kernels

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * @brief Per-quadrature point data, stored as array of doubles for compatibility with Sidre
+   */
+  detail::MaybeOwningPointer<mfem::QuadratureFunction> qfunc_;
+
+  std::vector<T> data_;
+  /**
+   * @brief The stride of the array
+   */
+  static constexpr double stride_ = sizeof(T) / static_cast<double>(sizeof(double));
```
samuelpmishLLNL: I think making `stride_` an integer type, rather than double, would be more intuitive and make for cleaner code-- it looks like every use of `stride_` in the PR is of the form:
```cpp
static_cast<some_integer_type>(std::ceil(stride_))
```
which suggests that the quantity we actually care about is the integer value of the ceiling itself.

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,197 @@
...
+  // A view into the quadrature point data
+  mfem::Vector view;
+  // Use the existing MFEM offset calculation logic instead of reimplementing it here
+  // Avoids making this code dependent on mfem::QuadratureSpace impl
+  detail::retrieve(qfunc_).GetElementValues(element_idx, q_idx, view);
+  double*    end_ptr   = view.GetData();
+  double*    start_ptr = detail::retrieve(qfunc_).GetData();
+  const auto idx       = static_cast<std::size_t>(end_ptr - start_ptr) / static_cast<std::size_t>(std::ceil(stride_));
+  return data_[static_cast<std::size_t>(idx)];
+}
```
samuelpmishLLNL: It's not clear to me how this (core) functionality of the QuadratureData class can be adapted to a GPU.

samuelpmishLLNL:
I think we need to modify the implementation a bit to ensure that this works on the GPU as well (I believe `QuadratureData<T>::operator()` is the only part of the interface that really needs to be used on a GPU). It does not need to test against the GPU `serac::Functional` kernels necessarily (since those have not been merged in yet), but it should pass a few basic tests like

```cpp
#include <iostream>                                                              
                                                                                 
template <typename T>                                                            
struct QuadratureData {                                                          
  QuadratureData(int nelem, int nquad) : stride(nquad) {                         
    cudaMallocManaged(&ptr, sizeof(T) * nelem * nquad);                          
  };                                                                             
                                                                                 
  void destroy() { cudaFree(ptr); }                                              
                                                                                 
  __host__ __device__ T & operator()(int e, int q) { return ptr[e * stride + q]; }
  T * ptr;                                                                       
  int stride;                                                                    
};                                                                               
                                                                                 
template< typename T >                                                           
__global__ void fill(QuadratureData<T> output, int num_elements, int num_quadrature_points) {
  int elem_id = threadIdx.x + blockIdx.x * blockDim.x;                           
  int quad_id = threadIdx.y;                                                     
  if (elem_id < num_elements && quad_id < num_quadrature_points) {               
    output(elem_id, quad_id) = elem_id * elem_id + quad_id;                      
  }                                                                              
}                                                                                
                                                                                 
template< typename T >                                                           
__global__ void copy(QuadratureData<T> destination, QuadratureData<T> source, int num_elements, int num_quadrature_points) {
  int elem_id = threadIdx.x + blockIdx.x * blockDim.x;                           
  int quad_id = threadIdx.y;                                                     
  if (elem_id < num_elements && quad_id < num_quadrature_points) {               
    destination(elem_id, quad_id) = source(elem_id, quad_id);                    
  }                                                                              
}                                                                                
                                                                                 
int main() {                                                                     
  constexpr int num_elements = 256;                                              
  constexpr int elements_per_block = 16;                                         
  constexpr int num_quadrature_points = 27;                                      
  QuadratureData<int> source(num_elements, num_quadrature_points);               
  QuadratureData<int> destination(num_elements, num_quadrature_points);          
                                                                                 
  dim3 blocks{elements_per_block, num_quadrature_points};                        
  dim3 grids{num_elements / elements_per_block};                                 
                                                                                 
  fill<<<grids, blocks>>>(source, num_elements, num_quadrature_points);          
  copy<<<grids, blocks>>>(destination, source, num_elements, num_quadrature_points);
  cudaDeviceSynchronize();                                                       
                                                                                 
  for (int i = 0; i < 30; i++) {                                                 
    std::cout << source.ptr[i] << " " << destination.ptr[i] << std::endl;        
  }                                                                              
                                                                                 
  source.destroy();                                                              
  destination.destroy();                                                         
}  
```

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -8,28 +8,38 @@
 
 namespace serac {
 
+namespace detail {
+
+MaybeOwningPointer<mfem::ParGridFunction> initialGridFunc(mfem::ParFiniteElementSpace* space, const bool alloc)
+{
+  if (alloc) {
```
jamiebramwell: These are both allocating memory. Maybe rename the variable `unique` or `owning`?

src/serac/physics/utilities/quadrature_data.hpp
joshessman-llnl: Good suggestion, I had initially set it as a double before actually propagating it through the code but this definitely cleans things up!

src/serac/physics/utilities/variant.hpp
```diff
@@ -0,0 +1,407 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file variant.hpp
+ *
+ * @brief This file contains the declaration of a two-element variant type
```
jamiebramwell, line 10: A comment about why this is necessary would be welcome.

src/serac/physics/utilities/functional/tests/functional_material_state_test.cpp
```diff
@@ -0,0 +1,292 @@
...
+  static void                 mutate(value_type& v)
+  {
+    v.x[0] = static_cast<char>(v.x[0] + 3);
+    v.x[1] = static_cast<char>(v.x[1] + 2);
+    v.x[2] = static_cast<char>(v.x[2] + 1);
+  }
+};
+
+using StateTypes = ::testing::Types<MultiFieldWrapper, IntWrapper, ThreeBytesWrapper>;
+TYPED_TEST_SUITE(QuadratureDataStateManagerTest, StateTypes, );
```
jamiebramwell, line 188: Why is this extra comma here?

src/serac/physics/utilities/functional/tests/functional_material_state_test.cpp
```diff
@@ -0,0 +1,292 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include <gtest/gtest.h>
```
jamiebramwell, line 7: All of these tests seem to be checking if the quadrature value at every point is updated to the same value. Would it be possible to check against a known quadrature value that varies across the mesh?

src/serac/physics/utilities/state_manager.hpp
```diff
@@ -44,6 +46,25 @@ class StateManager {
    */
   static FiniteElementState newState(FiniteElementState::Options&& options = {});
 
+  template <typename T>
+  static QuadratureData<T>& newQuadratureData(const std::string& name, const int p)
+  {
+    if (is_restart_) {
+      auto field = datacoll_->GetQField(name);
+      syncable_data_.push_back(std::make_unique<QuadratureData<T>>(*field));
+      // return {*field};
```
jamiebramwell: dead code

src/serac/physics/utilities/state_manager.hpp
```diff
@@ -90,7 +112,8 @@ class StateManager {
   /**
    * @brief Name of the Sidre DataCollection
    */
-  static std::string collection_name_;
+  static std::string                                collection_name_;
+  static std::vector<std::unique_ptr<SyncableData>> syncable_data_;
```
jamiebramwell, line 118: Needs doxygen comment.

src/serac/physics/utilities/state_manager.cpp
```diff
@@ -6,14 +6,16 @@
 
 #include "serac/physics/utilities/state_manager.hpp"
 
-#include "serac/infrastructure/logger.hpp"
-
 namespace serac {
 
+// Definition of extern variable
+QuadratureData<void> dummy_qdata;
```
jamiebramwell, line 12: What is this used for?

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,212 @@
...
+/**
+ * @brief A shim class for describing the interface of something that can be synced
+ *
+ * In practice this is used by StateManager so that the T[] stored by QuadratureData
+ * can be synced (copied) to the double[] owned by its underlying mfem::QuadratureFunction
+ * immediately prior to saving the data to disk.  That is, the interface for this class
+ * is intended to capture the T[] -> double[] action required to use QuadratureData
+ * with mfem::DataCollection's interface for quadrature point data (i.e., through mfem::QuadratureFunction)
+ */
+class SyncableData {
```
jamiebramwell, line 68: I'm confused why we need this in addition to `QuadratureData`.

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,212 @@
...
+ */
+template <typename T>
+class QuadratureData : public SyncableData {
+public:
+  /**
+   * @brief Constructs using a mesh and polynomial order
+   * @param[in] mesh The mesh for which quadrature-point data should be stored
+   * @param[in] p The polynomial order of the associated finite elements
+   */
+  QuadratureData(mfem::Mesh& mesh, const int p, const bool alloc = true);
```
jamiebramwell, line 88: This shouldn't be associated with the polynomial order of the FE space. There are times when you want a quadrature rule to be higher than what is strictly necessary for the underlying polynomial approximation. For example, analytic coefficients in your PDE.

jamiebramwell: It should be based on the quadrature rule used by `Functional`. This is hard coded right now, but we want to allow this to be user-modifiable in the future.

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+  std::vector<T> data_;
+  /**
+   * @brief The stride of the array
+   */
+  static constexpr int stride_ = detail::ceil(sizeof(T) / static_cast<double>(sizeof(double)));
+};
+
+/**
+ * @brief "Dummy" specialization, intended to be used as sentinel
```
jamiebramwell: I'm not sure what you mean by `sentinel`. Is this for error-checking?

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+// A dummy global so that lvalue references can be bound to something of type QData<void>
+// FIXME: There's probably a cleaner way to do this, it's technically a non-const global
+// but it's not really mutable because no operations are defined for it
+extern QuadratureData<void> dummy_qdata;
+
+// Hijacks the "vdim" parameter (number of doubles per qpt) to allocate the correct amount of storage
+template <typename T>
+QuadratureData<T>::QuadratureData(mfem::Mesh& mesh, const int p, const bool alloc)
+    : qspace_(std::make_unique<mfem::QuadratureSpace>(&mesh, p + 1)),
```
jamiebramwell, line 204: Ahhhh, mfem does this as well! I think this is fine for now and I'll add an issue to make more customizable quadrature spaces in the future.

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,212 @@
...
+      // When left unallocated, the allocation can happen inside the datastore
+      // Use a raw pointer here when unallocated, lifetime will be managed by the DataCollection
+      qfunc_(alloc ? detail::MaybeOwningPointer<mfem::QuadratureFunction>{std::make_unique<mfem::QuadratureFunction>(
```
jamiebramwell: I thought I remembered seeing a helper function for this...

src/serac/physics/utilities/quadrature_data.hpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+template <typename T>
+T& QuadratureData<T>::operator()(const int element_idx, const int q_idx)
+{
+  // A view into the quadrature point data
+  mfem::Vector view;
+  // Use the existing MFEM offset calculation logic instead of reimplementing it here
+  // Avoids making this code dependent on mfem::QuadratureSpace impl
+  detail::retrieve(qfunc_).GetElementValues(element_idx, q_idx, view);
```
jamiebramwell, line 225: @samuelpmishLLNL , is this consistent with the lexicographic ordering?

jamiebramwell: I guess in the end it doesn't matter if we're always accessing it in the same way. It could mess up visualization though.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -131,13 +131,16 @@ class Functional<test(trial)> : public mfem::Operator {
    * @tparam geometry_dim The dimension of the element (2 for quad, 3 for hex, etc)
    * @tparam spatial_dim The full dimension of the mesh
    * @tparam lambda the type of the integrand functor: must implement operator() with an appropriate function signature
+   * @tparam PointData The type of the data to store for each quadrature point
```
jamiebramwell: `QuadraturePointData` would be more appropriate.

jamiebramwell:
> 
> 
> For `QuadratureFunction` data I attempted to use a structure very similar to what the StateManager already uses for `GridFunction` data, with the exception being that StateManager always retains ownership of the QuadratureData objects. This is because we use a double-buffer approach (discussed with @samuelpmishLLNL) where the memory allocated for the `QuadratureFunction` (a `double*`) is only synced to immediately before saving - otherwise, we use a `std::vector<T>` for generality - though of course we could use any generic container.

Is this required? Quadrature data can really increase our overall memory footprint, so I'd like to avoid this if possible.

src/serac/physics/utilities/functional/tests/functional_material_state_test.cpp
joshessman-llnl: This was a clang-specific issue I ran into with variadic macros - apparently it's an error to leave the variadic argument unspecified so instead it's left empty

jamiebramwell: Weird! I'd add a comment just so people know it's not a typo.

joshessman-llnl:
> 
> 
> > For `QuadratureFunction` data I attempted to use a structure very similar to what the StateManager already uses for `GridFunction` data, with the exception being that StateManager always retains ownership of the QuadratureData objects. This is because we use a double-buffer approach (discussed with @samuelpmishLLNL) where the memory allocated for the `QuadratureFunction` (a `double*`) is only synced to immediately before saving - otherwise, we use a `std::vector<T>` for generality - though of course we could use any generic container.
> 
> Is this required? Quadrature data can really increase our overall memory footprint, so I'd like to avoid this if possible.

My initial approach used a single buffer (i.e., the one managed by MFEM) but in order to type-pun that memory into user-defined types a copy is required to not invoke UB.  Copying the state for each point twice (for the read/write) every single time it is accessed would likely have a significant performance impact so after some discussion with @samuelpmishLLNL I went with a different implementation.  It may be possible to only create the `mfem::QuadratureFunction` objects when they are needed (when saving) so although it wouldn't affect peak memory use it could still significantly reduce the average usage.

jamiebramwell:
> 
> 
> > > For `QuadratureFunction` data I attempted to use a structure very similar to what the StateManager already uses for `GridFunction` data, with the exception being that StateManager always retains ownership of the QuadratureData objects. This is because we use a double-buffer approach (discussed with @samuelpmishLLNL) where the memory allocated for the `QuadratureFunction` (a `double*`) is only synced to immediately before saving - otherwise, we use a `std::vector<T>` for generality - though of course we could use any generic container.
> > 
> > 
> > Is this required? Quadrature data can really increase our overall memory footprint, so I'd like to avoid this if possible.
> 
> My initial approach used a single buffer (i.e., the one managed by MFEM) but in order to type-pun that memory into user-defined types a copy is required to not invoke UB. Copying the state for each point twice (for the read/write) every single time it is accessed would likely have a significant performance impact so after some discussion with @samuelpmishLLNL I went with a different implementation. It may be possible to only create the `mfem::QuadratureFunction` objects when they are needed (when saving) so although it wouldn't affect peak memory use it could still significantly reduce the average usage.

Fair enough! I would add a comment stating this and we can revisit it if it becomes an issue.

src/serac/physics/utilities/quadrature_data.hpp
samuelpmishLLNL: I've never explicitly specified an ordering on the quadrature points in mfem. Empirically, they seem to be ordered lexicographically (so that's what I do on our quadrature rules too), I don't know if there are other options or not.

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: `QuadraturePointData` sounds like a good name for the parameter itself, but I think this line of code is documenting the name of the type itself. `data_type`, `data_t`, or just `T` might be good alternatives. We seem to be mixing snake_case (e.g. `geometry_dim`) and CamelCase (e.g. PointData) here.

src/serac/physics/utilities/state_manager.cpp
samuelpmishLLNL: I think this exists to allow for a default argument to `Functional::AddIntegral()`, since the `QuadratureData` is passed by (non-const) reference

src/serac/physics/utilities/quadrature_data.hpp
joshessman-llnl: Added a further comment here

https://github.com/LLNL/serac/blob/22cfc65c9cbce51e25010c8fdebd0f12c3ea10f8/src/serac/physics/utilities/quadrature_data.hpp#L37-L42

src/serac/physics/utilities/quadrature_data.hpp
joshessman-llnl: Added some extra documentation here

src/serac/physics/utilities/quadrature_data.hpp
joshessman-llnl: This will be addressed in a follow-up - tagging as #511 for now

src/serac/physics/utilities/functional/tests/functional_material_state_test.cpp
joshessman-llnl: I imagine you're wanting to make sure that the restart process preserves the order of the data - would it maybe be sufficient to set each point to something simple like the distance from the origin (since I think we have that info within the QFunc)?

src/serac/physics/utilities/functional/functional.hpp
joshessman-llnl: `qpt_data_type`?

jamiebramwell (approved):
Really awesome stuff!

jamiebramwell:
@samuelpmishLLNL @white238 , any other issues with this PR?

samuelpmishLLNL (approved):

scripts/llnl/common_build_functions.py
```diff
@@ -272,7 +272,7 @@ def build_and_test_host_config(test_root,host_config, report_to_stdout = False,
     bld_output_file =  pjoin(build_dir,"output.log.make.txt")
     print("[starting build]")
     print("[log file: %s]" % bld_output_file)
-    res = sexe("cd %s && make -j 16 VERBOSE=1 " % build_dir,
+    res = sexe("cd %s && make -j 8 VERBOSE=1 " % build_dir,
```
white238, line 275: what was the reason for this?

joshessman-llnl: This was what it took for the GCC image to stop imploding

PR #497, opened 2021-06-23
StateManager

white238:
This PR does two things:

* Breaks `StateManager` into it's own file
* Breaks a circular dependency between `infrastructure` and `physics/utilities` I mistakenly introduced

### DIFF
--- a/src/serac/infrastructure/output.cpp
+++ b/src/serac/infrastructure/output.cpp
@@ -9,7 +9,6 @@
 #include "serac/infrastructure/initialize.hpp"
 #include "serac/infrastructure/logger.hpp"
 #include "serac/infrastructure/terminator.hpp"
-#include "serac/physics/utilities/finite_element_state.hpp"
 
 #include "conduit/conduit.hpp"
 #include "ascent/ascent.hpp"
@@ -19,8 +18,8 @@
 
 namespace serac::output {
 
-void outputFields(const axom::sidre::DataStore& datastore, const std::string& file_name_prefix, double time,
-                  const Language language)
+void outputFields(const axom::sidre::DataStore& datastore, const std::string& data_collection_name,
+                  const std::string& file_name_prefix, double time, const Language language)
 {
   SLIC_INFO_ROOT(fmt::format("Outputting field data at time: {}", time));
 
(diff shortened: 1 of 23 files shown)


src/serac/infrastructure/output.hpp
```diff
@@ -38,10 +38,11 @@ enum class Language
...
  * @param[in] datastore Root of the Sidre datastore
  * @param[in] file_name_prefix Prefix used for generating output file name
+ * @param[in] data_collection_name Name of the Data Collection stored in Sidre
  * @param[in] time Current simulation time
  * @param[in] language The output language format
  */
-void outputFields(const axom::sidre::DataStore& datastore, const std::string& file_name_prefix, double time,
-                  const Language language = Language::JSON);
+void outputFields(const axom::sidre::DataStore& datastore, const std::string& data_collection_name,
+                  const std::string& file_name_prefix, double time, const Language language = Language::JSON);
```
joshessman-llnl: 5 params makes me a tiny bit nervous, especially the string parameters that are next to each other.  If we ever need to add further options possibly we could wrap all the parameters after the first one into a struct

src/serac/infrastructure/output.cpp
```diff
@@ -50,12 +49,10 @@ void outputFields(const axom::sidre::DataStore& datastore, const std::string& fi
...
   // Get domain Sidre group
-  // TODO: get this from StateManager directly?
-  const axom::sidre::Group* sidre_root      = datastore.getRoot();
-  const std::string         collection_name = StateManager::collectionName();
-  SLIC_ERROR_ROOT_IF(!sidre_root->hasGroup(collection_name),
-                     fmt::format("Expected a datacollection root at '{0}' but it was not found", collection_name));
-  const axom::sidre::Group* domain_grp = sidre_root->getGroup(collection_name);
+  const axom::sidre::Group* sidre_root = datastore.getRoot();
+  SLIC_ERROR_ROOT_IF(!sidre_root->hasGroup(data_collection_name),
+                     fmt::format("Expected a datacollection root at '{0}' but it was not found", data_collection_name));
```
jamiebramwell: Maybe adding `in sidre` or something to know what we are looking at?

joshessman-llnl (approved):

src/serac/physics/utilities/state_manager.cpp
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
```
jamiebramwell, line 1: Did you change anything in this other than moving the file?

src/serac/physics/utilities/state_manager.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * @file state_manager.hpp
+ *
+ * @brief This file contains the declaration of the StateManager class
+ */
+
+#pragma once
+
+//#include <functional>
```
jamiebramwell: Dead code.

src/serac/infrastructure/output.cpp
white238: No, the only functional change was the breaking the circular dependancy.  This happened in output.cpp/hpp and in serac.cpp source files

src/serac/physics/utilities/state_manager.cpp
white238: No, the only functional change was the breaking the circular dependancy. This happened in output.cpp/hpp and in serac.cpp source files

jamiebramwell (approved):

src/serac/infrastructure/output.hpp
white238: I have an idea.. change incoming.

joshessman-llnl (approved):

PR #501, opened 2021-06-24
Add adjoint solve options to the solid module

jamiebramwell:
This adds adjoint solve functionality to the solid mechanics module. It still needs a test, but I wanted to see if there is any feedback on the simple API.

- [x] Add tests
- [x] Add essential BC options
- [x] Figure out why the essential BCs aren't being set


jamiebramwell:
ping @pmm14006

src/serac/physics/solid.hpp
```diff
@@ -270,6 +270,14 @@ class Solid : public BasePhysics {
   void setMaterialParameters(std::unique_ptr<mfem::Coefficient>&& mu, std::unique_ptr<mfem::Coefficient>&& K,
                              bool material_nonlin = true);
 
+  /**
+   * @brief Set the adjoint load
+   * @note The ParLinearForm must be built on the same finite element space as the displacement
+   *
+   * @param adjoint_load The adjoint load to apply
+   */
+  void setAdjointLoad(mfem::ParLinearForm& adjoint_load);
```
jandrej: Do you think it is better to use Form instead of a GridFunction for the adjoint load?

IMHO: The adjoint load comes from another object, probably only knowing about the proper FE space. Meaning, the user (or accompanying code) has to en capsule the GridFunction it computes into a proper Form before passing it to this function.

jamiebramwell: This was the result of talking with @goxberry and the LiDO team. I believe the main point is they want to be able to differentiate between primal and dual variables.

jamiebramwell: ping @sethwatts @pmm14006

sethwatts: Yes, this is an area we're trying to be fastidious about in current LiDO. Also, from a practical standpoint, the adjoint load is going to come from integrating derivatives over the domain and/or boundaries, so it's already in a linear form format rather than a GridFunction.

sethwatts:
If I'm understanding correctly, this PR adds methods to provide the adjoint load and solve the adjoint problem, and to access the adjoint solution (adjoint displacements in this context). That all looks good. However, do you plan to include the terms for sensitivities of the residual w.r.t. the material properties (what Dan calls the pseudoloads) so that we can output the combined term from the Solid operator?

jamiebramwell:
> 
> 
> If I'm understanding correctly, this PR adds methods to provide the adjoint load and solve the adjoint problem, and to access the adjoint solution (adjoint displacements in this context). That all looks good. However, do you plan to include the terms for sensitivities of the residual w.r.t. the material properties (what Dan calls the pseudoloads) so that we can output the combined term from the Solid operator?

Yes, that's the next step and will be a follow up PR.

jamiebramwell:
I believe this is ready to go now. I will add the residual sensitivities in a follow-on PR.

src/serac/physics/solid.hpp
```diff
@@ -326,6 +338,20 @@ class Solid : public BasePhysics {
...
+  /**
+   * @brief Solve the adjoint problem
+   * @note It is expected that the forward analysis is complete and the current displacement state is valid
+   * @note If the essential boundary state is not specified, homogeneous essential boundary conditions are applied
+   *
+   * @param[in] adjoint_load_form The linear form that when assembled contains the right hand side of the adjoint system
+   * @param[in] state_with_essential_boundary A optional finite element state containing the non-homogenous essential
+   * boundary condition data for the adjoint problem
+   * @return The computed adjoint finite element state
+   *
```
white238: nitpick: line 350 can be removed

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -105,6 +105,14 @@ class FiniteElementState {
    */
   FiniteElementState(mfem::ParMesh& mesh, mfem::ParGridFunction& gf, const std::string& name = "");
 
+  /**
+   * @brief Minimal constructor for a FiniteElementState given an already-existing state
+   * @param[in] mesh The problem mesh (object does not take ownership)
+   * @param[in] fe_state The state for the new state to copy
```
white238, line 111: You have some mismatching parameter names here:

* doxygen comment: `fe_state`
* declaration: `gf`
* implementation: `original_state`

goxberry:
Insofar as I understand what's going on, it looks good to me.

Is there a reference you'd suggest for the solid mechanics in the theory guide? Holzapfel, maybe? I have no solid mechanics background, so I have to look up some of the terms from time to time.

goxberry (approved):
LGTM, as far as I understand it.

jamiebramwell:
[Nonlinear Finite Elements for Continua and Structures](https://www.wiley.com/en-us/Nonlinear+Finite+Elements+for+Continua+and+Structures%2C+2nd+Edition-p-9781118632703) is my go to reference for practical implementations. However, I might be biased because this is the book I originally learned this from.

goxberry:
> [Nonlinear Finite Elements for Continua and Structures](https://www.wiley.com/en-us/Nonlinear+Finite+Elements+for+Continua+and+Structures%2C+2nd+Edition-p-9781118632703) is my go to reference for practical implementations. However, I might be biased because this is the book I originally learned this from.

Thanks! Oddly enough, I may have the first edition of that book; a civil engineering friend was moving and gave me a copy, but I've never gotten around to reading it.

src/serac/physics/solid.hpp
```diff
@@ -326,6 +338,19 @@ class Solid : public BasePhysics {
...
+   * @note It is expected that the forward analysis is complete and the current displacement state is valid
+   * @note If the essential boundary state is not specified, homogeneous essential boundary conditions are applied
+   *
+   * @param[in] adjoint_load_form The linear form that when assembled contains the right hand side of the adjoint system
+   * @param[in] state_with_essential_boundary A optional finite element state containing the non-homogenous essential
+   * boundary condition data for the adjoint problem
+   * @return The computed adjoint finite element state
+   */
+  virtual const serac::FiniteElementState& solveAdjoint(mfem::ParLinearForm& adjoint_load_form,
+                                                        FiniteElementState*  state_with_essential_boundary = nullptr);
```
jonwong12, line 352: I forgot if we've had arrived at a convention whether `state_with_essential_boundary` should be `std::optional` or not? There are definitely some pros and cons.

src/serac/physics/solid.hpp
```diff
@@ -326,6 +338,19 @@ class Solid : public BasePhysics {
...
+   * @brief Solve the adjoint problem
+   * @note It is expected that the forward analysis is complete and the current displacement state is valid
+   * @note If the essential boundary state is not specified, homogeneous essential boundary conditions are applied
+   *
+   * @param[in] adjoint_load_form The linear form that when assembled contains the right hand side of the adjoint system
+   * @param[in] state_with_essential_boundary A optional finite element state containing the non-homogenous essential
+   * boundary condition data for the adjoint problem
+   * @return The computed adjoint finite element state
+   */
+  virtual const serac::FiniteElementState& solveAdjoint(mfem::ParLinearForm& adjoint_load_form,
```
jonwong12, line 351: Is the `adjoint_load_form` guaranteed to always be `LinearForm` or would it be more versatile possibly as a `NonlinearForm` and then use serac wrappers?

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -33,6 +33,18 @@ FiniteElementState::FiniteElementState(mfem::ParMesh& mesh, mfem::ParGridFunctio
...
 
+FiniteElementState::FiniteElementState(mfem::ParMesh& mesh, FiniteElementState& fe_state, const std::string& name)
+    : FiniteElementState(mesh, fe_state.gridFunc(), name)
+{
+}
+
+FiniteElementState& FiniteElementState::operator=(const double value)
+{
+  true_vec_ = value;
+  distributeSharedDofs();
```
jonwong12, line 60: Does the prolongation operator here zero out the non-truedofs() or does it leave them as is? If `value` is not the same on all ranks this operation might be a little tricky to interpret?

jonwong12 (changes requested):
Overall it looks good. Just wondering about a few conventions and some of the parallel operations and whether we might want to use nonlinearform instead of linearform.

src/serac/physics/solid.hpp
sethwatts: I believe the adjoint load will always be a linear form. However, I think the plan (perhaps as soon as the next PR?) is to change the interface to expect a more standard serac type, either a `Functional` or a `HypreParVec`. We used a `ParLinearForm` here since that is what LiDO is using currently. We plan to start wrapping and using serac types soon, though.

src/serac/physics/solid.hpp
jamiebramwell: The problem is there is no such thing as an optional reference and I would like to avoid copying if possible. Is there a preferred solution for this common problem @joshessman-llnl ?

src/serac/physics/solid.hpp
jamiebramwell: I also believe adjoint loads will always be linear forms since they are derived from the explicit derivative of the QoI with respect to the state variable. That said, the plan as @sethwatts states is to move this to a more standard `Functional` or `Vector` type.

src/serac/physics/utilities/finite_element_state.cpp
jamiebramwell: The prolongation operator picks up the correct true value from the owning process (it is MPI-aware).

jamiebramwell:
Fixed the boundary condition bug, so I believe this is ready to go now.

jamiebramwell:
@white238 , @jandrej , @jonwong12 , any other issues with this PR?

jamiebramwell:
@white238 @jonwong12 Any other issues with this PR? It's needed for the initial LiDO integration. I expect the interfaces to change a bit as that gets ironed out.

src/serac/physics/solid.cpp
```diff
@@ -28,13 +28,16 @@ Solid::Solid(int order, const SolverOptions& options, GeometricNonlinearities ge
           .order = order, .vector_dim = mesh_.Dimension(), .name = detail::addPrefix(name, "velocity")})),
       displacement_(StateManager::newState(FiniteElementState::Options{
           .order = order, .vector_dim = mesh_.Dimension(), .name = detail::addPrefix(name, "displacement")})),
+      adjoint_(StateManager::newState(FiniteElementState::Options{
+          .order = order, .vector_dim = mesh_.Dimension(), .name = detail::addPrefix(name, "adjoint_displacement")})),
```
white238, line 32: Should `adjoint_displacement` vs `adjoint`  match up?

white238 (approved):

src/serac/physics/solid.hpp
jonwong12: I think `Functional` has notions of being able to get the gradient while a plain old `Vector` does not.  I was mainly wonder if that notion was important (similar distinction between `LinearForm` and `NonLinearForm`.

jonwong12 (approved):

src/serac/physics/utilities/finite_element_state.cpp
jonwong12: I was concerned that it might be confusing for a user if they set the `FiniteElementState` to a particular value (that is not the same on all ranks), and the owning process sets it to a different value. Maybe it might be worth caveating that `operator=` mainly applies to the true d.o.fs.

jamiebramwell: I think semantically this is the right thing to do, as the "owning" process should set the DOF on non-owning processes. I'll add a not in the doxygen as you suggested.

PR #503, opened 2021-06-25
Add a norm convenience function

jamiebramwell:
This adds an Lp norm convenience function to the finite element state class.


jandrej (approved):

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -196,6 +196,14 @@ class FiniteElementState {
     return std::make_unique<Tensor>(&retrieve(space_));
   }
 
+  /**
+   * @brief Calculate the Lp norm of the underlying state grid function
+   *
+   * @param p Order of the norm
+   * @return The norm value
+   */
+  double norm(double p = 2) const;
```
samuelpmishLLNL: When practical, I prefer free functions and operator overloads over member functions for basic mathematical operations, since it is closer to the original mathematical notation (although mfem has mostly uses member functions). For example,
```cpp
double x_norm = norm(solid_solver.displacement());
```
is closer to ||u|| than 
```cpp
double x_norm = solid_solver.displacement().norm();
```
Making everything into member functions is more of a Java convention, and tends to hurt readability (in my opinion):
```java
double result = a.multiply(x.pow(2)).plus(b.multiply(x)).plus(c);
```
vs.
```cpp
double result = a * x * x + b * x + c;
```

This mostly just aesthetics, though. Does anyone else have a preference here?

jamiebramwell: I agree, this is nicer. I'll go fix it up.

tests/serac_solid_reuse_mesh.cpp
```diff
@@ -129,12 +125,12 @@ TEST(solid_solver, reuse_mesh)
   double dt = 1.0;
   solid_solver_3.advanceTimestep(dt);
 
-  double u_norm_3 = solid_solver_3.displacement().gridFunc().ComputeLpError(2.0, zerovec);
+  double u_norm_3 = solid_solver_3.displacement().norm();
   EXPECT_NEAR(0.0, u_norm_1 - u_norm_3, 0.001);
```
samuelpmishLLNL: we should probably be checking ||u_i - u_j|| rather than ||u_i|| - ||u_j||  in these tests

tests/serac_solid_reuse_mesh.cpp
```diff
@@ -129,12 +125,12 @@ TEST(solid_solver, reuse_mesh)
...
 
   solid_solver_3.resetToReferenceConfiguration();
-  EXPECT_NEAR(0.0, solid_solver_3.displacement().gridFunc().ComputeLpError(2.0, zerovec), 1.0e-8);
-  EXPECT_NEAR(0.0, solid_solver_3.velocity().gridFunc().ComputeLpError(2.0, zerovec), 1.0e-8);
+  EXPECT_NEAR(0.0, solid_solver_3.displacement().norm(), 1.0e-8);
+  EXPECT_NEAR(0.0, solid_solver_3.velocity().norm(), 1.0e-8);
```
samuelpmishLLNL: :+1: 
I prefer this inline implementation of the EXPECT_NEAR statement to some of the other uses we have that do something like
```cpp
// tmp's only purpose is to be put inside EXPECT_NEAR on the next statement,
// so it probably doesn't need to be an lvalue
double tmp = some_calculation(); 
EXPECT_NEAR(expected_value, tmp, tolerance);
```

tests/serac_thermal_structural_solver.cpp
```diff
@@ -133,14 +133,9 @@ TEST(dynamic_solver, dyn_solve)
...
-  mfem::Vector zero(dim);
-  zero = 0.0;
-  mfem::VectorConstantCoefficient zerovec(zero);
-
-  double v_norm    = ts_solver.velocity().gridFunc().ComputeLpError(2.0, zerovec);
-  double x_norm    = ts_solver.displacement().gridFunc().ComputeLpError(2.0, zerovec);
-  double temp_norm = ts_solver.temperature().gridFunc().ComputeLpError(2.0, zerovec);
+  double v_norm    = ts_solver.velocity().norm();
+  double x_norm    = ts_solver.displacement().norm();
+  double temp_norm = ts_solver.temperature().norm();
```
samuelpmishLLNL: see https://github.com/LLNL/serac/pull/503/files#r660788654

tests/test_utilities.cpp
```diff
@@ -111,37 +111,31 @@ std::string moduleName<ThermalConduction>()
...
 template <>
-void verifyFields(const ThermalConduction& phys_module, const axom::inlet::Inlet& inlet, const int)
+void verifyFields(const ThermalConduction& phys_module, const axom::inlet::Inlet& inlet)
 {
-  mfem::ConstantCoefficient zero(0.0);
   if (inlet.contains("expected_t_l2norm")) {
-    double t_norm = phys_module.temperature().gridFunc().ComputeLpError(2.0, zero);
+    double t_norm = phys_module.temperature().norm();
     EXPECT_NEAR(inlet["expected_t_l2norm"], t_norm, inlet["epsilon"]);
   }
```
samuelpmishLLNL: could this be equivalently implemented as 
```cpp
template <typename PhysicsModule>
void verifyFields(const PhysicsModule&, const axom::inlet::Inlet&) {
  if (inlet.contains("expected_u_l2norm")) {
    EXPECT_NEAR(inlet["expected_u_l2norm"], phys_module.displacement().norm(), inlet["epsilon"]);
  }
  if (inlet.contains("expected_v_l2norm")) {
    EXPECT_NEAR(inlet["expected_v_l2norm"], phys_module.velocity().norm(), inlet["epsilon"]);
  }
  if (inlet.contains("expected_t_l2norm")) {
    EXPECT_NEAR(inlet["expected_t_l2norm"], phys_module.temperature().norm(), inlet["epsilon"]);
  }
}
```
If so, then future multiphysics solvers (e.g. ThermalSolid) would likely already work without requiring additional overloads.

On an unrelated note, I think a lowercase "t" is an unconventional symbol for temperature, is it worth renaming that field in inlet? With user-facing interfaces I prefer more verbose names, maybe
`expected_displacement_l2norm`,  `expected_velocity_l2norm`, `expected_temperature_l2norm`.

jamiebramwell: I think it's kind of nice to keep these as separate instantiations to guard against physics modules that are not supported. I agree with your second point though, so I'll open an issue.

jamiebramwell:
@samuelpmishLLNL , I think I caught all of your comments. Let me know if there's other things you want changed.

white238 (approved):

PR #507, opened 2021-06-29
Switch to new Ubuntu 20-based Docker Images

joshessman-llnl:
See https://github.com/LLNL/axom-docker/pull/9


host-configs/docker/docker-linux-ubuntu20.04-ivybridge-clang@10.0.0.cmake
```diff
@@ -27,9 +27,9 @@ else()
 
 endif()
 
-set(CMAKE_C_FLAGS "-fPIC -fPIC" CACHE STRING "")
+set(CMAKE_C_FLAGS "-fPIC -pthread -fPIC -pthread" CACHE STRING "")
```
white238, line 30: why is this being doubled up?

scripts/docker/dockerfile_clang-10
```diff
@@ -22,14 +18,15 @@ USER serac
...
 
 # New containers will use Python3
-RUN cd serac && python3 ./scripts/uberenv/uberenv.py --spack-config-dir=./scripts/spack/configs/docker/ubuntu18/ \
-                                                     --project-json=.uberenv_config.json \
-                                                     --spec=%clang@10.0.0 --prefix=/home/serac/serac_tpls -k \
-             && rm -rf /home/serac/serac_tpls/builds
+RUN cd serac && python3 ./scripts/uberenv/uberenv.py --spack-config-dir=./scripts/spack/configs/docker/ubuntu20/ \
+                                                       --project-json=.uberenv_config.json \
+                                                       --spec=%clang@10.0.0 --prefix=/home/serac/serac_tpls -k
+# white238: commented out due to logs disapearing and this being saved due to layering
```
white238, line 20: This was probably wrong in hindsight but did make the problem go away if i remember correctly

white238:
Is anyone still on ubuntu 18? there were some WSL users on it but hopefully they updated to 20

joshessman-llnl:
> 
> 
> Is anyone still on ubuntu 18? there were some WSL users on it but hopefully they updated to 20

If this is in reference to the removed configs it's just the Docker-specific ones  in `scripts/spack/configs/docker/ubuntu18` that were deleted - the `scripts/spack/configs/linux_ubuntu_18` ones are still there

host-configs/docker/docker-linux-ubuntu20.04-ivybridge-clang@10.0.0.cmake
joshessman-llnl: https://github.com/spack/spack/blob/b8512983d999575d372e7ecaefa76896593bc94e/lib/spack/spack/build_systems/cached_cmake.py#L101

I think maybe we should be using `cxxflags` instead of `cppflags` as the latter might actually be referring to flags intended for the preprocessor?

jamiebramwell:
> 
> 
> Is anyone still on ubuntu 18? there were some WSL users on it but hopefully they updated to 20

I thought I was the only one and I just updated, but I believe @jonwong12 also uses ubuntu 18.

white238 (approved):

jamiebramwell (approved):
Thank you for fixing this!!

PR #509, opened 2021-07-07
Switch GCC Docker from 8.4 -> 9.3

### LINKED ISSUE #505: Split Github Actions Docker build in two
Keep running out of space when building both at the same time.

joshessman-llnl:
We already test GCC 8 on every commit via GitLab CI, so I thought it would be useful to test the GCC 9 that (most of?) the WSL folks are using.  Also I'm convinced GCC 8 on Docker might be cursed.

### DIFF
--- a/host-configs/docker/docker-linux-ubuntu20.04-ivybridge-clang@10.0.0.cmake
+++ b/host-configs/docker/docker-linux-ubuntu20.04-ivybridge-clang@10.0.0.cmake
@@ -27,9 +27,9 @@ else()
 
 endif()
 
-set(CMAKE_C_FLAGS "-fPIC -pthread -fPIC -pthread" CACHE STRING "")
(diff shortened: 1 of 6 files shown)


white238 (approved):

white238:
Thanks @joshessman-llnl

host-configs/docker/docker-linux-ubuntu20.04-ivybridge-clang@10.0.0.cmake
```diff
@@ -27,9 +27,9 @@ else()
 
 endif()
 
-set(CMAKE_C_FLAGS "-fPIC -pthread -fPIC -pthread" CACHE STRING "")
+set(CMAKE_C_FLAGS "-fPIC -pthread" CACHE STRING "")
```
jamiebramwell, line 30: I'm glad we can get rid of this!

jamiebramwell (approved):
Thanks for looking into this!

PR #510, opened 2021-07-07
Simplify expression profiling macros

joshessman-llnl:
A more elegant solution for expression profiling that just occurred to me...


cmake/SeracMacros.cmake
```diff
@@ -42,7 +42,7 @@ macro(serac_add_code_checks)
         foreach(_source ${_all_sources})
             set(_to_be_included FALSE)
             foreach(_include ${arg_INCLUDES})
-                if (${_source} MATCHES ${_include})
+                if (${_source} MATCHES "^${PROJECT_SOURCE_DIR}${_include}")
```
white238, line 45: 👍

white238 (approved):

jonwong12 (approved):
Thanks! It's beautiful now!

PR #515, opened 2021-07-19
Test reorganization

### LINKED ISSUE #222: Organize the unit tests

jamiebramwell:
This moves our unit tests into subdirectories closer to the objects they are testing.


src/serac/coefficients/tests/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+# other Serac Project Developers. See the top-level LICENSE file for
+# details.
+#
+# SPDX-License-Identifier: (BSD-3-Clause) 
+
+if (ENABLE_GTEST)
+
+    set(test_dependencies serac_physics serac_coefficients test_utils)
```
white238: Are these true at this point? since i assume they were just copy/pasted from the original place...

src/serac/infrastructure/tests/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+# other Serac Project Developers. See the top-level LICENSE file for
+# details.
+#
+# SPDX-License-Identifier: (BSD-3-Clause) 
+
+if (ENABLE_GTEST)
+
+    set(test_dependencies serac_physics serac_coefficients test_utils)
```
white238: These are wrong

src/serac/coefficients/tests/CMakeLists.txt
jamiebramwell: I know physics and coefficients are needed. I'll check test utils.

jamiebramwell: Physics and test_utils were both needed. Coefficients has been removed.

white238 (approved):

jamiebramwell:
I'm investigating why the lassen builds are failing. It's giving me a weird casting error.

tests/CMakeLists.txt
```diff
@@ -94,27 +48,17 @@ if (ENABLE_GTEST)
         blt_add_executable( NAME        serac_cuda_smoketest
                             SOURCES     serac_cuda_smoketest.cpp 
                             OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
-                            DEPENDS_ON  serac_cuda_smoketest_kernel ${test_dependencies} cuda_runtime
+                            DEPENDS_ON  serac_cuda_smoketest_kernel gtest
```
jamiebramwell, line 51: Thanks @white238 !!!

jamiebramwell:
Any issues @samuelpmishLLNL @joshessman-llnl ?

samuelpmishLLNL:
I think this is an improvement, but it's making me realize how much boilerplate we have in all of our CMakeLists.txt. For example, this snippet is used everywhere, verbatim:

```cmake
foreach(filename ${whatever_tests})
     get_filename_component(test_name ${filename} NAME_WE)

     blt_add_executable(NAME        ${test_name}
                        SOURCES     ${filename}
                        OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
                        DEPENDS_ON  ${test_dependencies}
                        FOLDER      serac/tests )
     blt_add_test(NAME          ${test_name}
                  COMMAND       ${test_name}
                  NUM_MPI_TASKS 1 )
 endforeach()
```

it might be worth making a function to avoid copy-and-pasting that same body in every single file, e.g.

```cmake
foreach(filename ${whatever_tests})
  add_serial_test(filename FOLDER serac/tests)
endforeach()
```

Additionally, it seems like there's another pattern where some_directory/CMakeLists.txt has
```cmake
...
if (enable_testing)
  add_subdirectory(tests)
endif()
...
```
and then some_directory/tests/CMakeLists.txt has another conditional guard
```cmake
if (ENABLE_GTEST)
...
# entire contents of file
...
endif()
```

Is there a case where we ever enable testing, but don't have gtest?

src/serac/physics/tests/CMakeLists.txt
```diff
@@ -0,0 +1,57 @@
...
+
+        blt_add_executable(NAME        ${test_name}
+                           SOURCES     ${filename}
+                           OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
+                           DEPENDS_ON  ${test_dependencies}
+                           FOLDER      serac/tests )
+        blt_add_test(NAME          ${test_name}
+                     COMMAND       ${test_name}
+                     NUM_MPI_TASKS 2 )
+    endforeach()
```
samuelpmishLLNL: this is another option to avoid us having to explicitly write out the names of each file in every CMakeLists.txt
```cmake
file(GLOB tests_in_this_directory "serac_*.cpp")

foreach(filename ${tests_in_this_directory})
  get_filename_component(test_name ${filename} NAME_WE)
  
  blt_add_executable(NAME        ${test_name}
                     SOURCES     ${filename}
                     OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
                     DEPENDS_ON  ${test_dependencies}
                     FOLDER      serac/tests )
  blt_add_test(NAME          ${test_name}
               COMMAND       ${test_name}
               NUM_MPI_TASKS 2 )
endforeach()
```
it might also be nice to have an additional function `add_mpi_test()` to replace this all with
```cmake
file(GLOB tests_in_this_directory "serac_*.cpp")
foreach(filename ${tests_in_this_directory})
  add_mpi_test(${filename} NUM_MPI_TASKS 2)
endforeach()
```

but I know some developers feel strongly about never using `file(GLOB ... )`

jamiebramwell: Thoughts on this @white238 ?

jamiebramwell:
I think these quality of life improvements make a lot of sense. I'll ask @white238 to make sure there isn't a good reason to not do this.

src/serac/physics/tests/CMakeLists.txt
white238: Globing is not recommended for source files by anyone in the CMake community (even the creators) it causes CMake to lose track of source dependencies.  They semi-recently added a `CONFIGURE_DEPENDS` option to `file(GLOB ..)` but they still recommend against it due to the rebuild time cost of rerunning that command everytime.

see the note here: https://cmake.org/cmake/help/latest/command/file.html#glob

white238:
As for the guarding against `ENABLE_GTEST`, I think its safe to assume that if our project has `ENABLE_TESTING` as on we can assume `gtest` is enabled.

white238:
Lets do the testing macro in a followup PR.  We can probably just have one macro something like:

```
serac_add_test(SOURCE file
                         DEPENDS_ON list
                         COMMAND_ARGS -something-here
                         NUM_MPI_TASKS 4)
```

where NUM_MPI_TASKS is optional and triggers the serial/parallel

samuelpmishLLNL (approved):

jonwong12 (commented):
Thanks! Looks good.

jonwong12 (approved):

PR #519, opened 2021-07-22
Material parameter sensitivities

jamiebramwell:
This is a work in progress for computing sensitivities of a quantity of interest defined by an adjoint load with respect to linear elastic material parameters, specifically the shear an bulk modulus. This will be cleaned up substantially when we convert the solid mechanics module to the new `Functional` machinery.


jamiebramwell:
ping @sethwatts @goxberry @pmm14006

src/serac/physics/solid.cpp
```diff
@@ -397,13 +398,69 @@ void Solid::advanceTimestep(double& dt)
...
+  return *bulk_sensitivity_form_;
 }
 
 const FiniteElementState& Solid::solveAdjoint(mfem::ParLinearForm& adjoint_load_form,
                                               FiniteElementState*  state_with_essential_boundary)
 {
   SLIC_ERROR_ROOT_IF(!is_quasistatic_, "Adjoint analysis only vaild for quasistatic problems.");
-  SLIC_ERROR_ROOT_IF(cycle_ == 0, "Adjoint analysis only valid following a forward solve.");
+  SLIC_ERROR_ROOT_IF(previous_solve_ == PreviousSolve::None, "Adjoint analysis only valid following a forward solve.");
+  SLIC_WARNING_ROOT_IF(previous_solve_ == PreviousSolve::Adjoint,
```
sethwatts, line 474: I believe that LiDO will often ask to solve multiple adjoint problems with different adjoint loads following a single forward solve, any time the solution from `Solid` is used in multiple QOIs. During our sensitivity calculation, we'd provide different adjoint loads, but both should have the same solver state as the (final iteration of) the forward solve. I understand the scenario you're guarding against here, and I assume this warning isn't fatal, so this isn't a huge problem, but I was wondering if we could suppress it from LiDO.

src/serac/physics/solid.cpp
```diff
@@ -397,13 +398,69 @@ void Solid::advanceTimestep(double& dt)
...
+                     "Sensitivities only valid following a forward and adjoint solve.");
+  SLIC_WARNING_ROOT_IF(
+      previous_solve_ == PreviousSolve::Forward,
+      "Sensitivities only valid following a forward and adjoint solve (in that order). The previous solve was a "
+      "forward analysis. Ensure that the correct displacement and adjoint states are set for sensitivies.");
+
+  if (!shear_sensitivity_coef_) {
+    LinearElasticMaterial* linear_mat = dynamic_cast<LinearElasticMaterial*>(material_.get());
+
+    SLIC_ERROR_ROOT_IF(!linear_mat, "Only linear elastic materials allowed for sensitivity analysis.");
```
sethwatts: I'm assuming this is temporary, pending further development of `Functional`?

src/serac/physics/solid.cpp
```diff
@@ -397,13 +398,69 @@ void Solid::advanceTimestep(double& dt)
...
+
+    shear_sensitivity_form_->AddDomainIntegrator(new mfem::DomainLFIntegrator(*shear_sensitivity_coef_));
+  }
+
+  shear_sensitivity_form_->Assemble();
+
+  return *shear_sensitivity_form_;
+}
+
+mfem::ParLinearForm& Solid::bulkModulusSensitivity(mfem::ParFiniteElementSpace& bulk_space)
```
sethwatts, line 444: I think `bulkModulusSensitivity` needs the same error checking `shearModulusSensitivity` has to make sure that an adjoint displacement has just been computed.

src/serac/physics/solid.hpp
```diff
@@ -351,6 +362,24 @@ class Solid : public BasePhysics {
   virtual const serac::FiniteElementState& solveAdjoint(mfem::ParLinearForm& adjoint_load_form,
                                                         FiniteElementState*  state_with_essential_boundary = nullptr);
 
+  /**
+   * @brief Compute the implicit sensitivity of the quantity of interest used in defining the load for the adjoint
```
sethwatts, line 366: It would probably be worth mentioning in the documentation here and for `bulkModulusSensitivitiy` that `solveAjoint` needs to be called first, and on the correct `adjoint_load_form`.

src/serac/physics/solid.cpp
jamiebramwell: Yeah, I wasn't sure how often you would do this (e.g. multiple QoIs). You're correct that this won't kill a run, but it will print a warning. I can remove this if you want. Another option is you pass us the appropriate displacement and adjoint fields for the adjoint and sensitivity calls so you always know what is being used. Any thoughts?

src/serac/physics/solid.cpp
jamiebramwell: Yeah, this is me being lazy. I could do these computations for neo-Hookean as well, but it will become trivially easy once we move the solid module to `Functional`, so I thought my time is better spent elsewhere.

src/serac/physics/solid.cpp
sethwatts: I can see how passing a displacement field along with the adjoint load would let you set the correct state for the solver for elasticity, but I think for plasticity or other models with internal state, it would be insufficient information, and I think we would be relying on Serac to already be in the right state. 

I also don't think this is a bad warning in the general case. Is there a way we could call it that communicates that we know what we're doing and just suppress the warning?

src/serac/physics/solid.cpp
```diff
@@ -397,13 +398,82 @@ void Solid::advanceTimestep(double& dt)
...
+                     "Sensitivities only valid following a forward and adjoint solve.");
+  SLIC_WARNING_ROOT_IF(
+      previous_solve_ == PreviousSolve::Forward,
+      "Sensitivities only valid following a forward and adjoint solve (in that order). The previous solve was a "
+      "forward analysis. Ensure that the correct displacement and adjoint states are set for sensitivies.");
+
+  if (!bulk_sensitivity_coef_) {
+    LinearElasticMaterial* linear_mat = dynamic_cast<LinearElasticMaterial*>(material_.get());
+
+    SLIC_ERROR_ROOT_IF(!linear_mat, "Only linear elastic materials allowed for sensitivity analysis.");
```
jandrej: That code is a duplicate from above. Should be moved to utility function.

jandrej (approved):
LGTM

src/serac/physics/solid.cpp
```diff
@@ -397,13 +398,82 @@ void Solid::advanceTimestep(double& dt)
...
+    shear_sensitivity_coef_ =
+        std::make_unique<mfem_ext::ShearSensitivityCoefficient>(displacement_, adjoint_displacement_, *linear_mat);
+  }
+
+  // Add a scalar linear form integrator using the shear sensitivity coefficient against the given shear modulus finite
+  // element space
+  if (!shear_sensitivity_form_ || shear_sensitivity_form_->FESpace() != &shear_space) {
+    shear_sensitivity_form_ = std::make_unique<mfem::ParLinearForm>(&shear_space);
+
+    shear_sensitivity_form_->AddDomainIntegrator(new mfem::DomainLFIntegrator(*shear_sensitivity_coef_));
```
jonwong12, line 435: I wasn't sure what equation you were implementing exactly in `shear_sensitivity_coef_`, but it seems between `shear_sensitivity_coef_` and `mfem::Domain:FIntegrator::AssembleRHSVector` some of the weights (`ip.weight`) might be used twice? Not sure if this is intended or not.

src/serac/coefficients/sensitivity_coefficients.cpp
```diff
@@ -0,0 +1,110 @@
...
+  displacement_.gridFunc().GetVectorGradient(T, du_dX_);
+
+  solid_util::calcDeformationGradient(du_dX_, F_);
+
+  double det_J = F_.Det();
+
+  // Compute the adjoint gradient and strain
+  adjoint_displacement_.gridFunc().GetVectorGradient(T, dp_dX_);
+
+  serac::solid_util::calcLinearizedStrain(dp_dX_, adjoint_strain_);
```
jonwong12, line 89: It seems like `adjoint_strain_`, `d_sigma_d_bulk_`, and `d_sigma_d_shear_` are temporaries based on the integration point and also `private` to the class.  Perhaps the code could be a little simpler if we just declared a temporary `DenseMatrix` here directly instead of the the `private` member.

jonwong12: Also similar considerations for `du_dX`, `dp_dx`, and `F`.

src/serac/physics/materials/hyperelastic_material.cpp
```diff
@@ -101,6 +101,44 @@ inline void LinearElasticMaterial::EvalCoeffs() const
...
+  d_sigma_d_shear.SetSize(dim);
+  epsilon_.SetSize(dim);
+
+  // Evaluate the linearized strain tensor from the displacement gradient
+  serac::solid_util::calcLinearizedStrain(du_dX, epsilon_);
+
+  d_sigma_d_shear      = 0.0;
+  double trace_epsilon = epsilon_.Trace();
+
+  // Calculate the sensitivity of the scress with respect to the shear moulus by differentiating Hooke's law
```
white238: scress to stress

same below

src/serac/physics/solid.cpp
goxberry: @sethwatts From a separation of concerns perspective, the issue of ignoring warnings seems like it could depend on Serac's philosophy? If Serac exposed the warning status to clients via a return value or pointer/reference argument, then that functionality would give clients more flexibility regarding how to handle that status. But there's also an argument to be made that in many programming contexts -- even in production codes, and especially in scientific applications -- error handling is deprioritized or ignored, and it may be more convenient for Serac's user base if Serac takes responsibility for error handling.

goxberry: In ODE-constrained optimization, the multiple QoI case arises often enough that the warnings would be kind of annoying. I imagine something similar is true of PDE-constrained optimization, and it's a legal operation, so it seems sort of weird to throw a warning.

I get the impression that the purpose of `PreviousSolve` is to avoid users performing operations on a `serac::Solid` object in an inconsistent state (e.g., solving an adjoint problem without solving a forward problem). Repeatedly solving the adjoint problem seems fine to me, but there might be other operations one would want to guard against (e.g., invoking the adjoint displacement getter before solving a forward problem or adjoint problem)?

src/serac/physics/solid.cpp
```diff
@@ -397,13 +398,82 @@ void Solid::advanceTimestep(double& dt)
   mesh_.NewNodes(*deformed_nodes_);
 
   cycle_ += 1;
+
+  previous_solve_ = PreviousSolve::Forward;
+}
+
+mfem::ParLinearForm& Solid::shearModulusSensitivity(mfem::ParFiniteElementSpace& shear_space)
+{
+  SLIC_ERROR_ROOT_IF(previous_solve_ == PreviousSolve::None,
```
goxberry, line 407: Could be worth refactoring the repeated error/warning logging code re: `previous_solve_` into an internal helper method/function?

goxberry (approved):
LGTM.

white238 (approved):

src/serac/physics/solid.cpp
jamiebramwell: Let's chat about more error handling at the next Thursday meeting.

src/serac/physics/solid.cpp
jamiebramwell: Nice catch! This was in fact double counting the weights.

src/serac/coefficients/sensitivity_coefficients.cpp
jamiebramwell: The issue is that it's fairly expensive to construct and destruct these dense matricies at each quadrature point as it uses dynamic allocations. This mutable private member solution is what is used in most of the MFEM integrators for that reason.

PR #521, opened 2021-07-22
Fix full lassen build

white238:
because it was broken.

### DIFF
--- a/.gitlab/build_lassen.yml
+++ b/.gitlab/build_lassen.yml
@@ -50,10 +50,10 @@ lassen-clang_10_0_1-src:
     HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-${COMPILER}.cmake"
   extends: [.src_build_on_lassen, .with_cuda]
 
+# Note: to reduce duplication SPEC is not defined here, if we move to more than one
+# spec on lassen add it back like quartz
 lassen-clang_10_0_1-full:
   variables:
     COMPILER: "clang@10.0.1"
-    SPEC: "%${COMPILER}+cuda"
-    EXTRA_SPEC: "cuda_arch=70"
     HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-${COMPILER}.cmake"
   extends: [.full_build_on_lassen, .with_cuda]
--- a/.gitlab/build_lassen.yml
+++ b/.gitlab/build_lassen.yml
@@ -38,7 +38,7 @@
 .full_build_on_lassen:
   stage: build
   variables:
-    ALLOC_COMMAND: lalloc 1 -W 45 -q pdebug -G ${ALLOC_BANK}
+    ALLOC_COMMAND: lalloc 1 -W 55 -q pdebug -G ${ALLOC_BANK}
   extends: [.full_build_script, .on_lassen, .full_workflow]
   needs: []
 
--- a/.gitlab/build_quartz.yml
+++ b/.gitlab/build_quartz.yml
@@ -70,11 +70,11 @@ quartz-gcc_8_3_1-src:
 quartz-clang_10_0_0-full:
   variables:
     COMPILER: "clang@10.0.0"
-    SPEC: "%${COMPILER}"
+    SPEC: "--spec=%${COMPILER}"
   extends: .full_build_on_quartz
 
 quartz-gcc_8_3_1-full:
   variables:
     COMPILER: "gcc@8.3.1"
-    SPEC: "%${COMPILER}"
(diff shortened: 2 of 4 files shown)


white238:
running full build via gitlab now.

jamiebramwell (approved):

white238:
This is ready for review.

scripts/spack/packages/serac/package.py
```diff
@@ -82,10 +82,8 @@ class Serac(CachedCMakePackage, CudaPackage):
     depends_on('python', when="+devtools")
     depends_on('py-sphinx', when="+devtools")
 
-    # Workaround for concretizer bug
-    #depends_on("mfem+sundials", when="+sundials")
-    #depends_on("sundials~shared", when="+sundials")
-    depends_on("sundials~shared+hypre+monitoring")
+    depends_on("sundials~shared+hypre+monitoring~examples-c~examples-f77~examples-install",
```
jamiebramwell, line 85: YIKES.

samuelpmish (approved):

.gitlab/build_lassen.yml
```diff
@@ -38,7 +38,7 @@
 .full_build_on_lassen:
   stage: build
   variables:
-    ALLOC_COMMAND: lalloc 1 -W 45 -q pdebug -G ${ALLOC_BANK}
+    ALLOC_COMMAND: lalloc 1 -W 55 -q pdebug -G ${ALLOC_BANK}
```
white238, line 41: Hit the time limit again

.gitlab/build_lassen.yml
```diff
@@ -50,10 +50,10 @@ lassen-clang_10_0_1-src:
...
   extends: [.src_build_on_lassen, .with_cuda]
 
+# Note: to reduce duplication SPEC is not defined here, if we move to more than one
+# spec on lassen add it back like quartz
 lassen-clang_10_0_1-full:
   variables:
     COMPILER: "clang@10.0.1"
-    SPEC: "%${COMPILER}+cuda"
-    EXTRA_SPEC: "cuda_arch=70"
     HOST_CONFIG: "lassen-blueos_3_ppc64le_ib_p9-${COMPILER}.cmake"
```
white238, line 58: This was out of sync with the one in `scripts/llnl/spack/specs.json` so was causing differences.  `scripts/llnl/build_tpls.py` naturally reads in the spec from the json file and since there is only one in this platform we don't need to list it here.  This is not the case for quartz.

.gitlab/build_quartz.yml
```diff
@@ -70,11 +70,11 @@ quartz-gcc_8_3_1-src:
 quartz-clang_10_0_0-full:
   variables:
     COMPILER: "clang@10.0.0"
-    SPEC: "%${COMPILER}"
+    SPEC: "--spec=%${COMPILER}"
```
white238, line 73: Pulled the commandline argument into this variable so it could be empty in lassen.

scripts/spack/packages/serac/package.py
white238: reason: sundials examples started failing to link at sometime on toss3 clang.  i do not know why.  we also do not need them so i turned them off via the given sundials variants: ~ for off + for on

scripts/spack/packages/serac/package.py
```diff
@@ -82,10 +82,8 @@ class Serac(CachedCMakePackage, CudaPackage):
...
+               when="+sundials")
```
white238, line 86: the `when` was slightly inconsistent.  we werent truly utilizing our sundials variant for this dependency.

samuelpmishLLNL (approved):

scripts/spack/packages/serac/package.py
goxberry: @white238 Just ran into this issue (twice; once with Serac and again with Smith). Has it been reported upstream to Spack and/or Sundials?

white238: it has not.. should but was busy to get this out the door.

white238: Also I'm still not clear what was the cause of this.  It used to work and as far as I know nothing changed to cause this. Plus we certainly don't depend on any of their examples and their library works just fine.

goxberry: Agreed; it used to work for us, too.

goxberry: I'm supposed to review the updates to MFEM's Spack package and I'm not sure what to tell them re: Sundials, so I'm doing some diagnostic work to see if I can get more info.

PR #523, opened 2021-07-23
Intermediate GPU kernel

jonwong12:
In this PR the elasticity and thermal evaluation and gradient kernels run on the GPU. The work is an amalgam of work between @samuelpmishLLNL , @joshessman-llnl , and @jonwong12 . 

Notable changes:
- serac::tuple implementation
- Introduces the notation of `cpu_policy` and `gpu_policy`
- Separates CPU and GPU kernels.
- Methods are separated into several files. Some are shared between boundary and domain integrals. Some are shared between CPU and GPU implementations of domain_integrals, and etc.
- Hardcoded indexing for derivatives_ptr has been abstracted in case we need to change the access pattern consistently on the GPU.
- GPU kernels have two implementations: element-based, and quadrature-based implementation. Both kernels use the same underlying quadrature method.

Caveats:
- There is an incorrect answer for the Hcurl case that needs to be investigated.
- Quadrature point data isn't supported on the GPU yet.

TODO:
I think most of the work that needs to be done is just in tidying up the code in it's current state.

- [x] Better comments and documentation. 
- [x] Deleting unnecessary debugging calls.
- [x] Reconsolidating quadrature calls into a `SERAC_HOST_DEVICE` to help with consistency and maintainability.
- [x] Make sure all tests pass


src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,309 @@
...
+      // once we've finished the element integration loop, write our element residuals
+      // out to memory, to be later assembled into global residuals by mfem
+      detail::Add(r, r_elem, e);
+    } // quadrature x element loop
+
+  }
+
+
+    template <Geometry g, typename test, typename trial, int Q, int blocksize,
+	  typename derivatives_type, typename lambda>
```
samuelpmishLLNL: unusual indentation

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,309 @@
...
+  __global__ void eval_cuda_element(const u_type u, r_type r, derivatives_type * derivatives_ptr, J_type J, X_type X, int num_elements, lambda qf) {
+
+    using test_element               = finite_element<g, test>;
+    using trial_element              = finite_element<g, trial>;
+    using element_residual_type      = typename trial_element::residual_type;
+    static constexpr auto rule       = GaussQuadratureRule<g, Q>();
+
+    // for each element in the domain
+    const int grid_stride = blockDim.x * gridDim.x;
+#pragma unroll
```
samuelpmishLLNL: does clang-format flush this all the way to the left? I usually prefer if `#pragma unroll` is aligned with its corresponding for loop

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,309 @@
...
+  auto   xi  = rule.points[q];
+  auto   dxi = rule.weights[q];
+  auto   J_q = make_tensor<dim, dim>([&](int i, int j) { return J(q, i, j, e); });
+  double dx  = det(J_q) * dxi;
+
+  // evaluate the (change in) value/derivatives at this quadrature point
+  auto darg = Preprocess<trial_element>(du_elem, xi, J_q);
+
+  // recall the derivative of the q-function w.r.t. its arguments at this quadrature point
+
```
samuelpmishLLNL: delete whitespace line

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,309 @@
...
+
+  // for each element in the domain
+  //  for (int e = 0; e < num_elements; e++) {
+  // int qe = blockIdx.x * blockDim.x + threadIdx.x;
+  // if (qe < num_elements * rule.size()) {
+
+  const int grid_stride = blockDim.x * gridDim.x;
+#pragma unroll
+  for (int qe = blockIdx.x * blockDim.x + threadIdx.x; qe < num_elements * rule.size(); qe += grid_stride) {
```
samuelpmishLLNL: it might be clearer to move the start/end values outside the loop
```suggestion
  auto thread_id = blockIdx.x * blockDim.x + threadIdx.x;
  auto num_quadrature_points = num_elements * rule.size();
  for (int qe = thread_id; qe < num_quadrature_points; qe += grid_stride) {
```

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,131 @@
...
+ * test function space.
+ *
+ * By default:
+ *  H1 family elements integrate std::get<0>(f) against the test space shape functions
+ *                           and std::get<1>(f) against the test space shape function gradients
+ *  Hcurl family elements integrate std::get<0>(f) against the test space shape functions
+ *                              and std::get<1>(f) against the curl of the test space shape functions
+ * TODO: Hdiv family elements integrate std::get<0>(f) against the test space shape functions
+ *                                  and std::get<1>(f) against the divergence of the test space shape functions
+ * TODO: L2 family elements integrate f against test space shape functions
```
samuelpmishLLNL: I think this isn't a `TODO:` anymore
```suggestion
 *  L2 family elements integrate f against test space shape functions
```

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,131 @@
...
+  // to get element residual contributions
+  r_elem += Postprocess<test_element>(get_value(qf_output), xi, J_q) * dx;
+
+  // here, we store the derivative of the q-function w.r.t. its input arguments
+  //
+  // this will be used by other kernels to evaluate gradients / adjoints / directional derivatives
+
+  // detail::AccessDerivatives(derivatives_ptr, e, q, rule, num_elements) = get_gradient(qf_output);
+  // Note: This pattern may result in non-coalesced access depend on how it executed.
+  detail::AccessDerivatives(derivatives_ptr, e, q, rule, num_elements) = get_gradient(qf_output);
```
samuelpmishLLNL, line 156: @joshessman-llnl 
this is another use case for a multidimensional array container, it would be more idiomatic if we could write
```cpp
derivatives(e, q) = get_gradient(qf_output);
```

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -72,6 +72,7 @@ class Functional;
  * Functional< H1(Hcurl) > my_residual;
  * @endcode
  *
+ * @tparam execution_policy the The space of test functions to use
```
```suggestion
 * @tparam execution_policy which kind of processor should be used to carry out calculations
```

src/serac/physics/utilities/functional/integral_utilities.hpp
```diff
@@ -230,6 +232,29 @@ struct qf_result<lambda_type, x_t, u_du_t, qpt_data_type, std::enable_if_t<!std:
                                     std::add_lvalue_reference_t<qpt_data_type>>;
 };
 
+/**
+ * @brief derivatives_ptr access
+ *
+ * Templating this will allow us to change the stride-access patterns more consistently
+ *
```
samuelpmishLLNL, line 285: maybe rename `quadrature_coalescing` to `row_major` or provide a `@tparam` description for people who aren't familiar with GPU memory access terminology.

src/serac/physics/utilities/functional/integral_utilities.hpp
```diff
@@ -230,6 +232,29 @@ struct qf_result<lambda_type, x_t, u_du_t, qpt_data_type, std::enable_if_t<!std:
...
+SERAC_HOST_DEVICE constexpr derivatives_type& AccessDerivatives(derivatives_type* derivatives_ptr, int e, int q,
+                                                                [[maybe_unused]] rule_type& rule,
+                                                                [[maybe_unused]] int        num_elements)
+{
+  if constexpr (quadrature_coalescing) {
+    return derivatives_ptr[e * int(rule.size()) + q];
+  } else {
+    return derivatives_ptr[q * num_elements + e];
+  }
+}
```
samuelpmishLLNL: we could just accept a stride as the last argument, and not have to have the `[[maybe_unused]]`, e.g.
```cpp
template <typename derivatives_type, typename rule_type, bool quadrature_coalescing = true>
 SERAC_HOST_DEVICE constexpr derivatives_type& AccessDerivatives(derivatives_type* derivatives_ptr, int e, int q, int stride)
 {
   if constexpr (quadrature_coalescing) {
     return derivatives_ptr[e * stride + q];
   } else {
     return derivatives_ptr[q * stride + e];
   }
 }
```

but this will get cleaner when we have a multidimensional array container

src/serac/physics/utilities/functional/test.cu
```diff
@@ -0,0 +1,33 @@
+#include <cuda_runtime.h>
+#include "../src/serac/physics/utilities/functional/tuple_arithmetic.hpp"
+
+using namespace serac;
+
+__global__ void kernel() {
+
```
samuelpmishLLNL: This doesn't appear to test anything other than that the code compiles, should it be deleted?

src/serac/physics/utilities/functional/integral_utilities.hpp
jonwong12: Maybe with a multidimensional array that will be better? Ideally we can enforce the strides properly the same way you have everything enforced in the `tensor` classes. I'd like to avoid specifying specifying `stride` as an argument since that can lead to error prone behavior.

src/serac/physics/utilities/functional/test.cu
jonwong12: Good catch.

src/serac/physics/utilities/functional/integral_utilities.hpp
samuelpmishLLNL: yep, we have to jump through these hoops to index things because we just don't have a proper container! I think mfem::DeviceTensor is pretty close to what we want, except that it's purely a view.

src/serac/infrastructure/accelerator.hpp
```diff
@@ -15,17 +15,47 @@
...
 /**
  * @brief Accelerator functionality
  */
-namespace serac::accelerator {
+namespace serac {
+
+/**
+ * @brief tag type for signaling that calculations that should be performed on the CPU
+ */
+struct cpu_policy {
```
joshessman-llnl, line 55: Could we look at reusing the execution space/policy stuff that Axom has: https://github.com/LLNL/axom/blob/develop/src/axom/core/execution/execution_space.hpp

src/serac/infrastructure/accelerator.cpp
```diff
@@ -24,16 +24,16 @@ std::unique_ptr<mfem::Device> device;
 void initializeDevice()
 {
   SLIC_ERROR_ROOT_IF(device, "serac::accelerator::initializeDevice cannot be called more than once");
-  device = std::make_unique<mfem::Device>();
+  // device = std::make_unique<mfem::Device>();
```
jamiebramwell: Why is this commented out?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -166,42 +92,36 @@ void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type*
...
-      // here, we store the derivative of the q-function w.r.t. its input arguments
-      //
-      // this will be used by other kernels to evaluate gradients / adjoints / directional derivatives
-      derivatives_ptr[e * int(rule.size()) + q] = get_gradient(qf_output);
+      // eval_quadrature is a SERAC_HOST_DEVICE quadrature point calculation
+      if constexpr (std::is_same_v<qpt_data_type, void>) {
+        eval_quadrature<g, test, trial, Q, derivatives_type, lambda>(e, q, u_elem, r_elem, derivatives_ptr, J, X,
+                                                                     num_elements, qf);
+      } else {
+        auto   xi  = rule.points[q];
```
jamiebramwell, line 104: Am I correct in thinking this switch is because the quadrature point data is not on the GPU yet? If so, a comment would be appropriate.

src/serac/infrastructure/accelerator.cpp
jamiebramwell: Is this due to the testing bug? If so, we should add a comment.

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -590,19 +493,45 @@ class DomainIntegral {
...
+
+      gradient_mat_ = [=](mfem::Vector& K_e) {
+        gradient_matrix_kernel<geometry, test_space, trial_space, Q>(K_e, qf_derivatives.get(), J_, num_elements);
+      };
+    }
+
+    if constexpr (std::is_same_v<execution_policy, serac::gpu_policy>) {
+      // TODO: add support for gradient_matrix_kernel
+#if defined(__CUDACC__)
+      evaluation_ = [=](const mfem::Vector& U, mfem::Vector& R) {
```
jamiebramwell, line 523: The code should error if a gpu policy is requested on a non-cuda build

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,163 @@
...
+      curl_W = dot(curl_W, transpose(J));
+    }
+    return (W * serac::get<0>(f) + curl_W * serac::get<1>(f));
+  }
+}
+
+// TODO: Add more comments. Quadrature level evaluation
+
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type, typename lambda,
+          typename u_elem_type, typename element_residual_type, typename J_type, typename X_type>
```
jamiebramwell, line 119: `J_type` and `X_type` need more descriptive names.

jamiebramwell: `Jacobian_type` and `position_type`?

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,292 @@
...
+    o << success_string << std::endl;
+  }
+}
+
+/**
+ * @brief Defines whether to loop by elements or quadrature points
+ */
+enum ThreadExecutionPolicy
+{
+  THREAD_PER_ELEMENT_QUADRATURE_POINT,
```
jamiebramwell: Would `THREAD_PER_QUADRATURE_POINT` be appropriate? I think it's more descriptive.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,292 @@
...
+                                                                   num_elements, qf);
+    }
+
+    // once we've finished the element integration loop, write our element residuals
+    // out to memory, to be later assembled into global residuals by mfem
+    detail::Add(r, r_elem, e);
+  }  // e loop
+}
+
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type, typename lambda, typename u_type,
```
jamiebramwell: While not much, this duplication is a bit worrisome. I assume this will be raja-ified and merged with the CPU version in a future PR.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,292 @@
...
+    int blocks_element = (num_elements + config.blocksize - 1) / config.blocksize;
+    gradient_cuda_element<g, test, trial, Q, derivatives_type>
+        <<<blocks_element, config.blocksize>>>(du, dr, derivatives_ptr, J, num_elements);
+  }
+
+  cudaDeviceSynchronize();
+  serac::detail::displayLastCUDAErrorMessage(std::cout);
+  dR.HostRead();
+
+  // J_.UseDevice(false);
```
jamiebramwell: dead code.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,292 @@
...
+  using test_element               = finite_element<g, test>;
+  using trial_element              = finite_element<g, trial>;
+  using element_residual_type      = typename trial_element::residual_type;
+  static constexpr int  test_ndof  = test_element::ndof;
+  static constexpr int  trial_ndof = trial_element::ndof;
+  static constexpr auto rule       = GaussQuadratureRule<g, Q>();
+  static constexpr int  dim        = dimension_of(g);
+
+  // Use the device (GPU)
+  // J_.UseDevice(true);
```
jamiebramwell: dead code.

src/serac/physics/utilities/functional/tests/CMakeLists.txt
```diff
@@ -49,10 +49,50 @@ endforeach()
...
-    set(functional_tests_cuda
-        tensor_unit_tests_cuda.cu)
+    set(functional_tests_cuda 
+        tensor_unit_tests_cuda.cu )
 
     serac_add_tests( SOURCES ${functional_tests_cuda}
                      DEPENDS_ON gtest serac_functional ${functional_depends} cuda)
 
+    # Due to GPU bug we build the test and then create ctests for each individual gtest
+    set(functional_comparisons_cuda_tests
```
jamiebramwell: 👍

src/serac/physics/utilities/functional/domain_integral.hpp
jonwong12: Quadrature point data isn't supported yet on the GPU since it was added during the PR request.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
jonwong12: Sorry, we're in the process of debugging this. So some of the comments may or may not come back.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
jonwong12: Sorry, I don't think I explained it well since this is an intermediate work PR. I'll remember to add in some comments here. `eval_cuda_element` and `eval_cuda_quadrature` are GPU kernels that launch over 1 thread per element and 1 thread per quadrature point. The kernels will be slightly different and you may want to use one over another depending on the actual lambda `qf`. `eval_quadrature` is the same for both the CPU and the GPU. Basically in `eval_cuda_quadrature` the loop over quadrature points is flattened and merged into the loop over elements. When you do that, you need to retrieve `e` and `q`. It may be possible to merge them into RAJA, but it might be a little harder to control the performance of the kernels.

`eval_cuda_kernel` is the actual call that launches the kernel on the GPU and converts `mfem::Vectors/DenseTensors` to something the kernel can understand.

src/serac/infrastructure/accelerator.cpp
jonwong12: Yes. Sorry. The current state of the PR is a little weird because of the mfem. 4.3 update.

jonwong12: For now I'm putting it back in case other tests rely on this behavior, but in the future we may have to keep an eye on how mfem decides `mfem::Device` should work.

src/serac/infrastructure/accelerator.hpp
```diff
@@ -15,17 +15,47 @@
...
+#else
+#define SERAC_SUPPRESS_NVCC_HOSTDEVICE_WARNING #pragma hd_warning_disable
+#endif
 #else
 /**
  * @brief Macro that toggles between decorating a function for host and device or noop's for non-accelated builds.
  */
 #define SERAC_HOST_DEVICE
+#define SERAC_DEVICE
+#define SERAC_HOST
```
joshessman-llnl, line 39: Extremely minor: Could these be in the same order in both the cuda and non-cuda ifdef blocks?

src/serac/physics/utilities/functional/CMakeLists.txt
```diff
@@ -33,8 +33,14 @@ set(functional_detail_headers
     detail/segment_Hcurl.inl
     detail/segment_L2.inl
     detail/metaprogramming.hpp
+    integral_utilities.hpp
+    domain_integral_shared.hpp
     )
 
+if(ENABLE_CUDA)
+list(APPEND functional_headers domain_integral_cuda.cuh)
+endif()
```
joshessman-llnl:
```suggestion
blt_list_append(TO functional_headers ELEMENTS domain_integral_cuda.cuh IF ENABLE_CUDA)
```

src/serac/infrastructure/accelerator.hpp
```diff
@@ -15,17 +15,47 @@
 
 #if defined(__CUDACC__)
 #define SERAC_HOST_DEVICE __host__ __device__
+#define SERAC_HOST __host__
+#define SERAC_DEVICE __device__
+
+#if __CUDAVER__ >= 75000
+#define SERAC_SUPPRESS_NVCC_HOSTDEVICE_WARNING #pragma nv_exec_check_disable
```
joshessman-llnl, line 28: Can we add a snippet (either here or where this gets used) explaining why we need to suppress this warning?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -590,19 +495,44 @@ class DomainIntegral {
...
+
+      gradient_ = [=](const mfem::Vector& dU, mfem::Vector& dR) {
+        serac::detail::ThreadExecutionConfiguration exec_config{.blocksize = 128};
+
+        domain_integral::gradient_kernel_cuda<geometry, test_space, trial_space, Q,
+                                              serac::detail::ThreadExecutionPolicy::THREAD_PER_QUADRATURE_POINT>(
+            exec_config, dU, dR, qf_derivatives.get(), J_, num_elements);
+      };
+#else
+      SLIC_ERROR_ROOT("serac::gpu_policy not supported on non-cuda platforms at the moment");
```
joshessman-llnl: Since this is within an `if constexpr` block can this be a compile-time error - either with https://en.cppreference.com/w/cpp/preprocessor/error or https://artificial-mind.net/blog/2020/10/03/always-false

src/serac/physics/utilities/functional/tests/CMakeLists.txt
```diff
@@ -49,10 +49,24 @@ endforeach()
...
                      DEPENDS_ON gtest serac_functional ${functional_depends} cuda)
 
+    # FIXME: Different linker is used depending on whether axom is a dependency??
+    # clang directly (with axom) or the MPI wrapper (without axom)
+    #serac_add_tests( SOURCES mfem_issue_reproducer.cu
+    #                 DEPENDS_ON mfem cuda mpi axom)
+
+    blt_add_executable(NAME        mfem_issue_reproducer
+                       SOURCES     mfem_issue_reproducer.cu
+                       OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
```
joshessman-llnl, line 67: I think we can get rid of this

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -0,0 +1,446 @@
...
+
+#include "axom/slic/core/SimpleLogger.hpp"
+#include "serac/infrastructure/input.hpp"
+#include "serac/serac_config.hpp"
+#include "serac/numerics/expr_template_ops.hpp"
+#include "serac/physics/operators/stdfunction_operator.hpp"
+#include "serac/physics/utilities/functional/functional.hpp"
+#include "serac/physics/utilities/functional/tensor.hpp"
+#include "serac/infrastructure/profiling.hpp"
+#include <gtest/gtest.h>
```
joshessman-llnl: Minor: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

In particular, the gtest header should be higher up

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -0,0 +1,446 @@
...
+int parallel_refinement = 0;
+
+int num_procs, myid;
+int nsamples = 1;  // because mfem doesn't take in unsigned int
+
+constexpr bool                 verbose = true;
+std::unique_ptr<mfem::ParMesh> mesh2D;
+std::unique_ptr<mfem::ParMesh> mesh3D;
+
+std::map<int, std::string> meshfiles = {{2, SERAC_REPO_DIR "/data/meshes/star.mesh"},
```
joshessman-llnl:
```suggestion
const std::map<int, std::string> meshfiles = {{2, SERAC_REPO_DIR "/data/meshes/star.mesh"},
```

src/serac/infrastructure/accelerator.hpp
jonwong12: I think that's a good next step to look at. There are some additional GPU related configurations I'd like to be able to access possibly both at compile time and run time.

src/serac/physics/utilities/functional/tests/CMakeLists.txt
jonwong12: I was going to keep it for now because there is still a strange cuda attribute error on lassen. It seems not catastrophic but it reproduces the error.

src/serac/infrastructure/accelerator.hpp
jonwong12: @joshessman-llnl @samuelpmishLLNL Can you mention why this suppression is needed? Thanks!

samuelpmishLLNL: see: https://godbolt.org/z/9M57xao1q
nvcc will sometimes emit a warning if a `__host__ __device__` function calls a `__host__`-only or `__device__`-only function. `make_tensor` is marked `__host__ __device__` and is used frequently in the code base, so it was emitting a lot of warnings. This #pragma directive suppresses the warning for a specific function.

src/serac/infrastructure/accelerator.hpp
samuelpmishLLNL: it looks like axom's policies are trying to do a lot more than the ones in this PR. Here,`cpu_policy` and `gpu_policy` only specify where the calculations should take place, but let the individual kernels decide their own configurations. Right now, none of our kernels support different loop arrangements, reduction strategies, etc.

src/serac/physics/utilities/functional/domain_integral.hpp
samuelpmishLLNL: I don't really understand what this `if constexpr` is doing. This function is never executed on the GPU anyway, so it should be able to support using `QuadratureData`.

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+namespace domain_integral {
+
+/**
+ * @brief Computes the arguments to be passed into the q-function (shape function evaluations)
+ * By default:
+ *  H1 family elements will compute {value, gradient}
+ *  Hcurl family elements will compute {value, curl}
+ *  TODO: Hdiv family elements will compute {value, divergence}
+ *  TODO: L2 family elements will compute value
```
samuelpmishLLNL:
```suggestion
 *  L2 family elements will compute value
```

src/serac/physics/utilities/functional/tuple_arithmetic.hpp
```diff
@@ -199,23 +159,24 @@ auto chain_rule_tuple_matvec(std::tuple<T...> df_dx, std::tuple<S...> dx)
...
- * is given by decltype(std::get<i>(A) + std::get<i>(B))
+ * is given by decltype(serac::get<i>(A) + serac::get<i>(B))
  */
-template <typename... S, typename... T>
-constexpr auto operator+(const std::tuple<S...>& A, const std::tuple<T...>& B)
-{
-  static_assert(sizeof...(S) == sizeof...(T), "Error in operator+: tuple sizes must match");
-  return detail::plus_helper(A, B, std::make_integer_sequence<int, int(sizeof...(S))>{});
-}
+// template <typename... S, typename... T>
```
white238: There is a lot of dead code in this file.  reason?

src/serac/physics/utilities/functional/tuple.hpp
```diff
@@ -0,0 +1,408 @@
+#pragma once
+
+#include <utility>
```
white238, line 14: We need some unit tests for these..

src/serac/physics/utilities/functional/tuple_arithmetic.hpp
samuelpmishLLNL: With the refactor to use the gpu-enabled `serac::tuple`, much of the arithmetic was put into the `serac::tuple` header directly. We could likely just remove this file completely (might have to move the remaining relevant parts into the `serac::tuple` header).

src/serac/physics/utilities/functional/tuple.hpp
samuelpmishLLNL: there are unit tests for `serac::tuple` here: 
https://github.com/LLNL/serac/blob/feature/jonwong12/gpu_functional_comparisons/src/serac/physics/utilities/functional/tests/tuple_arithmetic_unit_tests.cpp
but they may not be exhaustive

src/serac/physics/utilities/functional/tests/mfem_issue_reproducer.cu
```diff
@@ -0,0 +1,90 @@
...
+  u_global.Randomize();
+
+  mfem::Vector U(fespace.TrueVSize());
+  U.UseDevice(true);
+  u_global.GetTrueDofs(U);
+
+  // Compute the residual using standard MFEM methods
+  mfem::Vector r1(U.Size());
+  J->Mult(U, r1);
+  r1 -= *F;
```
samuelpmishLLNL, line 73: I'm not sure if we noted it explicitly, but our operator overloading expression templates don't work with the GPU (mfem's memory management model makes it difficult to reason about what those expressions should do).

src/serac/physics/utilities/functional/tuple.hpp
white238: missed that. should be good enough. just wanted a basic level of correctness checking before it blows up in larger things.

src/serac/infrastructure/accelerator.hpp
jamiebramwell: I think we should keep it as-is for now. Integrating with axom policies can be part of a follow on PR.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,259 @@
...
+ */
+struct ThreadExecutionConfiguration {
+  int blocksize;
+};
+
+}  // namespace detail
+
+namespace domain_integral {
+
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type, typename lambda,
```
jamiebramwell, line 69: These need doxygen.

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,163 @@
...
+    if constexpr (dim == 3) {
+      curl_W = dot(curl_W, transpose(J));
+    }
+    return (W * serac::get<0>(f) + curl_W * serac::get<1>(f));
+  }
+}
+
+// TODO: Add more comments. Quadrature level evaluation
+
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type, typename lambda,
```
jamiebramwell, line 118: Needs doxygen.

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,163 @@
...
+  // this will be used by other kernels to evaluate gradients / adjoints / directional derivatives
+
+  // detail::AccessDerivatives(derivatives_ptr, e, q, rule, num_elements) = get_gradient(qf_output);
+  // Note: This pattern may result in non-coalesced access depend on how it executed.
+  detail::AccessDerivatives(derivatives_ptr, e, q, rule, num_elements) = get_gradient(qf_output);
+}
+
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type, typename du_elem_type,
+          typename dr_elem_type>
+SERAC_HOST_DEVICE void gradient_quadrature(int e, int q, du_elem_type& du_elem, dr_elem_type& dr_elem,
```
jamiebramwell, line 192: needs doxygen.

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -0,0 +1,446 @@
...
+  [[maybe_unused]] static constexpr auto I = Identity<dim>();
+
+  using test_space  = decltype(test);
+  using trial_space = decltype(trial);
+
+  Functional<test_space(trial_space), gpu_policy> residual(&fespace, &fespace);
+
+  residual.AddDomainIntegral(Dimension<dim>{}, elastic_qfunction<dim>{}, mesh);
+
+  // Functional<test_space(trial_space)> residual_cpu(&fespace, &fespace);
```
jamiebramwell: dead code.

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -0,0 +1,446 @@
...
+    std::cout << "||g2||: " << g2.Norml2() << std::endl;
+    std::cout << "||g1-g2||/||g1||: " << mfem::Vector(g1 - g2).Norml2() / g1.Norml2() << std::endl;
+  }
+  EXPECT_NEAR(0., mfem::Vector(g1 - g2).Norml2() / g1.Norml2(), 1.e-13);
+
+  serac::profiling::terminateCaliper();
+}
+
+TEST(thermal, 2D_linear) { functional_test(H1<1>{}, H1<1>{}, Dimension<2>{}); };
+// TEST(thermal, 2D_quadratic) { functional_test(H1<2>{}, H1<2>{}, Dimension<2>{}); }
```
jamiebramwell: I thought we had a work around so these could all be run from this file?

src/serac/physics/utilities/functional/tests/mfem_issue_reproducer.cu
```diff
@@ -0,0 +1,90 @@
...
+int num_procs, myid;
+int nsamples = 1;  // because mfem doesn't take in unsigned int
+
+std::unique_ptr<mfem::ParMesh> mesh2D;
+std::unique_ptr<mfem::ParMesh> mesh3D;
+
+std::map<int, std::string> meshfiles = {{2, SERAC_REPO_DIR "/data/meshes/star.mesh"},
+                                        {3, SERAC_REPO_DIR "/data/meshes/beam-hex.mesh"}};
+
+void some_parametrized_test(int p, int dim)
```
jamiebramwell, line 26: Not part of this PR, but this should be upstreamed to MFEM.

jamiebramwell (commented):
Looks good! I think the only big thing left for me is doxygen-ing the cuda files.

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -166,42 +92,36 @@ void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type*
 
     // for each quadrature point in the element
     for (int q = 0; q < static_cast<int>(rule.size()); q++) {
-      // get the position of this quadrature point in the parent and physical space,
```
white238, line 169: You removed a comment here, is it no longer needed? valid? was this a mistake?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -166,42 +92,36 @@ void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type*
...
-      // integrate qf_output against test space shape functions / gradients
-      // to get element residual contributions
-      r_elem += Postprocess<test_element>(get_value(qf_output), xi, J_q) * dx;
-
-      // here, we store the derivative of the q-function w.r.t. its input arguments
-      //
-      // this will be used by other kernels to evaluate gradients / adjoints / directional derivatives
-      derivatives_ptr[e * int(rule.size()) + q] = get_gradient(qf_output);
+      // eval_quadrature is a SERAC_HOST_DEVICE quadrature point calculation
+      if constexpr (std::is_same_v<qpt_data_type, void>) {
```
white238, line 100: It might be helpful to add a comment why you are checking for a void type here.

src/serac/physics/utilities/functional/integral_utilities.hpp
```diff
@@ -294,4 +323,27 @@ using trial_space_t = typename detail::get_trial_space<spaces>::type;
...
                                                     Geometry::Hexahedron};
 
+/**
+ * @brief create shared_ptr to an array of `n` values of type `T`, either on the host or device
+ * @tparam T the type of the value to be stored in the array
+ * @tparam execution_policy the memory space where the data lives
+ * @param n how many entries to allocate in the array
+ */
+template <typename T, typename execution_policy>
+std::shared_ptr<T> make_shared_array(std::size_t n)
```
white238: Is this a good place for this? Seems generic...

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -591,19 +494,42 @@ class DomainIntegral {
...
+
+      gradient_mat_ = [=](mfem::Vector& K_e) {
+        gradient_matrix_kernel<geometry, test_space, trial_space, Q>(K_e, qf_derivatives.get(), J_, num_elements);
+      };
+    }
+
+#if defined(__CUDACC__)
+    if constexpr (std::is_same_v<execution_policy, serac::gpu_policy>) {
+      evaluation_ = [=](const mfem::Vector& U, mfem::Vector& R) {
+        serac::detail::ThreadExecutionConfiguration exec_config{.blocksize = 128};
```
white238: Please pull this magic number into a define with a reason why you chose it if possible

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,259 @@
...
+
+namespace detail {
+
+/**
+ * @brief utility method to display last cuda error message
+ *
+ * @param[in] o The output stream to post success or CUDA error messages
+ * @param[in] success_string A string to print if there are no CUDA error messages
+ */
+inline void displayLastCUDAErrorMessage(std::ostream& o, const char* success_string = "")
```
white238: Can this generic function go in infrastructure?

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,259 @@
...
+{
+  using test_element          = finite_element<g, test>;
+  using trial_element         = finite_element<g, trial>;
+  using element_residual_type = typename trial_element::residual_type;
+  static constexpr auto rule  = GaussQuadratureRule<g, Q>();
+
+  // for each element in the domain
+  const int grid_stride = blockDim.x * gridDim.x;
+  for (int e = blockIdx.x * blockDim.x + threadIdx.x; e < num_elements; e += grid_stride) {
+    // get the DOF values for this particular element
```
white238, line 82: A+ comments on this

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,259 @@
...
+
+    // once we've finished the element integration loop, write our element residuals
+    // out to memory, to be later assembled into global residuals by mfem
+    detail::Add(r, r_elem, e);
+  }  // quadrature x element loop
+}
+
+template <Geometry g, typename test, typename trial, int Q, serac::detail::ThreadExecutionPolicy policy,
+          typename derivatives_type, typename lambda>
+void evaluation_kernel_cuda(serac::detail::ThreadExecutionConfiguration config, const mfem::Vector& U, mfem::Vector& R,
```
white238: the inconsistency of naming on these function is slightly jarring... 

evaluation_kernal_cuda
eval_cuda_quadrature
gradient_cuda_element
gradient_kernel_cuda

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
```diff
@@ -0,0 +1,259 @@
...
+    gradient_quadrature<g, test, trial, Q, derivatives_type>(e, q, du_elem, dr_elem, derivatives_ptr, J, num_elements);
+
+    // once we've finished the element integration loop, write our element residuals
+    // out to memory, to be later assembled into global residuals by mfem
+    detail::Add(dr, dr_elem, e);
+  }
+}
+
+template <Geometry g, typename test, typename trial, int Q, serac::detail::ThreadExecutionPolicy policy,
+          typename derivatives_type>
```
white238, line 396: These long templates need to be in a very specific order to ease use.  should `derivatives_types` go right after `Q` like the other functions?

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,163 @@
...
+ * In the future, the user will be able to override these defaults
+ * to omit unused components (e.g. specify that they only need the gradient)
+ *
+ * @param[in] u The DOF values for the element
+ * @param[in] xi The position of the quadrature point in reference space
+ * @param[in] J The Jacobian of the element transformation at the quadrature point
+ * @tparam element_type The type of the element (used to determine the family)
+ */
+template <typename element_type, typename T, int dim>
+SERAC_HOST_DEVICE auto Preprocess(T u, const tensor<double, dim> xi, const tensor<double, dim, dim> J)
```
white238, line 26: Why are `Preprocess` and `Postprocess` capitalized?

src/serac/physics/utilities/functional/domain_integral_shared.hpp
```diff
@@ -0,0 +1,163 @@
...
+
+  // integrate qf_output against test space shape functions / gradients
+  // to get element residual contributions
+  r_elem += Postprocess<test_element>(get_value(qf_output), xi, J_q) * dx;
+
+  // here, we store the derivative of the q-function w.r.t. its input arguments
+  //
+  // this will be used by other kernels to evaluate gradients / adjoints / directional derivatives
+
+  // detail::AccessDerivatives(derivatives_ptr, e, q, rule, num_elements) = get_gradient(qf_output);
```
white238: why is this here?

src/serac/physics/utilities/functional/tests/CMakeLists.txt
white238: Can you log this as as issue so its not lost?

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -0,0 +1,446 @@
...
+// TEST(elasticity, 3D_linear) { functional_test(H1<1, 3>{}, H1<1, 3>{}, Dimension<3>{}); }
+// TEST(elasticity, 3D_quadratic) { functional_test(H1<2, 3>{}, H1<2, 3>{}, Dimension<3>{}); }
+// TEST(elasticity, 3D_cubic) { functional_test(H1<3, 3>{}, H1<3, 3>{}, Dimension<3>{}); }
+
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+
+  auto [num_procs, myid] = serac::initialize(argc, argv);
+
```
white238, line 423: You need to initialize a SLIC logger otherwise none of our messages will show.

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -0,0 +1,446 @@
...
+  args.AddOption(&serial_refinement, "-r", "--ref", "");
+  args.AddOption(&parallel_refinement, "-pr", "--pref", "");
+  args.AddOption(&nsamples, "-n", "--n-samples", "Samples per test");
+
+  args.Parse();
+  if (!args.Good()) {
+    if (myid == 0) {
+      args.PrintUsage(std::cout);
+    }
+    MPI_Finalize();
```
white238: This should be used here:

`serac::exitGracefully(true);`

src/serac/physics/utilities/functional/tests/tensor_unit_tests_cuda.cu
```diff
@@ -123,7 +125,7 @@ void navier_stokes_tests()
 
 int main()
 {
-  basic_tensor_tests<<<1,1>>>();
-  //elasticity_tests();
-  //navier_stokes_tests();
+  basic_tensor_tests<<<1, 1>>>();
+  // elasticity_tests();
```
white238: Please make a TODO why these are turned off...

src/serac/physics/utilities/functional/tuple.hpp
```diff
@@ -0,0 +1,408 @@
...
+{
+    return tuple<unwrap_decay_t<Types>...>(std::forward<Types>(args)...);
+}
+#endif
+
+}  // namespace serac
+
+// namespace std {
+//
+//  template < size_t i, typename ... T >
```
white238: dead code...

src/serac/physics/utilities/functional/tuple.hpp
```diff
@@ -0,0 +1,408 @@
+#pragma once
+
+#include <utility>
+
+#include "serac/infrastructure/accelerator.hpp"
+
+namespace serac {
+
+template <typename... T>
```
white238, line 27: This file needs documentation and a comment explaining why we wrote our own tuple....

src/serac/physics/utilities/functional/tuple_arithmetic.hpp
```diff
@@ -120,73 +79,74 @@ constexpr auto make_dual_helper(double arg, std::integer_sequence<int, i...>)
...
+                                               std::integer_sequence<int, i...>)
 {
-  return (chain_rule(std::get<i>(df_dx), std::get<i>(dx)) + ...);
+  return (chain_rule(serac::get<i>(df_dx), serac::get<i>(dx)) + ...);
 }
 
 // chain rule between a tuple-of-tuples, and another tuple, kind of like a "matrix vector product"
 // effectively equivalent to
 //
 // clang-format off
```
white238: Why is this here? was clang-format actually formatting a comment?

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
samuelpmishLLNL: I agree that the execution policy should be last, or removed completely (in the short term)

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
samuelpmishLLNL: I discussed this with Jonathan last week, our plan was to make them into overloads of the 3 fundamental kernels
```cpp
evaluation< execution_space >
directional_derivative< execution_space >
gradient_matrix< execution_space >
```
for consistency

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
samuelpmishLLNL: probably in accelerator.hpp / accelerator.cpp?

src/serac/physics/utilities/functional/tuple.hpp
samuelpmishLLNL:
```suggestion
// This is a class that mimics most of std::tuple's interface, except that it is usable in CUDA kernels and 
// admits some arithmetic operator overloads
template <typename... T>
```

src/serac/physics/utilities/functional/domain_integral_shared.hpp
samuelpmishLLNL:
```suggestion
```

src/serac/physics/utilities/functional/domain_integral_shared.hpp
samuelpmishLLNL: In this PR, they're capitalized for consistency with the earlier implementation (which should have been lowercase)

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
jamiebramwell: I say leave this for now and fix it in a follow up PR. I'll add a github issue.

src/serac/physics/utilities/functional/domain_integral_shared.hpp
jamiebramwell: I assumed this was because this will ideally be upstreamed to MFEM, so we're following their convention.

src/serac/physics/utilities/functional/domain_integral_shared.hpp
jamiebramwell: Were the resolved comments actually resolved? I don't see the changes.

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
jonwong12: Oops. I did not mean to commit, but I was trying to verify that the reproducer is still necessary. There's still a cupointer issue that corresponds to the reproducer in the test directory.

src/serac/physics/utilities/functional/domain_integral.hpp
jonwong12: Let me add some documentation, but I think the execution configuration launch needs to be refactored anyways, so I'm hesitant to add a define since this should get addressed soon anyways.

src/serac/physics/utilities/functional/domain_integral_shared.hpp
samuelpmishLLNL: I think it's just an oversight on my initial implementation

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
jonwong12: Sorry the naming of the policy type is confusing. I've changed it to be more reflective of what it actually is. It's a non-deduced value.

src/serac/physics/utilities/functional/domain_integral.hpp
jonwong12: I'll put it into the combined host_device function.

src/serac/physics/utilities/functional/domain_integral.hpp
jamiebramwell: This still needs a SLIC error if `__CUDACC__` is not defined and `serac::gpu_policy` is selected.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
jamiebramwell: Is there a reason to keep `derivatives_type` after the thread parallelization strategy policy? I agree it should be next to `Q`.

src/serac/physics/utilities/functional/domain_integral_shared.hpp
jamiebramwell: We should put all of the `Functional` stuff in an `mfem_extensions` namespace to make this clearer.

src/serac/physics/utilities/functional/domain_integral_cuda.cuh
jonwong12: Sorry it's a little unclear, but at the moment, `derivatives_type` is the first (and only) deduced type, while the parallelization strategy is explicitly supplied. So `Geometry` to the parallelization enum are all specified, but the `derivatives_type` is deduced.

src/serac/physics/utilities/functional/domain_integral.hpp
jonwong12: I don't know why github is not showing the correct code, but I think I've made the necessary changes https://github.com/LLNL/serac/blob/f76afe09c288f929f7f4abe16ac35009bc6f67a9/src/serac/physics/utilities/functional/domain_integral.hpp#L519. I put the `__CUDACC__` guard around the ` if constexpr(gpu_policy)` and I guarded `serac::gpu_policy`. This is related to the naming convention for the kernel. It should hopefully get resolved with future improvements to the naming, and execution policy stuff. Then we won't need to put the runtime error since it should be a compile time error.

src/serac/infrastructure/accelerator.hpp
```diff
@@ -39,4 +82,49 @@ void initializeDevice();
...
+#endif
+
+/**
+ * @brief create shared_ptr to an array of `n` values of type `T`, either on the host or device
+ * @tparam T the type of the value to be stored in the array
+ * @tparam execution_policy the memory space where the data lives
+ * @param n how many entries to allocate in the array
+ */
+template <typename T, typename execution_policy>
+std::shared_ptr<T> make_shared_array(std::size_t n)
```
joshessman-llnl: Can this be `shared_ptr<T[]>`?  This makes it possible to avoid custom deleters for the `cpu_policy` case - `operator delete[]` gets used per https://eel.is/c++draft/util.smartptr.shared.const#6

src/serac/physics/utilities/functional/tests/tuple_arithmetic_unit_tests.cpp
```diff
@@ -4,7 +4,9 @@
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
 
+#include "../tuple.hpp"
```
joshessman-llnl: Should these be relative to the root of the repository?

samuelpmishLLNL: good catch, having these relative includes is a bad practice (my fault!)

src/serac/infrastructure/accelerator.hpp
```diff
@@ -56,10 +56,12 @@ struct cpu_policy {
 };
 
 /**
- * @brief tag type for signaling that calculations that should be performed on the CPU
+ * @brief tag type for signaling that calculations that should be performed on the GPU
  */
+#if defined(__CUDACC__)
 struct gpu_policy {
 };
+#endif
```
samuelpmishLLNL: I don't believe we want this type to be conditionally defined.

src/serac/infrastructure/accelerator.hpp
samuelpmishLLNL: I'm not sure how to implement the CUDA case if we change the return type to 
`std::shared_ptr<T[]>`

samuelpmishLLNL: after playing with it for a few minutes, this does seem to compile (I am not too familiar with the distinctions between `T[]` and `T *`):
```cpp
template <typename T, typename execution_policy>                                    
std::shared_ptr<T[]> make_shared_array(std::size_t n) {                             
                                                                                    
  if constexpr (std::is_same_v<execution_policy, serac::cpu_policy>) {                     
    return std::shared_ptr<T[]>(new T[n]);                                          
  }                                                                                 
                                                                                    
  #if defined(__CUDACC__)                                                           
  if constexpr (std::is_same_v<execution_policy, serac::gpu_policy>) {                     
    T * data;                                                                       
    cudaMalloc(&data, sizeof(T) * n);                                               
    auto deleter = [](T* ptr) { cudaFree(ptr); };                                   
    return std::shared_ptr<T[]>(data, deleter);                                     
  }                                                                                 
  #endif                                                                            
}    
```
is it reasonable to just pass a `T *` as the first argument to the constructor for `std::shared_ptr<T[]>`?

white238 (approved):

src/serac/physics/utilities/functional/domain_integral.hpp
jamiebramwell: I think adding a TODO or comment clarifying this is good enough and we'll clean it up with the template version.

src/serac/infrastructure/accelerator.hpp
joshessman-llnl: Yes - the fact that it's `T[]` tells the deleter to call `op delete[]`.  This snippet is basically what I was expecting 👍

src/serac/infrastructure/accelerator.hpp
jonwong12: I'll add a documentation note that it's just a temporary guard and we can get rid of it after we make the 
evaluation_kernel<policy>

src/serac/infrastructure/accelerator.hpp
```diff
@@ -59,6 +59,8 @@ struct cpu_policy {
  * @brief tag type for signaling that calculations that should be performed on the GPU
  */
 #if defined(__CUDACC__)
+// TEMPORARY: Add temporary guard so gpu_policy cannot be used when there is no GPU.
```
jamiebramwell, line 62: 👍

samuelpmishLLNL (approved):

src/serac/physics/utilities/functional/domain_integral_shared.hpp
jamiebramwell: @jonwong12 , these functions still need doxygen.

jamiebramwell (approved):
Thank you so much for all of this work! It's a really excited feature.

PR #524, opened 2021-07-26
Fix parallel driver

joshessman-llnl:
Ascent already handles MPI-specific aspects of naming output directories/files, so we shouldn't be including the rank.

The extracts path is global so including the rank causes ascent to try to throw an error which results in an MPI deadlock.


white238 (approved):

white238:
thanks @joshessman-llnl !

samuelpmishLLNL (approved):

PR #525, opened 2021-07-27
Output curve field data

white238:
This outputs a new file, `serac_datacoll_curves.json`, that contains various curve data at each time step.  This is outputting the following values:

* L1 norm
* L2 norm
* L infinity norm
* average
* minimum
* maximum

once per time step, to the root mpi tasks Sidre tree under "serac_curves".  As well as the time step value, t.

At the end of the of the simulation run it outputs this to a json file.

Example output:

```
{
  "t": [0.25, 0.5, 0.75, 1.0],
  "temperature": 
  {
    "l1norms": [10.8455612450063, 10.8769266938811, 10.9262526637677, 10.9890569744431],
    "l2norms": [4.00325571439491, 3.98778843947879, 3.988193145272, 3.99834571962069],
    "linfnorms": [1.99858754555508, 1.99602762948269, 1.99349091868975, 1.9915520450815],
    "avgs": [1.36394528852752, 1.36810524509546, 1.374153349967, 1.38184017005875],
    "mins": [1.00067514812945, 1.00311180843846, 1.00816133998986, 1.01610881694094],
    "maxs": [2.0, 2.0, 2.0, 2.0]
  },
  "velocity": 
  {
    "l1norms": [0.0, 0.0, 0.0, 0.0],
    "l2norms": [0.0, 0.0, 0.0, 0.0],
    "linfnorms": [0.0, 0.0, 0.0, 0.0],
    "avgs": [0.0, 0.0, 0.0, 0.0],
    "mins": [0.0, 0.0, 0.0, 0.0],
    "maxs": [0.0, 0.0, 0.0, 0.0]
  },
  "displacement": 
  {
    "l1norms": [0.0, 0.924962166264643, 1.84283834206585, 2.74688270159081],
    "l2norms": [0.0, 0.4275976012174, 0.851821419489431, 1.26946055313981],
    "linfnorms": [0.0, 0.310353229084647, 0.618487529803454, 0.921952178076032],
    "avgs": [0.0, 0.0391142766414546, 0.0763267241134098, 0.111371682619804],
    "mins": [0.0, -0.0382275916996528, -0.0908270569236879, -0.156722895441363],
    "maxs": [0.0, 0.317086042420179, 0.632902614360887, 0.944383108329756]
  },
  "sidre_group_name": "serac_curves"
}
```

Future improvements:

* toggle which values output in input file
* change language the file is output (is this even something we should encourage?)

### DIFF
--- a/src/serac/infrastructure/initialize.cpp
+++ b/src/serac/infrastructure/initialize.cpp
@@ -6,6 +6,14 @@
 
 #include "serac/infrastructure/initialize.hpp"
 
+#ifdef WIN32
+#include <windows.h>
+#include <tchar.h>
+#else
+#include <unistd.h>
+#include <limits.h>
+#endif
+
 #include "serac/infrastructure/accelerator.hpp"
 #include "serac/infrastructure/logger.hpp"
 #include "serac/infrastructure/profiling.hpp"
@@ -25,6 +33,46 @@ std::pair<int, int> getMPIInfo(MPI_Comm comm)
     SLIC_ERROR("Failed to determine MPI rank");
   }
   return {num_procs, rank};
+}
+
+#define INFO_BUFFER_SIZE 32767
+
+std::string getHostName()
+{
+  std::string hostName = "";
+#ifdef WIN32
+  TCHAR infoBuf[INFO_BUFFER_SIZE];
+  DWORD bufCharCount = INFO_BUFFER_SIZE;
+  bufCharCount       = INFO_BUFFER_SIZE;
+  if (!GetComputerName(infoBuf, &bufCharCount)) {
+    hostName = std::string(infoBuf);
+  }
+#else
+  char infoBuf[HOST_NAME_MAX];
+  if (gethostname(infoBuf, HOST_NAME_MAX) == 0) {
+    hostName = std::string(infoBuf);
+  }
+#endif
+  return hostName;
+}
+
+std::string getUserName()
+{
+  std::string userName = "";
+#ifdef WIN32
+  TCHAR infoBuf[INFO_BUFFER_SIZE];
+  DWORD bufCharCount = INFO_BUFFER_SIZE;
+  bufCharCount       = INFO_BUFFER_SIZE;
+  if (GetUserName(infoBuf, &bufCharCount)) {
+    userName = std::string(infoBuf);
+  }
+#else
+  char infoBuf[LOGIN_NAME_MAX];
+  if (getlogin_r(infoBuf, LOGIN_NAME_MAX) == 0) {
+    userName = std::string(infoBuf);
+  }
+#endif
+  return userName;
 }
 
 std::pair<int, int> initialize(int argc, char* argv[], MPI_Comm comm)
--- a/src/serac/infrastructure/output.cpp
+++ b/src/serac/infrastructure/output.cpp
@@ -33,22 +33,20 @@ std::string file_format_string(const FileFormat file_format)
 }
 }  // namespace detail
 
-void outputCurves(const axom::sidre::DataStore& datastore, const std::string& data_collection_name,
-                  const FileFormat file_format)
+void outputSummary(const axom::sidre::DataStore& datastore, const std::string& data_collection_name,
+                   const FileFormat file_format)
 {
   auto [_, rank] = getMPIInfo();
   if (rank != 0) {
     return;
   }
 
-  if (file_format == FileFormat::HDF5) {
-    // FIXME: remove this error when implemented, do we even want hdf5 curves?
-    SLIC_ERROR_ROOT("hdf5 has not been implemented in outputCurves");
-  }
+  // FIXME: remove this error when implemented, do we even want hdf5 curves?
+  SLIC_ERROR_ROOT_IF(file_format == FileFormat::HDF5, "hdf5 has not been implemented in outputCurves");
   std::string file_format_string = detail::file_format_string(file_format);
 
-  const std::string file_name = fmt::format("{0}_curves.{1}", data_collection_name, file_format_string);
-  datastore.getRoot()->getGroup("serac_curves")->save(file_name, file_format_string);
+  const std::string file_name = fmt::format("{0}_summary.{1}", data_collection_name, file_format_string);
+  datastore.getRoot()->getGroup("serac_summary")->save(file_name, file_format_string);
 }
 
 void outputFields(const axom::sidre::DataStore& datastore, const std::string& data_collection_name, double time,
--- a/src/serac/infrastructure/output.cpp
+++ b/src/serac/infrastructure/output.cpp
@@ -18,6 +18,29 @@
 
 namespace serac::output {
 
+void outputCurves(const axom::sidre::DataStore& datastore, const std::string& data_collection_name,
+                  const Language language)
+{
+  auto [_, rank] = getMPIInfo();
+  if (rank != 0) {
+     return;
+  }
+
+  std::string output_language = "";
+  if (language == Language::JSON) {
+    output_language = "json";
+  } else if (language == Language::HDF5) {
+    // FIXME: remove this error when implemented, do we even want hdf5 curves?
+    SLIC_ERROR_ROOT("hdf5 has not been implemented in outputCurves");
+    output_language = "hdf5";
+  } else if (language == Language::YAML) {
+    output_language = "yaml";
+  }
+
+  const std::string file_name = fmt::format("{0}_curves.{1}", data_collection_name, output_language);
+  datastore.getRoot()->getGroup("serac_curves")->save(file_name, output_language);
+}
+
 void outputFields(const axom::sidre::DataStore& datastore, const std::string& data_collection_name, double time,
                   const Language language)
 {
--- a/src/serac/infrastructure/output.cpp
+++ b/src/serac/infrastructure/output.cpp
@@ -18,31 +18,41 @@
 
 namespace serac::output {
 
+namespace detail {
+std::string file_format_string(const FileFormat file_format)
+{
+  std::string value = "";
+  if (file_format == FileFormat::JSON) {
+    value = "json";
+  } else if (file_format == FileFormat::HDF5) {
+    value = "hdf5";
+  } else if (file_format == FileFormat::YAML) {
+    value = "yaml";
+  }
+  return value;
+}
+}
+
 void outputCurves(const axom::sidre::DataStore& datastore, const std::string& data_collection_name,
-                  const Language language)
+                  const FileFormat file_format)
 {
   auto [_, rank] = getMPIInfo();
   if (rank != 0) {
     return;
   }
 
-  std::string output_language = "";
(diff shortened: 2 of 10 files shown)


src/serac/infrastructure/output.cpp
```diff
@@ -18,6 +18,29 @@
...
+  } else if (language == Language::HDF5) {
+    // FIXME: remove this error when implemented, do we even want hdf5 curves?
+    SLIC_ERROR_ROOT("hdf5 has not been implemented in outputCurves");
+    output_language = "hdf5";
+  } else if (language == Language::YAML) {
+    output_language = "yaml";
+  }
+
+  const std::string file_name = fmt::format("{0}_curves.{1}", data_collection_name, output_language);
+  datastore.getRoot()->getGroup("serac_curves")->save(file_name, output_language);
```
samuelpmishLLNL: hmm, I don't really think of JSON or HDF5 as "languages", is this `Language` enum defined in serac or somewhere out of our control?

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -61,6 +65,28 @@ FiniteElementState& FiniteElementState::operator=(const double value)
   return *this;
 }
 
+double avg(const FiniteElementState& state)
+{
+  auto [count, _] = getMPIInfo(state.comm());
+  double sum, local_avg = state.trueVec().Sum() / state.trueVec().Size();
+  MPI_Allreduce(&local_avg, &sum, 1, MPI_DOUBLE, MPI_SUM, state.comm());
+  return sum / count;
+}
```
samuelpmishLLNL: Taking the average of the individual processor averages doesn't give the global average, in general. For example the vector {1, 2, 3, 4, 5} has a mean of 3, but if it is partitioned across two processors {1, 2} and {3, 4, 5}, averaging means of those two processors (1.5 and 4, respectively) gives 2.75 rather than 3.

I think the code below gets the averaging right (conceptually), but I'm not very proficient with MPI, so someone with more experience might want to double check these calls.
```cpp
double global_sum, local_sum = state.trueVec().Sum();
int global_size, local_size = state.trueVec().Size();
MPI_Allreduce(&local_sum, &global_sum, 1, MPI_DOUBLE, MPI_SUM, state.comm());
MPI_Allreduce(&local_size, &global_size, 1, MPI_INT, MPI_SUM, state.comm());
return global_sum / global_size;
```

src/serac/physics/base_physics.cpp
```diff
@@ -134,6 +135,114 @@ void BasePhysics::outputState() const
...
+    min_value      = min(state);
+
+    // Don't save curves on anything other than root node
+    if (rank == 0) {
+      // Group for each Finite Element State
+      axom::sidre::Group* state_group = curves_group->getGroup(state.name());
+
+      axom::sidre::View*         l1norms_view = state_group->getView("l1norms");
+      axom::sidre::Array<double> l1norms(l1norms_view);
+      l1norms.append(l1norm_value);
```
samuelpmishLLNL, line 233: This is not a criticism of this PR, but as someone who is not familiar with sidre, their Array/View semantics seem counterintuitive.

- typically, a "View" implies a type with reference semantics, so `getView()` returning a _pointer_ (which also has reference semantics) to a view feels either redundant or indicative that `axom::side::View` doesn't actually behave like a reference. Or perhaps `axom::sidre::View` is the base class of several kinds of view objects, and that's why it needs to be a pointer?

- Arrays typically have value semantics (construction from a reference implies copying), but it appears that is not happening here, or the code would do nothing. Do `axom::sidre::Array`s not own their data?

For example, following the same pattern (get a view, construct an value-type, and modify the value) using `std::string` and `std::string_view` from the standard library gives completely different behavior:
https://godbolt.org/z/Po8aTsesj

src/serac/infrastructure/output.cpp
white238: That enum is ours (we have a couple that slightly overlap), would you feel better out `Format`?  They are lumped in under language (in the most broadest sense of the CS word) in many places, well at least JSON/YAML.

src/serac/physics/utilities/finite_element_state.cpp
white238: This is a good point. I will change it.  Also global/local is more clear I will update all the functions.

src/serac/infrastructure/output.cpp
samuelpmishLLNL: For this particular function, it seems like `FileExtension`, `FileFormat` or similar would be natural names, but I don't know if those would make sense for other places the enum gets used.

src/serac/physics/base_physics.cpp
white238: just as a bit of knowledge here is the description of how Sidre is laid out:

* Datastore. The main access point to data managed by Sidre; it contains a collection of Buffers, a collection of default Attributes, and a tree structure of Groups.
* Buffer. Describes and holds a chunk of data in memory owned by Sidre.
* Group. Defines a tree structure like a filesystem, where Groups are like folders and Views are like files.
* View. Describes a conceptual layout of data in memory. Each View has a collection of its explicitly-set Attributes.
* Attribute. Provides an item of metadata describing a View.

I agree the Array has some confusing semantics but @joshessman-llnl is actively improving.

jonwong12:
This might be a dumb question, but can you tell from the json output what the configuration of the MPI ranks is and maybe machine information? I assume sidre has some of that information, I'm just not familiar with sidre to know if you can get the information easily as part of a post-processing or analysis step.

src/serac/infrastructure/output.cpp
white238: I'll review and i think `FileFormat` will work in all places.  I will swap it shortly

white238:
> 
> 
> This might be a dumb question, but can you tell from the json output what the configuration of the MPI ranks is and maybe machine information? I assume sidre has some of that information, I'm just not familiar with sidre to know if you can get the information easily as part of a post-processing or analysis step.

I like it.. i'll add some run information.

src/serac/infrastructure/initialize.cpp
```diff
@@ -27,6 +35,46 @@ std::pair<int, int> getMPIInfo(MPI_Comm comm)
   return {num_procs, rank};
 }
 
+#define INFO_BUFFER_SIZE 32767
+
+std::string getHostName()
```
white238, line 40: This is temporarily here until I move it to Axom and then Axom is updated

jamiebramwell: I was about to complain about the gross string handling, but it makes sense if the plan is to upstream it.

src/serac/infrastructure/output.cpp
```diff
@@ -18,8 +18,39 @@
...
+
+void outputSummary(const axom::sidre::DataStore& datastore, const std::string& data_collection_name,
+                   const FileFormat file_format)
+{
+  auto [_, rank] = getMPIInfo();
+  if (rank != 0) {
+    return;
+  }
+
+  // FIXME: remove this error when implemented, do we even want hdf5 curves?
```
jamiebramwell: We don't want HDF5 curves.

src/serac/physics/utilities/finite_element_state.cpp
```diff
@@ -61,6 +65,29 @@ FiniteElementState& FiniteElementState::operator=(const double value)
   return *this;
 }
 
+double avg(const FiniteElementState& state)
+{
+  double global_sum, local_sum   = state.trueVec().Sum();
```
jamiebramwell: Style quibble: I don't like multiple declarations on a single line. However, it passes style check so I can't complain too much.

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -158,6 +158,8 @@ class FiniteElementState {
    * Returns a non-owning reference to the vector of true DOFs
    */
   mfem::HypreParVector& trueVec() { return true_vec_; }
+  /// \overload
+  const mfem::HypreParVector& trueVec() const { return true_vec_; }
```
jamiebramwell, line 162: 👍

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -255,6 +257,32 @@ class FiniteElementState {
   std::string                                       name_ = "";
 };
 
+// FIXME: Should these go somewhere else?
```
jamiebramwell: I think it makes sense to keep these utility functions in the same file as the class they support.

jamiebramwell (approved):
This will be super useful!

joshessman-llnl (approved):

src/serac/infrastructure/output.cpp
white238: it makes even less sense with the other summary info.

src/serac/physics/utilities/finite_element_state.cpp
white238: I don't sure care so I'll go with what makes @jamiebramwell happy...

white238:
New updated summary file:

```
{
  "user_name": "white238",
  "host_name": "rzgenie23",
  "mpi_rank_count": 2,
  "curves": 
  {
    "t": [0.25, 0.5, 0.75, 1.0],
    "temperature": 
    {
      "l1norms": [10.8455612450063, 10.8769266990901, 10.9262526637545, 10.9890569749474],
      "l2norms": [4.0032557143949, 3.98778844023623, 3.98819314519665, 3.99834571964472],
      "linfnorms": [1.99858754555508, 1.99602762948269, 1.99349091889569, 1.99155204517398],
      "avgs": [1.36394528852752, 1.36810524509546, 1.37415334996515, 1.38184017007846],
      "mins": [1.00067514812945, 1.00311180843846, 1.00816133994727, 1.01610881693124],
      "maxs": [2.0, 2.0, 2.0, 2.0]
    },
    "velocity": 
    {
      "l1norms": [0.0, 0.0, 0.0, 0.0],
      "l2norms": [0.0, 0.0, 0.0, 0.0],
      "linfnorms": [0.0, 0.0, 0.0, 0.0],
      "avgs": [0.0, 0.0, 0.0, 0.0],
      "mins": [0.0, 0.0, 0.0, 0.0],
      "maxs": [0.0, 0.0, 0.0, 0.0]
    },
    "displacement": 
    {
      "l1norms": [0.0, 0.924966469035003, 1.84283500859848, 2.74688290675836],
      "l2norms": [0.0, 0.427599659305674, 0.851819882122632, 1.26946066146415],
      "linfnorms": [0.0, 0.310354814771331, 0.618486361369973, 0.921952252373591],
      "avgs": [0.0, 0.0391152700496381, 0.0763268640867222, 0.111371752822853],
      "mins": [0.0, -0.0382284827086096, -0.0908270386551352, -0.156722996902398],
      "maxs": [0.0, 0.317087633095987, 0.632901408765298, 0.944383176995519]
    }
  },
  "sidre_group_name": "serac_summary"
}
```

PR #527, opened 2021-07-28
Enable all output to go to a directory

white238:
This allows all output to be written/read to a user specified output directory via a new command line option `-o <dir>` or `--output_directory <dir>`.

This includes loading restarts as well.  So users will have to use the same output directory command line option if they want to restart.

Things that were moved:

- Inlet docs + input file value file
- StateManager state files
- Summary file
- Field data files

Note: I removed the SLIC logging stream that wrote to a file for two reasons:

- it's not a full capture of all output to the screen because not all things are going through SLIC
- it was created pre command line options and therefore slightly complicated to fix though doable.

### DIFF
--- a/src/docs/conf.py
+++ b/src/docs/conf.py
@@ -167,9 +167,10 @@ def setup(app):
 # using the given strftime format.
 #html_last_updated_fmt = '%b %d, %Y'
 
-# If true, SmartyPants will be used to convert quotes and dashes to
-# typographically correct entities.
-#html_use_smartypants = True
+# If true, the Docutils Smart Quotes transform, originally based on SmartyPants
+# (limited to English) and currently applying to many languages, will be used to convert
+# quotes and dashes to typographically correct entities. Default: True.
+smartquotes = False
 
 # Custom sidebar templates, maps document names to template names.
 #html_sidebars = {}
--- a/src/docs/sphinx/user_guide/command_line_options.rst
+++ b/src/docs/sphinx/user_guide/command_line_options.rst
@@ -1,0 +1,44 @@
+.. ## Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+.. ## other Serac Project Developers. See the top-level COPYRIGHT file for details.
+.. ##
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+
+====================
+Command Line Options
+====================
+
+Below is the documentation for Serac's command line options:
+
+.. list-table:: Options
+   :widths: 25 25 25 25
+   :header-rows: 1
+   :stub-columns: 1
+
+   * - Long form
+     - Short form
+     - Variable Type
+     - Description
+   * - --help
+     - -h
+     - N/A
+     - Print this help message and exit
+   * - --input-file
+     - -i
+     - Path
+     - Input file to use
+   * - --restart-cycle
+     - -c
+     - Integer
+     - Cycle to restart from
+   * - --create-input-file-docs
+     - -d
+     - N/A
+     - Writes Sphinx documentation for input file, then exits
+   * - --output-directory
+     - -o
+     - Path
+     - Directory to put outputted files
+   * - --output-fields
+     - 
+     - N/A
+     - Writes field data to file system
--- a/src/docs/sphinx/user_guide/command_line_options.rst
+++ b/src/docs/sphinx/user_guide/command_line_options.rst
@@ -12,7 +12,6 @@ Below is the documentation for Serac's command line options:
 .. list-table:: Options
    :widths: 25 25 25 25
    :header-rows: 1
-   :stub-columns: 1
 
    * - Long form
      - Short form
--- a/src/drivers/serac.cpp
+++ b/src/drivers/serac.cpp
@@ -94,7 +94,7 @@ int main(int argc, char* argv[])
 
   // Output directory
   std::string output_directory = "";
-  search          = cli_opts.find("output_directory");
+  search                       = cli_opts.find("output_directory");
   if (search != cli_opts.end()) {
     output_directory = search->second;
   }
--- a/src/drivers/serac.cpp
+++ b/src/drivers/serac.cpp
@@ -117,7 +117,7 @@ int main(int argc, char* argv[])
 
   // Intialize MFEMSidreDataCollection
   // If restart_cycle is non-empty, then this is a restart run and the data will be loaded here
-  serac::StateManager::initialize(datastore, "serac", restart_cycle);
+  serac::StateManager::initialize(datastore, "serac", output_directory, restart_cycle);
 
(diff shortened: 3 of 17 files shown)


src/drivers/serac.cpp
```diff
@@ -92,6 +92,15 @@ int main(int argc, char* argv[])
     input_file_path = search->second;
   }
 
+  // Output directory
```
jamiebramwell: I would add a comment about what is stored in this directory as potential users will probably look at this as a template.

src/serac/infrastructure/output.cpp
```diff
@@ -68,7 +70,9 @@ void outputFields(const axom::sidre::DataStore& datastore, const std::string& da
   conduit::Node extracts;
   // "relay" is the Ascents Extract type for saving data
   extracts["e1/type"]            = "relay";
-  extracts["e1/params/path"]     = fmt::format("{}_fields.{}", data_collection_name, file_format_string);
+  const std::string file_name    = fmt::format("{}_fields.{}", data_collection_name, file_format_string);
```
jamiebramwell, line 73: Now that we're saving histories directly to sidre, are these ascent extracts still useful?

src/serac/physics/base_physics.cpp
```diff
@@ -98,14 +100,17 @@ void BasePhysics::initializeOutput(const serac::OutputType output_type, const st
...
+void BasePhysics::outputState(const std::string output_directory) const
 {
   switch (output_type_) {
     case serac::OutputType::VisIt:
       [[fallthrough]];
     case serac::OutputType::ParaView:
       dc_->SetCycle(cycle_);
       dc_->SetTime(time_);
+      if (!output_directory.empty()) {
+        dc_->SetPrefixPath(output_directory);
```
jamiebramwell: Is this needed every output cycle? I think it would be nice if we could give this directory string in the initialize step and then we don't have to carry it around for every output.

src/serac/infrastructure/cli.cpp
```diff
@@ -29,6 +29,8 @@ std::unordered_map<std::string, std::string> defineAndParse(int argc, char* argv
                "Writes Sphinx documentation for input file, then exits");
   bool output_fields{true};
   app.add_flag("--output-fields,!--no-output-fields", output_fields, "Writes field data to file system.");
+  std::string output_directory;
+  app.add_option("-o, --output_directory", output_directory, "Directory to put outputted files.");
```
joshessman-llnl: Would the current working directory be a good default here?  It would have to be at the user's own risk if they're running multiple instances of serac simultaneously (which is what we ran into with the tests but seems like maybe a rare use case outside of that)

src/serac/infrastructure/cli.cpp
```diff
@@ -58,6 +60,9 @@ std::unordered_map<std::string, std::string> defineAndParse(int argc, char* argv
   if (output_fields) {
     cli_opts.insert({"output_fields", {}});
   }
+  if (output_directory != "") {
+    cli_opts.insert({std::string("output_directory"), output_directory});
```
joshessman-llnl: Did you run into issues just passing the `const char*` here?

src/serac/infrastructure/output.cpp
white238: I'll start with turning them off by default.  Then revisit this in a follow up pr.

src/serac/infrastructure/cli.cpp
white238: Changed to input file "stem" (this means "basename - file extension" per std::filesystem::path::stem) per webex converstation.

src/serac/infrastructure/cli.cpp
white238: no idea why i did that. it's my inherit hatred for `char *`.

src/drivers/serac.cpp
```diff
@@ -87,41 +87,57 @@ int main(int argc, char* argv[])
 
   // Read input file
   std::string input_file_path = "";
-  auto        search          = cli_opts.find("input_file");
+  auto        search          = cli_opts.find("input-file");
   if (search != cli_opts.end()) {
     input_file_path = search->second;
   }
 
+  // Output directory used for all files written to the file system.
```
jamiebramwell, line 95: 👍

jamiebramwell (approved):

src/serac/physics/base_physics.cpp
white238: done.

white238:
Comments should be addressed.  Really ready for review this time.

white238:
Added a new simple sphinx page for command line options here:

https://serac.readthedocs.io/en/bugfix-white238-fix_overlapping_output_files/sphinx/user_guide/command_line_options.html

src/docs/sphinx/user_guide/command_line_options.rst
```diff
@@ -0,0 +1,44 @@
...
+     - Description
+   * - --help
+     - -h
+     - N/A
+     - Print this help message and exit
+   * - --input-file
+     - -i
+     - Path
+     - Input file to use
+   * - --restart-cycle
```
jamiebramwell, line 28: These are rendering as long dashes in html. Is there a way to escape that?

white238: curses! no good deed goes unpunished....

white238: fixed by turning off smart quotes on sphinx.

src/docs/conf.py
```diff
@@ -167,9 +167,10 @@ def setup(app):
 # using the given strftime format.
 #html_last_updated_fmt = '%b %d, %Y'
 
-# If true, SmartyPants will be used to convert quotes and dashes to
-# typographically correct entities.
-#html_use_smartypants = True
+# If true, the Docutils Smart Quotes transform, originally based on SmartyPants
```
white238, line 170: The deleted option was deprecated in Sphinx 1.6

src/serac/infrastructure/cli.cpp
```diff
@@ -45,30 +48,56 @@ std::unordered_map<std::string, std::string> defineAndParse(int argc, char* argv
...
+  cli_opts.insert({"output-directory", output_directory});
+
   return cli_opts;
 }
 
+std::string cliValueToString(std::string value) { return value; }
+
+std::string cliValueToString(bool value) { return value ? "true" : "false"; }
+
+std::string cliValueToString(int value) { return std::to_string(value); }
```
joshessman-llnl, line 79: `namespace detail`?

joshessman-llnl (approved):

src/serac/infrastructure/cli.cpp
white238: done.

PR #528, opened 2021-07-28
Temporary fix to unblock automation

white238:
Still think there is a parallel file write that is causing a lock.  Trying to temporarily turn off this test.  This is a more serious problem that needs to be resolved.  I am running out of steam though and hopefully this unblocks people.


joshessman-llnl (approved):

jonwong12 (approved):

white238:
To clarify: I have a suspicion this is two drivers being run at the same time and writing the same files in the current working directory.  It isn't this specific test causing the problem but the combination of two. i think.

PR #530, opened 2021-07-29
Install Functional's implementation headers to detail directory

### LINKED ISSUE #529: Serac does not install detail headers for `physics/utilities/functional/finite_element.hpp` as of commit `a839f8dc`
An issue that came up with LiDO when installing Smith or Serac into a prefix is that `src/serac/physics/utilities/functional/finite_element.hpp` `#include`s header implementation files satisfying the glob `src/serac/physics/utilities/functional/detail/*.inl` (e.g., `src/serac/physics/utilities/functional/detail/segment_H1.inl`), but none of these header implementation files are installed when Smith or Serac is installed.

Looking at `https://github.com/LLNL/serac/blob/develop/src/serac/physics/utilities/functional/CMakeLists.txt`, it's not obvious to me why this behavior occurs. I can provide access to the installation prefixes on the LC for confirmation, though we may need to add accounts to some groups.

joshessman-llnl:
Resolves #529 (hopefully) - when convenient @goxberry, would you be able to confirm that fixes the issue?  Feel free to reach out if you run into any other issues upgrading your serac version.

### DIFF
--- a/src/serac/physics/utilities/functional/CMakeLists.txt
+++ b/src/serac/physics/utilities/functional/CMakeLists.txt
@@ -17,6 +17,9 @@ set(functional_headers
     polynomials.hpp
     quadrature.hpp
     finite_element.hpp
+    )
+
+set(functional_detail_headers
     detail/hexahedron_H1.inl
     detail/hexahedron_Hcurl.inl
     detail/hexahedron_L2.inl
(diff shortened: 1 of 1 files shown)


samuelpmish (approved):

samuelpmishLLNL (approved):

white238 (approved):

goxberry:
@joshessman-llnl Yes, this PR fixed the issue in #529. Thanks for making this PR so quickly!

PR #536, opened 2021-08-05
functional: add missing headers to CMakeLists.txt

goxberry:
This commit adds some missing headers from the functional directory to
CMakeLists.txt so these headers are installed when Serac is installed
into a prefix. For my own sanity and to avoid missing headers, I
alphabetized the header list I modified.


white238 (approved):

white238:
Thanks @goxberry !

jamiebramwell (approved):
Thanks @goxberry !

PR #537, opened 2021-08-05
docs/index.rst: fix use of possessive "its"

goxberry:
This commit replaces "It's" in the documentation index page with "Its"
because its use in that sentence is possessive -- referring to Serac
-- rather than the contraction "it is".

### DIFF
--- a/src/docs/index.rst
+++ b/src/docs/index.rst
@@ -7,8 +7,8 @@
 Serac
 =======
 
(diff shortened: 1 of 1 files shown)


white238 (approved):
Thanks @goxberry

jamiebramwell (approved):
Yeah, this definitely has my typo-riddled stink on it. Thanks for the catch!

PR #542, opened 2021-08-10
WIP: Add attribute lists to functional boundary integrals

jamiebramwell:
This allows boundary integrals to be defined on a specific set of mesh attributes. Ideally this would be handled by the `GetFaceRestriction` operators, but that would require a fairly major MFEM change. 

This work around calculates the boundary integrals on all external faces, but only accumulates the ones with the correct attribute marking.


jamiebramwell:
Nevermind! This isn't working.

jamiebramwell:
This fundamentally won't work.

PR #545, opened 2021-08-12
Fix the doxygen

jamiebramwell:
This fixes some broken doxygen links and quiets the warnings that have snuck in. 
Built docs [here](https://serac.readthedocs.io/en/docs-bramwell-fix_links/).

### DIFF
--- a/src/serac/physics/utilities/functional/detail/metaprogramming.hpp
+++ b/src/serac/physics/utilities/functional/detail/metaprogramming.hpp
@@ -120,7 +120,7 @@ inline constexpr void for_constexpr(lambda&& f)
 }
 
 /**
- * @brief overload
+ * @overload
  */
 template <int n1, typename lambda>
 void for_loop(lambda f)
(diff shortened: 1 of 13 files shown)


src/serac/physics/utilities/functional/detail/metaprogramming.hpp
```diff
@@ -164,6 +175,9 @@ void for_loop(lambda f)
   }
 }
 
+/**
+ * @brief overload
```
white238: We should use the overload doxygen command:

https://www.doxygen.nl/manual/commands.html#cmdoverload

white238 (approved):

src/serac/physics/utilities/functional/detail/metaprogramming.hpp
jamiebramwell: Yep, this was a typo.

white238: I figured after I saw you use it below but was too tired to go back and edit my comment. Thanks for fixing it.

joshessman-llnl (approved):

PR #553, opened 2021-08-19
Add runtime device memory query

jonwong12:
- Change serac::accelerator::displayLastCUDAErrorMessage to instead use SLIC for reporting.
- Add runtime device memory query to figure out how much memory has been used by MFEM or serac

Added into `tests/functional_comparisons_cuda` as an example.

### DIFF
--- a/src/serac/infrastructure/accelerator.hpp
+++ b/src/serac/infrastructure/accelerator.hpp
@@ -87,16 +87,22 @@ void initializeDevice();
 void terminateDevice();
 
 #if defined(__CUDACC__)
+
 /**
  * @brief Utility method to display last cuda error message
  *
  * @param[in] success_string A string to print if there are no CUDA error messages
+ * @param[in] exit_on_error Exit on CUDA error
  */
-inline void displayLastCUDAErrorMessage(const char* success_string = "")
+inline void displayLastCUDAErrorMessage(const char* success_string = "", bool exit_on_error = false)
 {
   auto error = cudaGetLastError();
   if (error != cudaError::cudaSuccess) {
-    SLIC_ERROR_ROOT(serac::profiling::concat("Last CUDA Error Message :", cudaGetErrorString(error)));
+    if (exit_on_error) {
+      SLIC_ERROR_ROOT(serac::profiling::concat("Last CUDA Error Message :", cudaGetErrorString(error)));
+    } else {
+      SLIC_WARNING_ROOT(serac::profiling::concat("Last CUDA Error Message :", cudaGetErrorString(error)));
+    }
   } else if (strlen(success_string) > 0) {
     SLIC_INFO_ROOT(success_string);
   }
@@ -117,6 +123,17 @@ inline std::tuple<std::size_t, std::size_t> getCUDAMemInfo()
   displayLastCUDAErrorMessage();
   return std::make_tuple(free_memory, total_memory);
 }
+
+/**
+ * @brief returns a string with GPU memory information
+ */
+
+std::string getCUDAMemInfoString()
+{
+  auto [free_memory, total_memory] = getCUDAMemInfo();
+  return fmt::format("Free memory: {} Total_memory: {}", free_memory, total_memory);
+}
+
 #endif
 
 /**
--- a/src/serac/infrastructure/accelerator.hpp
+++ b/src/serac/infrastructure/accelerator.hpp
@@ -41,8 +41,10 @@
 #define SERAC_SUPPRESS_NVCC_HOSTDEVICE_WARNING
 #endif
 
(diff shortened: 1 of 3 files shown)


src/serac/infrastructure/accelerator.hpp
```diff
@@ -88,18 +90,33 @@ void terminateDevice();
...
- * @param[in] o The output stream to post success or CUDA error messages
  * @param[in] success_string A string to print if there are no CUDA error messages
  */
-inline void displayLastCUDAErrorMessage(std::ostream& o, const char* success_string = "")
+inline void displayLastCUDAErrorMessage(const char* success_string = "")
 {
   auto error = cudaGetLastError();
   if (error != cudaError::cudaSuccess) {
-    o << "Last CUDA Error Message :" << cudaGetErrorString(error) << std::endl;
+    SLIC_ERROR_ROOT(serac::profiling::concat("Last CUDA Error Message :", cudaGetErrorString(error)));
```
joshessman-llnl: This will result in program termination (which is not the current behavior) - do we want this to just be a warning that doesn't immediately kill the program?

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -414,6 +426,11 @@ int main(int argc, char* argv[])
 {
   ::testing::InitGoogleTest(&argc, argv);
 
+  cudaDeviceSynchronize();
+  auto [functional_free_memory, functional_total_memory] = serac::accelerator::getCUDAMemInfo();
+  std::cout << "Initial: Free memory:" << functional_free_memory << " Total_memory:" << functional_total_memory
+            << std::endl;
```
joshessman-llnl: Do you think it's worth having an additional convenience function that returns a string so you could do something like:
```cpp
SLIC_INFO("Initial: " << serac::accelerator::getCUDAMemInfoString());
```

src/serac/infrastructure/accelerator.hpp
jonwong12: Ahh. I can see how it might be useful both ways. What is the best SLIC "stream" to use for soft errors?

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
jonwong12: I think that's a good idea. I can add a `<<` overload.

white238: you have INFO, WARNING, ERROR.  Only ERROR halts.

src/serac/infrastructure/accelerator.hpp
white238: you have INFO (for regular messages) , WARNING, ERROR. Only ERROR halts.

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
joshessman-llnl: Just to clarify - this doesn't need to be a `struct` with an `op<<` overload - just a function that returns a string is sufficient.

samuelpmishLLNL (approved):

src/serac/infrastructure/accelerator.hpp
```diff
@@ -85,21 +87,53 @@ void initializeDevice();
...
+
 /**
  * @brief Utility method to display last cuda error message
  *
- * @param[in] o The output stream to post success or CUDA error messages
  * @param[in] success_string A string to print if there are no CUDA error messages
+ * @param[in] exit_on_error Exit on CUDA error
  */
-inline void displayLastCUDAErrorMessage(std::ostream& o, const char* success_string = "")
+inline void displayLastCUDAErrorMessage(const char* success_string = "", bool exit_on_error = false)
```
white238: This doesn't display only errors at this point.  Should be changed to `displayLastCUDAMessage`.

src/serac/physics/utilities/functional/tests/functional_comparisons_cuda.cu
```diff
@@ -193,6 +198,9 @@ void functional_test(H1<p> test, H1<p> trial, Dimension<dim>)
   // Ensure the two methods generate the same result
   EXPECT_NEAR(0., mfem::Vector(g1 - g2).Norml2() / g1.Norml2(), 1.e-14);
 
+  //  auto [functional_free_memory, functional_total_memory] = serac::accelerator::getCUDAMemInfo();
```
white238: remove dead code

white238 (approved):

PR #554, opened 2021-08-19
Add a dual type for linear forms

### LINKED ISSUE #561: Add dual vectors to the state manager for plotting
> 
> 
> > Will you be able to visualize the reaction forces? Being able to visualize those things can be useful, also when you have convergence problems the nodal residuals.
> 
> Good point. I'll see if I can find a way to add this to the data store without making the grid function user-accessible.

I've thought about this, and I think it's possible but it would be a significant amount of work in the `StateManager` class. I propose we file an issue and look at this when it becomes necessary.

_Originally posted by @jamiebramwell in https://github.com/LLNL/serac/issues/554#issuecomment-912807946_

jamiebramwell:
This adds a type dual to `FiniteElementState` that represents a linear functional applied to a specific basis set. This is useful for denoting their topological space to ensure consistency in the function signatures as well as provide extra information when interpolations and projects are needed. Resolves https://github.com/LLNL/serac/issues/561


jamiebramwell:
ping @goxberry @pmm14006 @sethwatts

goxberry:
@jamiebramwell I'm wondering if it might make more sense to have an abstract base class for "finite element state" with two concrete subclasses: one for primal state and one for dual state. I'd think of `FiniteElementState::gridFunc`, `FiniteElementState::gridFuncCoef`, and so on as being appropriate for primal state, but not for dual state, yet `FiniteElementDual` will inherit those operations from `FiniteElementState`. Clearly, though, both would have common operations that should enable extracting underlying `HypreParVector` data, and it would make sense to have conversions from one type to the other via Riesz maps and their inverses. (Depending on what you want to support, move conversions or copy conversions would make sense.)

goxberry:
That all said, I agree with the aim of keeping Serac accessible to engineers, and don't know if the primal/dual distinction would violate that aim, even though it's sort of a generalization of a displacement/force pairing.

goxberry:
To supplement the discussion at yesterday's meeting, a couple papers for background:
* R. Kirby, "From Functional Analysis to Iterative Methods", SIAM Review 52 (2010), pp. 269-293 discusses how the primal-dual pairing can be used as a natural preconditioner for linear systems.
  preprint version: https://cpb-us-w2.wpmucdn.com/sites.baylor.edu/dist/6/32/files/2015/01/iterative-11k9q6j.pdf
  journal version: https://epubs.siam.org/doi/abs/10.1137/070706914
* T. Schwedes, D. A. Ham, S. W. Funke and M. D. Piggott, "Mesh Dependence in PDE-Constrained Optimization" in _Mesh Dependence in PDE-Constrained Optimization_, Mathematics of Planet Earth, D. Crisan and D. Holm, eds., Springer, Cham, Switzerland, 2017, pp. 53-78 discusses how the primal-dual pairing can be used as a natural preconditioner in optimization algorithms.
  preprint version of chapter, listed as "An iteration count estimate for a mesh-dependent steepest descent method based on finite elements and Riesz inner product representation" on arXiv: https://arxiv.org/abs/1606.08069
  book version: https://link.springer.com/book/10.1007/978-3-319-59483-5

The second reference is probably less mathematically demanding than the first.

jamiebramwell:
> 
> 
> @jamiebramwell I'm wondering if it might make more sense to have an abstract base class for "finite element state" with two concrete subclasses: one for primal state and one for dual state. I'd think of `FiniteElementState::gridFunc`, `FiniteElementState::gridFuncCoef`, and so on as being appropriate for primal state, but not for dual state, yet `FiniteElementDual` will inherit those operations from `FiniteElementState`. Clearly, though, both would have common operations that should enable extracting underlying `HypreParVector` data, and it would make sense to have conversions from one type to the other via Riesz maps and their inverses. (Depending on what you want to support, move conversions or copy conversions would make sense.)

Agreed about the dual state not really having a grid function. However, I've run into areas where I really want the ghost dofs on a dual vector, and the grid function is a nice solution for that. I could see having a completely separate class, but then I would be reimplementing a lot of the existing grid function stuff. Also, I'm hesitant to use an abstract base class as this is a fundamental data type and a "standard engineer" user might be confused why it's there. At the end of the day, I could go either way. Any thoughts @white238 @solberg2 @sethwatts @samuelpmishLLNL ?

goxberry:
>  However, I've run into areas where I really want the ghost dofs on a dual vector, and the grid function is a nice solution for that.

As opposed to getting ghost dof information from the `mfem::ParFiniteElementSpace`?

>  I could see having a completely separate class, but then I would be reimplementing a lot of the existing grid function stuff. Also, I'm hesitant to use an abstract base class as this is a fundamental data type and a "standard engineer" user might be confused why it's there. At the end of the day, I could go either way.

I don't have terribly strong opinions here. Another option would be to put all of the dual-related methods into `serac::FiniteElementState` and have a field indicating whether a type is primal or dual, which has its own tradeoffs. I support whatever makes sense to engineers.

sethwatts:
>  I'm hesitant to use an abstract base class as this is a fundamental data type and a "standard engineer" user might be confused why it's there.

I think this is a valid concern, since users without a background FE theory could become confused. However, I would assume that most "standard engineer" users would just be using the solver interfaces to solve their PDEs and then visualize the results. In that setting, I don't think the existence of the dual fields would really come up, nor would the fact that they and primal fields derive from a common abstract base class.

People using the adjoint solvers, writing their own Functionals, etc. I would assume would have enough background to understand the differences between the fields with some documentation (possibly pointing to MFEM's page on the topic). I think it's worth the risk of some confusion to avoid duplicate implementations.

solberg2:
I think it's fine to have this class.

Really this is the place where you store "reactions" and "residuals" and things like that - in electromagnetics this is where you would store the "nodal currents", which you use to construct the circuit model.

Having some examples of it's use in the code would help the "regular engineer" use it.

src/serac/physics/utilities/finite_element_dual.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include "mfem.hpp"
+
+#include "serac/physics/utilities/finite_element_state.hpp"
+
+namespace serac {
+
+/**
+ * @brief Class for encapsulating the topological dual space of a finite element space (i.e. the
+ * space of linear forms) as applied to a specific basis set
+ *
```
solberg2: There should be some examples here to help the reader.

And I think you want to get rid of the word "topological", it's confusing to the reader.  It's just the dual space to the finite element space, e.g. the space of reactions.

solberg2 (approved):

jamiebramwell:
I refactored this to remove the `GridFunction` view of a dual class. All it really needs is the distributed, local (or L-vector) view of the data and the partitioned, true (or T-vector) view of the data. It adds convenience functions attaching the appropriate mesh and finite element space so we can use the prolongation and restriction operators appropriately. Any thoughts @samuelpmishLLNL , @goxberry , @sethwatts , @pmm14006 ?

sethwatts:
> I refactored this to remove the `GridFunction` view of a dual class.

I think this is a valuable update since, per the MFEM documentation, the `GridFunction` is a primal representation. I agree that all we need are access to local and true DOFs.

solberg2:
Will you be able to visualize the reaction forces?  Being able to visualize those things can be useful, also when you have convergence problems the nodal residuals.

jamiebramwell:
> Will you be able to visualize the reaction forces? Being able to visualize those things can be useful, also when you have convergence problems the nodal residuals.

Good point. I'll see if I can find a way to add this to the data store without making the grid function user-accessible.

src/serac/physics/utilities/finite_element_dual.hpp
goxberry: The distinction between algebraic and topological dual spaces is only material in the infinite-dimensional case, and for an engineering audience, I can see the appeal in eliding that distinction.

jamiebramwell:
> 
> 
> > Will you be able to visualize the reaction forces? Being able to visualize those things can be useful, also when you have convergence problems the nodal residuals.
> 
> Good point. I'll see if I can find a way to add this to the data store without making the grid function user-accessible.

I've thought about this, and I think it's possible but it would be a significant amount of work in the `StateManager` class. I propose we file an issue and look at this when it becomes necessary.

solberg2:
> > > Will you be able to visualize the reaction forces? Being able to visualize those things can be useful, also when you have convergence problems the nodal residuals.
> > 
> > 
> > Good point. I'll see if I can find a way to add this to the data store without making the grid function user-accessible.
> 
> I've thought about this, and I think it's possible but it would be a significant amount of work in the `StateManager` class. I propose we file an issue and look at this when it becomes necessary.

Does making a grid function available get rid of the need for work in the State Manager class?  If so maybe I'd put the grid function back in, with requisit Caveat Emptor notices.

goxberry:
> > > > Will you be able to visualize the reaction forces? Being able to visualize those things can be useful, also when you have convergence problems the nodal residuals.
> > > 
> > > 
> > > Good point. I'll see if I can find a way to add this to the data store without making the grid function user-accessible.
> > 
> > 
> > I've thought about this, and I think it's possible but it would be a significant amount of work in the `StateManager` class. I propose we file an issue and look at this when it becomes necessary.
> 
> Does making a grid function available get rid of the need for work in the State Manager class? If so maybe I'd put the grid function back in, with requisite [sic] Caveat Emptor notices.

MFEM's visualization interface -- `mfem::DataCollection` -- requires `mfem::GridFunction` objects. Usually, if one is constructing a load from an `mfem::Coefficient`, this `mfem::Coefficient` is projected onto an `mfem::GridFunction` for plotting. This projection is the Riesz representer of the assembled load in a corresponding `mfem::LinearForm`, so if only the `mfem::LinearForm`-like information is available, I imagine that computing the Riesz representer might be what a user would want, unless a user really does want to plot the entries of an assembled load. But I'm not sure what engineering nomenclature corresponds to "get the Riesz representer" because, despite being an engineer, most of what I know about finite element methods is from math books -- chemical engineering curricula don't typically cover finite element methods (at any level).

jamiebramwell:
After thinking about this, I realized we could do this by using composition without inheritance. This avoids the duplication with the grid function capability but still uses the hardened state manager functionality under the hood. There's only one static cast in the state manager which abuses the fact that the "local dual vector" is really a grid function under the hood. I think this is the best of both worlds. Any thoughts @solberg2 @goxberry ?

jamiebramwell:
If this solution is acceptable, I believe this PR is ready to go.

goxberry:
> After thinking about this, I realized we could do this by using composition without inheritance. This avoids the duplication with the grid function capability but still uses the hardened state manager functionality under the hood. There's only one static cast in the state manager which abuses the fact that the "local dual vector" is really a grid function under the hood. I think this is the best of both worlds. Any thoughts @solberg2 @goxberry ?

I like using composition better than using inheritance in this situation. We have use cases where we'd want the entries of the underlying `ParLinearForm` and cases where we really want the Riesz representer of that `ParLinearForm`. From the perspective of keeping the `FiniteElementDual` API minimal, I think it makes sense to return the former and delegate to the user the responsibility of computing the latter, so your current implementation works for me.

goxberry (approved):

samuelpmishLLNL:
it looks like 90% of the interface in FiniteElementDual is shared with FiniteElementState, so it doesn't seem like composition is helping much:

```cpp
class FiniteElementState {
 public:
  // parts in common with FiniteElementDual
  struct Options {
    int order = 1;
    int vector_dim = 1;
    std::unique_ptr<mfem::FiniteElementCollection> coll = {};
    mfem::Ordering::Type ordering = mfem::Ordering::byVDIM;
    std::string name = "";
    bool alloc_gf = true;
  };

  FiniteElementState(mfem::ParMesh& mesh, Options&& options);
  FiniteElementState(mfem::ParMesh& mesh, mfem::ParGridFunction& gf, const std::string& name = "");
  FiniteElementState(mfem::ParMesh& mesh, mfem::ParFiniteElementSpace& space, const std::string& name = "");

  mfem::ParMesh& mesh();

  mfem::ParFiniteElementSpace& space();
  const mfem::ParFiniteElementSpace& space() const;

  mfem::HypreParVector& trueVec();
  const mfem::HypreParVector& trueVec() const;

  std::string name() const;

  void initializeTrueVec();
  void distributeSharedDofs();

  FiniteElementState& operator=(const double value);

  template <typename Tensor>
  std::unique_ptr<Tensor> createOnSpace()

  // parts different from FiniteElementDual
  mfem::ParGridFunction& gridFunc();
  const mfem::ParGridFunction& gridFunc() const;

  void project(const GeneralCoefficient& coef)
  void project(mfem::Coefficient& coef)
  void project(mfem::VectorCoefficient& coef)

  FiniteElementState(mfem::ParMesh& mesh, FiniteElementState& fe_state, const std::string& name = "");
};


class FiniteElementDual {
 public:
  // parts in common with FiniteElementState
  struct Options {
    int order = 1;
    int vector_dim = 1;
    std::unique_ptr<mfem::FiniteElementCollection> coll = {};
    mfem::Ordering::Type ordering = mfem::Ordering::byVDIM;
    std::string name = "";
    bool alloc_local = true;
  };

  FiniteElementDual(mfem::ParMesh& mesh, Options&& options);
  FiniteElementDual(mfem::ParMesh& mesh, mfem::ParGridFunction& gf, const std::string& name = "")
  FiniteElementDual(mfem::ParMesh& mesh, mfem::ParFiniteElementSpace& space, const std::string& name = "")

  mfem::ParMesh& mesh();

  mfem::ParFiniteElementSpace& space();
  const mfem::ParFiniteElementSpace& space() const 

  mfem::HypreParVector& trueVec();
  const mfem::HypreParVector& trueVec() const;

  std::string name() const;

  void initializeTrueVec();
  void distributeSharedDofs();

  FiniteElementDual& operator=(const double value)

  template <typename Tensor>
  std::unique_ptr<Tensor> createOnSpace()

  // parts different from FiniteElementState
  mfem::Vector& localVec();
  const mfem::Vector& localVec() const;
};
```

as opposed to 

```cpp
struct BaseClass {
  struct Options {
    int order = 1;
    int vector_dim = 1;
    std::unique_ptr<mfem::FiniteElementCollection> coll = {};
    mfem::Ordering::Type ordering = mfem::Ordering::byVDIM;
    std::string name = "";
    bool alloc = true;
  };

  BaseClass(mfem::ParMesh& mesh, Options&& options);
  BaseClass(mfem::ParMesh& mesh, mfem::ParGridFunction& gf, const std::string& name = "")
  BaseClass(mfem::ParMesh& mesh, mfem::ParFiniteElementSpace& space, const std::string& name = "")

  mfem::ParMesh& mesh() 

  mfem::ParFiniteElementSpace& space() 
  const mfem::ParFiniteElementSpace& space() const 

  mfem::HypreParVector& trueVec() 
  const mfem::HypreParVector& trueVec() const 

  std::string name() const 

  void initializeTrueVec() 
  void distributeSharedDofs() 

  FiniteElementDual& operator=(const double value)

  template <typename Tensor>
  std::unique_ptr<Tensor> createOnSpace()
};

struct FiniteElementState : public BaseClass {
  mfem::ParGridFunction& gridFunc();
  const mfem::ParGridFunction& gridFunc() const;

  void project(const GeneralCoefficient& coef);
  void project(mfem::Coefficient& coef);
  void project(mfem::VectorCoefficient& coef);
};

struct FiniteElementDual : public BaseClass {
  mfem::Vector& localVec();
  const mfem::Vector& localVec();
};
```
Note: might we also want `project` on the dual space?

samuelpmishLLNL:
also, if we're exposing members (by non-const reference) like this through methods,
```cpp
  mfem::ParMesh& mesh();

  mfem::ParFiniteElementSpace& space();
  const mfem::ParFiniteElementSpace& space() const 

  mfem::HypreParVector& trueVec();
  const mfem::HypreParVector& trueVec() const;

  std::string name() const;
```
is it different than just making them public member variables and letting the user access them directly?

goxberry:
> Note: might we also want `project` on the dual space?

What would be the argument(s) to such a method? `LinearFormIntegrator`s?

samuelpmishLLNL:
> > Note: might we also want `project` on the dual space?
> 
> What would be the argument(s) to such a method? `LinearFormIntegrator`s?

an mfem "Coefficient" like the other project method, I imagine. I can't think of a use case, but was just curious if there was one.

edit: on second thought, fitting a dual field to a function seems un-physical, so it probably doesn't make sense.

goxberry:
> it looks like 90% of the interface in FiniteElementDual is shared with `FiniteElementState`, so it doesn't seem like composition is helping much

@samuelpmishLLNL For what it's worth, I agree with you from a pure software engineering perspective -- I made a similar comment [earlier in the thread](https://github.com/LLNL/serac/pull/554#issuecomment-902896416). However, from the discussion so far, my impression has been that having a common base class for primal and dual states creates its own set of UI problems, e.g., explaining to an engineer why such a base class exists (see [this earlier comment](https://github.com/LLNL/serac/pull/554#issuecomment-903936231)).

jamiebramwell:
> 
> 
> > it looks like 90% of the interface in FiniteElementDual is shared with `FiniteElementState`, so it doesn't seem like composition is helping much
> 
> @samuelpmishLLNL For what it's worth, I agree with you from a pure software engineering perspective -- I made a similar comment [earlier in the thread](https://github.com/LLNL/serac/pull/554#issuecomment-902896416). However, from the discussion so far, my impression has been that having a common base class for primal and dual states creates its own set of UI problems, e.g., explaining to an engineer why such a base class exists (see [this earlier comment](https://github.com/LLNL/serac/pull/554#issuecomment-903936231)).

Exactly. I can either:

1. Inherit from `FiniteElementState` to encapsulate the dual nature in the type. However, this gives users direct access to the grid functions, which isn't a mathematically correct view of the data.
2. Include the `FiniteElementState` by composition. This allows all of the plotting niceties of the `FiniteElementState` without having direct access to the grid function, but it is a lot of boilerplate code.
3. Do not include `FiniteElementState` at all. This requires a lot of duplication in both the projection and restriction application as well as additional machinery needed to add to plot files. 

This current iteration is option 1, but I can see arguments for the other 2 as well.

samuelpmishLLNL:
> > > it looks like 90% of the interface in FiniteElementDual is shared with `FiniteElementState`, so it doesn't seem like composition is helping much
> > 
> > 
> > @samuelpmishLLNL For what it's worth, I agree with you from a pure software engineering perspective -- I made a similar comment [earlier in the thread](https://github.com/LLNL/serac/pull/554#issuecomment-902896416). However, from the discussion so far, my impression has been that having a common base class for primal and dual states creates its own set of UI problems, e.g., explaining to an engineer why such a base class exists (see [this earlier comment](https://github.com/LLNL/serac/pull/554#issuecomment-903936231)).
> 
> Exactly. I can either:
> 
> 1. Inherit from `FiniteElementState` to encapsulate the dual nature in the type. However, this gives users direct access to the grid functions, which isn't a mathematically correct view of the data.
> 2. Include the `FiniteElementState` by composition. This allows all of the plotting niceties of the `FiniteElementState` without having direct access to the grid function, but it is a lot of boilerplate code.
> 3. Do not include `FiniteElementState` at all. This requires a lot of duplication in both the projection and restriction application as well as additional machinery needed to add to plot files.
> 
> This current iteration is option 1, but I can see arguments for the other 2 as well.

The code I showed above does not have the gridfunction access problem described in point (1), it does not require the boilerplate introduced by forwarding accessors from composition in point (2), and does not have the duplication required by point (3).

jamiebramwell:
Whoops! Typo! It currently has option 2.

jamiebramwell:
Ahh, just looked at the code snippet. The only problem with that approach is it requires an "engineering" user to figure out why there is an abstract base class and following that to look up basic `FiniteElementState` functionality. I agree it is probably the cleanest solution though.

samuelpmishLLNL:
> Ahh, just looked at the code snippet. The only problem with that approach is it requires an "engineering" user to figure out why there is an abstract base class and following that to look up basic `FiniteElementState` functionality. I agree it is probably the cleanest solution though.

The base class is not abstract, it implements its interface (note: there is no `virtual` anywhere in my code snippet). 

I don't really understand the worry about someone asking the question "why is there a base class?". It was my understanding that inheritance was invented to address precisely this kind of problem (this collection of types sure have a lot of stuff in common, how could I avoid rewriting that same code in each one?). Admittedly, the savings is only about 40-50% in our case, since we will likely only ever have 2 derived classes, but that still seems worth it to me.

jamiebramwell:
> 
> 
> > Ahh, just looked at the code snippet. The only problem with that approach is it requires an "engineering" user to figure out why there is an abstract base class and following that to look up basic `FiniteElementState` functionality. I agree it is probably the cleanest solution though.
> 
> The base class is not abstract, it implements its interface (note: there is no `virtual` anywhere in my code snippet).
> 
> I don't really understand the worry about someone asking the question "why is there a base class?". It was my understanding that inheritance was invented to address precisely this kind of problem (this collection of types sure have a lot of stuff in common, how could I avoid rewriting that same code in each one?). Admittedly, the savings is only about 40-50% in our case, since we will likely only ever have 2 derived classes, but that still seems worth it to me.

A fair point. I will implement this and should have it done by mid-afternoon.

jamiebramwell:
I pushed @samuelpmishLLNL 's idea. After implementing it, I agree it's much cleaner and worth putting in inheritance for.

src/serac/physics/utilities/finite_element_dual.hpp
```diff
@@ -0,0 +1,86 @@
...
+      : FiniteElementVector(mesh, gf, name){};
+
+  /**
+   * @brief Minimal constructor for a FiniteElementDual given a finite element space
+   * @param[in] mesh The problem mesh (object does not take ownership)
+   * @param[in] space The space to use for the finite element dual. This space is deep copied into the new FE state
+   * @param[in] name The name of the field
+   */
+  FiniteElementDual(mfem::ParMesh& mesh, mfem::ParFiniteElementSpace& space, const std::string& name = "")
+      : FiniteElementVector(mesh, space, name){};
```
samuelpmishLLNL: I think these might be replaceable by 
`using FiniteElementVector::FiniteElementVector;`

see https://godbolt.org/z/xjh9enTnd

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -45,78 +46,49 @@ inline bool is_vector_valued(const GeneralCoefficient& coef)
...
   /**
-   * Returns the MPI communicator for the state
+   * @brief Minimal constructor for a FiniteElementState given an already-existing state
+   * @param[in] mesh The problem mesh (object does not take ownership)
+   * @param[in] fe_state The state for the new state to copy
+   * @param[in] name The name of the field
    */
-  MPI_Comm comm() const { return detail::retrieve(space_).GetComm(); }
+  FiniteElementState(mfem::ParMesh& mesh, FiniteElementState& fe_state, const std::string& name = "")
+      : FiniteElementVector(mesh, fe_state.gridFunc(), name){};
```
samuelpmishLLNL: [this](https://github.com/LLNL/serac/pull/554/files#r704811693) may apply here too

src/serac/physics/utilities/finite_element_state.hpp
```diff
@@ -193,94 +141,21 @@ class FiniteElementState {
...
-   */
-  void distributeSharedDofs() { detail::retrieve(gf_).SetFromTrueDofs(true_vec_); }
-
   /**
    * @brief Set a finite element state to a constant value
    *
-   * @param value The constant to set the finite element state to
-   * @return The modified finite element state
+   * @param value The constant to set the finite element dual to
+   * @return The modified finite element dual
```
samuelpmishLLNL: these should still be `state`s I think

src/serac/physics/utilities/finite_element_vector.hpp
```diff
@@ -0,0 +1,213 @@
...
+    mfem::Ordering::Type ordering = mfem::Ordering::byVDIM;
+    /**
+     * @brief The name of the field encapsulated by the state object
+     */
+    std::string name = "";
+    /**
+     * @brief Whether the GridFunction should be allocated (and owned by the FEState object)
+     * @note This should only be false for restart-based runs
+     */
+    bool alloc_local = true;
```
samuelpmishLLNL: if the value of this variable is entirely dependent on whether or not it's a restart, it might be clearer to make it something like `bool is_a_restart` (would have to be negated from its current interpretation) or an enum if there are more options

```
// this is a terrible name
enum class SimulationType { Normal, Restart, SomethingElse };
```

src/serac/physics/utilities/finite_element_dual.hpp
jamiebramwell: TIL! Thanks!

src/serac/physics/utilities/finite_element_vector.hpp
samuelpmishLLNL: it would be good to get other people's input on this name though (since I won't pretend to know what it's used for)

jamiebramwell: Yeah, this name is confusing. After digging a bit, I realized "only restart" was also misleading. This is really denoting whether the memory is managed by sidre or not. If it isn't, we prefer to use unique pointers. If it is, they expect a raw new. Would the name `managed_by_sidre` with appropriate comments help?

jamiebramwell: Fixed! Hopefully this makes more sense now.

jamiebramwell:
I think I captured all of @samuelpmishLLNL 's comments. Thank you for the review!

jamiebramwell:
> 
> 
> also, if we're exposing members (by non-const reference) like this through methods,
> 
> ```c++
>   mfem::ParMesh& mesh();
> 
>   mfem::ParFiniteElementSpace& space();
>   const mfem::ParFiniteElementSpace& space() const 
> 
>   mfem::HypreParVector& trueVec();
>   const mfem::HypreParVector& trueVec() const;
> 
>   std::string name() const;
> ```
> 
> is it different than just making them public member variables and letting the user access them directly?

I agree it's boilerplate, but I like having interface functions just in case we change the underlying implementation. Any thoughts @joshessman-llnl @white238 ?

white238:
> 
> 
> > also, if we're exposing members (by non-const reference) like this through methods,
> > ```c++
> >   mfem::ParMesh& mesh();
> > 
> >   mfem::ParFiniteElementSpace& space();
> >   const mfem::ParFiniteElementSpace& space() const 
> > 
> >   mfem::HypreParVector& trueVec();
> >   const mfem::HypreParVector& trueVec() const;
> > 
> >   std::string name() const;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is it different than just making them public member variables and letting the user access them directly?
> 
> I agree it's boilerplate, but I like having interface functions just in case we change the underlying implementation. Any thoughts @joshessman-llnl @white238 ?

I agree with @jamiebramwell .  It also allows you to place a breakpoint/print statement to aid in debugging.

src/serac/physics/solid.cpp
```diff
@@ -454,47 +463,61 @@ mfem::ParLinearForm& Solid::bulkModulusSensitivity(mfem::ParFiniteElementSpace&
 
   // Add a scalar linear form integrator using the shear sensitivity coefficient against the given bulk modulus finite
   // element space
-  if (!bulk_sensitivity_form_ || shear_sensitivity_form_->FESpace() != &bulk_space) {
-    bulk_sensitivity_form_ = std::make_unique<mfem::ParLinearForm>(&bulk_space);
+  if (!bulk_sensitivity_form_ || bulk_space) {
+    SLIC_ERROR_IF(!bulk_space, fmt::format("Finite element space is required for first sensitivity call."));
```
white238: Is it at all helpful to have this error mention that this is a bulk sensitivity/space?  Along with the shear one above.

src/serac/physics/solid.hpp
```diff
@@ -354,13 +355,13 @@ class Solid : public BasePhysics {
    * @note It is expected that the forward analysis is complete and the current displacement state is valid
    * @note If the essential boundary state is not specified, homogeneous essential boundary conditions are applied
    *
-   * @param[in] adjoint_load_form The linear form that when assembled contains the right hand side of the adjoint system
-   * @param[in] state_with_essential_boundary A optional finite element state containing the non-homogenous essential
+   * @param[in] adjoint_load_state The dual state that contains the right hand side of the adjoint system
```
white238: `adjoint_load_state` does not match `adjoint_load`.  If the doxygen comment is right don't forget to change the cpp file as well.

src/serac/physics/utilities/finite_element_dual.hpp
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * @brief Use the finite element vector constructors
+   */
+  using FiniteElementVector::FiniteElementVector;
+
+  /**
+   * @brief Returns a non-owning reference to the local degrees of freedom
+   *
+   * @return mfem::Vector& The local dof vector
+   * @note While this is a grid function for plotting and parallelization, we only return a vector
```
white238, line 39: A+ comment.

src/serac/physics/utilities/finite_element_vector.cpp
```diff
@@ -0,0 +1,97 @@
...
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "serac/physics/utilities/finite_element_vector.hpp"
+
+namespace serac {
+
+namespace detail {
+
+/**
+ * @brief Helper function for creating a GridFunction in both restart and not-restart scenarios
```
white238: non-restart?

white238 (approved):

src/serac/physics/utilities/finite_element_vector.hpp
```diff
@@ -0,0 +1,218 @@
...
+ * Namely: Mesh, FiniteElementCollection, FiniteElementVector,
+ * GridFunction, and a distributed vector of the solution
+ */
+class FiniteElementVector {
+public:
+  /**
+   * @brief Structure for optionally configuring a FiniteElementVector
+   */
+  // The optionals are explicitly default-constructed to allow the user to partially aggregrate-initialized
+  // with only the options they care about
```
samuelpmishLLNL: "optionals" seems like a typo

this might be good to include in the doxygen comment too, so it's apparent that the user can use aggregate initialization and only specify which values they want to override

src/serac/physics/utilities/finite_element_vector.hpp
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * @brief Returns a non-owning reference to the internal FESpace
+   * @return The underlying finite element space
+   */
+  mfem::ParFiniteElementSpace& space() { return detail::retrieve(space_); }
+  /// \overload
+  const mfem::ParFiniteElementSpace& space() const { return detail::retrieve(space_); }
+
+  /**
+   * @brief Returns a non-owning reference to the vector of true DOFs
```
samuelpmishLLNL, line 117: minor: I think reference variables in C++ are always "non-owning", so this description is a little bit redundant

since this is a class an analyst might interact with directly, it might be good to explain (or link to a document that does) what "true" dofs are

src/serac/physics/utilities/finite_element_vector.hpp
```diff
@@ -0,0 +1,218 @@
...
+  std::reference_wrapper<mfem::ParMesh> mesh_;
+  /**
+   * @brief Possibly-owning handle to the FiniteElementCollection, as it is owned
+   * by the FiniteElementVector in a normal run and by the MFEMSidreDataCollection
+   * in a restart run
+   * @note Must be const as FESpaces store a const reference to their FEColls
+   */
+  detail::MaybeOwningPointer<const mfem::FiniteElementCollection> coll_;
+  /**
+   * @brief Possibly-owning handle to the FiniteElementCollection, as it is owned
```
samuelpmishLLNL:
```suggestion
   * @brief Possibly-owning handle to the mfem::ParFiniteElementSpace, as it is owned
```

src/serac/physics/utilities/finite_element_vector.hpp
```diff
@@ -0,0 +1,218 @@
...
+   */
+  detail::MaybeOwningPointer<mfem::ParGridFunction> gf_;
+  mfem::HypreParVector                              true_vec_;
+  std::string                                       name_ = "";
+};
+
+/**
+ * @brief Find the average value of a finite element vector across all dofs
+ *
+ * @param state The state variable to compute a max of
```
samuelpmishLLNL:
```suggestion
 * @param state The state variable to compute the average of
```

src/serac/physics/utilities/finite_element_vector.hpp
```diff
@@ -0,0 +1,218 @@
...
+ */
+double max(const FiniteElementVector& fe_vector);
+
+/**
+ * @brief Find the min value of a finite element vector across all dofs
+ *
+ * @param state The state variable to compute a min of
+ * @return The min value
+ */
+double min(const FiniteElementVector& fe_vector);
```
samuelpmishLLNL: It might be worth adding that these operations act directly on the components of the field, rather than the actual field itself (e.g. the maximum displacement in a domain may not occur at a node)

samuelpmishLLNL (approved):

PR #556, opened 2021-08-24
Add doxygen/sphinx warnings CI check

white238:
This adds a new script that will find any log for `warning:` and `error:` and allows for a file filled with regex's to ignore specific lines.  Here is the usage:

```
usage: check_log.py [-h] -l LOG [-i IGNORE]

Checks log for warnings and errors

optional arguments:
  -h, --help            show this help message and exit
  -l LOG, --log LOG     Path to log file to be checked
  -i IGNORE, --ignore IGNORE
                        Path to file that includes regex's to ignore lines
```

 **This currently is forced to return true until we clean up the existing (82) warnings.** 

Docker changes:
- base axom images now have an editor: `vim`
- our images now have: doxygen, latex, sphinx, and graphviz
- I deleted serac's and spack's build directories after they successfully completed their builds and they are no longer apart of the image layers.  This reduced gcc's image from 4.9gb to 1.89gb


white238:
This is ready for review.  New check output can be seen here:

https://dev.azure.com/llnl-serac/serac/_build/results?buildId=5326&view=logs&j=1be3ff84-5b97-5584-c1a4-7ebbbeb3b802&t=d32b9d99-ea3e-5518-9381-52c611e33fea&l=1856

Notice that it is failing because there are warnings but I'm forcing it to pass until we fix it.

scripts/docker/dockerfile_clang-10
```diff
@@ -1,28 +1,26 @@
-FROM axom/compilers:ubuntu-20-clang-10_07-19-21_23h-07m
+FROM axom/compilers:ubuntu-20-clang-10_08-24-21_07h-18m
 LABEL maintainer="bramwell1@llnl.gov"
 ARG branch=develop
 
 SHELL ["/bin/bash", "-c"]
 RUN sudo apt-get update -y
-RUN sudo apt-get install ssh libopenblas-dev mpich gfortran -fy
+RUN sudo apt-get install doxygen gfortran graphviz libopenblas-dev mpich python3-sphinx ssh texlive-full -fy
```
joshessman-llnl, line 7: These changes should be propagated to `dockerfile.in` - not sure if anyone else uses that workflow (`build_new_dockerfile.sh`) for creating new images but for what it's worth I used it to create the GCC 9 image

joshessman-llnl (approved):

scripts/docker/dockerfile_clang-10
white238: done!  That file was very out of date.....

scripts/check_log.py
```diff
@@ -0,0 +1,101 @@
...
+    print("")
+
+    # Print summary info
+    print("~~~~~~~~~~~~~~~~~ Summary ~~~~~~~~~~~~~~~~~~")
+    print("Warning Count: {0}".format(len(warnings)))
+    print("Error Count:   {0}".format(len(errors)))
+    print("Ignored Count: {0}".format(ignored_count))
+
+    # Error out if any found
+    # TODO: use this return statement in follow up PR that fixes warnings/errors
```
jamiebramwell, line 96: 👍 I can do this follow on PR

scripts/docker/dockerfile_clang-10
jamiebramwell: Yeah, I don't even know the last time we used that directly...

jamiebramwell (approved):

PR #559, opened 2021-09-01
Refactor element gradient kernels and sparse matrix assembly

samuelpmishLLNL:
This PR is aimed at simplifying the element gradient kernels, and extending the sparse matrix assembly routines to include lookup tables for the boundary element contributions as well.


src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -104,12 +104,12 @@ template <Geometry g, typename test, typename trial, int Q, typename derivatives
           typename qpt_data_type = void>
 void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type* derivatives_ptr,
                        const mfem::Vector& J_, const mfem::Vector& X_, const mfem::Vector& N_, int num_elements,
-                       lambda qf, QuadratureData<qpt_data_type>& data = dummy_qdata)
+                       lambda qf)
 
 {
   using test_element               = finite_element<g, test>;
   using trial_element              = finite_element<g, trial>;
-  using element_residual_type      = typename trial_element::residual_type;
```
samuelpmishLLNL, line 112: this change isn't related to this PR in particular, this is just a bug that wasn't caught in previous tests

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -148,18 +148,7 @@ void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type*
...
-        if constexpr (std::is_same_v<qpt_data_type, void>) {
-          // [[maybe_unused]] not supported in captures
-          (void)data;
-          (void)e;
-          (void)q;
-          return qf(x_q, n_q, make_dual(arg));
-        } else {
-          return qf(x_q, n_q, make_dual(arg), data(e, q));
-        }
-      }();
```
samuelpmishLLNL, line 162: I don't believe we actually support user-defined quadrature data on the boundary yet, so this was reverted

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -257,6 +246,61 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
   }
 }
 
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type>
+void element_gradient_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr, const mfem::Vector& J_, int num_elements)
```
samuelpmishLLNL: Note: I believe this kernel is working correctly, but the assembly of the sparse matrix from these element matrices is complicated by the inconsistent dof numbering in mfem, see https://github.com/mfem/mfem/issues/2495

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -340,6 +387,14 @@ class BoundaryIntegral {
    */
   void GradientMult(const mfem::Vector& input_E, mfem::Vector& output_E) const { gradient_(input_E, output_E); }
 
+  /**
+   * @brief Computes the element stiffness matrices, storing them in an `mfem::Vector` that has been reshaped into a
+   * multidimensional array
+   * @param[inout] K_b The reshaped vector as a mfem::DeviceTensor of size (test_dim * test_dof, trial_dim * trial_dof,
+   * elem)
+   */
+  void ComputeElementMatrices(mfem::Vector& K_b) const { gradient_mat_(K_b); }
```
samuelpmishLLNL: should this be renamed to `ComputeElementGradients()`, given that they won't be element matrices in the QoI case? Also, do we want this to be a public method?

src/serac/physics/utilities/functional/dual.hpp
```diff
@@ -227,7 +227,7 @@ template <typename gradient_type>
 auto exp(dual<gradient_type> a)
 {
   using std::exp;
-  return dual<gradient_type>{exp(a.value), exp(a.value)};
```
samuelpmishLLNL, line 230: oops

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +388,246 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
   class Gradient : public mfem::Operator {
+
+    template < typename T >
+    struct Array3D{
+      Array3D() = default;
+      Array3D(int n1, int n2, int n3) : strides{n2 * n3, n3, 1}, data(n1 * n2 * n3) {}
+      auto & operator()(int i, int j, int k) { return data[i * strides[0] + j * strides[1] + k * strides[2]]; }
+      int strides[3];
+      std::vector < T > data;
+    };
```
samuelpmishLLNL: these are temporary containers being used until we have access to `axom::array`

src/serac/physics/utilities/functional/tensor.hpp
```diff
@@ -100,12 +100,58 @@ struct tensor<T> {
...
+  SERAC_HOST_DEVICE constexpr auto& operator[](int) { return value; };
+  SERAC_HOST_DEVICE constexpr auto  operator[](int) const { return value; };
+
+  operator tensor<T, 1>(){ return value; }
+  tensor() : value() {}
+  tensor(T v) : value(v) {}
+  tensor(tensor< T, 1 > v) : value(v) {}
+
+  tensor< T, 1 > value;
+};
```
samuelpmishLLNL: these specializations help handle some of the degenerate cases that occur in the element_gradient_kernel when the test and/or trial spaces are scalar-valued

src/serac/physics/utilities/functional/boundary_integral.hpp
joshessman-llnl: This is correct - supporting this would probably require coordination with MFEM but theoretically it'd mostly be a matter of tweaking the logic here: https://github.com/mfem/mfem/blob/master/fem/fespace.cpp#L3452

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -257,6 +246,61 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
+      } 
+
+    }
+
+    // once we've finished the element integration loop, write our element gradients
+    // out to memory, to be later assembled into the global gradient by mfem
+    // 
+    // Note: we "transpose" these values to get them into the layout that mfem expects
+    for_loop<test_ndof, test_dim, trial_ndof, trial_dim>([&](int i, int j, int k, int l) {
+      dk(i + test_ndof * j, k + trial_ndof * l, e) += K_elem[i][k][j][l];
```
joshessman-llnl: Definitely seeing the use case for being able to flexibly restride arrays here :)

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -197,7 +201,48 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
+  if constexpr (element_type::family == Family::HCURL) {
+    auto N = dot(element_type::shape_functions(xi), inv(J));
+    auto curl_N = element_type::shape_function_curl(xi) / det(J);
+    if constexpr (dim == 3) {
+      curl_N = dot(curl_N, transpose(J));
+    }
+
+    using pair_t = linear_approximation< 
+      typename std::remove_reference<decltype(N[0])>::type, 
+      typename std::remove_reference<decltype(curl_N[0])>::type
```
joshessman-llnl: Extremely minor: `std::remove_reference_t`?

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -23,45 +24,52 @@
...
+  inline int get_sign(int i) { return (i >= 0) ? 1 : -1; }
+  inline int get_index(int i) { return (i >= 0) ? i : - 1 - i; }
+
+  struct signed_index{
+    int index;
+    int sign;
+    operator int(){ return index; }
+  };
+
+  inline void apply_permutation(mfem::Array<int> & input, const mfem::Array<int> & permutation) {
```
joshessman-llnl: Minor: Can we switch the arg order to be consistent with https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs?

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -128,29 +138,28 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
-    // Generate the face restriction operators using the shared face mesh data
-    G_test_boundary_  = GetFaceRestriction<test>(test_fes);
-    G_trial_boundary_ = GetFaceRestriction<trial>(trial_fes);
-
-    input_L_.SetSize(P_trial_->Height(), mfem::Device::GetMemoryType());
-    input_E_.SetSize(G_trial_->Height(), mfem::Device::GetMemoryType());
-
-    output_E_.SetSize(G_test_->Height(), mfem::Device::GetMemoryType());
-    output_L_.SetSize(P_test_->Height(), mfem::Device::GetMemoryType());
+    if (test_space_) test_space_->ExchangeFaceNbrData();
```
joshessman-llnl: Minor/style: I think we try to put even single-statement things inside curly braces

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -196,8 +205,7 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
    * and @a spatial_dim template parameter
    */
   template <int dim, typename lambda, typename qpt_data_type = void>
-  void AddBoundaryIntegral(Dimension<dim>, lambda&& integrand, mfem::Mesh& domain,
-                           QuadratureData<qpt_data_type>& data = dummy_qdata)
+  void AddBoundaryIntegral(Dimension<dim>, lambda&& integrand, mfem::Mesh& domain)
```
joshessman-llnl, line 255: For a future PR: there's some internal inconsistency where sometimes `lambda` gets passed and sometimes `lambda&&`; I don't think it makes a huge difference since functors are usually pretty lightweight, but since the lambda gets passed down a fairly long call chain maybe the latter is preferable.

src/serac/physics/utilities/functional/functional.hpp
joshessman-llnl: just FYI - I think what's currently in axom develop is sufficient to replace this

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +388,246 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
+
+      if (form.domain_integrals_.size() > 0) {
+        bool on_boundary = false;
+
+        for (int e = 0; e < num_elements; e++) {
+          form.test_space_->GetElementDofs(e, test_dofs);
+          form.trial_space_->GetElementDofs(e, trial_dofs);
+
+          const mfem::Array<int> & test_native_to_lexicographic = dynamic_cast<const mfem::TensorBasisElement *>(form.test_space_->GetFE(0))->GetDofMap();
+          const mfem::Array<int> & trial_native_to_lexicographic = dynamic_cast<const mfem::TensorBasisElement *>(form.trial_space_->GetFE(0))->GetDofMap();
```
joshessman-llnl: We should either check to see if the `dynamic_cast` was successful or just use a `static_cast` here

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: Unfortunately, this needs to be done carefully. It seems like the quadrature point numbering they use in GetBdrGeomFactors, for example, is inconsistent with the boundary face ids in the FiniteElementSpace, so the quadrature point data (location, jacobian, etc) are something that I'm currently reimplementing now to make it agree.

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: Good point, these should be made consistent. I'm not sure if an rvalue reference can be passed in to a cuda kernel, but we can find out.

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: I believe we should be okay, since we're actually checking that all of the elements are of the appropriate type in `AddDomainIntegral`, `AddBoundaryIntegral`, but it's probably worth doing. It's also not clear to me whether or not GetDofMap actually depends on the instance or not. If this permutation is always the same for each element type, we should probably move it out of the loop.

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -104,12 +104,12 @@ template <Geometry g, typename test, typename trial, int Q, typename derivatives
           typename qpt_data_type = void>
 void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, derivatives_type* derivatives_ptr,
                        const mfem::Vector& J_, const mfem::Vector& X_, const mfem::Vector& N_, int num_elements,
-                       lambda qf, QuadratureData<qpt_data_type>& data = dummy_qdata)
+                       lambda qf)
```
jamiebramwell, line 107: This is because we decided to remove this while the until the GPU quadrature data branch is merged, correct?

src/serac/physics/utilities/functional/boundary_integral.hpp
jamiebramwell: Would it be worth testing on my MFEM example 1 thing? We could verify it by the eyeball norm at least.

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -257,6 +246,61 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
+template <Geometry g, typename test, typename trial, int Q, typename derivatives_type>
+void element_gradient_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr, const mfem::Vector& J_, int num_elements)
+{
+  using test_element               = finite_element<g, test>;
+  using trial_element              = finite_element<g, trial>;
+  static constexpr int  test_ndof  = test_element::ndof;
+  static constexpr int  test_dim   = test_element::components;
+  static constexpr int  trial_ndof = trial_element::ndof;
+  static constexpr int  trial_dim  = test_element::components;
+  static constexpr auto rule       = GaussQuadratureRule<g, Q>();
```
jamiebramwell, line 259: Is this gauss lobatto or gauss legendre?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -197,7 +201,48 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
-}  // namespace domain_integral
+template < typename T1, typename T2 >
+struct linear_approximation {
+  T1 value;
+  T2 derivative;
+};
+
+template < typename element_type, int dim >
+auto evaluate_shape_functions(tensor < double, dim > xi, tensor< double, dim, dim > J) {
+  if constexpr (element_type::family == Family::HCURL) {
```
white238: Have you run style over this?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -227,212 +272,70 @@ void gradient_matrix_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr
...
+  static constexpr int   test_dim   = test_element::components;
+  static constexpr int   trial_ndof = trial_element::ndof;
+  static constexpr int   trial_dim  = test_element::components;
+  static constexpr auto  rule       = GaussQuadratureRule<g, Q>();
 
   // mfem provides this information in 1D arrays, so we reshape it
   // into strided multidimensional arrays before using
   auto J  = mfem::Reshape(J_.Read(), rule.size(), dim, dim, num_elements);
   auto dk = mfem::Reshape(K_e.ReadWrite(), test_ndof * test_dim, trial_ndof * trial_dim, num_elements);
+  //auto derivatives = mfem::Reshape(derivatives_ptr, );
```
white238: dead code.

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -227,212 +272,70 @@ void gradient_matrix_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr
...
-  static constexpr int                  trial_ndof = trial_element::ndof;
-  static constexpr int                  trial_dim  = test_element::components;
-  static constexpr auto                 rule       = GaussQuadratureRule<g, Q>();
-  static constexpr int                  dim        = dimension_of(g);
-  [[maybe_unused]] static constexpr int curl_dim   = dim == 3 ? 3 : 1;
+  static constexpr int   dim        = dimension_of(g);
+  static constexpr int   test_ndof  = test_element::ndof;
+  static constexpr int   test_dim   = test_element::components;
+  static constexpr int   trial_ndof = trial_element::ndof;
+  static constexpr int   trial_dim  = test_element::components;
```
white238: I assume this is on purpose that `test_dim` and `trial_dim` are the same?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -227,212 +272,70 @@ void gradient_matrix_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr
...
-      auto                    J_q    = make_tensor<dim, dim>([&](int i, int j) { return J(q, i, j, e); });
-      auto                    detJ_q = det(J_q);
-      [[maybe_unused]] double dx     = detJ_q * dxi_q;
+      auto xi_q  = rule.points[q];
+      auto dxi_q = rule.weights[q];
+      auto J_q   = make_tensor<dim, dim>([&](int i, int j) { return J(q, i, j, e); });
+      double dx = det(J_q) * dxi_q;
 
       // recall the derivative of the q-function w.r.t. its arguments at this quadrature point
+      //auto dq_darg = detail::AccessDerivatives(derivatives_ptr, e, q, rule, num_elements);
```
white238: dead code.

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -313,15 +356,19 @@ class BoundaryIntegral {
     //
     // note: the qf_derivatives_ptr is copied by value to each lambda function below,
     //       to allow the evaluation kernel to pass derivative values to the gradient kernel
-    evaluation_ = [=, &data](const mfem::Vector& U, mfem::Vector& R) {
+    evaluation_ = [=](const mfem::Vector& U, mfem::Vector& R) {
```
jamiebramwell: Are the default value captures needed? I think it's best to explicitly state what we're capturing. See [this core guideline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f54-if-you-capture-this-capture-all-variables-explicitly-no-default-capture).

src/serac/physics/utilities/functional/boundary_integral.hpp
jamiebramwell: I agree `ComputeElementGradients` makes sense.

jamiebramwell: What is the use case for this being a public method?

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -23,45 +24,52 @@
 #include "serac/numerics/assembled_sparse_matrix.hpp"
 #include "serac/infrastructure/logger.hpp"
 
-namespace serac {
+namespace detail {
```
white238, line 29: This should be:

```
namespace serac::detail {
```

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -197,7 +201,48 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
 }
 
-}  // namespace domain_integral
+template < typename T1, typename T2 >
+struct linear_approximation {
+  T1 value;
+  T2 derivative;
+};
+
+template < typename element_type, int dim >
```
jamiebramwell: This needs doxygen.

src/serac/physics/utilities/functional/domain_integral.hpp
jamiebramwell: I believe it should be `trial_dim = trial_element::components`.

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -197,7 +201,48 @@ void gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivatives_type*
...
+    for (int i = 0; i < element_type::ndof; i++) {
+      output[i].value = N[i];
+      output[i].derivative = curl_N[i];
+    }
+    return output;
+  } else {
+    auto N = element_type::shape_functions(xi);
+    auto grad_N = dot(element_type::shape_function_gradients(xi), inv(J));
+
+    using pair_t = linear_approximation<
```
jamiebramwell: At first glance, it is difficult to understand what this does. I think this method could use some comments.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -310,11 +319,33 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
+    if (K_b_.Size() == 0) {
+      int num_boundary_elements = test_space_->GetNBE();
+      int dofs_per_test_boundary_element = test_space_->GetBE(0)->GetDof() * test_space_->GetVDim();
+      int dofs_per_trial_boundary_element = trial_space_->GetBE(0)->GetDof() * trial_space_->GetVDim();
+      K_b_.SetSize(dofs_per_test_boundary_element * dofs_per_trial_boundary_element * num_boundary_elements);
+    }
+    // zero out internal vector
+    K_b_ = 0.;
+    // loop through integrals and accumulate
+    for (auto & boundary : boundary_integrals_) boundary.ComputeElementMatrices(K_b_);
```
white238: Please put curly braces around this.  It reduces the chance of future bugs

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -227,212 +272,70 @@ void gradient_matrix_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr
...
+      auto & q11 = serac::get<1>(serac::get<1>(dq_darg)); // derivative of   flux term w.r.t. field derivative
+
+      auto M = evaluate_shape_functions< test_element >(xi_q, J_q);
+      auto N = evaluate_shape_functions< trial_element >(xi_q, J_q);
+
+      // clang-format off
+      for (int i = 0; i < test_ndof; i++) {
+        for (int j = 0; j < trial_ndof; j++) {
+          K_elem[i][j] += (
+            M[i].value      * q00 * N[j].value +
```
jamiebramwell: I really like this!

src/serac/physics/utilities/functional/dual.hpp
```diff
@@ -227,7 +227,7 @@ template <typename gradient_type>
...
+  return dual<gradient_type>{exp(a.value), exp(a.value) * a.gradient};
```
jamiebramwell, line 230: Math strikes again! 😄

src/serac/physics/utilities/functional/finite_element.hpp
```diff
@@ -62,6 +62,17 @@ constexpr int dimension_of(Geometry g)
   return -1;
 }
 
+/**
+ * @brief Element conformity
+ */
+enum class Family
```
jamiebramwell, line 68: 👍

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +388,246 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
+        }
+      }
+
+      sparsity_pattern_initialized = true;
+
+    }
+
+    operator mfem::SparseMatrix() {
+
+      if (!sparsity_pattern_initialized) initialize_sparsity_pattern();
```
white238: curly braces.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -23,45 +24,52 @@
...
- * @brief a type trait used to identify the Hcurl family
- */
-template <typename T>
-struct is_hcurl {
-  /**
-   * @brief Flag denoting the type contained is Hcurl
-   */
-  static constexpr bool value = false;
-};
+  struct elem_info{
```
jamiebramwell: Needs doxygen

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -564,13 +821,26 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
    * @brief Local internal AssembledSparseMatrix storage for ComputeElementMatrices
    *
    * If unique_ptr is empty, construct AssembledSparseMatrix.
    *
    */
   std::unique_ptr<serac::mfem_ext::AssembledSparseMatrix> assembled_spmat_;
+
+
+  template < typename T >
+  friend typename Functional<T>::Gradient & grad(Functional<T> &);
```
white238: doxygen

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
```diff
@@ -29,6 +29,65 @@ constexpr bool verbose     = true;
...
+  if (A.Width()  != B.Width())  return false;
+
+  double fnorm_A = 0.0;
+  double fnorm_A_minus_B = 0.0;
+
+  for (int r = 0; r < A.Height(); r++) {
+    auto columns = A.GetRowColumns(r);
+    for (int j = 0; j < A.RowSize(r); j++) {
+      int c = columns[j];
+      if (c < 0) std::cout << "??" << std::endl;
```
white238: curly braces and a more helpful message.. is this a failure?

src/serac/physics/utilities/functional/functional.hpp
jamiebramwell: Also, these methods are not capitalized by either our style guide (types in `PascalCase`, methods in `camelCase`) or MFEM's (types in `PascalCase`, methods in `PascalCase`).

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
```diff
@@ -159,11 +233,11 @@ TEST(boundary, 2D_quadratic) { boundary_test(*mesh2D, H1<2>{}, H1<2>{}, Dimensio
 TEST(boundary, 3D_linear) { boundary_test(*mesh3D, H1<1>{}, H1<1>{}, Dimension<3>{}); }
 TEST(boundary, 3D_quadratic) { boundary_test(*mesh3D, H1<2>{}, H1<2>{}, Dimension<3>{}); }
 
-TEST(boundary_L2, 2D_linear) { boundary_test(*mesh2D, L2<1>{}, L2<1>{}, Dimension<2>{}); }
-TEST(boundary_L2, 2D_quadratic) { boundary_test(*mesh2D, L2<2>{}, L2<2>{}, Dimension<2>{}); }
-
-TEST(boundary_L2, 3D_linear) { boundary_test(*mesh3D, L2<1>{}, L2<1>{}, Dimension<3>{}); }
-TEST(boundary_L2, 3D_quadratic) { boundary_test(*mesh3D, L2<2>{}, L2<2>{}, Dimension<3>{}); }
+//TEST(boundary_L2, 2D_linear) { boundary_test(*mesh2D, L2<1>{}, L2<1>{}, Dimension<2>{}); }
```
white238: Can you add a comment why these are commented out?

src/serac/physics/utilities/functional/functional.hpp
jamiebramwell: I think this can be a follow on PR. I'll make an issue.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +388,246 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
+          row_ptr[j+1] = nonzero_ids[i];
+        }
+      }
+
+      ++nnz;
+      for (int j = infos.back().global_row; j < nrows; j++) {
+        row_ptr[j+1] = nnz;
+      }
+
+      element_nonzero_LUT = Array3D<::detail::signed_index>(num_elements, dofs_per_test_element * test_vdim, dofs_per_trial_element * trial_vdim);
```
jamiebramwell: What does `LUT` mean? Comments would be appreciated.

src/serac/physics/utilities/functional/boundary_integral.hpp
jonwong12: I think in either case the `mfem::Vector` itself doesn't really contain enough information to give the "Matrices" or "Gradients" meaning.

Ideally in the future I think we should move away from returning reshaped vector since it could easily be reshaped incorrectly back into the element matrices. I'd prefer the result to be a "Matrix format" and something like a vector for the QoI case.

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -227,212 +272,70 @@ void gradient_matrix_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr
...
+      auto & q10 = serac::get<0>(serac::get<1>(dq_darg)); // derivative of   flux term w.r.t. field value
+      auto & q11 = serac::get<1>(serac::get<1>(dq_darg)); // derivative of   flux term w.r.t. field derivative
+
+      auto M = evaluate_shape_functions< test_element >(xi_q, J_q);
+      auto N = evaluate_shape_functions< trial_element >(xi_q, J_q);
+
+      // clang-format off
+      for (int i = 0; i < test_ndof; i++) {
+        for (int j = 0; j < trial_ndof; j++) {
+          K_elem[i][j] += (
```
jonwong12: Nice refactor. This is quite a bit cleaner!

src/serac/physics/utilities/functional/functional.hpp
jonwong12: Can you add a short doxygen annotation for Array3D and how it's strided?

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
jamiebramwell: Yes, this should be a `SLIC_ERROR`

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +388,246 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
 
     virtual void Mult(const mfem::Vector& x, mfem::Vector& y) const override { form.GradientMult(x, y); }
+    
+    mfem::Vector operator()(const mfem::Vector& x) const { 
+      mfem::Vector y(form.Height());
+      form.GradientMult(x, y); 
+      return y;
+    }
+
+    void initialize_sparsity_pattern() {
```
jonwong12: I think it might be helpful to add some more comments throughout this method and perhaps provide a short summary of what you're doing at the top.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +388,246 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
+        } else {
+          element_nonzero_LUT(element_id, local_row, local_col) = nonzero_ids[i];
+        }
+      }
+
+      sparsity_pattern_initialized = true;
+
+    }
+
+    operator mfem::SparseMatrix() {
```
jonwong12: doxygen

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +388,246 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
+    }
 
   private:
+
     /**
      * @brief The "parent" @p Functional to calculate gradients with
      */
     Functional<test(trial), execution_policy>& form;
+
+    std::vector< int > row_ptr;
```
jonwong12: I think it's worth documenting the `row_ptr`, `col_ind`, and the *`nonzero_LUT` variables.

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
```diff
@@ -74,21 +130,38 @@ void boundary_test(mfem::ParMesh& mesh, H1<p> test, H1<p> trial, Dimension<dim>)
...
     std::cout << "sum(r2):  " << r2.Sum() << std::endl;
     std::cout << "||r1||: " << r1.Norml2() << std::endl;
     std::cout << "||r2||: " << r2.Norml2() << std::endl;
     std::cout << "||r1-r2||/||r1||: " << mfem::Vector(r1 - r2).Norml2() / r1.Norml2() << std::endl;
+
+    std::ofstream outfile;
+
+    outfile.open(std::to_string(p) + "_" + std::to_string(dim) + "_A_mfem.mtx");
+    gradient1.SortColumnIndices();
+    gradient1.PrintMM(outfile);
```
jonwong12: Does the matrix printing work in parallel?

src/serac/physics/utilities/functional/domain_integral.hpp
samuelpmishLLNL: Updated. I do like the shorter name, but there's something about seeing an expression like 
```
std::remove_reference_t<decltype(N[0])>
```
ending with a `>` that looks unnatural to me at first glance.

src/serac/physics/utilities/functional/domain_integral.hpp
samuelpmishLLNL: Apparently, my `style` target was not being enabled since we introduced the SERAC_ENABLE_CODE_CHECKS option (prior to that, ENABLE_CLANGFORMAT=ON was sufficient to enable the style checking). After explicitly specifying `SERAC_ENABLE_CODE_CHECKS=ON`, it's working again (thanks Josh).

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
samuelpmishLLNL: an old debugging comment. Removed!

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: `QuadratureData` does not work on boundary integrals, regardless of CPU/GPU

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: In my experience, just saying "Gauss Quadrature" implies Gauss-Legendre, but I can clarify on the doxygen comments for that function.

jamiebramwell: That's fair. Full disclosure, I didn't look up the doxygen as I was only looking at it on GitHub.

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: are you recommending
```cpp
evaluation_ = [qf_derivatives, &J_, &X_, &normals_, &num_elements, qf](const mfem::Vector& U, mfem::Vector& R) {
```
,
```cpp
evaluation_ = [=, this](const mfem::Vector& U, mfem::Vector& R) {
```
or
```cpp
evaluation_ = [qf_derivatives, this](const mfem::Vector& U, mfem::Vector& R) {
```
or something else?

jamiebramwell: I like option 3 the best, but I'm fine with any of them. Basically, I just want it to portray that even though it is capturing by value, we are getting the private data by reference. Any thoughts @white238 @joshessman-llnl ?

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: I agree that mfem::Vector is clearly the wrong container for this, but I think we stick to it for the near future until the axom array stuff is ready. I'm renaming things these quantities from "Matrix" to "Gradient" where appropriate.

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: I mean to say that the elementwise gradients are likely correct but can't be assembled into a global sparse matrix properly due to those numbering inconsistencies discussed last week. How would you want to test it?

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
samuelpmishLLNL: I'm not sure. Right now, the SparseMatrix that I form isn't a HypreParMatrix, so it's just printing the local processor's contribution

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: LUT: "look-up table"

added more comments (in-line and Doxygen) to help clarify steps and intent

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: update: it sounds like we'll start testing this on the nonlinear thermal example in a future PR.

src/serac/physics/utilities/functional/boundary_integral.hpp
white238: 1 or 3 is fine but I will leave it up to you.  It is a good thing to avoid defaulting to value captures

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -531,12 +122,12 @@ class DomainIntegral {
             exec_config, U, R, qf_derivatives.get(), J_, X_, num_elements, qf);
       };
 
-      gradient_ = [=](const mfem::Vector& dU, mfem::Vector& dR) {
+      action_of_gradient_ = [=](const mfem::Vector& dU, mfem::Vector& dR) {
```
jamiebramwell: Explicit capture is needed.

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: thanks, I glossed over these because they were `#ifdef`d out at the time!

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
```diff
@@ -80,15 +136,36 @@ void boundary_test(mfem::ParMesh& mesh, H1<p> test, H1<p> trial, Dimension<dim>)
...
   if (verbose) {
     std::cout << "sum(r1):  " << r1.Sum() << std::endl;
     std::cout << "sum(r2):  " << r2.Sum() << std::endl;
     std::cout << "||r1||: " << r1.Norml2() << std::endl;
     std::cout << "||r2||: " << r2.Norml2() << std::endl;
     std::cout << "||r1-r2||/||r1||: " << mfem::Vector(r1 - r2).Norml2() / r1.Norml2() << std::endl;
+
+    std::ofstream outfile;
+
+    // re-enable after working around FaceRestriction numbering inconsistencies
```
jamiebramwell: I would add a TODO so it's grep-able.

src/serac/physics/utilities/functional/tests/functional_material_state_test.cpp
```diff
@@ -18,6 +18,9 @@
 
 #include "serac/physics/utilities/state_manager.hpp"
 
+// TODO reenable these tests
```
jamiebramwell: Add a note saying why they were disabled and when we can turn them back on.

jamiebramwell (approved):

src/serac/infrastructure/accelerator.hpp
```diff
@@ -139,18 +129,18 @@ std::string getCUDAMemInfoString()
 /**
  * @brief create shared_ptr to an array of `n` values of type `T`, either on the host or device
  * @tparam T the type of the value to be stored in the array
- * @tparam execution_policy the memory space where the data lives
+ * @tparam execution_space the memory space where the data lives
```
white238: this should be `exec`

src/serac/physics/utilities/functional/domain_integral_kernels.hpp
```diff
@@ -0,0 +1,273 @@
...
+/**
+ * @brief The base kernel template used to create different finite element calculation routines
+ *
+ * @tparam test The type of the test function space
+ * @tparam trial The type of the trial function space
+ * The above spaces can be any combination of {H1, Hcurl, Hdiv (TODO), L2 (TODO)}
+ *
+ * Template parameters other than the test and trial spaces are used for customization + optimization
+ * and are erased through the @p std::function members of @p DomainIntegral
+ * @tparam g The shape of the element (only quadrilateral and hexahedron are supported at present)
```
white238: You have an extra `qpt_data_type` as a `tparam`.

You have an extra `data` as `param`

Also can you please order these to match the actual order.

src/serac/physics/utilities/functional/domain_integral_kernels.hpp
```diff
@@ -0,0 +1,273 @@
...
+ * kernels associated with finite element calculations
+ *
+ * @tparam test The type of the test function space
+ * @tparam trial The type of the trial function space
+ * The above spaces can be any combination of {H1, Hcurl, Hdiv (TODO), L2 (TODO)}
+ *
+ * Template parameters other than the test and trial spaces are used for customization + optimization
+ * and are erased through the @p std::function members of @p DomainIntegral
+ * @tparam g The shape of the element (only quadrilateral and hexahedron are supported at present)
+ * @tparam Q Quadrature parameter describing how many points per dimension
```
white238, line 115: missing test and trial

src/serac/physics/utilities/functional/finite_element.hpp
```diff
@@ -152,6 +155,52 @@ struct is_finite_element<finite_element<g, Hcurl<p> > > {
   static constexpr bool value = true;  ///< whether or not type T is a finite_element
 };
 
+/**
+ * @brief a class that helps to extract the test space from a function signature template parameter
+ * @tparam space The function signature itself
```
white238: spaces

src/serac/physics/utilities/functional/finite_element.hpp
```diff
@@ -152,6 +155,52 @@ struct is_finite_element<finite_element<g, Hcurl<p> > > {
...
+ */
+template <typename spaces>
+struct get_test_space;  // undefined
+
+/**
+ * @brief a class that helps to extract the test space from a function signature template parameter
+ * @tparam space The function signature itself
```
white238: change to test_space and add trial_space

src/serac/physics/utilities/functional/finite_element.hpp
```diff
@@ -152,6 +155,52 @@ struct is_finite_element<finite_element<g, Hcurl<p> > > {
...
+/**
+ * @brief a type function that extracts the test space from a function signature template parameter
+ * @tparam space The function signature itself
+ */
+template <typename spaces>
+using test_space_t = typename get_test_space<spaces>::type;
+
+/**
+ * @brief a type function that extracts the trial space from a function signature template parameter
+ * @tparam space The function signature itself
```
white238: Please review all parameter lists in this file.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -25,45 +26,89 @@
...
+  int  global_col;
+  int  local_row;
+  int  local_col;
+  int  element_id;
+  int  sign;
+  bool on_boundary;
 };
 
+/// @brief for sorting lexicographically by {global_row, global_col}
+inline bool operator<(const ElemInfo& x, const ElemInfo& y)
```
white238, line 54: Missing a lot of doxygen parameter comments in this file.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -228,10 +277,10 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
    * @param[in] domain The mesh to evaluate the integral on
    * @param[in] data The data structure containing per-quadrature-point data
```
white238: Remove data please

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -242,10 +291,10 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
```
white238: remove data

src/serac/physics/utilities/functional/integral_utilities.hpp
```diff
@@ -306,21 +275,141 @@ SERAC_HOST_DEVICE constexpr derivatives_type& AccessDerivatives(derivatives_type
...
+ * to omit unused components (e.g. specify that they only need the gradient)
+ *
+ * @param[in] u The DOF values for the element
+ * @param[in] xi The position of the quadrature point in reference space
+ * @param[in] J The Jacobian of the element transformation at the quadrature point
+ * @tparam element_type The type of the element (used to determine the family)
  */
-template <typename spaces>
-using trial_space_t = typename detail::get_trial_space<spaces>::type;
+template <typename element_type, typename T, int dim>
```
white238, line 353: T and dim are not documented

src/serac/physics/utilities/functional/calculate_boundary_integration_points.hpp
```diff
@@ -0,0 +1,37 @@
+#pragma once
+
+#include "serac/physics/utilities/functional/tensor.hpp"
+#include "serac/physics/utilities/functional/quadrature.hpp"
+#include "serac/physics/utilities/functional/finite_element.hpp"
+
+template <Geometry g, typename function_space, int Q, int dim>
```
jonwong12: Not sure if this needs to be doxygened.

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -500,27 +90,28 @@ class DomainIntegral {
...
         domain_integral::evaluation_kernel<geometry, test_space, trial_space, Q>(U, R, qf_derivatives.get(), J_, X_,
-                                                                                 num_elements, qf, data);
+                                                                                 num_elements, qf);
       };
 
-      gradient_ = [=](const mfem::Vector& dU, mfem::Vector& dR) {
-        domain_integral::gradient_kernel<geometry, test_space, trial_space, Q>(dU, dR, qf_derivatives.get(), J_,
-                                                                               num_elements);
+      action_of_gradient_ = [this, qf_derivatives, num_elements](const mfem::Vector& dU, mfem::Vector& dR) {
+        domain_integral::action_of_gradient_kernel<geometry, test_space, trial_space, Q>(dU, dR, qf_derivatives.get(),
```
jonwong12, line 100: Nice. This looks much cleaner!

src/serac/physics/utilities/functional/domain_integral_kernels.cuh
```diff
@@ -155,9 +178,35 @@ __global__ void eval_cuda_quadrature(const solution_type u, residual_type r, der
     // this is where we will accumulate the element residual tensor
     element_residual_type r_elem{};
 
-    // for each quadrature point in the element
-    eval_quadrature<g, test, trial, Q, derivatives_type, lambda>(e, q, u_elem, r_elem, derivatives_ptr, J, X,
-                                                                 num_elements, qf);
+    //// for each quadrature point in the element
```
jonwong12: dead-ish code. Maybe just leave a comment that this section used to call `eval_quadrature` to act as a marker?

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
```diff
@@ -80,15 +136,36 @@ void boundary_test(mfem::ParMesh& mesh, H1<p> test, H1<p> trial, Dimension<dim>)
   mfem::Vector r1 = (*J) * U + (*F);
   mfem::Vector r2 = residual(U);
 
+  mfem::SparseMatrix gradient1 = B.SpMat();
+  // TODO: re-enable after working around FaceRestriction numbering inconsistencies
+  // mfem::SparseMatrix gradient2 = grad(residual);
```
jonwong12, line 141: Dead code? Same with the following few sections of `TODO` code.

jonwong12: Not sure what the proper convention is. Maybe we bound it with `// END TODO` ?

src/serac/physics/utilities/functional/calculate_boundary_integration_points.hpp
samuelpmishLLNL: oops, this file shouldn't be in here, this is part of a follow-on PR!

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
samuelpmishLLNL: it's not dead code, just disabled for a few days until the next PR gets in

src/serac/physics/utilities/functional/domain_integral_kernels.hpp
joshessman-llnl: 😆 I noticed this as well in some stuff already in `develop`, but it actually meant less copying and pasting for me in #518

src/serac/physics/utilities/functional/functional.hpp
samuelpmishLLNL: I'm not sure how to describe `x` and `y` in a way that isn't tautological, but I'll add the @param statements to these operators nonetheless

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +437,312 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
    * @brief Lightweight shim for mfem::Operator that produces the gradient of a @p Functional from a @p Mult
    */
   class Gradient : public mfem::Operator {
+    // this is a temporary array implementation to be replaced by axom::Array as soon as it's ready
```
joshessman-llnl, line 441: Minor: Is it worth throwing in a `FIXME` or `TODO` here for easier searchability later?

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -357,20 +437,312 @@ class Functional<test(trial), execution_policy> : public mfem::Operator {
...
+          }
+        }
+      }
+
+      return mfem::SparseMatrix(row_ptr.data(), col_ind.data(), values, Height(), Width(),
+                                sparse_matrix_frees_graph_ptrs, sparse_matrix_frees_values_ptr, col_ind_is_sorted);
+    }
 
   private:
+    /// @brief The "parent" @p Functional to calculate gradients with
```
joshessman-llnl, line 720: Minor: trailing underscore for these member variable names

src/serac/physics/utilities/functional/integral_utilities.hpp
```diff
@@ -306,21 +275,144 @@ SERAC_HOST_DEVICE constexpr derivatives_type& AccessDerivatives(derivatives_type
...
+ * that makes the element gradient evaluation simpler to implement
+ *
+ * @tparam element_type which type of finite element shape functions to evaluate
+ * @tparam dim the geometric dimension of the element
+ *
+ * @param[in] xi where to evaluate shape functions and their gradients
+ * @param[in] J the jacobian, dx_dxi, of the isoparametric map from parent-to-physical coordinates
+ */
+template <typename element_type, int dim>
+auto evaluate_shape_functions(tensor<double, dim> xi, tensor<double, dim, dim> J)
```
joshessman-llnl: Minor: `const tensor&` here? (and in Preprocess/Postprocess)

joshessman-llnl (approved):

white238 (approved):

PR #563, opened 2021-09-14
Fix the doxygen warnings and turn on the CI check

### LINKED ISSUE #562: Turn on the doxygen checks with the appropriate fixes
👍 I can do this follow on PR

_Originally posted by @jamiebramwell in https://github.com/LLNL/serac/pull/556#discussion_r707793575_

jamiebramwell:
Resolves https://github.com/LLNL/serac/issues/562

### DIFF
--- a/src/serac/infrastructure/accelerator.hpp
+++ b/src/serac/infrastructure/accelerator.hpp
@@ -33,7 +33,7 @@
 #include <cuda_runtime.h>
 #else  //__CUDACC__
 /**
- * @brief Macro that toggles between decorating a function for host_device or noop's for non-accelated builds.
+ * @brief Macro that evaluates to `__host__ __device__` when compiling with nvcc and does nothing on a host compiler.
  */
 #define SERAC_HOST_DEVICE
 /**
--- a/src/serac/infrastructure/accelerator.hpp
+++ b/src/serac/infrastructure/accelerator.hpp
@@ -37,11 +37,11 @@
  */
 #define SERAC_HOST_DEVICE
 /**
- * @brief Macro that evaluates to __host__ when compiling with nvcc and does nothing on a host compiler
+ * @brief Macro that evaluates to `__host__` when compiling with nvcc and does nothing on a host compiler
  */
 #define SERAC_HOST
 /**
- * @brief Macro that evaluates to __device__ when compiling with nvcc and does nothing on a host compiler
+ * @brief Macro that evaluates to `__device__` when compiling with nvcc and does nothing on a host compiler
  */
 #define SERAC_DEVICE
 /**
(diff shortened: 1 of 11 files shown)


white238 (approved):

src/serac/infrastructure/accelerator.hpp
```diff
@@ -33,11 +33,20 @@
 #include <cuda_runtime.h>
 #else  //__CUDACC__
 /**
- * @brief Macro that toggles between decorating a function for host and device or noop's for non-accelated builds.
+ * @brief Macro that toggles between decorating a function for host_device or noop's for non-accelated builds.
```
samuelpmishLLNL:
```suggestion
 * @brief Macro that evaluates to `__host__ __device__` when compiling with nvcc and does nothing on a host compiler.
```

src/serac/infrastructure/accelerator.hpp
```diff
@@ -33,11 +33,20 @@
...
  */
 #define SERAC_HOST_DEVICE
+/**
+ * @brief Macro that toggles between decorating a function for host or noop's for non-accelated builds.
+ */
 #define SERAC_HOST
+/**
+ * @brief Macro that toggles between decorating a function for device or noop's for non-accelated builds.
+ */
```
samuelpmishLLNL: the original wording makes it sound like it marks the function as being a no-op, rather than the macro itself doing nothing

I'm not sure what the right wording should be, this is consistent with the earlier suggestion:

Macro that evaluates to {`__host__`, `__device__`} when compiling with nvcc and does nothing on a host compiler

samuelpmishLLNL: also, for 1-liner doxygen comments, could we consider doing something like

```cpp
/// @brief lorem ipsum dolor si amet
#define SERAC_HOST
```

so that a 1-line function doesn't balloon into like 5 lines of code when adding doxygen comments?

jamiebramwell: I'm fine with this when appropriate. I honestly use the javadoc `/**` blocks because my vscode plugin can generate them automatically.

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -85,9 +85,9 @@ inline int get_index(int i) { return (i >= 0) ? i : -1 - i; }
  * {sign, index} int32_t encoding)
  */
 struct SignedIndex {
-  int index;
-  int sign;
-      operator int() { return index; }
+  int index;                            ///< The index values
+  int sign;                             ///< The sign of the index
+      operator int() { return index; }  ///< () operator returning the index
```
```suggestion
      operator int() { return index; }  ///< implicit conversion to type `int`, returning index member variable

samuelpmishLLNL (approved):

PR #565, opened 2021-09-21
Isotropic thermal expansion

jamiebramwell:


src/serac/physics/materials/thermal_expansion_material.hpp
```diff
@@ -0,0 +1,133 @@
...
+   *
+   */
+  mfem::ElementTransformation* parent_to_reference_transformation_;
+};
+
+/**
+ * @brief Neo-Hookean hyperelastic model with a strain energy density function given
+ *   by the formula: \f$(\mu/2)(\bar{I}_1 - dim) + (K/2)(det(F)/g - 1)^2\f$ where
+ *   F is the deformation gradient and \f$\bar{I}_1 = (det(F))^{-2/dim} Tr(F
+ *   F^t)\f$. The parameters \f$\mu\f$ and \f$\lambda\f$ are the Lame parameters.
```
sethwatts: Am I correct in understanding that the purpose of the `ThermalExpansionMaterial` is to  provide an appropriate `evalThermalDeformationGradient` method that can be used to augment the (total) thermal deformation gradient within the `DisplacementHyperelasticIntegrator`? If so, I think this Doxygen needs to be updated since it looks like the `ThermalExpansionMaterial` does not provide a complete SEDF on its own.

src/serac/physics/integrators/displacement_hyperelastic_integrator.cpp
```diff
@@ -31,6 +31,13 @@ void DisplacementHyperelasticIntegrator::CalcKinematics(const mfem::FiniteElemen
 
   solid_util::calcDeformationGradient(du_dX_, F_);
 
+  if (thermal_material_) {
+    F_copy_ = F_;
+    F_thermal_.SetSize(du_dX_.Width());
+    thermal_material_->evalThermalDeformationGradient(F_thermal_);
+    mfem::Mult(F_thermal_, F_copy_, F_);
```
sethwatts: I haven't done enough finite thermoelasticty to say which is more common, but in the META write up, we have the terms of the multiplicative decomposition in opposite order: F = F_elastic * F_thermal. In every case I can think of, F_thermal is going to be diagonal, so this is a distinction without a difference, but I thought I would mention it (and possibly update the write up if necessary).

jamiebramwell: Yep - this is a typo. Thanks for catching it!

src/serac/physics/materials/thermal_expansion_material.hpp
jamiebramwell: Oh yes - the doxygen is 100% wrong and simply an artifact of my cutting and pasting. I'm doing a pass now to correct it.

jamiebramwell:
The Lassen build is also failing on develop, so I'm guessing it's a system change issue.

src/serac/physics/materials/thermal_expansion_material.hpp
```diff
@@ -0,0 +1,160 @@
...
+ * \mathbf{F} = \mathbf{F}_\theta \mathbf{F}_M
+ * \f]
+ *
+ * where \f$\mathbf{F}\f$ is the total deformation gradient, \f$\mathbf{F}_\theta\f$ is the
+ * deformation gradient induced by thermal expansion, and \f$\mathbf{F}_M\f$ is the mechanical
+ * deformation gradient. The thermal deformation gradient of this model is given by
+ *
+ * \f[
+ * \mathbf{F}_\theta = \mathbf{I} + \alpha \left( \theta - \theta_\textrm{ref} \right)
+ * \f]
```
samuelpmishLLNL: this renders to
image
which doesn't make sense to me, do we want
```suggestion
 * \f[
 * \mathbf{F}_\theta = \alpha \left( \theta - \theta_\textrm{ref} \right) \mathbf{I}
 * \f]
```

image

 instead?

samuelpmishLLNL: oops, my comment above is bad, probably this instead (?)
image


```suggestion
 * \f[
 * \mathbf{F}_\theta = (1 + \alpha (\theta - \theta_\textrm{ref})) \mathbf{I}
 * \f]
```

src/serac/physics/materials/thermal_expansion_material.cpp
```diff
@@ -0,0 +1,37 @@
...
+void IsotropicThermalExpansionMaterial::modifyDisplacementGradient(mfem::DenseMatrix& du_dX)
+{
+  EvalCoeffs();
+
+  double expansion = coef_thermal_expansion_ * (reference_temp_ - temp_);
+
+  for (int i = 0; i < du_dX.Width(); ++i) {
+    du_dX(i, i) += expansion * (1.0 + du_dX(i, i));
+  }
+}
```
samuelpmishLLNL: Is this code doing the right thing? An isotropic scaling applied to the displacement gradient would affect all of its entries (not just diagonal ones). Or maybe this is actually calculating the thermal expansion deformation gradient, F_{\theta), in which case we should probably change the function/argument names.

src/serac/physics/materials/thermal_expansion_material.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * @brief Current temperature in constant form
+   */
+  mutable double temp_;
+
+  /**
+   * @brief Coefficient of thermal expansion in coefficient form
+   */
+  std::unique_ptr<mfem::Coefficient> c_coef_thermal_expansion_;
```
samuelpmishLLNL: do we need both `coef_thermal_expansion_` and `c_coef_thermal_expansion` to be members? It looks like maybe `EvalCoeffs()` could just return the thermal expansion coefficient at that point (`coef_thermal_expansion_`), rather than writing/reading to a member variable to access it.

src/serac/physics/integrators/displacement_hyperelastic_integrator.hpp
```diff
@@ -40,11 +41,12 @@ class DisplacementHyperelasticIntegrator : public mfem::NonlinearFormIntegrator
    * @brief The constructor for the displacement hyperelastic integrator
    *
    * @param[in] m  HyperelasticModel that will be integrated.
+   * @param[in] thermal_m Optional thermal expansion material model
```
samuelpmishLLNL: Could we clarify how "optional" enabling of thermal expansion works, e.g. "if `thermal_m` is a null pointer, thermal expansion is disabled"

Alternatively, we could just provide an additional constructor
```cpp
  explicit DisplacementHyperelasticIntegrator(HyperelasticMaterial& m,
                                               GeometricNonlinearities geom_nonlin = GeometricNonlinearities::On)
       : material_(m), thermal_material_{}, geom_nonlin_(geom_nonlin)
   {
   }
```
so the user doesn't have to explicitly write `nullptr` in one of the arguments if they don't want it

src/serac/physics/materials/thermal_expansion_material.hpp
jamiebramwell: A fair point. I was following the pattern of the hyperelastic model.

src/serac/physics/materials/thermal_expansion_material.cpp
jamiebramwell: I believe this is correct. We commonly want material models to act on the displacement gradient, so that's why I'm modifying it here. My thinking is:

<a href="https://www.codecogs.com/eqnedit.php?latex=\begin{align*}&space;\mathbf{F}_\textbf{total}&space;&=&space;\mathbf{F}_\textbf{thermal}\mathbf{F}_\textbf{mechanical}&space;\\&space;&=&space;(\mathbf{I}&plus;\mathbf{H}_\textbf{thermal})(\mathbf{I}&plus;\mathbf{H}_\textbf{mechanical})&space;\\&space;&=&space;\mathbf{I}&space;&plus;&space;\mathbf{H}_\textbf{thermal}&space;&plus;&space;\mathbf{H}_\textbf{mechanical}&space;&plus;&space;\mathbf{H}_\textbf{thermal}\mathbf{H}_\textbf{mechanical}&space;\\&space;&=&space;\mathbf{I}&space;&plus;&space;\mathbf{H}_\textbf{modified}&space;\\&space;\mathbf{H}_\textbf{thermal}&space;&=&space;\mathbf{F}_\textbf{thermal}&space;-&space;\mathbf{I}&space;\\&space;&=&space;\alpha&space;\left(&space;\theta&space;-&space;\theta_\textbf{ref}\right)\mathbf{I}&space;\end{aligh*}" target="_blank"></a>

jamiebramwell: The last term in F_total is due to geometric nonlinearities. I was considering adding a flag to turn in off.

samuelpmishLLNL: sounds good, let's try to document this assumption (neglecting the product term) in the doxygen comment for `modifyDisplacementGradient`, if possible

samuelpmishLLNL (approved):
Looks good! 

As someone who isn't very familiar with how thermal expansion actually gets implemented, I appreciated the in-line comments explaining the relevant theory.

goxberry (approved):
My solid mechanics is shaky at best, but from a quick skim of the relevant sections of Holzapfel, this implementation LGTM.

src/serac/infrastructure/initialize.cpp
```diff
@@ -95,7 +95,8 @@ std::pair<int, int> initialize(int argc, char* argv[], MPI_Comm comm)
   profiling::initializeCaliper();
 
   // Initialize GPU (no-op if not enabled/available)
-  accelerator::initializeDevice();
+  // TODO for some reason this causes errors on Lassen. We need to look into this ASAP.
+  // accelerator::initializeDevice();
```
jamiebramwell, line 99: The lassen builds are currently broken due to a device initialization call. We need to figure out the root cause and re-enable this.

PR #568, opened 2021-09-27
Add Quantities of Interest and Global Gradient Assembly for Boundary Integrals

samuelpmishLLNL:
The main contribution of this PR is a partial template specialization for `serac::Functional`, where the "return type" template argument is of type `double`. This is used to denote that the integrands provided in `AddBoundaryIntegral`, `AddDomainIntegral` are not integrated against test functions, but integrated directly instead (alternatively: they are integrated against a single test function: `f(x) := 1`). The intention is that this will allow `serac::Functional` to be used to calculate quantities of interest, objective functions, and constraint functions (as well as Galerkin-method residuals).

For example, the mathematical expression (where `u` is, say, a temperature field)

ql_7a7df75ce7f209cd5e6f31bdfa301fc0_l3
can be expressed in the following way using `serac::Functional`:

```cpp
Functional<double(trial_space)> f(&fespace);
f.AddVolumeIntegral([&](auto x, auto temperature) {
    auto [u, grad_u] = temperature;
    return x[0] * x[0] + sin(x[1]) + x[0] * u * u * u;
}, mesh);
f.AddSurfaceIntegral([&](auto x, auto /*n*/, auto u) { 
  return x[0] - x[1] + cos(u * x[1]); 
}, mesh);
```

Additionally, `serac::Functional::AddBoundaryIntegral` now properly supports assembling the gradient to an `mfem::Vector` (for the case of a quantity of interest) or an `mfem::SparseMatrix` (for the case of a finite-element-type residual). At present, our implementation is based on `mfem::FaceRestriction`, and as a result, we inherit some of the limitations of that object (mfem does not support some function spaces).


src/serac/physics/base_physics.cpp
```diff
@@ -205,21 +205,22 @@ void BasePhysics::saveSummary(axom::sidre::DataStore& datastore, const double t)
 {
   double l1norm_value, l2norm_value, linfnorm_value, avg_value, max_value, min_value;
 
-  axom::sidre::Group* sidre_root;
-  axom::sidre::Group* curves_group;
-  auto [_, rank] = getMPIInfo();
+  // this change was made to work around the following error in GCC:
```
jamiebramwell: I think this is OK. Any thoughts @white238 ?

src/serac/physics/utilities/functional/array.hpp
```diff
@@ -0,0 +1,376 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
```
jamiebramwell, line 2: In the quadrature data GPU branch, we're using chai array. Should we use that here? Is the ultimate goal axom array? Thoughts @joshessman-llnl ?

src/serac/physics/base_physics.cpp
white238: This is good but I don't think we should start putting comments in the code for non-complicated compiler errors.

Thanks for fixing this @samuelpmishLLNL.  Did this show up in a newer gcc? I never showed up for mine

samuelpmishLLNL: I've been using gcc 10.2.0 on my local machine. Apparently, GCC is notorious for emitting this kind of error when it's not actually an issue (like here).

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,264 @@
...
+/**
+ * @brief this type explicitly stores sign (typically used conveying edge/face orientation) and index values
+ *
+ * TODO: investigate implementation via bitfield (should have smaller memory footprint, better readability than mfem's
+ * {sign, index} int32_t encoding)
+ */
+struct SignedIndex {
+  int index_;
+  int sign_;
+      operator int() { return index_; }
```
white238: Did this come out of clang-format?

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,264 @@
...
+ *    we choose to use the Restriction operator as the "source of truth", since we are also using its
+ *    convention for quadrature point numbering.
+ */
+struct DofNumbering {
+  DofNumbering(const mfem::ParFiniteElementSpace& fespace)
+      : element_dofs_(fespace.GetNE(), fespace.GetFE(0)->GetDof() * fespace.GetVDim()),
+        boundary_element_dofs_(fespace.GetNFbyType(mfem::FaceType::Boundary),
+                               fespace.GetBE(0)->GetDof() * fespace.GetVDim())
+  {
+    {
```
white238, line 169: Is there a reason you added two new internal scopes to this?

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,264 @@
...
+ */
+struct GradientAssemblyLookupTables {
+  GradientAssemblyLookupTables(mfem::ParFiniteElementSpace& test_fespace, mfem::ParFiniteElementSpace& trial_fespace)
+      : element_nonzero_LUT(trial_fespace.GetNE(), test_fespace.GetFE(0)->GetDof() * test_fespace.GetVDim(),
+                            trial_fespace.GetFE(0)->GetDof() * trial_fespace.GetVDim()),
+        boundary_element_nonzero_LUT(trial_fespace.GetNFbyType(mfem::FaceType::Boundary),
+                                     test_fespace.GetBE(0)->GetDof() * test_fespace.GetVDim(),
+                                     trial_fespace.GetBE(0)->GetDof() * trial_fespace.GetVDim())
+  {
+    DofNumbering test_dofs(test_fespace);
```
white238, line 252: Could you add some short comments to this function to explain which section is doing?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -90,17 +91,17 @@ class DomainIntegral {
     //       to allow the evaluation kernel to pass derivative values to the gradient kernel
 
     if constexpr (exec == ExecutionSpace::CPU) {
-      evaluation_ = [this, qf_derivatives, num_elements, qf](const mfem::Vector& U, mfem::Vector& R) {
+      evaluation_ = [this, Q, qf_derivatives, num_elements, qf](const mfem::Vector& U, mfem::Vector& R) {
```
white238: How did this work before? I would have expected this to error out due to Q not being captured.

Though `Q` is only used in the template instantiation along with `test_space` and `trial_space`

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -299,12 +302,19 @@ void element_gradient_kernel(mfem::Vector& K_e, derivatives_type* derivatives_pt
...
-      auto N = trial_element::shape_functions(xi_q);
-
-      for (int i = 0; i < test_ndof; i++) {
+      if constexpr (std::is_same<test, QOI>::value) {
+        auto N = trial_element::shape_functions(xi_q);
         for (int j = 0; j < trial_ndof; j++) {
-          K_elem[i][j] += M[i] * dq_darg * N[j] * dx;
+          K_elem[0][j] += dq_darg * N[j] * dx;
+        }
+      } else {
```
white238: do we know why this doesnt hit the same constexpr compiler bug we were hitting before?

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -423,7 +291,9 @@ class Functional<test(trial), exec> : public mfem::Operator {
     trial_space_->GetEssentialTrueDofs(ess_attr, ess_tdof_list_);
   }
 
-private:
+  // TODO reenable privacy after debugging
+  // private:
```
white238: reenable this

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,517 @@
...
+ */
+
+namespace serac {
+
+struct QoIProlongation : public mfem::Operator {
+  QoIProlongation(MPI_Comm c) : mfem::Operator(1, 1), comm(c) {}
+
+  void Mult(const mfem::Vector&, mfem::Vector&) const override
+  {
+    std::cout << "QoIProlongation::Mult() is not defined, exiting..." << std::endl;
```
white238: Why is this not using SLIC and our exit gracefully function?

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,517 @@
...
+
+  MPI_Comm comm;
+};
+
+struct QoIElementRestriction : public mfem::Operator {
+  QoIElementRestriction(int num_elements) : mfem::Operator(num_elements, 1) {}
+
+  void Mult(const mfem::Vector&, mfem::Vector&) const override
+  {
+    std::cout << "QoIElementRestriction::Mult() is not defined, exiting..." << std::endl;
```
white238: Same here as above

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,517 @@
...
+   * @note The @p Dimension parameters are used to assist in the deduction of the @a geometry_dim
+   * and @a spatial_dim template parameter
+   */
+  template <int dim, typename lambda, typename qpt_data_type = void>
+  void AddDomainIntegral(Dimension<dim>, lambda&& integrand, mfem::Mesh& domain)
+  {
+    auto num_elements = domain.GetNE();
+    if (num_elements == 0) return;
+
+    SLIC_ERROR_ROOT_IF(dim != domain.Dimension(), "Error: invalid mesh dimension for domain integral");
```
white238: You do not need to put "Error: " here it's automatically added by SLIC

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,517 @@
...
+   * and @a spatial_dim template parameter
+   */
+  template <int dim, typename lambda, typename qpt_data_type = void>
+  void AddBoundaryIntegral(Dimension<dim>, lambda&& integrand, mfem::Mesh& domain)
+  {
+    // TODO: fix mfem::FaceGeometricFactors
+    auto num_boundary_elements = domain.GetNBE();
+    if (num_boundary_elements == 0) return;
+
+    SLIC_ERROR_ROOT_IF((dim + 1) != domain.Dimension(), "Error: invalid mesh dimension for boundary integral");
```
white238: Same.

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  /**
+   * @brief Adds a boundary integral term to the weak formulation of the PDE
+   * @tparam dim The dimension of the boundary element (1 for line, 2 for quad, etc)
+   * @tparam lambda the type of the integrand functor: must implement operator() with an appropriate function signature
+   * @param[in] integrand The user-provided quadrature function, see @p Integral
+   * @param[in] domain The domain on which to evaluate the integral
+   * @param[in] data The data structure containing per-quadrature-point data
+   * @note The @p Dimension parameters are used to assist in the deduction of the @a geometry_dim
```
white238, line 168: nitpick: lower case `Dimension`

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,517 @@
...
+   */
+  mutable mfem::Vector input_E_;
+
+  /**
+   * @brief The output set of per-element DOF values
+   */
+  mutable mfem::Vector output_E_;
+
+  /**
+   * @brief The input set of per-boundaryelement DOF values
```
white238: this comment and below are inconsistent `per-boundaryelement` vs `per-boundary-element`

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,517 @@
...
+                                                             mfem::Element::QUADRILATERAL, mfem::Element::HEXAHEDRON};
+
+  /// @brief The gradient object used to implement @p GetGradient
+  mutable Gradient grad_;
+
+  /// @brief array that stores each element's gradient of the residual w.r.t. trial values
+  Array<double, 3, exec> element_gradients_;
+
+  /// @brief array that stores each boundary element's gradient of the residual w.r.t. trial values
+  Array<double, 3, exec> belement_gradients_;
```
white238: `belement` is used a lot of in this PR.  I say this looks too much like a typo.  This should be `bdr_element` or something similiar. @jamiebramwell ?

src/serac/physics/utilities/functional/integral_utilities.hpp
```diff
@@ -427,4 +437,94 @@ SERAC_HOST_DEVICE auto Postprocess(const T& f, const tensor<double, dim>& xi, co
...
+
+/**
+ * @brief a (poorly named) tuple of quantities used to discover the sparsity
+ * pattern associated with element and boundary element matrices.
+ *
+ * It stores information about how the entries of an element "stiffness" matrix
+ * map to the global stiffness. The operator< definition allows us to sort them
+ * lexicographically, to facilitate creating the CSR matrix graph.
+ */
+struct ElemInfo {
```
white238: why is this struct duplicated?

src/serac/physics/utilities/functional/tests/array_test.cpp
```diff
@@ -0,0 +1,56 @@
...
+  for (uint32_t i = 0; i < values.size(0); i++) {
+    for (uint32_t j = 0; j < values.size(1); j++) {
+      values(i, j) = i + j;
+    }
+  }
+}
+
+int main()
+{
+  serac::CPUArray<double, 2> my_array(N1, N2);
```
white238, line 36: This and the cuda tests need to check actual values not just print values.

src/serac/physics/utilities/functional/dof_numbering.hpp
samuelpmishLLNL: It did-- apparently `clang-format` thinks the word `operator` is a name, and tries to align it with the other names. Adding a line of whitespace fixes it!

src/serac/physics/utilities/functional/dof_numbering.hpp
samuelpmishLLNL: When I was writing this code, the element and boundary cases are almost identical, except that the quantities have slightly different names (e.g. `num_elements` vs `num_boundary_elements`, `dofs_per_element` vs. `dofs_per_boundary_element`). So, when I copied and pasted the element implementation and modified it to change the variables, I managed to miss a few name changes, which introduced subtle bugs as a result of having the wrong sizes. 

So, when I implement the element and boundary element loops in this file, I put them in their own scopes to make sure that those temporary variables wouldn't be accessible from the other's implementation, in an effort to let the compiler notify if a variable name didn't get changed appropriately.

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: which bug are you referring to?

src/serac/physics/utilities/functional/functional_qoi.inl
samuelpmishLLNL: this note is referring to the first parameter, `Dimension<dim>`

src/serac/physics/utilities/functional/dof_numbering.hpp
white238: excellent! it wasn't a criticism at all i was just curious about your thought process.

src/serac/physics/utilities/functional/functional_qoi.inl
samuelpmishLLNL: I use the full word `boundary` much of the time, but it does add quite a few characters. I agree it should be made consistent. The three naming options that come to mind immediately are (corresponding to the boundary version of some variable `element_quantity`):

`belement_quantity`: shortest, but least clear
`bdr_element_quantity`: somewhere in the middle, similar to what mfem does some of the time
`boundary_element_quantity`: longest, but most clear

but other suggestions are welcomed

src/serac/physics/utilities/functional/integral_utilities.hpp
samuelpmishLLNL: oops, I thought I deleted original implementation when moving it to `DofNumbering.hpp`

src/serac/physics/utilities/functional/tests/array_test.cpp
samuelpmishLLNL: I had not planned to keep these tests in the long term (or even the near term, perhaps they should not even go in with this PR). If you prefer, I could flesh them out into proper tests and keep them, though.

src/serac/physics/utilities/functional/functional_qoi.inl
samuelpmishLLNL: Thanks for the clarification. There were similar mistakes made in `functional.hpp` (where this code was copied from) that have now been fixed, too.

src/serac/physics/utilities/functional/tests/array_test.cpp
white238: that's fine, especially since this is very temporary. just add a short comment.

src/serac/physics/utilities/functional/boundary_integral.hpp
joshessman-llnl: I believe Chris is referring to the spurious warnings NVCC sometimes issues for return statements within `if constexpr` blocks.  In this case I don't think that bug applies because the function doesn't return from within either block (or at all, because it's a `void` function).

src/serac/physics/utilities/functional/domain_integral.hpp
samuelpmishLLNL: I think this change can be reverted. I was getting some cryptic compiler errors when refactoring, and I (incorrectly) thought they might be related to a weird capture issue. The C++ standard says you shouldn't have to capture `constexpr` quantities to use them in a lambda, but some compilers get that wrong (MSVC, for example).

src/serac/physics/utilities/functional/boundary_integral.hpp
samuelpmishLLNL: ah, it's probably the nvcc `if constexpr` one. I think that bug shows up when returning from the body of an `if constexpr` statement, and nvcc (incorrectly) thinks that the function is missing a return statement. There are no return statements in this `if constexpr`, though.

white238: thanks for the clarification... it just tickled my brain

src/serac/physics/utilities/functional/array.hpp
jamiebramwell: Nevermind! Per our discussion earlier this week, let's merge these branches as-is and consolidate to axom::array as a follow up.

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -313,23 +323,30 @@ void element_gradient_kernel(mfem::Vector& K_e, derivatives_type* derivatives_pt
     // out to memory, to be later assembled into the global gradient by mfem
     //
     // Note: we "transpose" these values to get them into the layout that mfem expects
-    for_loop<test_ndof, test_dim, trial_ndof, trial_dim>(
-        [&](int i, int j, int k, int l) { dk(i + test_ndof * j, k + trial_ndof * l, e) += K_elem[i][k][j][l]; });
+    // clang-format off
+    for_loop<test_ndof, test_dim, trial_ndof, trial_dim>([&](int i, int j, int k, int l) {
```
jamiebramwell: Explicit capturing of variables in lambdas is safer.

src/serac/physics/utilities/functional/functional_qoi.inl
jamiebramwell: I vote for `bdr_element_quantity`. I think the 3 extra characters really improve readability.

src/serac/physics/utilities/functional/array.hpp
```diff
@@ -0,0 +1,376 @@
...
+#include <stddef.h>
+
+#include <iostream>
+#include <algorithm>
+
+#include "serac/infrastructure/accelerator.hpp"
+
+namespace serac {
+
+namespace impl {
```
joshessman-llnl: Minor: I think we try to stick with `detail` for these kinds of things

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,282 @@
...
+ * @brief operator determining inequality by {global_row, global_col}
+ * @param x the ElemInfo on the left
+ * @param y the ElemInfo on the right
+ */
+bool operator!=(const ElemInfo& x, const ElemInfo& y)
+{
+  return (x.global_row_ != y.global_row_) || (x.global_col_ != y.global_col_);
+}
+
+auto& operator<<(std::ostream& out, ElemInfo e)
```
joshessman-llnl: Minor: `const ElemInfo&` here

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -268,21 +272,20 @@ void action_of_gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivat
  * @param[in] num_elements The number of elements in the mesh
  */
 template <Geometry g, typename test, typename trial, int Q, typename derivatives_type>
-void element_gradient_kernel(mfem::Vector& K_e, derivatives_type* derivatives_ptr, const mfem::Vector& J_,
-                             int num_elements)
+void element_gradient_kernel(ArrayView<double, 3, ExecutionSpace::CPU> dk, derivatives_type* derivatives_ptr,
```
joshessman-llnl: Minor: `CPUView` alias here?

src/serac/physics/utilities/functional/domain_integral.hpp
```diff
@@ -100,7 +101,7 @@ class DomainIntegral {
                                                                                          J_, num_elements);
       };
 
-      element_gradient_ = [this, qf_derivatives, num_elements](mfem::Vector& K_e) {
+      element_gradient_ = [this, qf_derivatives, num_elements](ArrayView<double, 3, ExecutionSpace::CPU> K_e) {
```
joshessman-llnl: Minor: see above on using the `CPUView` alias here

src/serac/physics/utilities/functional/functional.hpp
```diff
@@ -692,56 +375,36 @@ class Functional<test(trial), exec> : public mfem::Operator {
...
-
     /**
-     * @brief whether or not the sparse matrix graph has been determined
-     *
-     * note: this is not done automatically, since a user might only require
-     * the action of the gradient (which does not need this sparsity pattern),
-     * so the sparsity pattern is only initialized the first time the sparse
-     * matrix is formed, and then cached. This assumes the sparsity pattern does
-     * not change dynamically.
+     * @brief TODO
```
joshessman-llnl: assume this just means docs needed here?

src/serac/physics/utilities/functional/functional_qoi.inl
joshessman-llnl: Note that `exitGracefully` is not needed here if slic is used - it will be called automatically from a `SLIC_ERROR`

src/serac/physics/utilities/functional/boundary_integral_kernels.hpp
```diff
@@ -0,0 +1,322 @@
...
+      auto qf_output = qf(x_q, n_q, make_dual(arg));
+
+      // integrate qf_output against test space shape functions / gradients
+      // to get element residual contributions
+      r_elem += Postprocess<test_element>(get_value(qf_output), xi) * dx;
+
+      // here, we store the derivative of the q-function w.r.t. its input arguments
+      //
+      // this will be used by other kernels to evaluate gradients / adjoints / directional derivatives
+      derivatives_ptr[e * int(rule.size()) + q] = get_gradient(qf_output);
```
jamiebramwell: Is there a reason we can't use `detail::AccessDerivatives` here?

src/serac/physics/utilities/functional/boundary_integral_kernels.hpp
```diff
@@ -0,0 +1,322 @@
...
+      // and calculate the measure of that point in physical space.
+      auto   xi  = rule.points[q];
+      auto   dxi = rule.weights[q];
+      double dx  = J(q, e) * dxi;
+
+      // evaluate the (change in) value/derivatives at this quadrature point
+      auto darg = Preprocess<trial_element>(du_elem, xi);
+
+      // recall the derivative of the q-function w.r.t. its arguments at this quadrature point
+      auto dq_darg = derivatives_ptr[e * int(rule.size()) + q];
```
jamiebramwell: `detail::AccessDerivatives`?

src/serac/physics/utilities/functional/boundary_integral_kernels.hpp
```diff
@@ -0,0 +1,322 @@
...
+    // for each quadrature point in the element
+    for (int q = 0; q < static_cast<int>(rule.size()); q++) {
+      // get the position of this quadrature point in the parent and physical space,
+      // and calculate the measure of that point in physical space.
+      auto   xi_q  = rule.points[q];
+      auto   dxi_q = rule.weights[q];
+      double dx    = J(q, e) * dxi_q;
+
+      // recall the derivative of the q-function w.r.t. its arguments at this quadrature point
+      auto dq_darg = derivatives_ptr[e * int(rule.size()) + q];
```
jamiebramwell: `detail::AccessDerivatives`?

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,356 @@
...
+/**
+ * @brief a (poorly named) tuple of quantities used to discover the sparsity
+ * pattern associated with element and boundary element matrices.
+ *
+ * It stores information about how the entries of an element "stiffness" matrix
+ * map to the global stiffness. The operator< definition allows us to sort them
+ * lexicographically, to facilitate creating the CSR matrix graph.
+ */
+struct ElemInfo {
+  uint32_t global_row_;   ///< The global row number
```
jamiebramwell, line 18: What is the current best practice for `uint32_t` vs `size_t`? Thoughts @joshessman-llnl @white238 ?

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,356 @@
...
+  return (fes.FEColl()->GetContType() == mfem::FiniteElementCollection::DISCONTINUOUS);
+}
+
+/**
+ * @param fes the finite element space in question
+ *
+ * @brief attempt to characterize which FiniteElementSpaces
+ * mfem::FaceRestriction actually works with
+ */
+bool supports_bdr_stuff(const mfem::ParFiniteElementSpace& fes)
```
jamiebramwell: I would be more explicit in the function name. Something like `compatible_with_face_restriction`.

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,356 @@
...
+ *               getting dofs/ids for boundary elements in 2D,3D w/ L2 spaces
+ */
+template <typename T, ExecutionSpace exec>
+serac::Array<T, 3, exec> guard_against_unimplemented_bdr_stuff(const mfem::ParFiniteElementSpace& trial_fes,
+                                                               const mfem::ParFiniteElementSpace& test_fes)
+{
+  if (supports_bdr_stuff(test_fes) && supports_bdr_stuff(trial_fes)) {
+    auto* test_BE  = test_fes.GetBE(0);
+    auto* trial_BE = trial_fes.GetBE(0);
+    return serac::Array<T, 3, exec>(static_cast<size_t>(trial_fes.GetNFbyType(mfem::FaceType::Boundary)),
```
jamiebramwell, line 127: What is this actually returning? The arguments and return data needs documentation.

src/serac/physics/utilities/functional/dof_numbering.hpp
```diff
@@ -0,0 +1,356 @@
...
+bool operator!=(const ElemInfo& x, const ElemInfo& y)
+{
+  return (x.global_row_ != y.global_row_) || (x.global_col_ != y.global_col_);
+}
+
+/**
+ * @brief mfem will frequently encode {sign, index} into a single int32_t.
+ * This function decodes the sign from such a type.
+ */
+int mfem_sign(int i) { return (i >= 0) ? 1 : -1; }
```
jamiebramwell: Methods should be in `camelCase`

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
```diff
@@ -236,11 +236,12 @@ TEST(boundary, 2D_quadratic) { boundary_test(*mesh2D, H1<2>{}, H1<2>{}, Dimensio
 TEST(boundary, 3D_linear) { boundary_test(*mesh3D, H1<1>{}, H1<1>{}, Dimension<3>{}); }
 TEST(boundary, 3D_quadratic) { boundary_test(*mesh3D, H1<2>{}, H1<2>{}, Dimension<3>{}); }
 
-TEST(boundary_L2, 2D_linear) { boundary_test(*mesh2D, L2<1>{}, L2<1>{}, Dimension<2>{}); }
-TEST(boundary_L2, 2D_quadratic) { boundary_test(*mesh2D, L2<2>{}, L2<2>{}, Dimension<2>{}); }
-
-TEST(boundary_L2, 3D_linear) { boundary_test(*mesh3D, L2<1>{}, L2<1>{}, Dimension<3>{}); }
-TEST(boundary_L2, 3D_quadratic) { boundary_test(*mesh3D, L2<2>{}, L2<2>{}, Dimension<3>{}); }
+// TODO: reenable when mfem implement support for L2
+// TEST(boundary_L2, 2D_linear) { boundary_test(*mesh2D, L2<1>{}, L2<1>{}, Dimension<2>{}); }
```
jamiebramwell, line 241: How did this work before but not now?

src/serac/physics/utilities/functional/tests/functional_comparisons.cpp
```diff
@@ -137,9 +140,7 @@ void functional_test(mfem::ParMesh& mesh, H1<p> test, H1<p> trial, Dimension<dim
   // Compute the gradient using functional
   mfem::Operator& grad2 = SERAC_PROFILE_EXPR(concat("functional_GetGradient", postfix), residual.GetGradient(U));
 
-  // Test fully assembled matrix
-  mfem::Array<int> dofs;
-  fespace.GetElementDofs(0, dofs);
+#if enable_assembled_sparsematrix_stuff
```
jamiebramwell: Why is the assembled sparse matrix turned off? GPU reasons?

src/serac/physics/utilities/functional/boundary_integral_kernels.hpp
samuelpmishLLNL: `detail::AccessDerivatives` is a weird function: it takes a lot of parameters and is less clear than just indexing into the pointer. Also, these are going to be imminently removed in favor of the axom::Array stuff, so I'd prefer to just not worry about this right now (these will all be refactored away in the variadic branch anyway, since the derivative info will be modified)

src/serac/physics/utilities/functional/dof_numbering.hpp
samuelpmishLLNL: I'll just comment briefly on the motivation here: these indices come from mfem, which uses `int` for everything. So, there's not much of a benefit to storing them in a larger integer type, since the data will never use the extra space. I did make the change to unsigned to avoid the `signed vs unsigned` conversion warnings, but I think it also is clearer than just `int`, since these values should never be negative.

src/serac/physics/utilities/functional/tests/functional_comparisons.cpp
samuelpmishLLNL: Actually, the entire `AssembledSparseMatrix` class is completely unused now, it could probably be removed. Should I do that now, or on a future PR?

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
samuelpmishLLNL: When this test was introduced, the boundary integrals didn't support calculating gradients (only evaluation and action of gradient). Now, `Functional` does more work in its constructor to set up all of the lookup tables it needs to assemble element/bdr_element gradient contributions. `mfem` treats the L2 finite element space differently from the rest w.r.t. what they call "boundary elements", so some of the calls to set up those lookup tables segfault.

src/serac/physics/utilities/functional/tests/restriction_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+  fespace.ExchangeFaceNbrData();
+
+  const mfem::IntegrationRule& ir = mfem::IntRules.Get(mfem::Element::QUADRILATERAL, 4);
+  constexpr auto flags            = mfem::FaceGeometricFactors::COORDINATES | mfem::FaceGeometricFactors::DETERMINANTS |
+                         mfem::FaceGeometricFactors::NORMALS;
+
+  // despite what their documentation says, mfem doesn't actually support the JACOBIANS flag.
+  // this is currently a dealbreaker, as we need this information to do any calculations
+  [[maybe_unused]] auto geom = mesh3D->GetFaceGeometricFactors(ir, flags, mfem::FaceType::Boundary);
```
joshessman-llnl: Is this test just intended to make sure that this line doesn't crash?

src/serac/physics/utilities/functional/tests/functional_qoi.cpp
```diff
@@ -0,0 +1,216 @@
...
+  mesh3D = mesh::refineAndDistribute(buildMeshFromFile(meshfile3D), serial_refinement, parallel_refinement);
+
+  functional_qoi_test(*mesh2D, H1<2>{}, Dimension<2>{});
+  functional_qoi_test(*mesh3D, H1<1>{}, Dimension<3>{});
+
+#if 0
+  int result = RUN_ALL_TESTS();
+  MPI_Finalize();
+  return result;
+#else
```
joshessman-llnl: Would it be possible to set these up as "regular" gtests so we can get more granular reporting?

src/serac/physics/utilities/functional/tests/functional_comparisons.cpp
jamiebramwell: If it's unused and will not be used in the future, it should be removed.

src/serac/physics/utilities/functional/tests/functional_boundary_test.cpp
jamiebramwell: ahhh gotcha. Could you capture this comment in the TODO?

src/serac/physics/utilities/functional/tests/functional_comparisons.cpp
```diff
@@ -17,10 +17,12 @@
 #include "serac/physics/operators/stdfunction_operator.hpp"
 #include "serac/physics/utilities/functional/functional.hpp"
 #include "serac/physics/utilities/functional/tensor.hpp"
-#include "serac/numerics/assembled_sparse_matrix.hpp"
 #include "serac/infrastructure/profiling.hpp"
 #include <gtest/gtest.h>
 
+// TODO remove AssembledSparseMatrix completely and refactor these tests
```
joshessman-llnl: This is all gone now right?

src/serac/physics/utilities/functional/tests/functional_basic.cpp
```diff
@@ -0,0 +1,113 @@
...
+  double relative_error1 = df1.DistanceTo(df2) / df1.Norml2();
+  double relative_error2 = df1.DistanceTo(df3) / df1.Norml2();
+
+  EXPECT_NEAR(0., relative_error1, 1.e-6);
+  EXPECT_NEAR(0., relative_error2, 1.e-6);
+
+  delete dfdU_matrix;
+}
+
+int main(int argc, char* argv[])
```
joshessman-llnl: Would it be worth refactoring this into a `gtest`?

src/serac/physics/utilities/functional/tests/array_test_cuda.cu
```diff
@@ -0,0 +1,20 @@
...
+{
+  serac::GPUArray<double, 2> my_array(32, 32);
+
+  fill_kernel<<<32, 32>>>(my_array);
+
+  serac::CPUArray<double, 2> my_array_h = my_array;
+
+  for (int i = 0; i < 4; i++) {
+    std::cout << my_array_h(4, i) << std::endl;
+  }
```
joshessman-llnl: Suggestion: could we make this test a little more thorough by checking all the values are as expected?

Probably not a big deal either way (per the discussion on the CPU array test) because this class is temporary

src/serac/physics/utilities/functional/tests/CMakeLists.txt
```diff
@@ -49,9 +56,13 @@ endforeach()
 
 if(ENABLE_CUDA)
 
+    # TODO: serac::initializeDevice is currently broken,
+    # until it gets fixed, these tests are disabled to avoid CI
+    # preventing pull requests from going through
```
joshessman-llnl: This block can be removed now I think

joshessman-llnl (commented):
Looks good, just some minor cleanup comments!

jamiebramwell:
ping @sethwatts

src/serac/physics/utilities/functional/dof_numbering.hpp
tepperly: I think size_t needs to be defined to be big enough to hold the largest possible data structure for the architecture, For example on a 64-bit processor, I think it basically has to be 64-bit. Otherwise, one can't have a std::vector<char> that's bigger than 4GB even if you're machine has 32GB.
The choice for what type to use for global_row_ should be decided based on how big you expect that kind of data size to get. I think that fixing the data sizes for different types was one of the things that Java got right.

src/serac/physics/utilities/functional/tests/functional_basic.cpp
jamiebramwell: I agree: this should be a gtest.

src/serac/physics/utilities/functional/tests/hypreparvector_test.cpp
```diff
@@ -0,0 +1,100 @@
...
+        "x_global * x_global: %f\n"
+        "x_local * x_local: %f\n\n",
+        InnerProduct(x_global, x_global), xTx);
+  }
+
+  {
+    mfem::Vector U = x_global;
+
+    U(0) = 42.0;
+    if (myid == 0) std::cout << U(0) << std::endl;
```
jamiebramwell: I don't think this needs to be changed here, but the `SLIC_INFO_ROOT` macro is designed for this purpose.

src/serac/physics/utilities/functional/tests/hypreparvector_test.cpp
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
```
jamiebramwell: What is the purpose of this test?

samuelpmishLLNL: I hope this one will eventually become an mfem unit test, it highlights inconsistencies between the behaviors of `mfem::Vector` and `mfem::HypreParVector`. Namely, copy semantics are different, and when an `mfem::HypreParVector` is upcast to `mfem::Vector` (e.g. when it is passed to a function that takes in an `mfem::Vector &`), the behavior changes from "global" to "local", which seems like a bug in the interface design.

samuelpmishLLNL: perhaps it doesn't make sense to keep it in this branch though

src/serac/physics/utilities/functional/tests/hypreparvector_test.cpp
samuelpmishLLNL: I should have explained this in a comment, but I'm trying to make this file have no external dependencies, so that it might be easily added to mfem in a future PR (related to interface changes in `mfem::HypreParVector`)

jamiebramwell: ahhh that makes sense.

src/serac/physics/utilities/functional/tests/hypreparvector_test.cpp
jamiebramwell: Yeah, I think it's good to keep around  (maybe in an MFEM branch?) but it doesn't really help with our testing.

jamiebramwell (approved):

jamiebramwell:
Looks good to me! Can the other reviewers take a look and point out anything that still needs work? Thanks!

src/serac/physics/utilities/functional/finite_element.hpp
```diff
@@ -109,6 +116,17 @@ struct L2 {
   static constexpr Family family     = Family::L2;  ///< the family of the basis functions
 };
 
+/**
```
sethwatts, line 119: Perhaps I'm mistaken, but this looks like a cut/paste from the L2 struct. Does the Doxygen for the QOI struct need to be updated?

src/serac/physics/utilities/functional/functional_qoi.inl
```diff
@@ -0,0 +1,548 @@
...
+  /// @brief destructor: deallocate the mfem::Operators that we're responsible for
+  ~Functional()
+  {
+    delete P_test_;
+    delete G_test_;
+    delete G_test_boundary_;
+  }
+
+  /**
+   * @brief Adds a domain integral term to the weak formulation of the PDE
```
sethwatts: Probably just me not understanding code structure, but if this is specialized to QOIs, I'm wondering if the statement that a term is added to the PDE's weak form is (necessarily) correct, in the sense that I typically wouldn't want my definition/creation of a QOI to affect the solution of the PDE.

src/serac/physics/utilities/functional/finite_element.hpp
samuelpmishLLNL: good catch!

src/serac/physics/utilities/functional/functional_qoi.inl
samuelpmishLLNL: Defining a QoI will not change the solution of the PDE. In fact, `Functional` doesn't actually solve PDEs at all, it just gives you tools to evaluate QoIs and residuals (as well as their derivatives), and then the user (e.g. serac) implements their own solution algorithm from there.

samuelpmishLLNL: I will also go back and update the docstrings to hopefully be more clear about this!

jamiebramwell: You can imagine `Functional` as a replacement for `NonlinearForm`. Setting up and solving the specific PDE is still up to the user, or in our case the physics module developer.

sethwatts: That was my understanding. The comment was just pointing out the the Doxygen made it sound like this method is adding to the weak for for you.

src/serac/physics/utilities/functional/boundary_integral.hpp
```diff
@@ -364,27 +75,39 @@ class BoundaryIntegral {
     using u_du_t          = typename detail::lambda_argument<trial_space, dim, dim + 1>::type;
     using derivative_type = decltype(get_gradient(qf(x_t{}, x_t{}, make_dual(u_du_t{}))));
 
-    std::shared_ptr<derivative_type[]> qf_derivatives(new derivative_type[num_quadrature_points]);
+    // allocate memory for the derivatives of the q-function at each quadrature point
+    //
+    // Note: ptr's lifetime is managed in an unusual way! It is captured by-value in one of the
+    // lambda functions below to augment the reference count, and extend its lifetime to match
+    // that of the DomainIntegral that allocated it.
```
goxberry: Should this `DomainIntegral` be `BoundaryIntegral`, or am I missing something?

src/serac/physics/utilities/functional/boundary_integral_kernels.hpp
```diff
@@ -0,0 +1,322 @@
...
+  }
+}
+
+/**
+ * @brief The base kernel template used to create create custom element stiffness matrices
+ * associated with finite element calculations
+ *
+ * @tparam test The type of the test function space
+ * @tparam trial The type of the trial function space
+ * The above spaces can be any combination of {H1, Hcurl, Hdiv (TODO), L2 (TODO)}
```
goxberry: Looks like `test` can also be `QOI`?

src/serac/physics/utilities/functional/domain_integral_kernels.hpp
```diff
@@ -194,16 +199,15 @@ void action_of_gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR, derivat
  * @tparam derivatives_type Type representing the derivative of the q-function w.r.t. its input arguments
  *
  *
- * @param[inout] K_e The full set of per-element element tangents [test_ndofs x test_dim, trial_ndofs x trial_dim]
- * @param[in] derivatives_ptr The address at which derivatives of the q-function with
- * respect to its arguments are stored
+ * @param[inout] dK 3-dimensional array storing the element gradient matrices
```
goxberry: Looks like this parameter should be `dk`?

src/serac/physics/utilities/functional/domain_integral_kernels.hpp
```diff
@@ -233,35 +236,60 @@ void element_gradient_kernel(mfem::Vector& K_e, derivatives_type* derivatives_pt
...
 
       // recall the derivative of the q-function w.r.t. its arguments at this quadrature point
-      auto dq_darg = detail::AccessDerivatives(derivatives_ptr, e, q, rule, num_elements);
+      auto dq_darg = qf_derivatives(static_cast<size_t>(e), static_cast<size_t>(q));
 
-      auto& q00 = serac::get<0>(serac::get<0>(dq_darg));  // derivative of source term w.r.t. field value
-      auto& q01 = serac::get<1>(serac::get<0>(dq_darg));  // derivative of source term w.r.t. field derivative
-      auto& q10 = serac::get<0>(serac::get<1>(dq_darg));  // derivative of   flux term w.r.t. field value
-      auto& q11 = serac::get<1>(serac::get<1>(dq_darg));  // derivative of   flux term w.r.t. field derivative
+      if constexpr (std::is_same<test, QOI>::value) {
```
goxberry, line 241: Should `QOI` also be listed as a valid argument for `test` in the Doxygen comments for this function?

goxberry:
LGTM so far. I have 5 files left to review, all of which look to be tests of `Functional`; I probably won't be able to get to them until tomorrow. If you guys want to merge it before then, please do; I'm excited to use this capability.

joshessman-llnl (approved):
Looks good, glad we were able to sidestep the CHAI issues!

PR #570, opened 2021-10-04
docfix: update clang 9 to clang 10 in quickstart

goxberry:
This commit updates clang version in the quickstart documentation from
`%clang@9.0.0` to `%clang@10.0.0` because the former is not listed in
`serac/scripts/spack/configs/toss_3_x86_64_ib/{packages,compilers}.yaml`,
but the latter is.

A quick Linux shell one-liner in the repo to confirm that
`%clang@9.0.0` is missing from those files is

```bash
find $(git rev-parse --show-toplevel) -name "*.yaml" -print0 | xargs -0 grep -Hin "clang@9.0.0"
```

Replacing `9.0.0` with `10.0.0` in the above command confirms that
`clang@10.0.0` is present, hence the update.


white238 (approved):
Thanks @goxberry

jamiebramwell (approved):

PR #571, opened 2021-10-04
serac spack pkg: add +asan variant

### LINKED ISSUE #531: Add `asan` variant to serac package
Downstream customers have asked to be able to toggle our ability to build our code with AddressSanitizer turned on.  Having this turned on in many packages allows better memory error detection. This should be easy in the serac spack package.

goxberry:
This commit adds a `+asan` variant to the Serac Spack package to
expose more readily the `ENABLE_ASAN` option to users building Serac
through Spack and Uberenv. This change is probably more convenient for
downstream users of Serac that build it as a dependency of their
projects via Spack/Uberenv.

Resolves #531.


scripts/spack/packages/serac/package.py
```diff
@@ -47,6 +47,8 @@ class Serac(CachedCMakePackage, CudaPackage):
             description='Enable runtime safety and debug checks')
     variant('shared',   default=False,
             description='Enable build of shared libraries')
+    variant('asan', default=False,
```
white238, line 51: Can you add a conflict if the compiler is not clang or gcc?

goxberry: Good catch! Totally forgot about that restriction.

goxberry: Added checks in [`8916e60f`](https://github.com/LLNL/serac/pull/571/commits/8916e60f5852ee7247e62886bb762fcbebc2d9bc). The checks might be overkill -- comments explain the approach.

white238 (approved):

jamiebramwell:
I'm going to go investigate why this is failing CI

jamiebramwell (approved):
Thanks @goxberry ! Sorry for the delay. CI woes. 😭

Issue #567, opened 2021-09-23
investigate using precompiled headers to accelerate compilation


samuelpmishLLNL:
Profiling the compilation of sourcefiles in serac reveals a considerable amount of time is being spent parsing headers. 

Some of the more expensive ones include (no pun intended):

```cpp
#include <fstream>        // ~ 300 ms
#include "mfem.hpp"       // ~1250 ms (!)
#include "axom/inlet.hpp" // ~2200 ms (!!)
#include "functional.hpp" // ~ 150 ms

```

The values above were taken from a profile of compiling `functional_comparisons.cpp` with clang-10.0.0 (raw data available [here](https://github.com/LLNL/serac/files/7214646/functional_comparisons_profiling.zip), view flamegraph with `chrome://tracing`). These numbers represent the time to merely parse the header, even if it is unused in the rest of the translation unit (note: the times are even worse when compiling a sourcefile with NVCC, as each file is processed more than once). These headers are present in many of our translation units, meaning that these compilation costs are incurred dozens of times within serac. 

CMake 3.16+ supports defining precompiled headers that help to avoid paying this parsing cost in every translation unit. This small project [here](https://github.com/LLNL/serac/files/7214620/pch_test.zip) demonstrates the potential gain, using mfem's monolithic `mfem.hpp` header as an example:

without using precompiled headers:
```
sam@provolone:~/code/pch_test/build$ touch ../*.cpp && time make
Consolidate compiler generated dependencies of target mfem
[ 95%] Built target mfem
Consolidate compiler generated dependencies of target pch
[ 97%] Building CXX object CMakeFiles/pch.dir/foo.cpp.o
[ 97%] Building CXX object CMakeFiles/pch.dir/bar.cpp.o
[ 97%] Building CXX object CMakeFiles/pch.dir/baz.cpp.o
[ 97%] Building CXX object CMakeFiles/pch.dir/qux.cpp.o
[100%] Building CXX object CMakeFiles/pch.dir/main.cpp.o
[100%] Linking CXX executable pch
[100%] Built target pch

real	0m3.917s
user	0m3.537s
sys	0m0.374s
```

and after precompiling `mfem.hpp`:
```
sam@provolone:~/code/pch_test/build$ touch ../*.cpp && time make
Consolidate compiler generated dependencies of target mfem
[ 95%] Built target mfem
Consolidate compiler generated dependencies of target pch
[ 97%] Building CXX object CMakeFiles/pch.dir/foo.cpp.o
[...]

PR #572, opened 2021-10-05
modify `serac_add_tests` to support pch

samuelpmishLLNL:
see: https://github.com/LLNL/serac/issues/567

This PR adds a new optional argument, `PRECOMPILED_HEADER` to the cmake macro `serac_add_tests`, that lets the user specify a header file to be precompiled and applied to each of the sourcefiles passed to `serac_add_tests`. This feature is opt-in: 
code that does not pass this optional argument to `serac_add_tests` behaves as it did before (where each translation unit is compiled "from scratch").

Using `functional_boundary_test.cpp` as an arbitrary example, compilation times decreased significantly (6.0 seconds -> 3.0 seconds) by precompiling `pch.hpp`.

Some thoughts:
- Each call to `serac_add_tests` that specifies `PRECOMPILED_HEADER` will compile that header file once. Since serac makes many calls to `serac_add_tests`, some headers may be precompiled multiple times. This is still an improvement to compile times, as those headers are processed once per call to `serac_add_tests`, as opposed to once per translation unit. 
- As a result of the point above, Serac would compile even faster if we made few calls to `serac_add_tests` with many sourcefiles as opposed to many calls to `serac_add_tests` with a couple of sourcefiles. Is it worth restructuring our tests to do this or not?
- Right now, the specified header to be precompiled can be different for each call to `serac_add_tests`. Is this helpful, or should we just have a single precompiled header for all of serac, and make it globally applied to each translation unit?


cmake/SeracMacros.cmake
```diff
@@ -166,6 +168,30 @@ macro(serac_add_tests)
         blt_add_test(NAME          ${test_name}
                      COMMAND       ${test_name}
                      NUM_MPI_TASKS ${arg_NUM_MPI_TASKS} )
+
+        if(DEFINED arg_PRECOMPILED_HEADER)
+
+            # only precompile the header for the first test in this group
+            if (NOT DEFINED first_test) 
+
+                #message(${test_name} " creating PCH from " ${arg_PRECOMPILED_HEADER})
```
jamiebramwell: dead code

src/serac/physics/utilities/functional/tests/pch.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
jamiebramwell: I would explicitly rename this to "precompiled_headers.hpp" just so it's obvious what it is in the directory tree.

jamiebramwell: This also needs a copyright notice.

jamiebramwell (commented):
I like this as a first pass. We can obviously make further optimizations, but I think we should merge this as a first step. Thoughts @white238 ?

jamiebramwell:
Just noticed the larger questions in the description. Whoops!

> * As a result of the point above, Serac would compile even faster if we made few calls to `serac_add_tests` with many sourcefiles as opposed to many calls to `serac_add_tests` with a couple of sourcefiles. Is it worth restructuring our tests to do this or not?

I could go either way and defer to the groups' opinion. I think we have higher priorities right now though.
 
> * Right now, the specified header to be precompiled can be different for each call to `serac_add_tests`. Is this helpful, or should we just have a single precompiled header for all of serac, and make it globally applied to each translation unit?

Most of the time, all of the precompiled stuff will be the same. I think this is an elegant example and I doubt we would have much use for a more configurable list. I don't feel too strongly though.

PR #576, opened 2021-10-12
Hopefully fix intermittent HDF5 issue

white238:
We are not being careful about writing to the same directory/files with sidre/spio.

This test seems to fail often and intermittently.  This will ensure that this specific test goes to a unique output directory and stop this from happening:

https://dev.azure.com/llnl-serac/serac/_build/results?buildId=5471&view=logs&j=2742912e-f547-59b4-f336-65c28301f30c&t=cd7bfce2-9b19-5c4f-5f3a-913e42776d4e&l=3475

More work needs to happen to go through the other places that `serac::StateManager::initialize` is called to make sure we aren't writing to the same location.


joshessman-llnl (approved):
Thank you for tracking this down!

PR #580, opened 2021-10-12
Prohibit saving of DataColl info to cwd, report output dirs to user

### LINKED ISSUE #577: Do not allow writing to current directory with `serac::StateManager::initialize`
This keeps causing problems where we have intermittent failures due to reading/writing to the same files. We should probably move away from allowing "" as a valid `output_directory`.

Shown in #576

### LINKED ISSUE #578: Output which datacollection file we are about to save
Outputting the name of the root of the datacollection file we are writing to would help a lot when debugging file system issues.

Something like the following in `void StateManager::save`:

```
 // this could be expanded on i think
  std::string file_path = axom::utilities::filesystem::joinPath(datacoll_->GetPrefixPath(), datacoll_->GetCollectionName());
  SLIC_INFO_ROOT(fmt::format("Saving data collection: {}", file_path));
```

joshessman-llnl:
Resolves #578 
Resolves #577


src/serac/physics/utilities/state_manager.cpp
```diff
@@ -45,7 +45,10 @@ void StateManager::initialize(axom::sidre::DataStore& ds, const std::string& col
   datacoll_.emplace(collection_name_, bp_index_grp, domain_grp, owns_mesh_data);
   datacoll_->SetComm(MPI_COMM_WORLD);
 
-  if (!output_directory.empty()) {
+  if (output_directory.empty()) {
+    SLIC_ERROR_ROOT(
+        "DataCollection output directory is empty - this may result in problems if executables are run in parallel");
```
white238: Since we are failing in this state, we should be more blunt in the error. "may" implies that we allow them to do this.

src/serac/physics/utilities/state_manager.hpp
```diff
@@ -37,8 +37,8 @@ class StateManager {
    * @param[in] output_directory The directory to output files to
```
white238: Can you put this cannot be empty here.

white238 (approved):
Approved on the basis the two minor things get fixed.  Thanks for taking care of this @joshessman-llnl

jamiebramwell (approved):
Sorry - I missed that this was ready. Looks good!

PR #589, opened 2021-11-03
Fix the dual space transfer operators

### LINKED ISSUE #588: Wondering about distributing/assembling dual vectors
Hi All,

I've been taking a look at the serac code after the MFEM workshop, and I really like the idea of the FiniteElementVector/State/Dual classes, but I've got a question about the implementation of `distributeSharedDofs` and `initializeTrueVec` for the case of the `FiniteElementDual` vectors. Based on this link: https://mfem.org/pri-dual-vec/, it looks like the conversion between dof and tdof vectors should be different for duals than for primal vectors. Here's the current implementation for all `FiniteElementVectors`:
https://github.com/LLNL/serac/blob/bd229c48c5ddf4c6e3c1dcaa815a40caaa670094/src/serac/physics/state/finite_element_vector.hpp#L135-L144

And I'm essentially wondering why this function isn't different for duals, something like
```
void distributeSharedDofs() { detail::retrieve(gf_).???(true_vec_); }

void initializeTrueVec() { detail::retrieve(gf_).ParallelAssemble(true_vec_); }
```
With the question marks in `distributeSharedDofs` since I couldn't find any `GridFunction` method that would multiply `gf = R^T tv`.

Thanks!

jamiebramwell:
- Corrects the operator applications for the `FiniteElementDual` object per @tuckerbabcock 's suggestion. Essentially, we were using primal prolongation and restrictions when we should have been using the transpose operators.
- Changes the `FiniteElementVector` method `initializeTrueVec` to `setTrueVec` as it is more appropriate.
- Adds a move constructor to `FiniteElementVector`

Resolves https://github.com/LLNL/serac/issues/588 .

### DIFF
--- a/src/serac/physics/state/finite_element_vector.hpp
+++ b/src/serac/physics/state/finite_element_vector.hpp
@@ -103,7 +103,7 @@ public:
    *
    * @param[in] input_vector The input vector used for construction
    */
-  FiniteElementVector(FiniteElementVector&& input_vector) = default;
+  FiniteElementVector(FiniteElementVector&& input_vector);
 
   /**
    * @brief Returns the MPI communicator for the state
--- a/src/serac/physics/state/finite_element_vector.hpp
+++ b/src/serac/physics/state/finite_element_vector.hpp
@@ -133,17 +133,6 @@ public:
   std::string name() const { return name_; }
 
   /**
-   * @brief Set the internal grid function using the true DOF values
-   */
-  void distributeSharedDofs() { detail::retrieve(gf_).SetFromTrueDofs(true_vec_); }
-
-  /**
-   * @brief Initialize the true DOF vector by extracting true DOFs from the internal
-   * grid function/local into the internal true DOF vector
-   */
-  void initializeTrueVec() { detail::retrieve(gf_).GetTrueDofs(true_vec_); }
-
-  /**
    * @brief Set a finite element state to a constant value
    *
    * @param value The constant to set the finite element state to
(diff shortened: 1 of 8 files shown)


src/serac/physics/state/finite_element_vector.hpp
```diff
@@ -93,6 +93,18 @@ class FiniteElementVector {
...
+   * @brief Delete the default copy constructor
+   */
+  FiniteElementVector(const FiniteElementVector&) = delete;
+
+  /**
+   * @brief Move construct a new Finite Element Vector object
+   *
+   * @param[in] input_vector The input vector used for construction
+   */
+  FiniteElementVector(FiniteElementVector&& input_vector);
```
joshessman-llnl, line 106: I think this can just be `= default` so you don't have to implement it manually

joshessman-llnl (approved):

src/serac/physics/state/finite_element_vector.hpp
tuckerbabcock: Since I'm getting notifications for this, I don't think this work as expected, since `mfem::HypreParVector` doesn't have a proper defined move-constructor.

I think you'd have to define a move constructor like:
```
FiniteElementVector::FiniteElementVector(FiniteElementVector&& input_vector)
    : mesh_(input_vector.mesh()),
      coll_(std::move(input_vector.coll_)),
      space_(std::move(input_vector.space_)),
      gf_(std::move(input_vector.gf_)),
      name_(std::move(input_vector.name_))
{
  auto *par_vec = input_vector.true_vec.StealParVector();
  true_vec.WrapHypreParVector(par_vec);
}
```
Or, define a proper move constructor for `mfem::HypreParVector`

jamiebramwell: Yes, you are right! Thanks! (and sorry for the spam)

tuckerbabcock: No worries!

white238:
Because overloading the very common CS verb `set` tweaks me.

Brainstorming better verb than `set`:

* initialize
* populate
* refresh

white238 (approved):

PR #590, opened 2021-11-05
Add a functional-enabled version of thermal conduction

jamiebramwell:
This adds a new physics module that implements thermal conduction via `Functional`. While I tried to keep this as close to the original implementation as possible, there are some key differences: 

- The module is templated on polynomial order and mesh dimension.
- Material models, source terms, and loads are implemented via static polymorphism with compile-time error checking.
- Flux boundary conditions must be applied on the entire boundary (the subdomain work in progress in MFEM will enable this).
- There is currently no current input file interface.
- Generic std::functions (and therefore lua functions) for defining sources and boundary fluxes are not allowed due to GPU considerations.


jamiebramwell:
@samuelpmishLLNL @joshessman-llnl It's still seg faulting, but this is probably ready for a cursory view in terms of overall structure. Any feedback is greatly appreciated!

src/serac/physics/tests/serac_thermal_functional.cpp
```diff
@@ -0,0 +1,76 @@
...
+void functional_test()
+{
+  MPI_Barrier(MPI_COMM_WORLD);
+
+  int serial_refinement   = 1;
+  int parallel_refinement = 0;
+
+  std::string meshfile2D = SERAC_REPO_DIR "/data/meshes/star.mesh";
+  auto        mesh2D = mesh::refineAndDistribute(buildMeshFromFile(meshfile2D), serial_refinement, parallel_refinement);
+  mesh2D->ExchangeFaceNbrData();
```
samuelpmishLLNL: I think `mfem::ParMesh::ExchangeFaceNbdrData()` is already being called here:
https://github.com/LLNL/serac/blob/d7cce1d1b788e7e859d21ad0559ccb741c501c8e/src/serac/mesh/mesh_utils.cpp#L433

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+  double rho;
+  double cp;
+  double kappa;
+
+  template <typename T1, typename T2>
+  SERAC_HOST_DEVICE T2 thermalFlux([[maybe_unused]] T1& u, [[maybe_unused]] T2& du_dx) const
+  {
+    return kappa * du_dx;
+  }
+};
```
samuelpmishLLNL: I think we should rename this to something like `LinearIsotropicConductor` or put it in a namespace and name it something like `Thermal::LinearIsotropicMaterial` (otherwise `LinearIsotropicMaterial` might be ambiguous for different physics). I'm not sure what the "Constant" implies in this name (properties don't vary with spatial position => "Homogeneous", kappa doesn't change with temperature => "Linear").

I'm more accustomed to calling it "heat flux" than "thermal flux", as far as the response member function name is concerned, but this is mostly aesthetics. Actually, I'm leaning toward defining the q-function as `MaterialType::operator()`, so that we don't have to introduce a bunch of names for the users/developers to keep track of.

e.g. for a thermomechanical material model, all of the following might be reasonable names for the response function:
1. StressAndHeatFlux
2. HeatFluxAndStress
3. ThermalFluxAndStress
4. ThermalFluxAndCauchyStress
5. ...

and whichever one we chose would be arbitrary, and a user would have to look it up, whereas `MaterialType::operator()` is universal.

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+     * @brief The timestepping method to be applied
+     *
+     */
+    TimestepMethod timestepper;
+
+    /**
+     * @brief The essential boundary enforcement method to use
+     *
+     */
+    DirichletEnforcementMethod enforcement_method;
```
samuelpmishLLNL: minor: could we consider adopting a convention where we use a single-line or same-line doxygen comments for short descriptions like this? e.g. 
```suggestion
  /// @brief A timestep method and config for the M solver
  struct TimesteppingOptions {
  
    /// @brief The timestepping method to be applied
    TimestepMethod timestepper;

    /// @brief The essential boundary enforcement method to use
    DirichletEnforcementMethod enforcement_method;
```

I think its valuable to be able to see more of a sourcefile at once, and these multi-line doxygen expressions for short messages occupy a lot of vertical space in an editor.

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+      }
+      if constexpr (dim > 2) {
+        x_tensor[2] = x[2];
+      }
+
+      return temp(x_tensor, t);
+    });
+
+    bcs_.addEssential(temp_bdr, temp_bdr_coef_, temperature_);
+  }
```
samuelpmishLLNL: It seems more natural for this function to accept an "`mfem::Coefficient`" or a `std::function < double(const mfem::Vector&, double) >` for its second argument.

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+        x_tensor[2] = x[2];
+      }
+
+      return temp_function(x_tensor, t);
+    });
+
+    temp_coef.SetTime(time_);
+    temperature_.project(temp_coef);
+    gf_initialized_[0] = true;
+  }
```
samuelpmishLLNL: same comment as in https://github.com/LLNL/serac/pull/590/files#r745172147

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+          auto [u, du_dx] = temperature;
+
+          auto flux = du_dx * 0.0;
+
+          auto source = source_function(x, time_);
+
+          // Return the source and the flux as a tuple
+          return serac::tuple{source, flux};
+        },
+        mesh_);
```
samuelpmishLLNL: `x` looks like it's always used, so its `[[maybe_unused]]` attribute can be removed, I think.

Here, the source function is a function of space and time, is it worthwhile to let it depend on temperature as well (I imagine space + time account for 90+ % of the uses).

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: I agree. I admit I default to this because my IDE autofills a lot of this.

src/serac/physics/thermal_conduction_functional.hpp
joshessman-llnl: +1 for `operator()` here

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+                                                               .vector_dim = 1,
+                                                               .ordering   = mfem::Ordering::byNODES,
+                                                               .name       = detail::addPrefix(name, "temperature")})),
+        M_functional_(&temperature_.space(), &temperature_.space()),
+        K_functional_(&temperature_.space(), &temperature_.space()),
+        residual_(temperature_.space().TrueVSize()),
+        ode_(temperature_.space().TrueVSize(), {.u = u_, .dt = dt_, .du_dt = previous_, .previous_dt = previous_dt_},
+             nonlin_solver_, bcs_)
+  {
+    SLIC_ERROR_ROOT_IF(mesh_.Dimension() != dim,
```
joshessman-llnl, line 136: :+1:

src/serac/physics/thermal_conduction_functional.hpp
joshessman-llnl: Yep, I think when we allow for arguments of arbitrary type it might also be worth introducing corresponding traits classes/static_asserts so we can provide reasonable compile error messages if the user screws something up.

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+      // Step the time integrator
+      ode_.Step(temperature_.trueVec(), time_, dt);
+    }
+
+    temperature_.distributeSharedDofs();
+    cycle_ += 1;
+  }
+
+  template <typename MaterialType>
+  void setMaterial(MaterialType material)
```
joshessman-llnl, line 212: e.g. here we could document the preconditions for `MaterialType` and enforce it via a static_assert

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,481 @@
...
+    // Project the coefficient onto the grid function
+    mfem::FunctionCoefficient temp_coef([temp_function](const mfem::Vector& x, double t) -> double {
+      tensor<double, dim> x_tensor;
+      x_tensor[0] = x[0];
+      if constexpr (dim > 1) {
+        x_tensor[1] = x[1];
+      }
+      if constexpr (dim > 2) {
+        x_tensor[2] = x[2];
+      }
```
joshessman-llnl: I had to do something somewhat similar in Inlet to convert to/from `mfem::Vector` - if this is something we anticipate being needed in multiple places it might be worth implementing a

```cpp
template <int dim>
tensor<double, dim> vectorToTensor(const mfem::Vector&);
```

or more broadly perhaps a tensor constructor accepting a `T*`?

samuelpmishLLNL: there is also 
```cpp
mfem::Vector foo = ... ;
auto x = make_tensor<dim>([&](int i){ return foo[i]; });
```

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: I'm trying to avoid `Coefficients` whenever possible because the ownership semantics are weird. I think std::function makes sense.

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: Question: I don't think I can use `std::function` for this since it is injecting directly into the kernels. Does this need to be a type template?

jamiebramwell:
@sethwatts @tepperly @goxberry Any feedback on the interface (particularly with regards to the material and source functions) is greatly appreciated.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,163 @@
...
+ */
+
+#pragma once
+
+#include "serac/numerics/functional/functional.hpp"
+
+/// ThermalConductionFunctional helper structs
+namespace serac::Thermal {
+
+/// Linear isotropic thermal conduction material model
```
sethwatts, line 20: It looks like the relevant properties are assumed to be constant over the domain (or at least over the material). Is it possible to instead define the properties via Coefficient?

sethwatts: Follow up question, are tensor-valued conductivities on the menu?

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,163 @@
...
+struct LinearIsotropicConductor {
+  /// Density
+  double density_;
+
+  /// Specific heat capacity
+  double specific_heat_capacity_;
+
+  /// Constant isotropic thermal conductivity
+  double conductivity_;
```
sethwatts, line 88: Possibly naive question, is it worth doing bad data checks to ensure that the material properties are non-negative to ensure we obey conversation laws?

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Unfortunately no. The coefficient virtual function interface is not amenable to GPUs. We can add a `FiniteElementState`-type interface or a CPU-only coefficient style one. Any thoughts @samuelpmishLLNL ?

jamiebramwell: We can easily implement a tensor valued conductivity using this interface.

sethwatts:
Just to check my understanding, if I wanted to provide my own non-constant source, non-constant flux term, etc., would I derive from the existing structs and re-implement their `operator()` methods?

jamiebramwell:
> 
> 
> Just to check my understanding, if I wanted to provide my own non-constant source, non-constant flux term, etc., would I derive from the existing structs and re-implement their `operator()` methods?

Exactly. I'll add some documentation outlining the process.

src/serac/physics/materials/thermal_functional_material.hpp
sethwatts: We would probably be pushing `FiniteElementState`s into `GridFunctionCoefficient`s anyway, so I think either of these approaches would work, but we'd definitely want a way to provide spatially-varying property fields (and loads, etc.) as a way to inject design, uncertainty, and so on.

samuelpmishLLNL: I don't see any reason why a general anisotropic linear model can't be implemented immediately as:

```cpp
template < int dim >
struct LinearConductor {
   double density_;
   double specific_heat_capacity_;
   tensor<double, dim, dim> thermal_conductivity_;

   template <typename T1, typename T2>
   SERAC_HOST_DEVICE auto operator()(T1& /* temperature */, T2& temperature_gradient) const
   {
     return -thermal_conductivity_ * temperature_gradient;
   }
};
```

(note: this implementation is a homogeneous material)

If you wanted a spatially-varying thermal conductivity tensor, you could introduce it by passing a `QuadratureData< tensor<double, dim, dim> >` into a call to `Functional::AddDomainIntegral`

samuelpmishLLNL: > we'd definitely want a way to provide spatially-varying property fields (and loads, etc.) as a way to inject design, uncertainty, and so on.

If your goal is material response as a function of design parameter, it seems more natural (to me) to express that with the multi-trial-space flavor of Functional, something like

```cpp
template < int dim >
struct ParametrizedLinearConductor {
   double density_;
   double specific_heat_capacity_;
   
   tensor< double, dim, dim > some_parametrized_function(double design_parameter) {
       // you implement this
   }

   template <typename T1, typename T2>
   SERAC_HOST_DEVICE auto operator()(T1& /* temperature */, T2& temperature_gradient, T3 & design_parameter) const
   {
     tensor< double, dim, dim > thermal_conductivity = some_parametrized_function(design_parameter);
     return -thermal_conductivity_ * temperature_gradient;
   }
};
```

samuelpmishLLNL:
> Just to check my understanding, if I wanted to provide my own non-constant source, non-constant flux term, etc., would I derive from the existing structs and re-implement their `operator()` methods?

You wouldn't derive from them in the sense of inheritance, you'd just implement a functor (struct w/ operator() definition), or a lambda with the appropriate inputs/outputs.

Essentially, that source functor / lambda is just being passed down into the kernel and called in the quadrature loop, so anything that has the right function signature should work.

src/serac/physics/materials/thermal_functional_material.hpp
sethwatts: This could work in at least some cases, but I think it violates our separation of concerns, in that I would need a serac object to know about design parameters, and would need to derive different serac objects depending on what I want to do in LiDO , e.g.,  am I interpolating properties based on design fields? Am I just assembling a spatially heterogeneous response due to different material regions? Am I passing you instances of a stochastic property field? Whereas all these use cases would be addressed if I can pass a FiniteElementState representation of the field.

samuelpmishLLNL: I didn't explain clearly in my previous comment, but the design parameter in this snippet:
```cpp
template < int dim >
struct ParametrizedLinearConductor {
   double density_;
   double specific_heat_capacity_;
   
   tensor< double, dim, dim > some_parametrized_function(double design_parameter) {
       // you implement this
   }

   template <typename T1, typename T2>
   SERAC_HOST_DEVICE auto operator()(T1& /* temperature */, T2& temperature_gradient, T3 & design_parameter) const
   {
     tensor< double, dim, dim > thermal_conductivity = some_parametrized_function(design_parameter);
     return -thermal_conductivity_ * temperature_gradient;
   }
};
```
is a finite element field evaluated at a particular quadrature point. If you want a nodally interpolated design parameter, then you use the H1 function space, if you want discontinuous, elementwise values, then L2 may be more appropriate.

The type `ParametrizedLinearConductor` is just expressing how that parameter controls the physical response of your material at a given quadrature point, not how that quadrature point value of the design parameter is calculated.

sethwatts: After talking with @jamiebramwell , I believe this interface will work for LiDO's needs.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,239 @@
...
+   *
+   * @tparam T1 type of the temperature (e.g. tensor or dual type)
+   * @tparam T2 type of the temperature gradient (e.g. tensor or dual type)
+   * @param temperature_gradient Gradient of the temperature (du_dx)
+   * @return The thermal flux of the material model
+   */
+  template <typename T1, typename T2>
+  SERAC_HOST_DEVICE T2 operator()(T1& /* temperature */, T2& temperature_gradient) const
+  {
+    SLIC_ASSERT_MSG(conductivity_ > 0.0,
```
white238: A+ to helpful asserts

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,239 @@
...
+};
+
+template <typename T, int dim>
+struct has_thermal_flux_boundary<
+    T, dim,
+    std::void_t<decltype(std::declval<T&>()(std::declval<tensor<double, dim>&>(), std::declval<tensor<double, dim>&>(),
+                                            std::declval<tensor<double, 1>&>()))>> : std::true_type {
+};
+
+}  // namespace serac::Thermal
```
white238: newline

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: Does `SLIC_ASSERT_MSG` actually work in a `__device__` context?

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,239 @@
...
+   * @tparam T2 Type of the temperature variable
+   */
+  template <typename T1, typename T2>
+  SERAC_HOST_DEVICE double specificHeatCapacity(T1& /* x */, T2& /* temperature */) const
+  {
+    SLIC_ASSERT_MSG(specific_heat_capacity_ > 0.0,
+                    "Specific heat capacity must be positive in the linear isotropic conductor material model.");
+
+    return specific_heat_capacity_;
+  }
```
samuelpmishLLNL: the type of the spatial coordinate is always just `tensor<double, dim>`. We only need `auto` or a type-template for the arguments that will be potentially differentiated with AD.

In the future, materials with manually-written derivatives (i.e. no AD) can explicitly specify the types all of the input parameters.

samuelpmishLLNL: Also, I'm not sure how to best use Doxygen for this kind of pattern. These materials are all implementations of a certain concept, so documenting every single material with the same boilerplate Doxygen comments seems wasteful (but I don't know an alternative).

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,239 @@
...
+   * @param temperature_gradient Gradient of the temperature (du_dx)
+   * @return The thermal flux of the material model
+   */
+  template <typename T1, typename T2>
+  SERAC_HOST_DEVICE auto operator()(T1& /* temperature */, T2& temperature_gradient) const
+  {
+    for (int i = 0; i < dim; ++i) {
+      SLIC_ASSERT_MSG(conductivity_(i, i) > 0.0,
+                      "Conductivity tensor must be positive definite in linear conductor material model.");
+    }
```
samuelpmishLLNL: I think having positive diagonal entries is necessary, but not sufficient for a symmetric matrix to be positive-definite, e.g.

```cpp
tensor A = {{1.0, -4.0, -4.0}, {-4.0, 1.0, -40}, {-4.0, -4.0, 1.0}};
tensor x = {1.0, 1.0, 1.0};

x * A * x == -21

```

samuelpmishLLNL: Also, maybe this could be checked elsewhere, as this implementation is putting a conditional + assertion in the most deeply-nested loop of the simulation. Since the conductivity is homogeneous for this material, it only needs to be checked once.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,239 @@
...
+                                      std::declval<tensor<double, dim>&>(), std::declval<tensor<double, 1>&>()))>>
+    : std::true_type {
+};
+
+template <typename T, int dim, typename = void>
+struct has_thermal_flux : std::false_type {
+};
+
+template <typename T, int dim>
+struct has_thermal_flux<
```
samuelpmishLLNL, line 188: These checks are useful for checking the existence of an `operator()` that accepts certain arguments, but it's worth noting that there are invalid material definitions that will pass this test, and there are valid material definitions that will not pass it (although, this case is unlikely to happen in practice).

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,239 @@
...
+template <int dim>
+struct LinearConductor {
+  /// Density
+  double density_ = -1.0;
+
+  /// Specific heat capacity
+  double specific_heat_capacity_ = -1.0;
+
+  /// Constant isotropic thermal conductivity
+  tensor<double, dim, dim> conductivity_ = {-1.0};
```
samuelpmishLLNL: these seem like unusual choices for default values

I'm also not sure what `tensor<double, dim, dim> conductivity_ = {-1.0};` actually does

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: That's a good question. Thoughts @white238 ?

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: I left the position arguments as `auto` as I wasn't sure if that would be needed for shape optimization. Under that context, would a plain `tensor` still be good enough?

Also I agree about the doxygen. Not really sure what the best solution is, although I would prefer boilerplate to letting missing doxygen parameters through.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Yeah, I wasn't sure the best place to put this. I used `ASSERT` so it wouldn't be compiled under release versions. Maybe a `isValid` method that we can call during the setup phase?

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: I wanted a default value that would fail if uninitialized. I assume the `= double` would just fill all of the entries with the single value, but I admit I have not checked.

src/serac/physics/tests/serac_thermal_functional.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    serac::StateManager::setMesh(std::move(mesh2D));
+  }
+
+  if constexpr (dim == 3) {
+    std::string meshfile3D = SERAC_REPO_DIR "/data/meshes/beam-hex.mesh";
+    auto mesh3D = mesh::refineAndDistribute(buildMeshFromFile(meshfile3D), serial_refinement, parallel_refinement);
+
+    serac::StateManager::setMesh(std::move(mesh3D));
+  }
```
samuelpmishLLNL: I don't think `if constexpr` is necessary here, maybe something like 

```suggestion

  std::map<int, std::string> meshfiles = {
    {2, SERAC_REPO_DIR "/data/meshes/star.mesh"},
    {3, SERAC_REPO_DIR "/data/meshes/beam-hex.mesh"}  
  };
  
  auto mesh = mesh::refineAndDistribute(buildMeshFromFile(meshfiles[dim]), serial_refinement, parallel_refinement);
  serac::StateManager::setMesh(std::move(mesh));
```

or 

```cpp
std::string filename = (dim == 2) ? 
  SERAC_REPO_DIR "/data/meshes/star.mesh" : 
  SERAC_REPO_DIR "/data/meshes/beam-hex.mesh";
auto mesh = mesh::refineAndDistribute(buildMeshFromFile(filename), serial_refinement, parallel_refinement);
serac::StateManager::setMesh(std::move(mesh));

File: src/serac/physics/materials/thermal_functional_material.hpp (hunk 12 above)
Comment (samuelpmishLLNL) [reply src/serac/physics/materials/thermal_functional_material.hpp]: TIL: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#assertion

I'm not sure if this macro uses `assert(...)` or not, but that's a useful feature

https://godbolt.org/z/e4PK7e4Gz

File: src/serac/physics/materials/thermal_functional_material.hpp (hunk 17 above)
Comment (samuelpmishLLNL) [reply src/serac/physics/materials/thermal_functional_material.hpp]: the tensor class doesn't define any constructors, so it initializes kind of like a multidimensional array in C++. That means the following code 
```cpp
std::cout << serac::tensor< double, 3, 3>{-1.0};
```
will print `{{-1, 0, 0}, {0, 0, 0}, {0, 0, 0}}`, in much the same way as https://godbolt.org/z/5hzfzfvb3

src/serac/physics/materials/thermal_functional_material.hpp
white238: SLIC does not work on CUDA devices.  It's on Axom's agenda for this year.  We could toggle between the two internally though since they are just macros.

#593

src/serac/physics/tests/serac_thermal_functional.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "serac/physics/thermal_conduction_functional.hpp"
+
+#include <fstream>
+
+#include <gtest/gtest.h>
+#include "mfem.hpp"
+
+#include "serac/mesh/mesh_utils.hpp"
+#include "serac/physics/state/state_manager.hpp"
+#include "serac/serac_config.hpp"
```
white238: `serac_config.hpp` should always go above all of our headers in case it toggles behavior in any of them.

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -0,0 +1,450 @@
...
+    /// The timestepping method to be applied
+    TimestepMethod timestepper;
+
+    /// The essential boundary enforcement method to use
+    DirichletEnforcementMethod enforcement_method;
+  };
+
+  /**
+   * @brief A configuration variant for the various solves
+   * Either quasistatic, or time-dependent with timestep and M options
```
white238: What is `M` in this context?

white238:
Should this be added to the serac driver? Or is that going to be in a followup PR after the input file is augmented to handle this case?

src/serac/physics/materials/thermal_functional_material.hpp
sethwatts: I agree that the diagonal check isn't sufficient in the general anisotropic case. Checking for positive determinant would be a nice, dimension-agnostic one-liner. Should there also be a symmetry check?

samuelpmishLLNL: positive determinant also isn't quite right either (e.g. det({{-1, 0, 0}, {0, -1, 0}, {0, 0, 1}}) == 1, but the matrix isn't positive definite). I think I have a routine for closed-form eigenfactorization of symmetric 2x2 and 3x3 matrices that may be useful here (is it safe to assume symmetry for the conductivity?)

sethwatts: Ah, you're right on the determinant. I agree checking individual eigenvalues is the right approach. 

Googling, I found a paper [(link)](https://www.inderscienceonline.com/doi/abs/10.1504/IJMSI.2014.067111) that offers a proof that the conductivity tensor must be symmetric for linear conduction, even in the case of an anisotrpic, heterogeneous material. So I think it would be safe to use a symmetric eigenvalue approach, possibly after asserting the tensor the user provides is indeed symmetric.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: What is a valid material that wouldn't pass? I mostly added these checks for human-readable compiler output if a user defined material is ill-formed.

src/serac/physics/tests/serac_thermal_functional.cpp
jamiebramwell: I'll add the second option.

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: the actual arguments that get passed to a q-function will be tensors of dual numbers, so you could explicitly write out the actual types to be passed in (I don't recommend this, as it is verbose and prevents the material from being composable in multiphysics simulations):

```cpp
struct ThisMaterialWorksWithFunctionalButFailsCheck {
  ...
  auto operator()(dual < double > u, tensor< dual < tensor < double, 3 > >, 3 > du_dx) {
    // implement the heat flux calculation
  }
};
```

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: fixed the comment because this one stinks.

jamiebramwell: I should add that *my* code comment stinks, not the github one above. 😆

jamiebramwell:
> 
> 
> Should this be added to the serac driver? Or is that going to be in a followup PR after the input file is augmented to handle this case?

Yeah, I want to do the input file stuff in another PR. After that it can get moved to the driver.

src/serac/physics/thermal_conduction_functional.hpp
white238: Too late.. super offended.

white238 (approved):

src/serac/physics/materials/thermal_functional_material.hpp
goxberry: I'd think checking for a Hermitian matrix and using [Sylvester's criterion](https://en.wikipedia.org/wiki/Sylvester%27s_criterion) would work here: all of the leading principal minors must be positive, which only requires extracting submatrices and computing determinants. Computing eigenvalues seems like overkill.

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: I believe the spatial coordinate can still be a plain `tensor`. In the shape optimization case, you could take a derivative w.r.t. a "design displacement" field on the surface, but that's a separate quantity.

src/serac/physics/base_physics.cpp
```diff
@@ -23,7 +23,7 @@ namespace serac {
 BasePhysics::BasePhysics()
     : mesh_(StateManager::mesh()),
       comm_(mesh_.GetComm()),
-      output_type_(serac::OutputType::VisIt),
+      output_type_(serac::OutputType::SidreVisIt),
```
joshessman-llnl, line 26: Haha, this one bit me (briefly) in #597 :)

joshessman-llnl (approved):

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: I moved these checks to the thermal material constructor. Let me know what you think!

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: I put a reasonable default of `1.0` in for now.

jamiebramwell:
@samuelpmishLLNL , let me know if you're OK with the current changes.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,278 @@
...
+    SLIC_ERROR_ROOT_IF(specific_heat_capacity_ < 0.0,
+                       "Specific heat capacity must be positive in the linear conductor material model.");
+
+    // Check that the conductivity tensor is symmetric
+    for (int i = 0; i < dim; ++i) {
+      for (int j = i + 1; j < dim; ++j) {
+        SLIC_ERROR_ROOT_IF(std::abs(conductivity_(i, j) - conductivity_(j, i)) > 1.0e-7,
+                           "Conductivity tensor must be symmetric for the linear conductor material model.");
+      }
+    }
```
samuelpmishLLNL: rather than implement these checks in the ctor (and later duplicate them in sister material models), we could expand the tensor library with an appropriately-named function:

```cpp
template < int n >
bool is_symmetric(tensor< double, n, n > A, double tolerance = 1.0e-8) {
  for (int i = 0; i < n; ++i) {
    for (int j = i + 1; j < n; ++j) {
      if (std::abs(A(i, j) - A(j, i)) > tolerance) { return false; };
    }
  }
  return true;
}
```

and that would let the code document itself, without relying on an inline comment to explain what is going on:

```cpp
SLIC_ERROR_ROOT_IF(!is_symmetric(conductivity_), "error message");
```

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,278 @@
...
+                       "Conductivity tensor must be positive definite for the linear conductor material model.");
+
+    SLIC_ERROR_ROOT_IF(det(conductivity_) < 0.0,
+                       "Conductivity tensor must be positive definite for the linear conductor material model.");
+
+    if (dim == 3) {
+      auto subtensor_2D = make_tensor<2, 2>([this](int i, int j) { return conductivity_(i, j); });
+      SLIC_ERROR_ROOT_IF(det(subtensor_2D) < 0.0,
+                         "Conductivity tensor must be positive definite for the linear conductor material model.");
+    }
```
samuelpmishLLNL: Same comment as https://github.com/LLNL/serac/pull/590/files#r758817476,

```cpp
// in tensor.hpp
template < int n >
bool is_positive_definite(tensor< double, n, n > A) {
  static_assert(n < 4, "this check is only implemented for square matrices with 3 or fewer columns");
  if (!is_symmetric(A)) { return false; }
  if (A(0, 0) < 0.0) { return false; }
  if (det(A) < 0.0) { return false; }
  if constexpr (dim == 3) {
    if (det(make_tensor<2, 2>([&](int i, int j) { return A(i, j); })) < 0.0) { return false; }
  }
  return true;
}


// in thermal_functional_material.hpp
LinearConductor(double density = 1.0, double specific_heat_capacity = 1.0, ...) {


  ...

  SLIC_ERROR_ROOT_IF(!is_positive_definite(conductivity_), "error message");

  ...

}

File: src/serac/physics/materials/thermal_functional_material.hpp [hunk 24]
```diff
@@ -0,0 +1,278 @@
...
+public:
+  /**
+   * @brief Construct a new Linear Isotropic Conductor object
+   *
+   * @param density Density of the material (mass/volume)
+   * @param specific_heat_capacity Specific heat capacity of the material (energy / (mass * temp))
+   * @param conductivity Thermal conductivity of the material (power / (length * temp))
+   */
+  LinearConductor(double density = 1.0, double specific_heat_capacity = 1.0,
+                  tensor<double, dim, dim> conductivity = {{{1.0, 0.0, 0.0}, {0.0, 1.0, 0.0}, {0.0, 0.0, 1.0}}})
```
samuelpmishLLNL: Does `tensor<double, dim, dim> conductivity = {{{1.0, 0.0, 0.0}, {0.0, 1.0, 0.0}, {0.0, 0.0, 1.0}}}` compile when dim == 2?

```suggestion
                  tensor<double, dim, dim> conductivity = Identity<dim>())
```

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,278 @@
...
+   * @tparam T1 type of the temperature (e.g. tensor or dual type)
+   * @tparam T2 type of the temperature gradient (e.g. tensor or dual type)
+   * @param temperature_gradient Gradient of the temperature (du_dx)
+   * @return The thermal flux of the material model
+   */
+  template <typename T1, typename T2>
+  SERAC_HOST_DEVICE T2 operator()(T1& /* temperature */, T2& temperature_gradient) const
+  {
+    return -1.0 * conductivity_ * temperature_gradient;
+  }
```
samuelpmishLLNL: I think these arguments (and others in similar materials/member functions) should be `const T &`. This shouldn't change the behavior, but the current implementation might give the impression that one is allowed to modify them and have `serac::Functional` do something with the mutated values (it doesn't).

The only ones that should be non-`const` references are material state data, since the q-function will need to be able to update that information in e.g. a plasticity model and save the updated values.

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: (minor) also, does SLIC have an assertion (i.e. check that the predicate is true, rather than false)? I think it is more natural to phrase these sort of checks as `assert(is_symmetric(A))` than the double-negative of `error_if(!is_symmetric(A))`

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,278 @@
...
+
+private:
+  /// Density
+  double density_ = 1.0;
+
+  /// Specific heat capacity
+  double specific_heat_capacity_ = 1.0;
+
+  /// Constant thermal conductivity
+  tensor<double, dim, dim> conductivity_ = {{{1.0, 0.0, 0.0}, {0.0, 1.0, 0.0}, {0.0, 0.0, 1.0}}};
```
samuelpmishLLNL: making these private members disables aggregate initialization, so I don't think that the "default" values specified here do anything.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -0,0 +1,278 @@
...
+   * @tparam T2 Type of the temperature
+   * @tparam dim The dimension of the problem
+   * @return The flux applied to the boundary
+   */
+  template <typename T1, typename T2, int dim>
+  SERAC_HOST_DEVICE double operator()(tensor<double, dim>& /* x */, T1& /* n */, T2& /* u */) const
+  {
+    return flux_;
+  }
+};
```
samuelpmishLLNL: this is beyond the scope of this PR, but it might be worth adding a feature to Functional that allows it to do surface / domain integrals that do not depend on the trial spaces, so we won't be interpolating the finite element fields and then just discarding the values in the q-function.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: @white238 , are there any SLIC calls (other than assert) that do this? I would like to keep these checks in release builds.

white238: There is not.

white238: To expand: 

We could add one on our side though via a macro that just negates the expression.  Though I personally prefer `SLIC_ERROR_IF(!is_symmetric(A))` but I admit this is just a personal preference.

Asserts (throws an error)/checks (throws a warning) are both in SLIC but are turned into noops when Axom is not built debug.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Nope! Thanks for the catch!

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Yeah, this is just me being probably too cautious. I'll remove them.

samuelpmishLLNL (approved):

PR #596, opened 2021-11-17
Switch over to axom::Array

joshessman-llnl:


src/serac/infrastructure/accelerator.hpp
```diff
@@ -75,6 +77,87 @@ enum class ExecutionSpace
...
+template <>
+struct execution_to_memory<ExecutionSpace::GPU> {
+  static constexpr axom::MemorySpace value = axom::MemorySpace::Device;
+};
+
+/// @overload
+template <>
+struct execution_to_memory<ExecutionSpace::Dynamic> {
+  static constexpr axom::MemorySpace value = axom::MemorySpace::Unified;
+};
```
samuelpmishLLNL, line 108: What is the interpretation of "Dynamic" and "Unified" here? Does this correspond to nvidia's notion of a unified memory model, or is it something else?

src/serac/infrastructure/accelerator.hpp
```diff
@@ -75,6 +77,87 @@ enum class ExecutionSpace
...
+namespace detail {
+
+/**
+ * @brief Trait for "translating" between serac::ExecutionSpace and axom::MemorySpace
+ */
+template <ExecutionSpace space>
+struct execution_to_memory {
+  /// @brief The corresponding axom::MemorySpace
+  static constexpr axom::MemorySpace value = axom::MemorySpace::Dynamic;
+};
```
samuelpmishLLNL, line 89: We might be able to just replace the `serac::ExecutionSpace` type in favor of `axom::MemorySpace` too, if there really is just a 1-to-1 mapping between their options

src/serac/numerics/functional/functional.hpp
```diff
@@ -384,14 +390,14 @@ class Functional<test(trial), exec> : public mfem::Operator {
...
-          boundary.ComputeElementGradients(view(K_belem));
+          boundary.ComputeElementGradients(K_belem);
         }
 
-        for (size_t e = 0; e < K_belem.size(0); e++) {
-          for (size_t i = 0; i < K_belem.size(1); i++) {
-            for (size_t j = 0; j < K_belem.size(2); j++) {
+        for (axom::IndexType e = 0; e < K_belem.shape()[0]; e++) {
+          for (axom::IndexType i = 0; i < K_belem.shape()[1]; i++) {
+            for (axom::IndexType j = 0; j < K_belem.shape()[2]; j++) {
```
samuelpmishLLNL, line 400: this is beyond the scope of this PR, but given that this pattern shows up a lot, it might be worth defining an iterator option:

```cpp
for (auto i : K_belem.range(0)) {
  ...
}
```

src/serac/infrastructure/accelerator.hpp
joshessman-llnl: Yes - previously ExecutionSpace::Dynamic corresponded to a cudaMallocManaged in `serac::Array`, so the new correspondence maintains that same functionality.

src/serac/infrastructure/accelerator.hpp
joshessman-llnl: For context, here's the definition:
https://github.com/LLNL/axom/blob/e2ed0670e1cd326a07eb00b0ed367fdb5e5edde5/src/axom/core/memory_management.hpp#L36-L46

I think in some sense there isn't a 1-1 mapping because `MemorySpace` allows for multiple kinds of device memory

src/serac/numerics/functional/functional.hpp
joshessman-llnl: Added to https://github.com/LLNL/axom/issues/633 for tracking purposes

src/serac/numerics/functional/tests/quadrature_data_cuda.cu
```diff
@@ -275,8 +275,8 @@ struct state_manager_varying_qfunction {
     mutated_data[idx++] = state;
     return u;
   }
-  DeviceArray<typename wrapper_t::value_type>& mutated_data;
-  int                                          idx = 0;
+  axom::Array<typename wrapper_t::value_type, 1, axom::MemorySpace::Unified>& mutated_data;
```
white238: Is there a reason you didnt just use the ~~`DynamicArray`~~ `UnifiedArray` alias here and below?

white238 (approved):

samuelpmishLLNL (approved):

PR #597, opened 2021-11-17
Add support for multiple meshes/datacollections

joshessman-llnl:
The user interface is currently just intended to minimize breaking changes.

I also threw in a bunch of comments on assumptions that I'm making

https://serac.readthedocs.io/en/feature-essman-multimesh_datacoll/sphinx/dev_guide/state_manager.html


src/serac/physics/state/state_manager.hpp
```diff
@@ -34,27 +35,31 @@ class StateManager {
...
 
   /**
    * @brief Factory method for creating a new FEState object, signature is identical to FEState constructor
    * @param[in] options Configuration options for the FEState, if a new state is created
    * @see FiniteElementState::FiniteElementState
    * @note If this is a restart then the options (except for the name) will be ignored
    */
-  static FiniteElementState newState(FiniteElementVector::Options&& options = {});
+  static FiniteElementState newState(FiniteElementVector::Options&& options = {},
+                                     const std::string&             tag     = primary_mesh_name_);
```
jamiebramwell: I think `default` instead of `primary` would be clearer.

src/serac/physics/state/state_manager.hpp
```diff
@@ -131,8 +142,18 @@ class StateManager {
...
    * @brief A set of @p QuadratureData<T> objects that need to be synchronized before saving to disk
+   * FIXME Need one set per datacoll since we can save at different times?
    */
   static std::vector<std::unique_ptr<SyncableData>> syncable_data_;
+
+  /// @brief Pointer (non-owning) to the datastore
+  static axom::sidre::DataStore* ds_;
+  /// @brief Output directory to which all datacollections are saved
+  static std::string output_dir_;
+  /// @brief Cycle to load - for restarts - FIXME can this be different between datacollections?
```
jamiebramwell: While it is possible someone might want to save and restart at different cycles, it is definitely a niche case. Any thoughts @sethwatts @junkudo?

src/serac/physics/state/state_manager.hpp
```diff
@@ -131,8 +142,18 @@ class StateManager {
...
+  static std::optional<int> cycle_to_load_;
+  /// @brief Default name for the mesh - mostly for backwards compatibility
+  const static std::string primary_mesh_name_;
```
jamiebramwell: I would call this `default_mesh_name_` instead of `primary` because it is closer to the intended purpose.

src/serac/physics/state/state_manager.hpp
jamiebramwell: I can imagine this becoming a bookkeeping nightmare.

src/serac/physics/state/state_manager.hpp
```diff
@@ -115,12 +123,15 @@ class StateManager {
...
+   * @brief Creates a new datacollection based on a registered mesh
+   */
+  static void newDataCollection(const std::string& name);
+
+  /**
+   * @brief The datacollection instances
    * The object is constructed when the user calls StateManager::initialize.
    */
-  static std::optional<axom::sidre::MFEMSidreDataCollection> datacoll_;
+  static std::unordered_map<std::string, axom::sidre::MFEMSidreDataCollection> datacolls_;
```
jamiebramwell, line 165: 🎉

jamiebramwell (commented):
I think this is an excellent first cut at a multiple mesh functionality. Thanks!!

src/serac/physics/state/state_manager.hpp
joshessman-llnl: Instead of keeping track of the cycle to load (set via a param to StateManager::initialize) I think it would be clearer to have a set of roughly symmetrical functions:

```cpp
StateManager::save(double t, int cycle, <datacoll tag>); // already exists
StateManager::load(int cycle, <datacoll tag>);
```

This provides additional user flexibility when needed and also eliminates the need to somehow restore what datacollections need to be restored (their names, specifically) before restoring the datacollections.

sethwatts: Apologies for not fully understanding the context - would this be loading a different place within a solver than one was at previously? I could see this being one possible way to support adjoint analysis of transient problems, if "cycle" can be taken to be synonymous with "timestep."

jamiebramwell: > 
> 
> Instead of keeping track of the cycle to load (set via a param to StateManager::initialize) I think it would be clearer to have a set of roughly symmetrical functions:
> 
> ```c++
> StateManager::save(double t, int cycle, <datacoll tag>); // already exists
> StateManager::load(int cycle, <datacoll tag>);
> ```
> 
> This provides additional user flexibility when needed and also eliminates the need to somehow restore what datacollections need to be restored (their names, specifically) before restoring the datacollections.

That makes a lot of sense to me.

src/serac/physics/state/state_manager.cpp
```diff
@@ -21,90 +21,99 @@ QuadratureData<void> dummy_qdata;
...
   if (is_restart_) {
-    auto field = datacoll_->GetParField(name);
+    auto field = datacoll.GetParField(name);
     return {mesh(), *field, name};
   } else {
-    SLIC_ERROR_ROOT_IF(datacoll_->HasField(name),
+    SLIC_ERROR_ROOT_IF(datacoll.HasField(name),
                        axom::fmt::format("Serac's datacollection was already given a field named '{0}'", name));
     options.managed_by_sidre = true;
     FiniteElementState state(mesh(), std::move(options));
```
junkudo: Should this `FiniteElementState` refer to the mesh associated with `tag`?

junkudo:
This looks good to me!  Thanks for the quick turn around.

src/serac/physics/state/state_manager.cpp
```diff
@@ -21,90 +21,99 @@ QuadratureData<void> dummy_qdata;
 QuadratureDataView<void> dummy_qdata_view;
 
 // Initialize StateManager's static members - both of these will be fully initialized in StateManager::initialize
```
white238: both is no longer accurate

src/serac/physics/state/state_manager.cpp
joshessman-llnl: Good catch, thanks!

src/docs/sphinx/dev_guide/state_manager.rst
```diff
@@ -0,0 +1,125 @@
+.. ## Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
+.. ## other Serac Project Developers. See the top-level COPYRIGHT file for details.
+.. ##
+.. ## SPDX-License-Identifier: (BSD-3-Clause)
+
+.. _statemanager-label:
+
+============
+StateManager
```
jamiebramwell, line 9: 🎉

jamiebramwell: This is great!!

src/docs/sphinx/dev_guide/state_manager.rst
```diff
@@ -0,0 +1,125 @@
...
+     a user of a physics module would find it more intuitive to pass the mesh pointer they wish to use.
+     Specifically, the meaning of a mesh parameter is much easier to discern than a string parameter.
+
+  #. The physics module creates its fields (e.g., temperature for a thermal conduction module) via
+     calls to ``StateManager::newState()``.  In addition to the ``serac::FiniteElementState`` constructor
+     arguments, this method also accepts a string-valued tag for the mesh with which the field is
+     associated.  The appropriate tag is a member of ``BasePhysics`` and initialized in the previous step.
+     FIXME: Should we provide a protected helper method in ``BasePhysics`` so derived modules don't need
+     to reference the member explicitly? Or perhaps ``StateManager::newState()`` et al should just take
+     a mesh pointer instead of a tag?
```
jamiebramwell, line 71: How can it take a mesh pointer? If I recall correctly, `mfem::Mesh` does not have an equality check.

src/docs/sphinx/dev_guide/state_manager.rst
```diff
@@ -0,0 +1,125 @@
...
+
+  #. The ``FiniteElementState`` is then constructed and registered in the corresponding ``MFEMSidreDataCollection``.
+     The only tricky part about this process is the need for the underlying ``GridFunction`` to be allocated within
+     Sidre.  This required an additional option to the ``FiniteElementState`` constructor that leaves the vector
+     data uninitialized (aka ``nullptr``).  After we create the ``FiniteElementState`` we register its underlying
+     ``GridFunction`` in the ``MFEMSidreDataCollection`` and zero it out.
+
+  #. When the user wishes to save simulation state to disk, they can call ``outputState()`` on their physics module.
```
jamiebramwell, line 79: We could delete the `outputState` method from the `BasePhysics` class. It makes sense that this would be associated with the state manager instead of the individual physics modules for the reason you described. I believe `outputState` predates the state manager anyway.

src/serac/physics/state/state_manager.cpp
```diff
@@ -20,139 +20,166 @@ QuadratureData<void> dummy_qdata;
...
   return static_cast<mfem::ParMesh&>(*mesh);
 }
 
+std::string StateManager::collectionID(mfem::ParMesh* pmesh)
+{
+  if (!pmesh) {
+    return default_mesh_name_;
+  } else {
+    for (auto& [name, datacoll] : datacolls_) {
+      if (datacoll.GetMesh() == pmesh) {
```
jamiebramwell, line 176: This pointer comparison makes me a little uneasy. Is it ever possible they could be moved?

src/docs/sphinx/dev_guide/state_manager.rst
joshessman-llnl: The current logic uses raw pointer comparisons, anticipating the use case where the user does something like:
```cpp
Mesh m = /* ... */;
StateManager::setMesh(m, "primary");
// ... pointers would be the same
ThermalConduction tc(..., m);
Solid s(..., m);

File: src/serac/physics/state/state_manager.cpp (hunk 10 above)
Comment (joshessman-llnl) [reply src/serac/physics/state/state_manager.cpp:176]: The mesh pointer owned by the datacollection does not change, if that's what you're asking

File: src/docs/sphinx/dev_guide/state_manager.rst (hunk 9 above)
Comment (joshessman-llnl) [reply src/docs/sphinx/dev_guide/state_manager.rst:79]: `outputState` would still be useful for non-sidre output I think, e.g., via ParaView or something

File: src/docs/sphinx/dev_guide/state_manager.rst (hunk 8 above)
Comment (jamiebramwell) [reply src/docs/sphinx/dev_guide/state_manager.rst:71]: I think this is a fair assumption.

File: src/docs/sphinx/dev_guide/state_manager.rst (hunk 9 above)
Comment (jamiebramwell) [reply src/docs/sphinx/dev_guide/state_manager.rst:79]: That's true. We'll have to think about if we still want to support non-sidre output.

File: src/serac/physics/state/state_manager.cpp (hunk 10 above)
Comment (jamiebramwell) [reply src/serac/physics/state/state_manager.cpp:176]: As long as the raw pointer comparison is documented, I'm OK with this.

**Review by jamiebramwell — APPROVED** [review]

File: scripts/spack/packages/axom/package.py [hunk 11]
```diff
@@ -38,7 +38,7 @@ class Axom(CachedCMakePackage, CudaPackage):
     git      = "https://github.com/LLNL/axom.git"
 
     # SERAC EDIT START
-    version('0.6.0serac', commit='c2b8dee53a15392bb09cbfaa919afa1525f7c7d5', submodules="True")
+    version('0.6.0serac', commit='65d8e67bcf629bbfacb837ba6202d26f3d960364', submodules="True")
```
white238, line 41: Please bump the axom submodule as well to this

src/serac/physics/base_physics.hpp
```diff
@@ -31,16 +31,20 @@ class BasePhysics {
...
-  BasePhysics();
+  BasePhysics(mfem::ParMesh* pmesh = nullptr);
 
   /**
    * @brief Constructor that creates n entries in state_ of order p
    *
    * @param[in] n Number of state variables
    * @param[in] p Order of the solver
+   * @param[in] pmesh An optional mesh reference, must be provided to configure the module
+   * when a mesh other than the primary mesh is used
```
white238:
```suggestion
   * when a mesh other than the default mesh is used
```

src/serac/physics/state/state_manager.cpp
```diff
@@ -20,139 +20,166 @@ QuadratureData<void> dummy_qdata;
...
-void StateManager::initialize(axom::sidre::DataStore& ds, const std::string& collection_name_prefix,
-                              const std::string output_directory, const std::optional<int> cycle_to_load)
+// Initialize StateManager's static members - these will be fully initialized in StateManager::initialize
+std::unordered_map<std::string, axom::sidre::MFEMSidreDataCollection> StateManager::datacolls_;
+bool                                                                  StateManager::is_restart_      = false;
+std::string                                                           StateManager::collection_name_ = "";
+std::vector<std::unique_ptr<SyncableData>>                            StateManager::syncable_data_;
+axom::sidre::DataStore*                                               StateManager::ds_                = nullptr;
+std::string                                                           StateManager::output_dir_        = "";
+const std::string                                                     StateManager::default_mesh_name_ = "primary";
```
white238:
```suggestion
const std::string                                                     StateManager::default_mesh_name_ = "default";
```

src/serac/physics/state/state_manager.hpp
```diff
@@ -33,50 +34,54 @@ class StateManager {
...
   /**
    * @brief Factory method for creating a new FEState object, signature is identical to FEState constructor
    * @param[in] options Configuration options for the FEState, if a new state is created
+   * @param[in] tag A string that uniquely identifies the mesh on which the field is to be defined
    * @see FiniteElementState::FiniteElementState
    * @note If this is a restart then the options (except for the name) will be ignored
    */
-  static FiniteElementState newState(FiniteElementVector::Options&& options = {});
+  static FiniteElementState newState(FiniteElementVector::Options&& options = {},
+                                     const std::string&             tag     = default_mesh_name_);
```
white238: Is `tag` clear enough here?  I am concerned it may be confused as a tag for something else.  Since it is a mesh tag and not State, Dual, or QuadratureData tag.

I suggest all of these be changed to `mesh_tag`

src/serac/physics/thermal_conduction.hpp
```diff
@@ -206,8 +206,11 @@ class ThermalConduction : public BasePhysics {
    * @param[in] order The order of the thermal field discretization
    * @param[in] options The system solver parameters
    * @param[in] name An optional name for the physics module instance
+   * @param[in] pmesh An optional mesh reference, must be provided to configure the module
+   * when a mesh other than the primary mesh is used
```
white238:
```suggestion
   * when a mesh other than the default mesh is used
```

src/serac/physics/tests/serac_thermal_structural_solver.cpp
```diff
@@ -140,6 +140,278 @@ TEST(dynamic_solver, dyn_solve)
...
+
+    // What's awkward about using a mesh as the param is that we have to then extract the mesh
+    // instead of using the string ID we just used in the call to load()
+    // Note that just the 'auto' will not compile but only because ParMesh move ctor is explicitly deleted
+    auto& primary_mesh   = serac::StateManager::mesh(primary_datacoll_name);
+    auto& secondary_mesh = serac::StateManager::mesh(secondary_datacoll_name);
+
+    // initialize the dynamic solver object
+    ThermalSolid ts_solver_primary(1, therm_options, default_dynamic, "coupled_primary", &primary_mesh);
+    ts_solver_primary.setDisplacementBCs(ess_bdr, deform);
```
white238, line 333: Outside the scope of this PR but pretty sure we agreed to remove set and get from our function calls.

jamiebramwell: Yeah, I think this is a holdover from before that decision.

white238: #611

white238 (approved):

PR #599, opened 2021-11-23
multiple trial spaces

samuelpmishLLNL:
This PR aims to extend `serac::Functional` to support multiple trials spaces, for example a static thermomechanical problem might use the following:

```cpp
constexpr int dim = 2;

using displacement_space = H1< 2, dim >; // vector-valued displacement field
using temperature_space = H1< 2 >;       // scalar-valued temperature field

// create the Functional object, with a template parameter indicating that it will be using 
// vector-valued nodal test functions (i.e. computing nodal forces), and it will depend on a
// vector-valued nodal field (displacements) in the first position, and a scalar-valued nodal field (temperatures)
// in the second position
serac::Functional< displacement_space(displacement_space, temperature_space) > nodal_forces(...);

// for the case of multiple trial spaces, the q-function is passed the spatial position 
// of the quadrature point in the first argument, and then as many additional 
// arguments as there are trial spaces (following the order of the template arguments)
nodal_forces.AddAreaIntegral([=](auto x, auto displacement, auto temperature) {

  auto [u, du_dx] = displacement;
  auto [phi, unused] = temperature;
  
  // this material has a temperature-dependent elasticity model
  auto mu = mu0 - phi * dmu_dphi;

  auto strain = 0.5 * (du_dx + transpose(du_dx));

  auto stress = lambda * tr(strain) * I + 2.0 * mu * strain;
  auto body_force = zero{};

  return serac::tuple{body_force, stress};

}, domain_of_integration);

nodal_forces.AddBoundaryIntegral([=](auto x, auto n, auto displacement, auto temperature) {
  auto [u, du_dx] = displacement;
  auto [phi, unused] = temperature;
  return (x[0] + x[1] - cos(u[0])) * n;
}, domain_of_integration);

// to carry out the actual calculations, use the call operator `Functional::operator()`
auto residual = nodal_forces(displacements, temperatures);

// to get derivatives w.r.t. an input argument, decorate that argument with `differentiate_wrt(...)`.
// the output will be a tuple of the residual evaluation and an object that represents the derivative
// w.r.t. the specified argument:
auto [f, dfdu] = nodal_forces(differentiate_wrt(displacements), temperatures);

// if used directly, dfdu can be used to compute the action-of-gradient in a matrix-free way
mfem::Vector du = ...;       // a "small" change in the nodal displacement vector
mfem::Vector df = dfdu * du; // the corresponding "small" change in the nodal force vector

// alternatively, dfdu can be assembled into a sparse matrix by assigning it to `mfem::HypreParMatrix`
mfem::HypreParMatrix * K = dfdu;
mfem::Vector df = (*K) * du;
```


src/serac/infrastructure/accelerator.hpp
```diff
@@ -148,7 +148,7 @@ std::string getCUDAMemInfoString()
  * @tparam exec the memory space where the data lives
  * @param n how many entries to allocate in the array
  */
-template <typename T, ExecutionSpace exec>
+template <ExecutionSpace exec, typename T>
```
white238, line 257: Was there a reason that this was flipped? I know this is minor but scanning through most `ExecutionSpace exec` are the last template parameter.

src/serac/numerics/functional/boundary_integral.hpp
```diff
@@ -51,11 +53,14 @@ class BoundaryIntegral {
...
-  BoundaryIntegral(int num_elements, const mfem::Vector& J, const mfem::Vector& X, const mfem::Vector& normals,
-                   Dimension<dim>, lambda_type&& qf)
+  // BoundaryIntegral(int num_elements, const mfem::Vector& J, const mfem::Vector& X, const mfem::Vector& normals,
+  //                 Dimension<dim>, lambda_type&& qf)
+  BoundaryIntegral(int, const mfem::Vector& J, const mfem::Vector& X, const mfem::Vector& normals, Dimension<dim>,
+                   lambda_type&&)
       : J_(J), X_(X), normals_(normals)
 
   {
+#if 0
```
white238: Can you add a comment why this is ifdef'd out or remove it?

src/serac/infrastructure/accelerator.hpp
samuelpmishLLNL: These arguments were rearranged so that it would be consistent with the variadic overload of the function defined immediately below this one. For C++ parsing reasons, variadic packs can only appear at the end of the explicitly-specified template argument list, so the ExecutionSpace has to come first.

src/serac/numerics/functional/boundary_integral.hpp
samuelpmishLLNL: it's just a temporary `#ifdef` to get some of the tests compiling before the variadic support is implemented for the boundary integral kernels. This code should be updated and re-enabled before this PR is merged in.

src/serac/infrastructure/accelerator.hpp
white238: Excellent thanks for the explanation!

src/serac/numerics/functional/tests/array_test.cpp
```diff
@@ -55,4 +60,36 @@ int main()
...
   }
+
+  {
+    using element_type = serac::finite_element<serac::Geometry::Quadrilateral, serac::H1<2, 4> >;
+
+    serac::EVectorView<serac::ExecutionSpace::CPU, element_type> foo({nullptr}, 5);
+
+    mfem::DeviceTensor<3, double> bar{nullptr, 9, 4, 5};
+
+    std::cout << &serac::get<0>(foo.data)(0, 0, 1) - &serac::get<0>(foo.data)(0, 0, 0) << std::endl;
```
white238: Are these lines ultimately going to be checked against values?

src/serac/numerics/functional/tests/functional_multiphysics.cpp
```diff
@@ -0,0 +1,161 @@
...
+  double kappa = 1.0;
+
+  // Construct the new functional object using the known test and trial spaces
+  Functional<test_space(trial_space, trial_space)> residual(&fespace, {&fespace, &fespace});
+
+  residual.AddVolumeIntegral(
+      [=](auto x, auto temperature, auto dtemperature_dt) {
+        auto [u, du_dx]      = temperature;
+        auto [du_dt, unused] = dtemperature_dt;
+        auto source          = rho * cp * du_dt * du_dt - (100 * x[0] * x[1]);
```
jamiebramwell, line 123: This is really nice!

src/serac/numerics/functional/tests/functional_material_state_test.cpp
```diff
@@ -40,7 +40,7 @@ class QuadratureDataTest : public ::testing::Test {
     mesh                = &new_mesh;
     festate             = std::make_unique<FiniteElementState>(*mesh);
     festate->gridFunc() = 0.0;
-    residual            = std::make_unique<Functional<test_space(trial_space)>>(&festate->space(), &festate->space());
+    residual = std::make_unique<Functional<test_space(trial_space)>>(&festate->space(), std::array{&festate->space()});
```
jamiebramwell: Why do you have to explicitly declare `std::array` here? I assume obscure C++ reasons.

src/serac/numerics/functional/tests/functional_nonlinear.cpp
```diff
@@ -69,16 +66,20 @@ void check_gradient(Functional<T>& f, mfem::Vector& U)
   df1 -= f(U_minus);
   df1 /= (2 * epsilon);
 
+  auto [value, dfdU] = f(differentiate_wrt(U));
```
jamiebramwell, line 66: I really like this syntax as is also gets rid of the implicit caching and state we had before.

samuelpmishLLNL: there is still some internal state (the derivatives are evaluated at a specific point)-- this change is mostly cosmetic. If the internal state is undesirable to you, we could consider a way to remove it.

src/serac/numerics/functional/tests/functional_material_state_test.cpp
samuelpmishLLNL: I'm not actually sure if it's necessary here, I'll try removing it and see if the compiler yells at me.

samuelpmishLLNL: It does appear to be necessary in this context. It seems that the compiler can't deduce the type of `{&festate->space()}` when indirectly going through `std::make_unique`. Calling the constructor directly with something like
```cpp
Functional<test_space(trial_space, trial_space)> residual(&fespace, {&fespace, &fespace});
```
works (without specifying `std::array` in front)

jamiebramwell:
@sethwatts @tepperly @goxberry @junkudo , any comments on the interface would be appreciated. See [this multiphysics test](https://github.com/LLNL/serac/blob/mish2/variadic_trial_spaces/src/serac/numerics/functional/tests/functional_multiphysics.cpp) for an example of how it would be used in code. Thanks!

src/serac/infrastructure/accelerator.hpp
```diff
@@ -247,6 +253,29 @@ std::shared_ptr<T[]> make_shared_array(std::size_t n)
...
+  }
+
+#if defined(__CUDACC__)
+  if constexpr (exec == ExecutionSpace::GPU) {
+    T* data;
+    cudaMalloc(&data, sizeof(T) * n);
+    auto deleter = [](T* ptr) { cudaFree(ptr); };
+    return std::shared_ptr<T[]>(data, deleter);
+  }
+#endif
```
jamiebramwell: What happens if you ask for `ExecutionSpace:GPU` and it is not compiled with CUDA? Should we add an error?

src/serac/numerics/functional/tests/functional_boundary_test.cpp
```diff
@@ -57,35 +57,43 @@ bool operator==(const mfem::SparseMatrix& A, const mfem::SparseMatrix& B)
...
-  mfem::GridFunction U_plus = U;
-  U_plus.Add(epsilon, dU);
-  mfem::Vector f_plus = f(U_plus);
+  auto [value, dfdU] = f(differentiate_wrt(U));
+  mfem::Vector df2   = dfdU(dU);
 
-  mfem::GridFunction U_minus = U;
-  U_minus.Add(-epsilon, dU);
-  mfem::Vector f_minus = f(U_minus);
+  // mfem::HypreParMatrix* dfdU_matrix = dfdU;
```
jamiebramwell: Is there a reason we can't test the matrix assembly yet?

jamiebramwell: ohhhh is this the face restriction issue? If so, we should add a comment.

src/serac/numerics/functional/tests/functional_nonlinear.cpp
jamiebramwell: No - I think the internal state is fine. It's more the old implicit requirement that a user manually call the residual evaluation first.

src/serac/numerics/functional/tests/functional_multiphysics.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  {
+    mfem::Vector df1 = f(U_plus, dU_dt);
+    df1 -= f(U_minus, dU_dt);
+    df1 /= (2 * epsilon);
+
+    auto [value, dfdU] = f(differentiate_wrt(U), dU_dt);
+    mfem::Vector df2   = dfdU(dU);
+
+    mfem::HypreParMatrix* dfdU_matrix = dfdU;
```
jamiebramwell: I assume we're keeping the raw pointer interface to conform with MFEM standards for possible upstreaming.

src/serac/numerics/functional/tests/get_qf_derivative_type.cpp
```diff
@@ -0,0 +1,50 @@
+// these tests will be removed once axom::Array is ready
```
jamiebramwell, line 1: Is this still needed? If so, we should add a copyright notice.

src/serac/numerics/functional/tuple.hpp
```diff
@@ -172,6 +172,12 @@ struct tuple<T0, T1, T2, T3, T4, T5, T6, T7> {
 template <typename... T>
 tuple(T...) -> tuple<T...>;
 
+template <typename... T>
```
jamiebramwell, line 179: Needs doxygen

src/serac/numerics/functional/tuple_arithmetic.hpp
```diff
@@ -158,16 +158,104 @@ SERAC_HOST_DEVICE auto chain_rule_tuple_matvec(serac::tuple<T...> df_dx, serac::
  * @param[in] args The set of values
  * The gradients for each value will be set to 1 (or its tensor equivalent)
  */
-template <typename... T>
-constexpr auto make_dual(T... args)
+// template <typename... T>
```
jamiebramwell: Dead code.

src/serac/numerics/functional/tuple_arithmetic.hpp
```diff
@@ -158,16 +158,104 @@ SERAC_HOST_DEVICE auto chain_rule_tuple_matvec(serac::tuple<T...> df_dx, serac::
...
+//}
+
+///// @overload
+// template <typename... T>
+// constexpr auto make_dual(serac::tuple<T...> args)
+//{
+//  return detail::make_dual_helper(args, std::make_integer_sequence<int, int(sizeof...(T))>{});
+//}
+
+template <int i, typename S, typename T>
```
jamiebramwell, line 155: Needs doxygen.

src/serac/numerics/functional/tuple_arithmetic.hpp
```diff
@@ -158,16 +158,104 @@ SERAC_HOST_DEVICE auto chain_rule_tuple_matvec(serac::tuple<T...> df_dx, serac::
...
+    return make_dual(x);
+  } else {
+    return x;
+  }
+}
+
+template <int n, typename... T, int... i>
+constexpr auto make_dual_helper(const serac::tuple<T...>& args, std::integer_sequence<int, i...>)
+{
+  // Sam: it took me longer than I'd like to admit to find this issue, so here's an explanation
```
jamiebramwell, line 265: 👍

jamiebramwell (commented):
I still need to go through the implementation details, but I figure it was good to push these comments as a first start.

src/serac/infrastructure/accelerator.hpp
samuelpmishLLNL: I believe this will never happen with the current implementation, since the GPU specialization of the code that calls this function is gated behind some conditional compilation to turn it off when serac is not configured with cuda support.

However, we could do something like

```cpp
/**
 * @brief enum used for signalling whether or not to perform certain calculations on the CPU or GPU
 */
enum class ExecutionSpace
{
  CPU,
#ifdef SERAC_CUDA_ENABLED
  GPU,
  Dynamic  // Corresponds to execution that can "legally" happen on either the host or device
#endif
};
```

To explicitly fail everywhere if a user tries to ask for a GPU execution space without cuda support.

src/serac/numerics/functional/tests/functional_multiphysics.cpp
samuelpmishLLNL: I don't know what the best thing is to do here-- I don't really like the raw pointer. We could return a `std::unique_ptr< mfem::HypreParMatrix >`, or put an `mfem::HypreParMatrix` class member on the gradient object, and then it could just return a reference to that (this is what the QoI specialization is doing right now).

Other suggestions are welcome, too!

src/serac/numerics/functional/tests/functional_boundary_test.cpp
samuelpmishLLNL: oops, I commented this stuff out initially when rewriting the boundary kernels, but this feature is working (at least the tests we have are passing with it re-enabled)

src/serac/infrastructure/accelerator.hpp
jamiebramwell: I think that's a good idea just as a failsafe. Any thoughts @white238 @adayton1 ?

src/serac/numerics/functional/tests/functional_multiphysics.cpp
jamiebramwell: I like the second option. This also means we could reuse the sparsity pattern during reassembly operations. When the upstreaming negotiations happen, we can discuss changing this if required.

src/serac/numerics/functional/boundary_integral.hpp
```diff
@@ -51,107 +54,106 @@ class BoundaryIntegral {
...
+        using derivative_type   = decltype(get_derivative_type<i, dim, trials...>(qf));
+        auto ptr = accelerator::make_shared_array<exec, derivative_type>(num_elements * quadrature_points_per_element);
+        ExecArrayView<derivative_type, 2, exec> qf_derivatives(ptr.get(), num_elements, quadrature_points_per_element);
+
+        evaluation_with_AD_[i] =
+            EvaluationKernel{DerivativeWRT<i>{}, eval_config, qf_derivatives, J, X, N, num_elements, qf};
+
+        // note: this lambda function captures ptr by-value to extend its lifetime
+        //                        vvv
+        action_of_gradient_[i] = [ptr, qf_derivatives, num_elements, J](const mfem::Vector& dU, mfem::Vector& dR) {
```
jandrej, line 92: sneaky

jandrej (commented):
Looks great. I still like to test it thoroughly in terms of "interface usability" when I find some more time.

src/serac/numerics/functional/dual.hpp
```diff
@@ -222,6 +222,14 @@ auto cos(dual<gradient_type> a)
   return dual<gradient_type>{cos(a.value), -a.gradient * sin(a.value)};
 }
 
+/** @brief implementation of cosine for dual numbers */
```
jamiebramwell:
```suggestion
/** @brief implementation of sine for dual numbers */
```

src/serac/numerics/functional/evector_view.hpp
```diff
@@ -0,0 +1,71 @@
+#pragma once
```
jamiebramwell, line 7: Needs a copyright notice.

jamiebramwell: And doxygen.

src/serac/numerics/functional/integral_utilities.hpp
```diff
@@ -141,6 +141,28 @@ SERAC_HOST_DEVICE void Add(const mfem::DeviceTensor<3, double>& r_global, tensor
   }
 }
 
+#if 0
```
jamiebramwell, line 144: Why is this commented out?

src/serac/numerics/functional/functional.hpp
```diff
@@ -25,6 +25,46 @@
 
 namespace serac {
 
+// this type exists as a way to signal to `serac::Functional` that the function
+// serac::Functional::operator()` should differentiate w.r.t. a specific argument
+//
+// TODO: better name
+struct dual_vector {
```
jamiebramwell: `vector_to_differentiate`?

src/serac/numerics/functional/functional.hpp
```diff
@@ -69,66 +109,77 @@ class Functional;
...
+      G_trial_[i] = trial_space_[i]->GetElementRestriction(mfem::ElementDofOrdering::LEXICOGRAPHIC);
+      SLIC_ERROR_IF(!G_trial_[i], "Couldn't retrieve element restriction operator for trial space");
 
-    // Ensure the mesh has the appropriate neighbor information before constructing the face restriction operators
-    if (test_space_) {
-      test_space_->ExchangeFaceNbrData();
-    }
-    if (trial_space_) {
-      trial_space_->ExchangeFaceNbrData();
+      if (compatibleWithFaceRestriction(*trial_space_[i])) {
```
jamiebramwell, line 154: What happens if a fe space is not compatible with face restrictions? I assume it generates and error when the boundary integral is requested.

src/serac/numerics/functional/functional.hpp
```diff
@@ -237,49 +288,135 @@ class Functional<test(trial), exec> : public mfem::Operator {
     AddBoundaryIntegral(Dimension<2>{}, integrand, domain);
   }
 
-  /**
-   * @brief Implements mfem::Operator::Mult
-   * @param[in] input_T The input vector
-   * @param[out] output_T The output vector
-   */
-  void Mult(const mfem::Vector& input_T, mfem::Vector& output_T) const override
+  void ActionOfGradient(const mfem::Vector& input_T, mfem::Vector& output_T, size_t which) const
```
jamiebramwell, line 313: I appreciate this more explicit name.

src/serac/numerics/functional/functional.hpp
```diff
@@ -237,49 +288,135 @@ class Functional<test(trial), exec> : public mfem::Operator {
...
    */
-  mfem::Operator& GetGradient(const mfem::Vector& x) const override
+  template <typename... T>
+  typename operator_paren_return<T...>::type operator()(const T&... args)
   {
-    Mult(x, dummy_);  // this is ugly
-    return grad_;
-  }
+    constexpr int num_dual_arguments = (std::is_same_v<T, dual_vector> + ...);
+    static_assert(num_dual_arguments <= 1,
```
jamiebramwell: 👍

src/serac/numerics/functional/functional.hpp
```diff
@@ -288,49 +425,52 @@ class Functional<test(trial), exec> : public mfem::Operator {
    *
    * @note This gets more interesting when having more than one trial space
    */
-  void SetEssentialBC(const mfem::Array<int>& ess_attr)
+  void SetEssentialBC(const mfem::Array<int>& ess_attr, size_t which)
   {
-    static_assert(std::is_same_v<test, trial>, "can't specify essential bc on incompatible spaces");
-    trial_space_->GetEssentialTrueDofs(ess_attr, ess_tdof_list_);
+    // TODO check that it actually makes sense to apply bcs to this trial space
```
jamiebramwell, line 459: We should have a discussion as to whether we want `Functional` to be essential BC aware at all. I'll add a discussion point to the meeting agenda.

src/serac/numerics/functional/functional.hpp
```diff
@@ -413,144 +554,68 @@ class Functional<test(trial), exec> : public mfem::Operator {
...
+     *    grad<0>(f) == df_dtrial0
+     *    grad<1>(f) == df_dtrial1
+     *    grad<2>(f) == df_dtrial2
+     */
+    uint32_t which_argument;
 
-    // scatter-add to compute global residuals
-    P_test_->MultTranspose(output_L_, output_T);
+    /// @brief shallow copy of the test space from the associated Functional
+    mfem::ParFiniteElementSpace* test_space_;
```
jamiebramwell, line 606: Is there a reason you are storing pointers instead of references?

src/serac/numerics/functional/functional.hpp
```diff
@@ -413,144 +554,68 @@ class Functional<test(trial), exec> : public mfem::Operator {
...
-   */
+  /// @brief Manages DOFs for the test space
   mfem::ParFiniteElementSpace* test_space_;
 
-  /**
-   * @brief Manages DOFs for the trial space
-   */
-  mfem::ParFiniteElementSpace* trial_space_;
+  /// @brief Manages DOFs for the trial space
+  std::array<mfem::ParFiniteElementSpace*, num_trial_spaces> trial_space_;
```
jamiebramwell, line 643: I can see that this is the reason for pointers, although we could use reference wrappers.

src/serac/numerics/functional/domain_integral.hpp
```diff
@@ -141,17 +137,24 @@ class DomainIntegral {
...
+  }
 
   /**
    * @brief Applies the integral, i.e., @a output_E = gradient( @a input_E )
    * @param[in] input_E The input to the evaluation; per-element DOF values
    * @param[out] output_E The output of the evalution; per-element DOF residuals
    * @see gradient_kernel
    */
-  void GradientMult(const mfem::Vector& input_E, mfem::Vector& output_E) const
+  void GradientMult(const mfem::Vector& input_E, mfem::Vector& output_E, size_t which = 0) const
```
jamiebramwell: It might be more explicit (although more verbose) to change `which` to `which_trial` or `which_trial_space`.

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -8,86 +8,101 @@
...
+struct QFunctionArgument<Hcurl<p>, Dimension<2> > {
+  using type = tuple<tensor<double, 2>, double>;
+};
+
+template <int p>
+struct QFunctionArgument<Hcurl<p>, Dimension<3> > {
+  using type = tuple<tensor<double, 3>, tensor<double, 3> >;
+};
+
+// template <int dim, typename... trials, typename lambda>
```
jamiebramwell: Dead code.

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -101,10 +116,274 @@ void evaluation_kernel(const mfem::Vector& U, mfem::Vector& R, CPUArrayView<deri
 
     // once we've finished the element integration loop, write our element residuals
     // out to memory, to be later assembled into global residuals by mfem
-    detail::Add(r, r_elem, e);
+    detail::Add(r, r_elem, int(e));
   }
 }
 
+template <int i>
```
jamiebramwell, line 69: This section needs comments as to what all of these types do.

src/serac/numerics/functional/functional_qoi.inl
```diff
@@ -58,64 +58,81 @@ struct QoIElementRestriction : public mfem::Operator {
...
   using test = QOI;
+  static constexpr tuple<trials...> trial_spaces{};
+  static constexpr uint32_t         num_trial_spaces = sizeof...(trials);
+
+  class Gradient;
+
+  template <typename... T>
+  struct operator_paren_return {
+    using type = typename std::conditional<
+      (std::is_same_v<T, dual_vector> + ...) == 1,  // if the there is a dual number in the pack
```
jamiebramwell: I really like the readability of this metaprogramming. Thank you for the comments!

jamiebramwell (commented):
Just some minor comments. I think once these are resolved (and the PR passes the doxygen CI check), it should be good to go!

src/serac/numerics/functional/integral_utilities.hpp
samuelpmishLLNL: The poorly-named header "integral_utilities.hpp" was originally going to be for tools that were used by both the boundary and domain integrals. These specializations are specific to one type of integral only, so I moved them out of here and closer to kernels they're used by.

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: I'm leaning more and more toward moving essential BCs out of `serac::Functional`. They sort of make sense in the single-physics case (only when test and trial space match), but in all other cases (different test/trial spaces, multiple trial spaces, quantities of interest), it's unclear what it means to enforce boundary conditions.

src/serac/numerics/functional/integral_utilities.hpp
jamiebramwell: Sounds good. I would delete this code.

src/serac/numerics/functional/functional.hpp
jamiebramwell: For the record, we discussed this and determined it was best to move the essential boundary condition operations outside of `Functional`.

src/serac/numerics/functional/tests/functional_multiphysics.cpp
samuelpmishLLNL: After trying to make it return `mfem::HypreParMatrix &` for a bit, it seems this doesn't work out so well after all (my bad). It seems like it makes more sense to really pass ownership of the matrix when assembling it, rather than returning a reference. I'm going to try implementing the interface below instead:

```cpp
struct Functional{
  ...
  struct Gradient{
    friend std::unique_ptr< mfem::HypreParMatrix > assemble(Gradient & grad);
  };
  ...
};
```

It would also be nice if the relevant mfem types (`mfem::HypreParMatrix`, `mfem::HypreParVector`) supported move constructors, so that we wouldn't need to wrap things in `std::unique_ptr`.

CMakeLists.txt
```diff
@@ -162,6 +162,12 @@ install(TARGETS              serac
         )
 
 if (SERAC_ENABLE_CODEVELOP)
+    # This really shouldn't be needed but it keeps getting dropped from axom.
```
white238: This should go away once you merge/rebase onto develop

src/serac/infrastructure/accelerator.hpp
```diff
@@ -247,6 +254,29 @@ std::shared_ptr<T[]> make_shared_array(std::size_t n)
 #endif
 }
 
+/**
+ * @brief create shared_ptr to an array of `n` values of type `T`, either on the host or device
```
white238, line 275:
```suggestion
 * @brief create shared_ptr to an array of \p n values of type \p T, either on the host or device
```

src/serac/infrastructure/accelerator.hpp
white238: I think adding another define would be confusing to figure out how it differs from the ones we have already.  This should be fine with just this:

```
#if defined(AXOM_USE_CUDA) && defined(__CUDACC__)
```

src/serac/numerics/functional/functional.hpp
```diff
@@ -25,6 +25,54 @@
...
+ *     mfem::Vector just_the_value = my_functional(arg0, arg1);
+ *     auto [value, gradient_wrt_arg1] = my_functional(arg0, differentiate_wrt(arg1));
+ * @endcode
+ */
+dual_vector differentiate_wrt(const mfem::Vector& v) { return dual_vector{v}; }
+
+/**
+ * @tparam T a list of types, containing at most 1 `dual_vector`
+ *
+ * @brief given a list of types, this function returns the index that corresponds to the type `dual_vector`.
```
white238:
```suggestion
 * @brief given a list of types, this function returns the index that corresponds to the type `dual_vector`.
 *
 * @tparam T a list of types, containing at most 1 `dual_vector`
```

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -127,8 +127,8 @@ class ThermalConductionFunctional : public BasePhysics {
                                                                .vector_dim = 1,
                                                                .ordering   = mfem::Ordering::byNODES,
                                                                .name       = detail::addPrefix(name, "temperature")})),
-        M_functional_(&temperature_.space(), &temperature_.space()),
-        K_functional_(&temperature_.space(), &temperature_.space()),
+        M_functional_(&temperature_.space(), {&temperature_.space()}),
```
white238, line 130: Having to have the brackets is awkward here.  Can we add a constructor to avoid it?

samuelpmishLLNL: Are you requesting an extra constructor when there is only 1 trial-space case that looks like

```cpp
serac::Functional(test_space, trial_space)
```

or asking that the existing constructor use a variadic list, rather than `std::array` for its trial spaces? e.g.

```cpp
serac::Functional(test_space, trial_space_0, trial_space_1, ... trial_space_n)
```

(instead of the current):

```cpp
serac::Functional(test_space, {trial_space_0, trial_space_1, ... trial_space_n})
```

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: Does this change the doxygen output, or is it a stylistic choice? 

If it's a style thing, maybe we could put the expected doxygen comment format [here](https://serac.readthedocs.io/en/latest/sphinx/dev_guide/style_guide.html#documentation).

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: Do you think this change should be added to this PR, or come at a later date?

src/serac/physics/thermal_conduction_functional.hpp
white238: Okay I understand now.  I missed the first parameter was a `test_space` and the second is an array of `trial_space`'s.

This may not make sense but if the test and trial spaces are the same would it be helpful to just have a single argument constructor?

src/serac/numerics/functional/functional.hpp
white238: Style/consistency.  We can handle it after this PR.

#629

src/serac/numerics/functional/tuple_arithmetic.hpp
```diff
@@ -151,23 +151,146 @@ SERAC_HOST_DEVICE auto chain_rule_tuple_matvec(serac::tuple<T...> df_dx, serac::
 }
 
 }  // namespace detail
+
+template <int i, typename S, typename T>
+struct one_hot_helper;
```
white238, line 156: What is `hot` in this context?

src/serac/numerics/functional/tests/tuple_arithmetic_unit_tests.cpp
```diff
@@ -144,11 +144,9 @@ TEST(ChainRuleTests, scalar_output_with_matrix_input)
   EXPECT_NEAR(((f(x + epsilon * dx) - f(x)) / epsilon) - chain_rule(dfdx, dx), 0.0, 1.e-6);
 }
 
-TEST(ChainRuleTests, tuple_output_with_tuple_input)
+TEST(ChainRuleTests, tensor_output_with_tuple_input)
```
white238: It's very helpful to have the filename as the first parameter so when one fails you know which file to go into to fix it.

src/serac/numerics/functional/tuple_arithmetic.hpp
samuelpmishLLNL: it's a reference to [one-hot encoding](https://en.wikipedia.org/wiki/One-hot), where you have a list of entries, all of which are "off" except for one

white238: k. Can you add this to the doxygen comments please?

src/serac/numerics/functional/tests/functional_multiphysics.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+    mfem::Vector df3 = (*df_ddU_dt_matrix) * ddU_dt;
+
+    double relative_error1 = df1.DistanceTo(df2) / df1.Norml2();
+    double relative_error2 = df1.DistanceTo(df3) / df1.Norml2();
+
+    EXPECT_NEAR(0., relative_error1, 5.e-5);
+    EXPECT_NEAR(0., relative_error2, 5.e-5);
+
+    std::cout << relative_error1 << " " << relative_error2 << std::endl;
```
white238: nitpick. This could be a slic call.

src/serac/numerics/functional/tests/functional_multiphysics.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "serac/infrastructure/input.hpp"
+#include "serac/serac_config.hpp"
+#include "serac/mesh/mesh_utils_base.hpp"
+#include "serac/numerics/expr_template_ops.hpp"
+#include "serac/numerics/stdfunction_operator.hpp"
+#include "serac/numerics/functional/functional.hpp"
+#include "serac/numerics/functional/tensor.hpp"
+
+using namespace serac;
+using namespace serac::profiling;
```
white238: Is there any profiling in this?

src/serac/numerics/functional/tests/functional_comparisons.cpp
```diff
@@ -71,24 +66,19 @@ void functional_test(mfem::ParMesh& mesh, H1<p> test, H1<p> trial, Dimension<dim
   A.AddDomainIntegrator(new mfem::DiffusionIntegrator(b_coef));
 
   // Assemble the bilinear form into a matrix
-  {
-    SERAC_PROFILE_SCOPE(concat("mfem_localAssemble", postfix));
-    A.Assemble(0);
-  }
+  A.Assemble(0);
```
white238, line 69: `A` should be renamed to `Avengers` for obvious reasons.

src/serac/numerics/functional/functional_qoi.inl
```diff
@@ -58,63 +58,80 @@ struct QoIElementRestriction : public mfem::Operator {
...
-    }
+    output_T_.SetSize(1, mfem::Device::GetMemoryType());
 
-    {
-      auto num_bdr_elements           = static_cast<size_t>(trial_space_->GetNFbyType(mfem::FaceType::Boundary));
-      auto ndof_per_test_bdr_element  = static_cast<size_t>(1);
-      auto ndof_per_trial_bdr_element = static_cast<size_t>(trial_space_->GetBE(0)->GetDof() * trial_space_->GetVDim());
-      bdr_element_gradients_ =
-          ExecArray<double, 3, exec>(num_bdr_elements, ndof_per_test_bdr_element, ndof_per_trial_bdr_element);
+    for (uint32_t i = 0; i < num_trial_spaces; i++) {
```
white238, line 118: Used to be the case, pretty sure it still is... `++i` was faster than `i++`

src/serac/numerics/functional/tuple_arithmetic.hpp
samuelpmishLLNL: this is explained in more detail on `one_hot_t` (which is the type that is actually intended to be used):
```
/**
 * @brief a tuple type with n entries, all of which are of type `serac::zero`,
 * except for the i^{th} entry, which is of type T
 *
 *  e.g. one_hot_t< 2, 4, T > == tuple<zero, zero, T, zero>
 */
template <int i, int n, typename T>
using one_hot_t = typename one_hot<i, n, T>::type;
```
Unfortunately, these other types (e.g. `one_hot`, `one_hot_helper`) are boilerplate that are practically required by the language to implement this sort of thing (`one_hot_t`). 

Up until now, I've just been documenting the main type (e.g. `one_hot_t`) to avoid redundant doxygen comments, but I could start including those additional comments if you think it's helpful.

src/serac/numerics/functional/tests/functional_multiphysics.cpp
samuelpmishLLNL: nope, just removed it!

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: `num_trial_spaces` will likely never be greater than 4 anyway, so I don't think there would any measurable difference in runtime

CMakeLists.txt
samuelpmishLLNL: I think it's a good idea, but then we'd have to amend all of our copyright statements to
```
// Copyright (c) 2019-2022, Disney, the Marvel Cinematic Universe, Lawrence Livermore National Security, LLC and
// other Serac Project Developers. See the top-level LICENSE file for ...
```
so we'll probably stick with `A`

jandrej (approved):
Besides the already stated comments, the code looks good to me. I tested a few simple and more complicated examples and the bare user experience seems very pleasant!

src/serac/numerics/functional/domain_integral.hpp
```diff
@@ -50,59 +52,54 @@ class DomainIntegral {
...
-
-      element_gradient_ = [qf_derivatives, num_elements, &J](CPUArrayView<double, 3> K_e) {
-        domain_integral::element_gradient_kernel<geometry, test_space, trial_space, Q>(K_e, qf_derivatives, J,
-                                                                                       num_elements);
-      };
+        };
+      });
     }
 
+#if 0
```
jamiebramwell, line 105: This needs an in-code comment stating why it is if-def'd out

src/serac/numerics/functional/tests/CMakeLists.txt
```diff
@@ -54,8 +56,8 @@ if(ENABLE_CUDA)
 
     set(functional_tests_cuda 
         tensor_unit_tests_cuda.cu 
-        functional_comparisons_cuda.cu
-        quadrature_data_cuda.cu
+#        functional_comparisons_cuda.cu
```
jamiebramwell, line 62: Comments are needed why this is commented out.

src/serac/numerics/functional/tests/get_qf_derivative_type.cpp
jamiebramwell: This comments stands. If we need this, it should have a copyright. If not, it should be deleted.

src/serac/numerics/functional/tuple_arithmetic.hpp
jamiebramwell: While it involves duplication, I think its useful to have the doxygen here as well so developers don't have to go digging for the explanation.

src/serac/numerics/functional/domain_integral.hpp
samuelpmishLLNL:
```suggestion
// some of the GPU functionality is temporarily disabled to 
// help incrementally roll-out the variadic implementation of Functional
// TODO: re-enable GPU kernels in a follow-up PR
#if 0
```

src/serac/numerics/functional/tests/CMakeLists.txt
samuelpmishLLNL:
```suggestion
#        some of the GPU functionality is temporarily disabled to 
#        help incrementally roll-out the variadic implementation of Functional
#        TODO: re-enable GPU kernels in a follow-up PR
#        functional_comparisons_cuda.cu
```

jamiebramwell (approved):

src/serac/numerics/functional/tuple_arithmetic.hpp
white238: Or just say go look at this other struct

src/serac/numerics/functional/tests/functional_multiphysics.cpp
white238: did you remove it?

white238 (approved):

PR #610, opened 2021-12-06
Get rid of Sidre warnings from `BasePhysics::saveSummary()`

white238:
This function was accessing a sidre group that doesnt exist on non-root nodes.  This fixes that.

Basically `SLIC_ERROR_ROOT_IF` was only failing if the root node didn't have the sidre group, then `getGroup` would throw a warning on all other nodes.


jamiebramwell (approved):

adayton1 (approved):
LGTM

PR #612, opened 2021-12-08
Fix the default BC in the adjoint solid solve

jamiebramwell:
This correctly sets the RHS to zero for the default homogenous essential boundary condition in the solid module adjoint solve. Previously, the RHS was not set while the matrix entries were eliminated.

This is needed for the LiDO project.


jamiebramwell:
ping @junkudo @sethwatts

samuelpmishLLNL (approved):

white238 (approved):

PR #613, opened 2021-12-13
Fix the MFEM/Axom co-develop documentation

jamiebramwell:
Built docs: https://serac.readthedocs.io/en/doc-bramwell-fix_codevelop/sphinx/dev_guide/mfem_dev.html


white238 (approved):

src/docs/sphinx/dev_guide/mfem_dev.rst
```diff
@@ -3,58 +3,27 @@
...
-		
-    $ mkdir <build>
-    $ cd <build>
-    $ cmake -C <serac/host-config.cmake> ..
+   # Initialize the submodules for MFEM and Axom.
+   $ git submodule update --init --recursive
 
+   # Configure our CMake build with the co-develop option using a local host-config (CMake configuration file) generated by spack.
+   # This uses CMake to configure Axom, MFEM, and Serac in one step.
+   $ ./config-build -hc <host config file> -DSERAC_ENABLE_CODEVELOP=On
```
samuelpmishLLNL, line 25: I understand that this is probably the preferred way of building on LC machines, but can't we also just configure `cmake` normally? e.g. I build serac without ever invoking `config-build`, with

```
/path/to/serac $ cmake . -C path/to/spack/generated/cache.cmake -B build -DSERAC_ENABLE_CODEVELOP=On
```

I imagine using `cmake` directly is probably more intuitive for people outside of LLNL

samuelpmishLLNL (approved):

src/docs/sphinx/dev_guide/mfem_dev.rst
jamiebramwell: I tentatively want to say yes, as I think I was doing this with the vscode plugin. Any thoughts @white238 ?

white238: I am fine with providing both.  I'll do a small PR shortly.

Side Note @samuelpmishLLNL : always put the path to the directory your `CMakeLists.txt` at the end of your cmake command.  Unfortunately CMake's command line arguments are order dependent and I have found very very hard to figure out behavior differences between putting it at the front or end of the command line.

```
mkdir build && cd build
cmake -C path/to/initialcache.cmake  ..
```

PR #615, opened 2021-12-16
Fix the solid sensitivity calculations

jamiebramwell:
This fixes a bug in the solid mechanics sensitivity calculations and adds a test using a fake adjoint pseudoload.

### DIFF
--- a/src/serac/physics/tests/serac_solid_sensitivity.cpp
+++ b/src/serac/physics/tests/serac_solid_sensitivity.cpp
@@ -1,0 +1,169 @@
+#include <gtest/gtest.h>
+#include <mpi.h>
+#include <serac/physics/solid.hpp>
+#include <serac/physics/state/state_manager.hpp>
+#include <serac/physics/state/finite_element_state.hpp>
+#include <serac/physics/state/finite_element_dual.hpp>
+#include <mfem.hpp>
+#include <memory>
+
+TEST(SeracFEState, finiteDiff)
+{
+  axom::sidre::DataStore datastore;
+  serac::StateManager::initialize(datastore, "does_not_matter");
+
+  // Build a mesh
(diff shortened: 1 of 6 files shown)


jamiebramwell:
ping @junkudo

src/serac/physics/tests/serac_solid_sensitivity.cpp
```diff
@@ -0,0 +1,169 @@
...
+    std::cout << "shearModulusSensitivity " << shearModulusSensitivity.localVec()(ix) << std::endl;
+    EXPECT_NEAR(dqoi_dshear, shearModulusSensitivity.localVec()(ix), 1.0e-4);
+  }
+}
+
+int main(int argc, char** argv)
+{
+  MPI_Init(&argc, &argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  const auto result = RUN_ALL_TESTS();
```
white238: You need to initialize a SLIC logger otherwise all output from through SLIC (axom and serac) will be noops.

jandrej (approved):

src/serac/physics/tests/serac_solid_sensitivity.cpp
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    // Compute numerical value of sensitivity of qoi with respect to bulk modulus
+    // by taking the inner product between adjoint load and displacement sensitivity
+    double dqoi_dshear = adjointLoad(du_dbulk);
+
+    // See if these are similar
+    std::cout << "dqoi_dshear " << dqoi_dshear << std::endl;
+    std::cout << "shearModulusSensitivity " << shearModulusSensitivity.localVec()(ix) << std::endl;
+    EXPECT_NEAR(dqoi_dshear, shearModulusSensitivity.localVec()(ix), 1.0e-4);
```
samuelpmishLLNL: I think it would be preferable to use a relative error measure (if the values being compared are nonzero) or scale the tolerance to an appropriate characteristic dimension for the problem. Otherwise, it's not clear if the absolute error of `1.0e-4` is big or small.

samuelpmishLLNL (approved):

PR #616, opened 2021-12-23
Remove Ascent

white238:
The very little we were using Ascent for has been replaced by the summary file.  We should re-add this if we ever add in-situ vis support.


samuelpmishLLNL (approved):

jamiebramwell (approved):
Removing unused TPLs is a vital emergency. 😄

PR #614, opened 2021-12-15
ATS

white238:
This PR handles adding ATS as a devtool and adds a basic ATS test workflow.

Important files:

`tests/integration/ats.sh.in`: Helper script that gets generated at the CMake step to run ats from the build directory
`tests/integration/test.ats`: Base ATS test file that exposes helpful variables and functions.  This also loads all subdirectory's `*.ats` files that it finds.
`tests/integration/default/default.ats`: adds a serial and parallel integration test for `data/input_files/default.lua`
`tests/integration/default/default_serial.json`: serial baseline summary file
`tests/integration/default/default_parallel.json`: parallel baseline summary file

Test repo PR: https://github.com/LLNL/serac_tests/pull/1

Relevant docs:
https://serac.readthedocs.io/en/task-white238-ats_devtool/sphinx/dev_guide/codevelop.html
https://serac.readthedocs.io/en/task-white238-ats_devtool/sphinx/dev_guide/testing.html

~~ATS Spack build needs this PR (just using the fork for now until this gets merged):
https://github.com/spack/spack/pull/27798~~
~~which also forces this to be fixed:
https://github.com/LLNL/uberenv/issues/73~~

- [x] Update TPL spack sha to past where the new uberenv needs
- [x] Build devtools on all platforms in shared location
- [x] Rebuild/hack regular host-configs to use new devtools
- [ ] Add rest of input-files
- [ ] Force @jamiebramwell to approve baselines
- [x] Figure out gitlab allocation scheme

PR #617, opened 2021-12-23
Bugfix: fix summary file name and remove now unneeded collection name from `StateManager`

white238:
Needed for #614

### DIFF
--- a/src/drivers/serac.cpp
+++ b/src/drivers/serac.cpp
@@ -230,7 +230,7 @@ int main(int argc, char* argv[])
   }
 
   // Output summary file (basic run info and curve data)
-  serac::output::outputSummary(datastore, serac::StateManager::collectionName(), output_directory);
+  serac::output::outputSummary(datastore, output_directory);
 
   serac::exitGracefully();
 }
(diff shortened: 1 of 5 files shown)


src/serac/infrastructure/output.cpp
```diff
@@ -68,15 +68,16 @@ void outputFields(const axom::sidre::DataStore& datastore, const std::string& da
   conduit::Node extracts;
   // "relay" is the Ascents Extract type for saving data
   extracts["e1/type"]            = "relay";
-  const std::string file_name    = axom::fmt::format("{}_fields.{}", data_collection_name, file_format_string);
+  const std::string file_name    = axom::fmt::format("fields.{0}", file_format_string);
```
white238: This whole function gets removed in #616

src/drivers/serac.cpp
```diff
@@ -232,9 +232,9 @@ int main(int argc, char* argv[])
     last_step = (t >= t_final - 1e-8 * dt);
   }
 
-  serac::output::outputSummary(datastore, serac::StateManager::collectionName(), output_directory);
```
white238, line 233: The setting of this variable got removed in #597. I decided instead of re-adding that line to just make the file more generically named due to it living in a directory we create.

`serac_datacoll_summary.json` -> `_summary.json` -> `summary.json`

jamiebramwell (approved):

samuelpmishLLNL (approved):

PR #622, opened 2022-01-18
fix missing include directory for axom_smoketest

samuelpmishLLNL:
This is a small workaround to an error related to a missing include directory when building `axom_smoketest.cpp` when using axom as a submodule. Thanks to @white238 for proposing the fix!

I don't believe we fully understand why this workaround is necessary, so it's still worth investigating why one of the `axom` target's include directories is not being propagated to targets linked to it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,6 +164,14 @@ install(TARGETS              serac
 if (SERAC_ENABLE_CODEVELOP)
     # Add Axom's targets to our export set
     # TODO: we should ask Axom for this list in the future
+
+
+    # This really shouldn't be needed but it keeps getting dropped from axom.
+    # It certainly shouldn't go here either.
+    target_include_directories(axom SYSTEM PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/axom/src>
+    )
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -164,6 +164,14 @@ install(TARGETS              serac
...
     # Add Axom's targets to our export set
     # TODO: we should ask Axom for this list in the future
+
+
+    # This really shouldn't be needed but it keeps getting dropped from axom.
+    # It certainly shouldn't go here either.
+    target_include_directories(axom SYSTEM PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/axom/src>
+    )
+
```
white238:
```suggestion
    # This really shouldn't be needed but it keeps getting dropped from axom.
    # It certainly shouldn't go here either.
    target_include_directories(axom SYSTEM PUBLIC
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/axom/src>
    )

    # Add Axom's targets to our export set
    # TODO: we should ask Axom for this list in the future
```

white238:
This is a very frustrating behavior that is coming out of CMake.  Congrats on being the 3rd person to try to fix it =) We should have a counter!

white238 (approved):

jamiebramwell (approved):

PR #623, opened 2022-01-19
Change the example output directory

jamiebramwell:
When the examples were run in the same directory as the executable, the output would fail as we are using the same name.


jamiebramwell:
ping @greene30

white238 (approved):

samuelpmishLLNL (approved):

PR #626, opened 2022-01-20
Performance tracking with Adiak/Caliper/Spot

adayton1:
* Added Adiak as a dependency
* Benchmarked thermal backed by functional
* Marked regions for profiling


adayton1:
@white238, can you point me to where I should start with getting adiak to build with spack?

src/serac/physics/benchmarks/benchmark_serac_thermal_functional.cpp
```diff
@@ -0,0 +1,191 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
```
adayton1: This is copied from physics/tests/serac_thermal_functional.cpp. In order to avoid code duplication for benchmarks, it probably makes sense to just do benchmarks for integrated tests and reuse the input decks.

src/serac/physics/benchmarks/benchmark_serac_thermal_functional.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+int main(int argc, char* argv[])
+{
+  MPI_Init(&argc, &argv);
+
+  axom::slic::SimpleLogger logger;
+
+  // Initialize Adiak
+  MPI_Comm comm = MPI_COMM_WORLD;
+  adiak::init(&comm);
```
adayton1: TODO: Wrap all this adiak code in a function called something like serac::profile_init.

src/serac/physics/benchmarks/benchmark_serac_thermal_functional.cpp
```diff
@@ -0,0 +1,191 @@
...
+  // Initialize Adiak metadata
+  adiak::launchdate();
+  adiak::executable();
+  adiak::cmdline();
+  adiak::clustername();
+  adiak::jobsize();
+  adiak::walltime();
+  adiak::cputime();
+  adiak::systime();
+
```
adayton1: TODO: Add the name of the test as Adiak metadata.

src/serac/numerics/functional/domain_integral.hpp
```diff
@@ -53,6 +53,8 @@ class DomainIntegral {
   DomainIntegral(int num_elements, const mfem::Vector& J, const mfem::Vector& X, Dimension<dim>, lambda_type&& qf,
                  QuadratureData<qpt_data_type>& data = dummy_qdata)
   {
+    SERAC_MARK_START("Domain Integral Set Up");
```
adayton1: These are the profiling regions.

adayton1: TODO: Rename SERAC_MARK_START to SERAC_MARK_BEGIN to match CALIPER_MARK_BEGIN

white238:
> @white238, can you point me to where I should start with getting adiak to build with spack?

We should sit down and talk about this. It's probably easier to do a 10 minute webex than try to point you at everything.

Good news there is an existing Adiak spack package:

https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/adiak/package.py

Basically we should change the current `caliper` spack variant to be something more generic like `performance` and have it install adiak as well.

This would happen in this file: 

https://github.com/LLNL/serac/blob/develop/scripts/spack/packages/serac/package.py

adayton1:
Thanks @white238, I just sent you a webex invite.

src/serac/infrastructure/profiling.cpp
```diff
@@ -20,6 +19,57 @@ std::optional<cali::ConfigManager> mgr;
...
+  } else {
+    SLIC_WARNING_ROOT("Caliper options invalid, ignoring: " << check_result);
+  }
+
+  // Defaults, should probably always be enabled
+  mgr->add("event-trace,runtime-report,spot");
+  mgr->start();
+#else
+  // Silence warnings
+  static_cast<void>(options);
```
white238: You can use this since we are C++17:

https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

src/serac/physics/benchmarks/benchmark_serac_thermal_functional.cpp
adayton1: Done.

src/serac/numerics/functional/domain_integral.hpp
adayton1: Done.

src/serac/infrastructure/profiling.cpp
adayton1: Done. Thanks!

src/serac/physics/benchmarks/benchmark_serac_thermal_functional.cpp
adayton1: This will be considered in a follow on pull request.

src/serac/physics/benchmarks/benchmark_serac_thermal_functional.cpp
adayton1: Done.

scripts/spack/packages/serac/package.py
```diff
@@ -141,7 +141,7 @@ class Serac(CachedCMakePackage, CudaPackage):
     # Libraries that do not have a debug variant
     depends_on("conduit@0.7.2serac~shared~python~test")
     depends_on("adiak@0.2.1~shared+mpi", when="+profiling")
-    depends_on("caliper@master~shared+mpi+adiak~papi", when="+profiling")
+    depends_on("caliper@2.7.0~shared+mpi+adiak~papi", when="+profiling")
```
jamiebramwell, line 144: 👍

src/docs/sphinx/dev_guide/profiling.rst
```diff
@@ -3,37 +3,79 @@
...
+
+To provide custom metadata for comparing program runs, call ``SERAC_SET_METADATA(name, data)``
+after ``serac::profiling::initialize()`` and before ``serac::profiling::finalize``.
+This will add extra metadata into the ``.cali`` file. Supported metadata types are
+integrals, floating points, and strings. Note that this macro is a no-op if the
+``profiling`` variant is not used.
+
+.. code-block:: c++
+		
+   SERAC_SET_METADATA("dimensions", 3);
```
jamiebramwell: The star mesh is dimension 2, so we should probably change this.
```suggestion
   SERAC_SET_METADATA("dimensions", 2);
```

jamiebramwell (approved):
Looks good! Thanks @adayton1 !

samuelpmishLLNL (approved):
This is not a criticism of the PR per se, but it is a little bit concerning to me how much work (permanent code annotations, additional dependencies, complexity of deciphering caliper database files, etc) is required to do this kind of profiling. Some other methods require significantly less invasive modification to the code base.

I hope that the benefits outweigh the additional effort required.

src/serac/numerics/functional/domain_integral.hpp
```diff
@@ -150,9 +152,13 @@ class DomainIntegral<test(trials...), exec> {
             int which_trial_space) const
   {
     if (which_trial_space == -1) {
+      SERAC_MARK_BEGIN("Domain Integral Evaluation");
       evaluation_(input_E, output_E);
+      SERAC_MARK_END("Domain Integral Evaluation");
```
samuelpmishLLNL, line 157: If a `SERAC_MARK_BEGIN(foo)` has no matching `SERAC_MARK_END(foo)`, I imagine the code still compiles without error (and just contains some mislabeled/useless timings)? It would be nice if we had a better way to catch small errors like that, but I doubt it will be a big issue.

jamiebramwell:
> This is not a criticism of the PR per se, but it is a little bit concerning to me how much work (permanent code annotations, additional dependencies, complexity of deciphering caliper database files, etc) is required to do this kind of profiling. Some other methods require significantly less invasive modification to the code base.
> 
> I hope that the benefits outweigh the additional effort required.

I agree we should keep an eye on this and upstream any possible solutions or simplifications to the caliper/adiak/spot teams. I think the biggest win for this framework is the performance monitoring capability. However, that could change so we should monitor other solutions as development progresses (google benchmarks is the obvious example).

src/docs/sphinx/dev_guide/profiling.rst
```diff
@@ -3,37 +3,79 @@
...
+=======================================
+Profiling Serac using Adiak and Caliper
+=======================================
+
+Introduction to Adiak
+---------------------
+
+`Adiak <https://github.com/LLNL/Adiak>`_ is a library developed at LLNL for collecting
+metadata that can be used to compare multiple runs across programs.  The full documentation
+can be found `here <https://github.com/LLNL/Adiak/blob/master/docs/Adiak%20API.docx>`_,
```
white238: Unfortunately these need to be uniquely named and will cause a warning, seen here:

https://dev.azure.com/llnl-serac/serac/_build/results?buildId=6359&view=logs&j=1be3ff84-5b97-5584-c1a4-7ebbbeb3b802&t=d32b9d99-ea3e-5518-9381-52c611e33fea&l=2432

src/docs/sphinx/dev_guide/profiling.rst
```diff
@@ -3,37 +3,79 @@
...
+`SPOT https://software.llnl.gov/news/2021/01/07/spot-new/`_ is a framework developed at
+LLNL for vizualizing performance data.  SPOT is an external tool and does not need to be
+linked into Serac.
+
+Build Instructions
+------------------
 
-To use Caliper with Serac, install the ``caliper`` variant of ``serac`` with Spack, i.e., ``serac+caliper``.
+To use Adiak and Caliper with Serac, install the ``profiling`` variant of ``serac``
+with Spack, i.e., ``serac+profiling``.
```
white238: Add a note that these are pre-built as part of the installed libraries on LC.

src/serac/numerics/functional/domain_integral.hpp
adayton1: Yeah, the code will compile without error. I will check and see if there is at least a run time warning. I'm not sure there's a way to do it at compile time and allow arbitrary strings as the argument, but I will think about that.

src/docs/sphinx/dev_guide/profiling.rst
adayton1: Fixed!

src/docs/sphinx/dev_guide/profiling.rst
adayton1: Done!

scripts/spack/packages/serac/package.py
adayton1: @white238, are you OK with this change? I prefer having the fixed version so that we don't get surprise updates. I guess I would need to rebuild the pre-built libraries on LC?

src/docs/sphinx/dev_guide/profiling.rst
adayton1: Oh, I didn't know that about rst. I believe I fixed it.

adayton1:
Yes, this solution is a bit more invasive. But we can certainly re-evaluate after trying it out for a bit and also keep an eye out for other solutions (e.g. perhaps even add support for google benchmarks to output to a sql database).

scripts/spack/packages/serac/package.py
white238: yes I prefer fixed versions.. Thanks for catching this...

PR #628, opened 2022-01-27
update devtool packages

white238:
I hacked the devtools into the previous host-configs in the ATS PR and missed these.  This is a noop change but will cause problems if we don't fix them for the next TPL build.


jamiebramwell (approved):

adayton1 (approved):
LGTM

PR #632, opened 2022-02-03
add precompiled header support

samuelpmishLLNL:
see: https://github.com/LLNL/serac/issues/567

This PR revisits adding support for precompiled headers in serac. They tend to reduce compilation time a bit (tested on `serac::Functional` tests):

|    |  without precompiled headers |  with precompiled headers |
|:---: |:---: |:---: |
|`make`| 65s | 52s |
|`make -j`| 12.0s | 9.5s |
|`ninja`|11.5s| 9.1s|

Some things worth looking into:
- many of our tests have the flag `-fPIE`, but I'm not sure where it's getting added. The precompiled headers must have the same flags as the targets that use them, so for now position-independent code is turned off.
- right now, precompiled headers are not tested with CUDA, as the CUDA TPL builds seem to be broken. I'm not sure how well the NVCC toolchain supports this feature (if at all)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,14 @@ include(${BLT_SOURCE_DIR}/SetupBLT.cmake)
 # Setup Macros and dependencies
 #------------------------------------------------------------------------------
 
+blt_add_library(
+    NAME        serac_precompiled_headers
+    SOURCES     ${PROJECT_SOURCE_DIR}/src/serac/infrastructure/pch.cpp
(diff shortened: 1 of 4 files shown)


white238:
This was auto marked as merged and closed due to these commits existing in #782.  There was a commit that reverted the changes but if PR's have overlapping commits Github is "smart" enough to know this work got into develop.

I suggest you manually create a new branch with the changes relevant to this PR and re create it.

PR #638, opened 2022-02-24
Thermal sensitivities with `Functional`

jamiebramwell:
This is the initial draft of a working thermal module with sensitivities for arbitrary parameterized user-defined materials, flux boundary, and volumetric heat sources. To use this, a user must define the parameterized material/flux/load data type, and implement the appropriate paren operators. It still needs cleaning and documentation, but any and all feedback is appreciated!

### DIFF
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -408,14 +408,14 @@ public:
    * note: it accepts a vector of mfem::Vectors that must be of length `num_trial_spaces`. This interface
    * assumes no derivative information is needed.
    *
-   * @param trial_inputs an array of trial space dofs used to carry out the calculation.
+   * @param input_T an array of trial space dofs used to carry out the calculation.
    */
-  mfem::Vector& operator()(std::vector<std::reference_wrapper<const mfem::Vector>> trial_inputs)
+  mfem::Vector& operator()(std::vector<std::reference_wrapper<const mfem::Vector>> input_T)
   {
-    SLIC_ERROR_IF(trial_inputs.size() != num_trial_spaces,
+    SLIC_ERROR_IF(input_T.size() != num_trial_spaces,
                   "The input vector of trial spaces is not equal to the number of trial spaces defined in the "
                   "Functional constructor");
-    return (*this)(trial_inputs, Index<-1>{});
+    return (*this)(input_T, Index<-1>{});
   }
 
   /**
@@ -424,7 +424,7 @@ public:
    * note: it accepts a vector of mfem::Vectors that must be of length `num_trial_spaces`.
    *
    * @tparam wrt The index of the input trial vector to additional compute derivatives with respect to
-   * @param trial_inputs an array of trial space dofs used to carry out the calculation.
+   * @param input_T an array of trial space dofs used to carry out the calculation.
    */
   template <int wrt>
   typename operator_paren_return_index<wrt>::type operator()(
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -369,11 +369,6 @@ public:
 
     // scatter-add to compute global residuals
     P_test_->MultTranspose(output_L_, output_T);
-
-    output_T.HostReadWrite();
-    for (int i = 0; i < ess_tdof_list_.Size(); i++) {
-      output_T(ess_tdof_list_[i]) = input_T(ess_tdof_list_[i]);
-    }
   }
 
   /**
@@ -473,11 +468,6 @@ public:
     // scatter-add to compute global residuals
     P_test_->MultTranspose(output_L_, output_T_);
 
-    output_T_.HostReadWrite();
-    for (int i = 0; i < ess_tdof_list_.Size(); i++) {
-      output_T_(ess_tdof_list_[i]) = 0.0;
-    }
-
     if constexpr (wrt >= 0) {
       // if the user has indicated they'd like to evaluate and differentiate w.r.t.
       // a specific argument, then we return both the value and gradient w.r.t. that argument
@@ -495,21 +485,6 @@ public:
       // e.g. mfem::Vector value = my_functional(arg0, arg1);
       return output_T_;
     }
-  }
-
-  /**
-   * @brief Applies an essential boundary condition to the attributes specified by @a ess_attr
-   * @param[in] ess_attr The mesh attributes to apply the BC to
-   * @param[in] which which trial space the specified attributes apply to
-   *
-   * @note This gets more interesting when having more than one trial space
-   *
-   * TODO: remove this interface completely
-   */
-  void SetEssentialBC(const mfem::Array<int>& ess_attr, std::size_t which)
-  {
-    // TODO check that it actually makes sense to apply bcs to this trial space
-    trial_space_[which]->GetEssentialTrueDofs(ess_attr, ess_tdof_list_);
   }
 
 private:
@@ -627,8 +602,6 @@ private:
 
       std::unique_ptr<mfem::HypreParMatrix> K(mfem::RAP(R, A, P));
 
-      K->EliminateRowsCols(form_.ess_tdof_list_);
-
       delete A;
 
       return K;
@@ -704,9 +677,6 @@ private:
 
   /// @brief Manages DOFs for the trial space
   std::array<mfem::ParFiniteElementSpace*, num_trial_spaces> trial_space_;
-
-  /// @brief The set of true DOF indices to which an essential BC should be applied
-  mfem::Array<int> ess_tdof_list_;
 
   /**
    * @brief Operator that converts true (global) DOF values to local (current rank) DOF values
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -331,7 +331,7 @@ public:
    * arguments may be a dual_vector, to indicate that Functional::operator() should not only evaluate the
    * element calculations, but also differentiate them w.r.t. the specified dual_vector argument
    */
-  void ActionOfGradient(const mfem::Vector& input_T, mfem::Vector& output_T, size_t which) const
+  void ActionOfGradient(const mfem::Vector& input_T, mfem::Vector& output_T, std::size_t which) const
   {
     P_trial_[which]->Mult(input_T, input_L_[which]);
 
@@ -506,7 +506,7 @@ public:
    *
    * TODO: remove this interface completely
    */
-  void SetEssentialBC(const mfem::Array<int>& ess_attr, size_t which)
+  void SetEssentialBC(const mfem::Array<int>& ess_attr, std::size_t which)
   {
     // TODO check that it actually makes sense to apply bcs to this trial space
     trial_space_[which]->GetEssentialTrueDofs(ess_attr, ess_tdof_list_);
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -399,16 +399,16 @@ public:
     [[maybe_unused]] constexpr int                          wrt = index_of_differentiation<T...>();
     std::vector<std::reference_wrapper<const mfem::Vector>> input_T{args...};
 
-    return this->evaluate_index(input_T, Index<wrt>{});
+    return (*this)(input_T, Index<wrt>{});
   }
 
-  mfem::Vector& evaluate_index(std::vector<std::reference_wrapper<const mfem::Vector>> input_T)
+  mfem::Vector& operator()(std::vector<std::reference_wrapper<const mfem::Vector>> input_T)
   {
-    return this->evaluate_index(input_T, Index<-1>{});
+    return (*this)(input_T, Index<-1>{});
   }
 
   template <int wrt>
-  typename operator_paren_return_index<wrt>::type evaluate_index(
+  typename operator_paren_return_index<wrt>::type operator()(
       std::vector<std::reference_wrapper<const mfem::Vector>> input_T, Index<wrt>)
   {
     // get the values for each local processor
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -402,11 +402,30 @@ public:
     return (*this)(input_T, Index<wrt>{});
   }
 
-  mfem::Vector& operator()(std::vector<std::reference_wrapper<const mfem::Vector>> input_T)
+  /**
+   * @brief this function lets the user evaluate the serac::Functional with the given trial space values
+   *
+   * note: it accepts a vector of mfem::Vectors that must be of length `num_trial_spaces`. This interface
+   * assumes no derivative information is needed.
+   *
+   * @param trial_inputs an array of trial space dofs used to carry out the calculation.
+   */
+  mfem::Vector& operator()(std::vector<std::reference_wrapper<const mfem::Vector>> trial_inputs)
   {
-    return (*this)(input_T, Index<-1>{});
+    SLIC_ERROR_IF(trial_inputs.size() != num_trial_spaces,
+                  "The input vector of trial spaces is not equal to the number of trial spaces defined in the "
+                  "Functional constructor");
+    return (*this)(trial_inputs, Index<-1>{});
   }
 
+  /**
+   * @brief this function lets the user evaluate the serac::Functional with the given trial space values
+   *
+   * note: it accepts a vector of mfem::Vectors that must be of length `num_trial_spaces`.
+   *
+   * @tparam wrt The index of the input trial vector to additional compute derivatives with respect to
+   * @param trial_inputs an array of trial space dofs used to carry out the calculation.
+   */
   template <int wrt>
   typename operator_paren_return_index<wrt>::type operator()(
       std::vector<std::reference_wrapper<const mfem::Vector>> input_T, Index<wrt>)
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -74,6 +74,17 @@ constexpr int index_of_differentiation()
   return -1;
 }
 
+/**
+ * @brief Compile-time alias for index of differentiation
+ */
+template <int ind>
+struct Index {
+  /**
+   * @brief Returns the index
+   */
+  constexpr operator int() { return ind; }
+};
+
 /// @cond
 template <typename T, ExecutionSpace exec = serac::default_execution_space>
 class Functional;
@@ -134,6 +145,16 @@ class Functional<test(trials...), exec> {
         mfem::Vector&                                           // otherwise, we just return the value
         >::type;
   };
+
+  template <int indx>
+  struct operator_paren_return_index {
+    using type = typename std::conditional<
+        indx >= 0,                                              // if the derivative index is valid
+        serac::tuple<mfem::Vector&, Gradient&>,                 // then we return the value and the derivative
+        mfem::Vector&                                           // otherwise, we just return the value
+        >::type;
+  };
+
   // clang-format on
 
 public:
@@ -375,12 +396,24 @@ public:
     static_assert(sizeof...(T) == num_trial_spaces,
                   "Error: Functional::operator() must take exactly as many arguments as trial spaces");
 
-    [[maybe_unused]] constexpr int                                           wrt = index_of_differentiation<T...>();
-    std::array<std::reference_wrapper<const mfem::Vector>, num_trial_spaces> input_T{args...};
+    [[maybe_unused]] constexpr int                          wrt = index_of_differentiation<T...>();
+    std::vector<std::reference_wrapper<const mfem::Vector>> input_T{args...};
 
+    return this->evaluate_index(input_T, Index<wrt>{});
+  }
+
+  mfem::Vector& evaluate_index(std::vector<std::reference_wrapper<const mfem::Vector>> input_T)
+  {
+    return this->evaluate_index(input_T, Index<-1>{});
+  }
+
+  template <int wrt>
+  typename operator_paren_return_index<wrt>::type evaluate_index(
+      std::vector<std::reference_wrapper<const mfem::Vector>> input_T, Index<wrt>)
+  {
     // get the values for each local processor
     for (uint32_t i = 0; i < num_trial_spaces; i++) {
-      P_trial_[i]->Mult(input_T[i].get(), input_L_[i]);
+      P_trial_[i]->Mult(input_T[i], input_L_[i]);
     }
 
     output_L_ = 0.0;
@@ -426,7 +459,7 @@ public:
       output_T_(ess_tdof_list_[i]) = 0.0;
     }
 
-    if constexpr (num_differentiated_arguments == 1) {
+    if constexpr (wrt >= 0) {
       // if the user has indicated they'd like to evaluate and differentiate w.r.t.
       // a specific argument, then we return both the value and gradient w.r.t. that argument
       //
@@ -435,7 +468,7 @@ public:
       // e.g. auto [value, gradient_wrt_arg1] = my_functional(arg0, differentiate_wrt(arg1));
       return {output_T_, grad_[wrt]};
     }
-    if constexpr (num_differentiated_arguments == 0) {
+    if constexpr (wrt == -1) {
       // if the user passes only `mfem::Vector`s then we assume they only want the output value
       //
       // mfem::Vector arg0 = ...;
--- a/src/serac/numerics/functional/tensor.hpp
+++ b/src/serac/numerics/functional/tensor.hpp
@@ -392,9 +392,9 @@ SERAC_HOST_DEVICE zero& get(zero& x)
 
 /** @brief let `zero` be accessed like a tuple */
 template <int i>
-SERAC_HOST_DEVICE const zero& get(const zero& x)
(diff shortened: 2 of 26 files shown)


src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -384,6 +487,79 @@ class ThermalConductionFunctional : public BasePhysics {
...
+  virtual const serac::FiniteElementState& solveAdjoint(FiniteElementDual& adjoint_load,
+                                                        FiniteElementDual* dual_with_essential_boundary = nullptr)
+  {
+    adjoint_load.initializeTrueVec();
+
+    // note: The assignment operator must be called after the copy constructor because
+    // the copy constructor only sets the partitioning, it does not copy the actual vector
+    // values
+    mfem::HypreParVector adjoint_load_vector(adjoint_load.trueVec());
+    adjoint_load_vector = adjoint_load.trueVec();
```
jandrej: Is the second assignment necessary?

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -384,6 +487,79 @@ class ThermalConductionFunctional : public BasePhysics {
...
+
+    adjoint_temperature_.distributeSharedDofs();
+
+    // Reset the equation solver to use the full nonlinear residual operator
+    nonlin_solver_.SetOperator(residual_);
+
+    return adjoint_temperature_;
+  }
+
+  template <int parameter_field>
```
jandrej, line 523: What is `parameter_field`? From the code below I see it's an index to a field. My first guess would be "which field index should the sensitivity be wrt", but that's just a guess. A clarification would be good.

In fact, I'm a little confused by the whole function. I can't seem to get the mathematical relation to a sensitivity directly.

jandrej (changes requested):

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: Yes, I should clarify. This PR is very much a WIP and the documentation/naming schemes need a lot of work. It's probably hard to read at this point, but I wanted some early feedback on the workflow. I'll also be going over it in the meeting today, which will likely be much more understandable.

jamiebramwell:
This is now working. It still needs some cleaning and documentation, but I think the interface is set. Let me know if you have any thoughts!

src/serac/numerics/functional/functional.hpp
```diff
@@ -375,12 +396,43 @@ class Functional<test(trials...), exec> {
     static_assert(sizeof...(T) == num_trial_spaces,
                   "Error: Functional::operator() must take exactly as many arguments as trial spaces");
 
-    [[maybe_unused]] constexpr int                                           wrt = index_of_differentiation<T...>();
-    std::array<std::reference_wrapper<const mfem::Vector>, num_trial_spaces> input_T{args...};
+    [[maybe_unused]] constexpr int                          wrt = index_of_differentiation<T...>();
```
tepperly, line 394: I've been wanting `maybe_unused` for years.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -88,6 +88,89 @@ class LinearIsotropicConductor {
...
+/// Linear isotropic thermal conduction material model
+class ParameterizedLinearIsotropicConductor {
+public:
+  /**
+   * @brief Construct a new Parameterized Linear Isotropic Conductor object
+   *
+   * @param density Density of the material (mass/volume)
+   * @param specific_heat_capacity Specific heat capacity of the material (energy / (mass * temp))
+   * @param conductivity Thermal conductivity of the material (power / (length * temp))
+   */
```
tepperly: This is more of an example than practical code, right?

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -88,6 +88,89 @@ class LinearIsotropicConductor {
...
+   *
+   * @tparam T1 type of the temperature (e.g. tensor or dual type)
+   * @tparam T2 type of the temperature gradient (e.g. tensor or dual type)
+   * @tparam T3 type of the parameter field (e.g. tensor or dual type)
+   * @param temperature_gradient Gradient of the temperature (du_dx)
+   * @param parameter The user-defined parameter used to calculate the thermal conductivity
+   * @return The thermal flux of the material model
+   */
+  template <typename T1, typename T2, typename T3>
+  SERAC_HOST_DEVICE auto operator()(const T1& /* temperature */, const T2& temperature_gradient,
```
tepperly: Can `maybe_unused` be used with arguments like temperature?

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -88,6 +88,89 @@ class LinearIsotropicConductor {
...
+                                    const T3& parameter) const
```
tepperly: I imagine the intent is that `T1`, `T2` and `T3` are short for Type1, Type2, and Type3. I was briefly confused because I thought `T1` was chosen because it was the type of the temperature, and `T2` was the type of the temperature gradient.  Why isn't the third argument `P1` because it's the first parameter?

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -88,6 +88,89 @@ class LinearIsotropicConductor {
...
+
+  /**
+   * @brief The specific heat capacity (heat capacity per unit mass) of the material model
+   *
+   * @tparam dim The dimension of the problem
+   * @tparam T Type of the temperature variable
+   */
+  template <typename T1, typename T2, int dim>
+  SERAC_HOST_DEVICE double specificHeatCapacity(const tensor<double, dim>& /* x */, const T1& /* temperature */,
+                                                const T2&) const
```
tepperly: I think that `T2` is the parameter argument, but the comment on line 147 suggests otherwise. It's hard to know for sure since no argument name is given.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -212,19 +263,6 @@ struct ConstantSource {
   }
```
tepperly, line 182: For line 208/259, little `t` is time? The quantities in `ParameterizedLinearIsotropicConductor` are not time dependent, so it's method don't have little `t`, right?

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -119,23 +120,47 @@ class ThermalConductionFunctional : public BasePhysics {
...
     SLIC_ERROR_ROOT_IF(mesh_.Dimension() != dim,
                        axom::fmt::format("Compile time dimension and runtime mesh dimension mismatch"));
 
+    // Create a pack of the primal field and parameter finite element spaces
+    std::array<mfem::ParFiniteElementSpace*, sizeof...(parameter_space) + 1> trial_spaces;
+    trial_spaces[0] = &temperature_.space();
+
+    functional_call_args_.emplace_back(temperature_.trueVec());
+
+    for (long unsigned int i = 0; i < sizeof...(parameter_space); ++i) {
```
tepperly: `egrep -r "long  *unsigned  *int" serac` doesn't show any other uses of `long unsigned int` in the serac source. It seems that `std::size_t` is more common in cases like this. Of course, `unsigned char` is probably big enough to handle counting up to `sizeof...(parameter_space)`, but it's troublesome knowing that `sizeof...()` returns a `std::size_t`.

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -211,40 +236,80 @@ class ThermalConductionFunctional : public BasePhysics {
   template <typename MaterialType>
   void setMaterial(MaterialType material)
   {
+    /*
```
tepperly: Presumably this commented out code will either be reinstated or deleted before merging?

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: `T` is to `typename` template arguments as `i` is to loop indices: a not-actually-useful, but conventional choice for a name. See:
https://en.cppreference.com/w/cpp/container/vector
https://en.cppreference.com/w/cpp/container/array

Another common convention is to use `S`, `T`, `U`, `V` for a sequence of types .. in the same way nested loops often get `i`, `j`, `k`, `l`, ...

An alternate implementation could be
```cpp
  template <typename temperature_type, typename temperature_gradient_type, typename parameter_type>
  SERAC_HOST_DEVICE auto operator()(const temperature_type&  temperature,
                                    const temperature_gradient_type& temperature_gradient,
                                    const parameter_type& parameter) const { ... }
```
but that looks overly verbose to my eyes.

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: It would be syntactically okay to decorate function parameters with `[[maybe_unused]]`, but I don't think it makes sense in this particular case, since `temperature` is _always_ unused here-- there's no "maybe" about it!

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: Could you clarify which aspects seem impractical to you? Jamie and I have been discussing a different way to express these materials that cuts out a lot of the boilerplate present in this implementation, but if you have other constructive criticisms, we'd like to hear them!

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: there are a handful of helpful attributes:
https://en.cppreference.com/w/cpp/language/attributes

`[[nodiscard]]` is another one that I think is really useful.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -390,6 +390,13 @@ SERAC_HOST_DEVICE zero& get(zero& x)
   return x;
 }
 
+/** @brief let `zero` be accessed like a tuple */
+template <int i>
+SERAC_HOST_DEVICE const zero& get(const zero& x)
+{
+  return x;
+}
```
samuelpmishLLNL, line 398:
```suggestion
template <int i>
SERAC_HOST_DEVICE zero get(zero)
{
  return zero{};
}
```

I don't expect this change to actually make an appreciable difference, but it seems like original implementation is saying "you have to return the same `zero` that was passed in", but it doesn't actually matter where the return value comes from. 

They generate slightly different assembly:
https://godbolt.org/z/3GrvobKo1
I think `xor eax, eax` is slightly preferred, but I'm splitting hairs.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -88,6 +88,89 @@ class LinearIsotropicConductor {
...
+  SERAC_HOST_DEVICE auto density(const tensor<double, dim>& /* x */, const T1& parameter) const
+  {
+    return density_ + 0.0 * parameter;
+  }
+
+  /**
+   * @brief The specific heat capacity (heat capacity per unit mass) of the material model
+   *
+   * @tparam dim The dimension of the problem
+   * @tparam T Type of the temperature variable
```
samuelpmishLLNL:
```suggestion
   * @tparam T1 Type of the temperature variable
   * @tparam T2 Type of the temperature gradient variable
```
note: I think the word "parameter" is more correct than "variable" for the term describing the inputs in a function declaration, but we're using "parameter" to mean something else in this context.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -88,6 +88,89 @@ class LinearIsotropicConductor {
...
+  /**
+   * @brief The density (mass per volume) of the material model
+   *
+   * @tparam dim The dimension of the problem
+   * @return The density
+   */
+  template <int dim, typename T1>
+  SERAC_HOST_DEVICE auto density(const tensor<double, dim>& /* x */, const T1& parameter) const
+  {
+    return density_ + 0.0 * parameter;
```
samuelpmishLLNL: do we still need the `0.0 * parameter`?

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: I think `t` is time here. The material models probably don't get `t` because they don't usually depend explicitly on the absolute time (there are time-dependent materials, but the only ones I know of depend on a time increment, rather than the absolute time). Source terms, on the other hand, frequently are functions of the absolute time.

src/serac/numerics/functional/functional.hpp
```diff
@@ -375,12 +396,43 @@ class Functional<test(trials...), exec> {
...
 
+  /**
+   * @brief this function lets the user evaluate the serac::Functional with the given trial space values
+   *
+   * note: it accepts a vector of mfem::Vectors that must be of length `num_trial_spaces`. This interface
+   * assumes no derivative information is needed.
+   *
+   * @param input_T an array of trial space dofs used to carry out the calculation.
+   */
+  mfem::Vector& operator()(std::vector<std::reference_wrapper<const mfem::Vector>> input_T)
```
samuelpmishLLNL, line 408: Can we use `std::array` instead of `std::vector` here? The number of entries is known at compile time.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -88,6 +88,89 @@ class LinearIsotropicConductor {
...
+  {
+    return specific_heat_capacity_;
+  }
+
+  /**
+   * @brief The number of parameters associated with this material model
+   *
+   * @return The number of material model parameters
+   */
+  static constexpr int numParameters() { return 1; }
```
samuelpmishLLNL:
```suggestion
  static constexpr size_t numParameters() { return 1; }
```
it looks like we're a little inconsistent on whether or not to use `size_t` or `std::size_t` in serac

src/serac/numerics/functional/functional.hpp
jamiebramwell: I tried to make an array at some point and got tangled in an issue of default values in a constructor. I've cleaned things up a bit since then, so I'll give it another shot.

jamiebramwell:
After talking with @samuelpmishLLNL about cleaning the load functions, I think the approach described [here](https://godbolt.org/z/srvxxdWad) under `ParameterizedLinearIsotropicConductor2`. It makes the under interface clearer, reduces boilerplate, and reduces the need for obscure SFINAE techniques. Any thoughts? If people are OK with this, changes incoming!

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Yes - this material isn't particularly useful. I'll add some materials where the conductivity/loads are directly specified by a parameter.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Yeah, I like doing this because it shows explicitly that it is unused.

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: Yeah, I'll do a pass to clean this up.

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: Yes - I was looking at how to do these SFINAE checks with parameter lists. No good solution exists, so I was considering changing the interface to look something more like [this](https://godbolt.org/z/6drYsh1cK).

src/serac/numerics/functional/tensor.hpp
jamiebramwell: Seems reasonable to me!

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: No - this was testing code paths. I'll remove it in the next pass.

src/serac/physics/materials/thermal_functional_material.hpp
tepperly: > Could you clarify which aspects seem impractical to you? Jamie and I have been discussing a different way to express these materials that cuts out a lot of the boilerplate present in this implementation, but if you have other constructive criticisms, we'd like to hear them!

In a practical example, I would expect:

- semantic naming of the parameter to make it's meaning clear to the engineers who would use it
- presumably the units or that it's dimensionless should be given for the parameter
- the use of the parameter in the calculation of function values should make physical sense

src/serac/physics/materials/thermal_functional_material.hpp
tepperly: Would `[[maybe_unused]]` work with Doxygen?

src/serac/physics/materials/thermal_functional_material.hpp
tepperly: For me, there is also a `T` is to temperature as `l` is to length or `m` is to mass going on here. It leads to a little cognitive dissonance. `T1` -> temperature, `T2` -> temperature gradient, `T3` -> parameter one :scream_cat:

samuelpmishLLNL: Unfortunately, the computer-science naming convention is conflicting with the physics naming convention here, so whichever one we choose will make someone angry! I think serac is sort of leaning more toward the computer science conventions, because they tend to be more self documenting (e.g. calling something `temperature` rather than `u` or `T`, "conductivity" rather than `k`, etc.), which I think is beneficial when you consider that a lot of the people that work on serac are not subject matter experts on the actual physics.

tepperly: I won't be angry.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -390,6 +390,13 @@ SERAC_HOST_DEVICE zero& get(zero& x)
   return x;
 }
 
+/** @brief let `zero` be accessed like a tuple */
+template <int i>
+SERAC_HOST_DEVICE zero get(zero)
+{
+  return zero{};
```
jamiebramwell: For some reason, this produces obscure `get` resolution errors.

samuelpmishLLNL: interesting, I'd like to hear more info about this at some point

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -39,52 +49,50 @@ class LinearIsotropicConductor {
...
-   */
-  template <typename T1, typename T2>
-  SERAC_HOST_DEVICE T2 operator()(const T1& /* temperature */, const T2& temperature_gradient) const
+  template <typename T1, typename T2, typename T3>
+  SERAC_HOST_DEVICE auto operator()(const T1&, const T2&, const T3& temperature_gradient) const
   {
-    return -1.0 * conductivity_ * temperature_gradient;
+    return MaterialResponse{.density                = density_,
+                            .specific_heat_capacity = specific_heat_capacity_,
+                            .heat_flux              = -conductivity_ * temperature_gradient};
```
tepperly: I don't have a strong sense about the relative benefits/costs of switching to the MaterialResponse versus individual calls for each property. On the plus side, I see:
* provides consistency across different modeling physics. Heat conduction has three parameters. Structural modeling might have a different number. There is always one consistent operator() return a struct with the right number of material quantities.
* all parameters are treated equally one isn't treated with operator() and the others with individual methods
* potentially better performance because more computation per function call (guessing a very small effect if any)
Potential costs:
* another type to remember (albeit a single one)
* aggressive adoption of C++ 20 syntactic sugar for struct initialization (I think LiDO is at C++ 17)
* potential overhead of struct construction/return
* what are the missing arguments T1 and T2?

tepperly: I think the benefits outweigh the potential costs.

src/serac/physics/materials/thermal_functional_material.hpp
```diff
@@ -39,52 +49,50 @@ class LinearIsotropicConductor {
...
-   * @brief The specific heat capacity (heat capacity per unit mass) of the material model
-   *
-   * @tparam dim The dimension of the problem
-   * @tparam T Type of the temperature variable
-   */
-  template <typename T, int dim>
-  SERAC_HOST_DEVICE double specificHeatCapacity(const tensor<double, dim>& /* x */, const T& /* temperature */) const
+  template <typename T1, typename T2, typename T3, typename T4, typename T5>
+  SERAC_HOST_DEVICE auto operator()(const T1&, const T2&, const T3& temperature_gradient, const T4& parameter_1,
+                                    const T5& parameter_2) const
```
tepperly: Is T1 time and T2 temperature?

tepperly: Or I bet T1 is the location.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: So we are technically still on C++17, but our compilers (gcc and clang) both support designated initializers despite it not being required until C++20.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Yep, `T1` is position. I'll go and add the placeholder comments to denote this.

src/serac/physics/materials/thermal_functional_material.hpp
samuelpmishLLNL: >     + potentially better performance because more computation per function call (guessing a very small effect if any)

I mocked up a little example showing the assembly generated for doing a calculation with a struct that requires 3 member function calls to get separate quantities (like the "old" way), and an equivalent calculation that uses a struct that returns a "tuple" of those 3 values in one call (like the "new" way):

https://godbolt.org/z/xasvhcs4o

It appears that, when compiling with optimization, the generated assembly is practically identical.

>     - potential overhead of struct construction/return

C++17 actually introduced some language-level changes that make this sort of thing less expensive (or "free", in many cases). Previously, copy elision was a compiler optimization, so it was left up to the compiler to decide on a case-by-case basis when things would be elided or not. This meant that a lot of developers just assumed the worst case scenario (that return statements incur additional overhead) and avoided returning objects altogether in the name of "performance". But it seems like that's really not the case any more, there are a number of ways to get the clarity of a return value (instead of mixing inputs and outputs in the argument list) without the performance drawbacks!

samuelpmishLLNL:
this is an example of the new operator<< overloads:

Screenshot from 2022-03-04 10-49-53

Now, arbitrarily nested and composed combinations of {`tensor`, `dual`, `tuple`, `zero`} can be written to ostreams like `std::cout`, `std::ofstream`. These might belong in a separate PR, but they were relevant for debugging the earlier sign error.

jamiebramwell:
Thank you for the help @samuelpmishLLNL !

src/serac/physics/thermal_conduction_functional.hpp
jamiebramwell: I'll check - but I believe yes. The copy constructor only copied the partitioning and not the data (or at least did at one point).

jamiebramwell: No - this is no longer needed. I will remove this.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: A good question! I'll look into this and we can discuss it in the weekly meeting.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: Yeah, since templates are very CS-y, I'm leaning toward the C++ conventions here.

tepperly (approved):
I looked through this again, and I don't have any additional comments. I imagine I'll have a better sense about it once I try wrapping it in LiDO.

src/serac/physics/materials/thermal_functional_material.hpp
jamiebramwell: I think variable/parameter is redundant here anyway, so I will delete it.

jamiebramwell:
This how now been cleaned and verified with a finite difference test. Any final comments and review notes would be appreciated!

jamiebramwell:
Any other comments on this? @jandrej , it looks like you have a change requested pending, but I think I've covered everything.

src/serac/physics/materials/parameterized_thermal_functional_material.hpp
```diff
@@ -0,0 +1,146 @@
...
+   * offset + parameter)
+   * @return The density, specific heat capacity, and heat flux of the material.
+   */
+  template <typename T1, typename T2, typename T3, typename T4>
+  SERAC_HOST_DEVICE auto operator()(const T1& /* x */, const T2& /* temperature */, const T3& temperature_gradient,
+                                    const T4& parameter) const
+  {
+    using FluxType = decltype((conductivity_offset_ + parameter) * temperature_gradient);
+
+    return MaterialResponse<double, double, FluxType>{
```
samuelpmishLLNL, line 62: Did this explicit type specification end up being required? I remember we tried this when debugging that earlier problem

jandrej (approved):

samuelpmishLLNL (approved):

PR #641, opened 2022-03-07
Add missing <optional> header

bendudson:
Required when compiling with GCC 11.2.0

Fixes error "‘optional’ in namespace ‘std’ does not name a template type; did you forget to ‘#include <optional>’?"


white238:
Thanks @bendudson !  I have pulled it into a local branch so that it can run our CI checks.  As soon as we this happens and we have two approvals, we will handle merging it.

jamiebramwell:
Thanks @bendudson !

PR #656, opened 2022-03-23
Add check for existence of mesh tag in data store operations

jamiebramwell:
A user reported that non-default tags were broken in the `Solid` module. This is due to confusion between the mesh and physics tags. This PR clarifies the documentation and adds checks to ensure the given mesh tag exists when interacting with the data store.


chapman39 (approved):

white238 (approved):

src/serac/physics/solid.hpp
```diff
@@ -200,6 +200,9 @@ class Solid : public BasePhysics {
    * @param[in] keep_deformation Flag to keep the deformation in the underlying mesh post-destruction
    * @param[in] name An optional name for the physics module instance
    * @param[in] pmesh The mesh to conduct the simulation on, if different than the default mesh
+   *
+   * @note @p name is the physics module name and not the mesh tag to be used. If a non-default mesh is desired, the
+   * mesh pointer must be provided. Otherwise, it attempts to find the "default" mesh in the data store.
    */
   Solid(int order, const SolverOptions& options, GeometricNonlinearities geom_nonlin = GeometricNonlinearities::On,
         FinalMeshOption keep_deformation = FinalMeshOption::Deformed, const std::string& name = "",
```
goxberry, line 208: re: the confusion, it might also be worth changing the `name` argument to `instance_name` or similar.

goxberry: Semantically, is there a distinction between the terms "name" and "tag" (e.g., `mesh_tag`) used in the API?

goxberry (commented):
LGTM; feel free to disregard the argument rename suggestion if you think it doesn't make sense.

Thanks for taking a look at this issue!

Issue #653, opened 2022-03-22
address inconsistencies in options structs for physics modules, and their defaults


samuelpmishLLNL:
It looks like the thermal module (both the original and functional implementation) have equivalent `SolverOptions` types, but they are incompatible, as a result of being defined inside of their respective classes.

the `SolidFunctional` object has the same problem, and it also does not provide any convenient way to get "default" options.

PR #658, opened 2022-03-25
Add parameterizations and sensitivities to `SolidFunctional`

jamiebramwell:
This adds arbitrary parameterizations of load and materials to the `SolidFunctional` physics model. Included are two parameterized models: linear elastic and neo-Hookean responses with user-parameterized bulk and shear modulus. The interface could easily be extended to include both new material models as well as new parameterizations ready for sensitivity analysis.

### DIFF
--- a/src/serac/physics/materials/functional_material_utils.hpp
+++ b/src/serac/physics/materials/functional_material_utils.hpp
@@ -34,17 +34,18 @@ struct TriviallyParameterizedMaterial {
   /**
    * @brief Material response wrapper for an unparameterized material in a parameterized context
    *
-   * @tparam T1 Spatial position type
-   * @tparam T2 Temperature type
-   * @tparam T3 Temperature gradient type
-   * @tparam S Unused parameter pack type
+   * @tparam PositionType Spatial position type
+   * @tparam StateType State type
+   * @tparam StateGradType State gradient type
+   * @tparam ParamTypes Unused parameter pack type
    * @param x Spatial position
-   * @param u Temperature
-   * @param du_dx Temperature gradient
+   * @param u State
+   * @param du_dx State gradient
    * @return Material response of the unparameterized material
    */
-  template <typename T1, typename T2, typename T3, typename... S>
-  SERAC_HOST_DEVICE auto operator()(const T1& x, const T2& u, const T3& du_dx, S...) const
+  template <typename PositionType, typename StateType, typename StateGradType, typename... ParamTypes>
+  SERAC_HOST_DEVICE auto operator()(const PositionType& x, const StateType& u, const StateGradType& du_dx,
+                                    ParamTypes...) const
   {
     return mat(x, u, du_dx);
   }
@@ -56,22 +57,22 @@ struct TriviallyParameterizedMaterial {
 /**
  * @brief Template deduction guide for the trivially parameterized material
  *
- * @tparam T The unparameterized material type
+ * @tparam MaterialType The unparameterized material type
  */
-template <typename T>
-TriviallyParameterizedMaterial(T) -> TriviallyParameterizedMaterial<T>;
+template <typename MaterialType>
+TriviallyParameterizedMaterial(MaterialType) -> TriviallyParameterizedMaterial<MaterialType>;
 
 /**
  * @brief Convert an unparameterized material to one which accepts parameter values in the paren operator
  *
- * @tparam T The unparameterized material type
+ * @tparam MaterialType The unparameterized material type
  * @param material The unparameterized material
  * @return The parameterized material
  */
-template <typename T>
-auto parameterizeMaterial(T& material)
+template <typename MaterialType>
+auto parameterizeMaterial(MaterialType& material)
 {
-  if constexpr (is_parameterized<T>::value) {
+  if constexpr (is_parameterized<MaterialType>::value) {
     return material;
   } else {
     return TriviallyParameterizedMaterial{material};
@@ -88,18 +89,19 @@ struct TriviallyParameterizedSource {
   /**
    * @brief Wrapper for an unparameterized source in a parameterized context
    *
-   * @tparam T1 Spatial position type
-   * @tparam T2 Temperature type
-   * @tparam T3 Temperature gradient type
-   * @tparam S Unused parameter pack type
+   * @tparam PositionType Spatial position type
+   * @tparam StateType State type
+   * @tparam StateGradType State gradient type
+   * @tparam ParamTypes Unused parameter pack type
    * @param x Spatial position
    * @param t Time
-   * @param u Temperature
-   * @param du_dx Temperature gradient
+   * @param u State
+   * @param du_dx State gradient
    * @return Volumetric source for the unparameterized source
    */
-  template <typename T1, typename T2, typename T3, typename... S>
-  SERAC_HOST_DEVICE auto operator()(const T1& x, double t, const T2& u, const T3& du_dx, S...) const
+  template <typename PositionType, typename StateType, typename StateGradType, typename... ParamTypes>
+  SERAC_HOST_DEVICE auto operator()(const PositionType& x, double t, const StateType& u, const StateGradType& du_dx,
+                                    ParamTypes...) const
   {
     return source(x, t, u, du_dx);
   }
@@ -111,22 +113,22 @@ struct TriviallyParameterizedSource {
 /**
  * @brief Template deduction guide for the trivially parameterized source
  *
- * @tparam T The unparameterized source type
+ * @tparam SourceType The unparameterized source type
  */
-template <typename T>
-TriviallyParameterizedSource(T) -> TriviallyParameterizedSource<T>;
+template <typename SourceType>
+TriviallyParameterizedSource(SourceType) -> TriviallyParameterizedSource<SourceType>;
 
 /**
  * @brief Convert an unparameterized source to one which accepts parameter values in the paren operator
  *
- * @tparam T The unparameterized source type
+ * @tparam SourceType The unparameterized source type
  * @param source The unparameterized source
  * @return The parameterized source
  */
-template <typename T>
-auto parameterizeSource(T& source)
+template <typename SourceType>
+auto parameterizeSource(SourceType& source)
 {
-  if constexpr (is_parameterized<T>::value) {
+  if constexpr (is_parameterized<SourceType>::value) {
     return source;
   } else {
     return TriviallyParameterizedSource{source};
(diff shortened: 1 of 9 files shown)


src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -278,8 +277,99 @@ void functional_solid_test_boundary(double expected_disp_norm, TestType test_mod
...
+      StateManager::newState(FiniteElementState::Options{.order = 1, .name = "parameterized_shear"}));
+
+  user_defined_shear_modulus = 1.0;
+
+  FiniteElementState user_defined_bulk_modulus(
+      StateManager::newState(FiniteElementState::Options{.order = 1, .name = "parameterized_bulk"}));
+
+  user_defined_bulk_modulus = 1.0;
+
+  // Construct a functional-based thermal conduction solver
```
barreracruz: structural/nonlinear mechanics solver?

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -278,8 +277,99 @@ void functional_solid_test_boundary(double expected_disp_norm, TestType test_mod
...
+  SolidFunctional<p, dim, H1<1>, H1<1>> solid_solver(default_static, GeometricNonlinearities::On,
+                                                     FinalMeshOption::Reference, "solid_functional",
+                                                     {user_defined_bulk_modulus, user_defined_shear_modulus});
+
+  solid_util::ParameterizedNeoHookeanSolid<dim> mat(1.0, 0.0, 0.0);
+  solid_solver.setMaterial(mat);
+
+  // Define the function for the initial temperature and boundary condition
```
barreracruz: initial displacement

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -278,8 +277,99 @@ void functional_solid_test_boundary(double expected_disp_norm, TestType test_mod
...
+  auto bc = [](const mfem::Vector&, mfem::Vector& bc_vec) -> void { bc_vec = 0.0; };
+
+  // Set the initial temperature and boundary condition
```
barreracruz: Ditto

src/serac/physics/tests/serac_solid_functional.cpp
jamiebramwell: Arg! My copy/paste leavings! I thought I got all of them but I'll do another pass.

src/serac/physics/solid_functional.hpp
```diff
@@ -24,6 +24,38 @@
...
+  NonlinearSolverOptions H_nonlin_options;
+
+  /**
+   * @brief The optional ODE solver parameters
+   * @note If this is not defined, a quasi-static solve is performed
+   */
+  std::optional<TimesteppingOptions> dyn_options = std::nullopt;
+};
+}  // namespace solid_util
+
```
samuelpmishLLNL, line 58: these appear to be effectively identical types to https://github.com/LLNL/serac/blob/develop/src/serac/physics/thermal_conduction_functional.hpp#L29-L57, except that they can't be used interchangeably because they're declared in separate namespaces.

Do these options have anything to do with the specific physics? If not, I think there should be a single definition, outside the physics module namespaces, and be shared by all of the different modules.

src/serac/physics/materials/functional_material_utils.hpp
```diff
@@ -198,50 +188,57 @@ auto parameterizeFlux(T& flux)
...
+ * @brief Wrapper to treat an unparameterized source as a parameterized one
+ *
+ * @tparam UnparameterizedSourceType The unparameterized source
+ */
+template <typename UnparameterizedPressureType>
+struct TriviallyParameterizedPressure {
+  /**
+   * @brief Wrapper for an unparameterized source in a parameterized context
+   *
+   * @tparam T1 Spatial position type
```
goxberry: If `T2` and `T3` aren't arguments to this template, is it necessary to document them?

src/serac/physics/materials/parameterized_solid_functional_material.hpp
```diff
@@ -0,0 +1,168 @@
...
+                                    double bulk_modulus_offset = 1.0)
+      : density_(density), bulk_modulus_offset_(bulk_modulus_offset), shear_modulus_offset_(shear_modulus_offset)
+  {
+    SLIC_ERROR_ROOT_IF(density_ < 0.0, "Density must be positive in the linear isotropic elasticity material model.");
+  }
+
+  /**
+   * @brief Material response call for a linear isotropic solid
+   *
+   * @tparam T1 Spatial position type
```
goxberry: As someone potentially using this API, I'd find it easier to understand and use if the `T1` through `T5` parameters were replaced with, e.g., `SpatialPositionType`, particularly if the intent is for users to emulate this pattern.

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -52,30 +76,24 @@ class LinearIsotropicSolid {
...
-   * @brief Function defining the Kirchoff stress (constitutive response)
+   * @brief Material response call for a linear isotropic solid
    *
-   * @tparam T type of the Kirchoff stress (i.e. second order tensor)
+   * @tparam T1 Spatial position type
+   * @tparam T2 Displacement type
+   * @tparam T3 Displacement gradient type
    * @param du_dX displacement gradient with respect to the reference configuration (du_dX)
-   * @return The Kirchoff stress (det(deformation gradient) * Cauchy stress) for the constitutive model
+   * @return The calculated material response (density, kirchoff stress) for the material
```
goxberry: kirchoff -> Kirchhoff

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -119,14 +137,16 @@ class NeoHookeanSolid {
...
-   * @brief Function defining the Kirchoff stress (constitutive response)
+   * @brief Material response call for a neo-Hookean solid
    *
-   * @tparam T type of the Kirchoff stress (i.e. second order tensor)
+   * @tparam T1 Spatial position type
+   * @tparam T2 Displacement type
+   * @tparam T3 Displacement gradient type
    * @param du_dX displacement gradient with respect to the reference configuration (du_dX)
-   * @return The Kirchoff stress (det(deformation gradient) * Cauchy stress) for the constitutive model
+   * @return The calculated material response (density, kirchoff stress) for the material
```
goxberry: See https://github.com/LLNL/serac/pull/658/files#r841976690.

goxberry:
AFAICT, LGTM.

goxberry (approved):

src/serac/physics/solid_functional.hpp
jamiebramwell: This will be resolved in a follow on PR resolving https://github.com/LLNL/serac/issues/653 .

src/serac/physics/materials/functional_material_utils.hpp
jamiebramwell: Good point. I'm surprised our doxygen bot didn't notice this.

src/serac/physics/materials/parameterized_solid_functional_material.hpp
jamiebramwell: I'm OK with either. Do you have an opinion @samuelpmishLLNL @btalamini @white238?

jamiebramwell: I just think this convention should be consistent. I followed the `T` style after tensor/functional, but the types here definitely have a physical meaning.

jamiebramwell: I remember @tepperly having a similar question, so I think two potential users bring this up means we should change it.

samuelpmishLLNL: I think it does make sense to use more descriptive template parameter names here, since these quantities aren't so abstract. For some of the lower-level types (e.g. `tensor`, `tuple`, etc) those template parameters can represent practically anything (so it's harder to attach a descriptive name), whereas these ones are much more specific.

Also, we only need to use a template parameter for arguments that might be promoted to AD-enabled versions, so I think the spatial position could just be `tensor< double, dim > x`, rather than a template parameter `SpatialPositionType x`.

The naming of arguments here is also a little inconsistent, we're using `x`, `du_dX` (mathematical notation) for some arguments, and then `displacement`, `bulk_parameter`, `shear_parameter` (descriptive words) for the others. I'm not sure which is better, maybe a compromise like:

```cpp
  template <typename displacement_type, 
            typename displacement_gradient_type, 
            typename bulk_modulus_type, 
            typename shear_modulus_type >
   SERAC_HOST_DEVICE auto operator()(const tensor< double, dim > & /* x */, 
                                     const displacement_type & /* u */, 
                                     const displacement_gradient_type & du_dX, 
                                     const bulk_modulus_type & K, 
                                     const shear_modulus_type & G) const { ... }
```

src/serac/physics/materials/parameterized_solid_functional_material.hpp
```diff
@@ -0,0 +1,168 @@
...
+    auto shear_modulus = shear_parameter + shear_modulus_offset_;
+
+    auto I      = Identity<dim>();
+    auto lambda = bulk_modulus - (2.0 / dim) * shear_modulus;
+    auto strain = 0.5 * (du_dX + transpose(du_dX));
+    auto stress = lambda * tr(strain) * I + 2.0 * shear_modulus * strain;
+
+    using StressType = decltype(stress);
+
+    return MaterialResponse<double, StressType>{.density = density_, .stress = stress};
```
samuelpmishLLNL: The `decltype` here looks unpleasant, does
```cpp
return MaterialResponse{.density = density_, .stress = stress};
```
or
```cpp
return MaterialResponse{density_, stress};
```
work just the same? 

If not, I think we might be able to change the implementation of `MaterialResponse` type-- I really don't want the user to be required to manually deduce certain types in their q-functions.

jamiebramwell: Let me check - I had to do a lot of compiler error chasing to get this to work. I'll see if it was something else I needed.

jamiebramwell: If I recall correctly, clang couldn't resolve the class template argument deduction properly. I'll test this though.

jamiebramwell: Well this seems to be working now, so I'll remove the explicit template types.

src/serac/physics/materials/parameterized_solid_functional_material.hpp
goxberry: > For some of the lower-level types (e.g. tensor, tuple, etc) those template parameters can represent practically anything (so it's harder to attach a descriptive name)

I agree. I'm not terribly concerned about terse template arguments for low-level, algebraic data types; I'm mostly concerned with arguments that have physical significance.

re: `x`, `u` versus `position`, `displacement`, as someone without a solid mechanics background, I find the latter more helpful due to collisions in notation namespaces for different disciplines (e.g., `mu` could be a shear modulus, chemical potential, coefficient of static friction, or equality-constraint duality multiplier, depending on context). The longer names are also easier to `grep` when making changes, easier to Google if I'm unfamiliar with the physics, and less likely to become stale than explanatory comments.

src/serac/physics/materials/parameterized_solid_functional_material.hpp
jamiebramwell: I spoke too soon! The linear version works, but the neo-Hookean model doesn't. I will investigate.

samuelpmishLLNL: if something like
```cpp
return MaterialResponse{.density = density_, .stress = stress};
```
doesn't work, I don't think it should hold up this PR. We can just put a `// TODO` on it and make an issue, to revisit later.

jamiebramwell:
@samuelpmishLLNL @btalamini any other issues in this PR?

jamiebramwell:
@samuelpmishLLNL @sethwatts @btalamini any other comments on this PR?

samuelpmishLLNL (approved):

jamiebramwell:
ARG! GCC 11 got me! 😆 I'll investigate.

PR #663, opened 2022-03-31
Add shape sensitivities to the `SolidFunctional` module

jamiebramwell:
This adds shape velocity fields to the solid mechanics module. Specifically, we consider a new configuration in the solid mechanics pipeline:

image

where `X` is the original reference configuration, `p` is the shape displacement field describing the movement of mesh nodes under shape optimization, and `X'` is the perturbed reference configuration. Additionally, we have:

image

where `x'` is the current configuration and `u` is the displacement field. Note that the displacement gradient that is passed to the material models is now:

image

As a verification problem, I performed a body load simulation using a fixed and non-affine transformed mesh. However, the boundary nodes remain fixed. The displacement field was equivalent.

MicrosoftTeams-image (3)

MicrosoftTeams-image (2)


jamiebramwell:
@barreracruz FYI

jamiebramwell:
Wait, I just realized an error in my calculation. Fix incoming tomorrow.

src/serac/physics/solid_functional.hpp
```diff
@@ -54,6 +55,28 @@ struct SolverOptions {
    */
   std::optional<TimesteppingOptions> dyn_options = std::nullopt;
 };
+
+/**
+ * @brief Adjust the displacement and displacement gradient with a shape displacement field
+ *
+ * @note If shape_index = -1, the original displacement is returned.
```
goxberry: Could be worth adding a named `constexpr int` parameter at namespace scope, e.g.,
```c++
constexpr int serac::solid_util::undeformed_index = -1; // I don't have strong feelings about this name
```

goxberry (approved):
AFAICT, LGTM.

jamiebramwell:
Thanks for the fix @samuelpmishLLNL ! Adding some tests now...

jamiebramwell:
@btalamini @samuelpmishLLNL @sethwatts @barreracruz Any other thoughts? Now that I have more reasonable examples and documentation, I was going to give another stab at explaining this PR during the meeting tomorrow.

src/serac/physics/solid_functional.hpp
```diff
@@ -299,14 +408,28 @@ class SolidFunctional : public BasePhysics {
...
-          auto source           = response.density * u * (1.0 + geom_factor * (det(deformation_grad) - 1.0));
+          // This deformation gradient is the volumetric transform to get us back to the original
+          // reference configuration dx/dX = I + du/dX + dp/dX. If we are not including geometric
+          // nonlinearities, we ignore the du/dX factor
+          auto deformation_grad = geom_factor * du_dX + dshape_dX + I_;
+
+          auto source = response.density * u * det(deformation_grad);
 
           return serac::tuple{source, flux};
         },
```
samuelpmishLLNL: it looks like there are some notation inconsistencies here:
1. "shape" (in the code) vs. "p" (in the comments/LaTeX)
2. the comments use `x'` for the current configuration, but `x` in the code appears to be the reference coordinates

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -364,12 +364,120 @@ void functional_parameterized_solid_test(double expected_disp_norm)
...
+
+  // Check the final displacement norm
+  EXPECT_NEAR(expected_disp_norm, norm(solid_solver.displacement()), 1.0e-5);
 }
 
 TEST(solid_functional, 2D_linear_static) { functional_solid_test_static<1, 2>(1.511052595); }
 TEST(solid_functional, 2D_quad_static) { functional_solid_test_static<2, 2>(2.18604855); }
 TEST(solid_functional, 2D_quad_parameterized_static) { functional_parameterized_solid_test<2, 2>(2.18604855); }
+TEST(solid_functional, 2D_quad_shape_static) { functional_shape_solid_test<2, 2>(0.00011278491, false); }
+TEST(solid_functional, 2D_quad_shape_static_move_nodes) { functional_shape_solid_test<2, 2>(0.000112885, true); }
```
samuelpmishLLNL: Devil's advocate: do we have reason to believe these answers are correct?

jamiebramwell: The moved node calculation is correct when compared to the original one. However, this is not an analytical solution test (which we need).

src/serac/physics/tests/serac_solid_functional_finite_diff.cpp
```diff
@@ -169,6 +169,146 @@ TEST(solid_functional_finite_diff, finite_difference)
...
+    }
+
+    // Compute numerical value of sensitivity of qoi with respect to bulk modulus
+    // by taking the inner product between adjoint load and displacement sensitivity
+    double dqoi_dshape = adjoint_load_form(ddisp_dshape);
+
+    // See if these are similar
+    SLIC_INFO(axom::fmt::format("dqoi_dshape: {}", dqoi_dshape));
+    SLIC_INFO(axom::fmt::format("sensitivity: {}", sensitivity.trueVec()(i)));
+    EXPECT_NEAR((sensitivity.trueVec()(i) - dqoi_dshape) / std::max(dqoi_dshape, 1.0e-2), 0.0, 5.0e-2);
```
samuelpmishLLNL: this tolerance seems really loose, can we be confident we're not making mistakes in the implementation?

jamiebramwell: I think so - this QoI is extremely sensitive to nodal coordinate changes. It's probably worthwhile to come up with a more reasonable QoI for test.

samuelpmishLLNL: what is the QoI in this problem?

jamiebramwell: I use a pseudo-load vector of 1 for all entries. This is equivalent to the L1 norm of the displacement vector (hence the large sensitivity).

src/serac/physics/tests/serac_solid_functional.cpp
samuelpmishLLNL: I'm not sure why these answers are only in agreement to a few sig figs, it seems like the shape displacement is a biquadratic function (which should be exactly approximated by the shape displacement function space). I wouldn't expect floating point arithmetic reordering alone to result in such a noticeable difference.

Also: the argument name "move_nodes" for these tests confused me initially, it looks like they both have nonzero shape displacements and yet neither is actually moving the nodes of the mesh (?). 

It might be a good test to compare answers from: 
1. a mesh with a perturbation applied directly to the mesh nodes (evaluated with zero shape displacement)
2. a mesh with no modification to the nodal positions, but evaluated with a shape displacement consistent with the nodal perturbation in 1

jamiebramwell:
I found the last subtle bug (thanks @samuelpmishLLNL !) so this branch should now be ready for a final review.

jamiebramwell:
I believe this refactor is now done. It has an "always on" version of shape sensitivities as it was getting very messy to change function signatures. I propose we add this as an initial step, but I am open to suggestions. Any thoughts @samuelpmishLLNL @btalamini @sethwatts @kswartz92 ?

src/serac/physics/state/finite_element_state.hpp
```diff
@@ -47,12 +47,84 @@ inline bool is_vector_valued(const GeneralCoefficient& coef)
...
+   * @brief Main constructor for building a new finite element state
+   * @param[in] mesh The problem mesh (object does not take ownership)
+   * @param[in] options The options specified, namely those relating to the order of the problem,
+   * the dimension of the FESpace, the type of FEColl, the DOF ordering that should be used,
+   * and the name of the field
    */
-  using FiniteElementVector::FiniteElementVector;
-  using FiniteElementVector::operator=;
-  using mfem::Vector::Print;
+  FiniteElementState(
```
jamiebramwell: The assignment operators were not actually getting called correctly, so I had to manually instantiate them. Due to compiler complaints, this meant I also had to manually instantiate the other constructors.

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -372,12 +372,12 @@ void functional_parameterized_solid_test(double expected_disp_norm)
   EXPECT_NEAR(expected_disp_norm, norm(solid_solver.displacement()), 1.0e-6);
 }
 
-TEST(SolidFunctional, 2DLinearStatic) { functional_solid_test_static<1, 2>(1.5110557316182822); }
-TEST(SolidFunctional, 2DQuadStatic) { functional_solid_test_static<2, 2>(2.1864815661936112); }
-TEST(SolidFunctional, 2DQuadParameterizedStatic) { functional_parameterized_solid_test<2, 2>(2.1864815661936112); }
+TEST(SolidFunctional, 2DLinearStatic) { functional_solid_test_static<1, 2>(1.4863114635060113); }
```
jamiebramwell: The norms changed due to the transpose error in the stress calculation also found in the patch test PR.

src/serac/physics/common.hpp
```diff
@@ -54,4 +54,23 @@ struct SolverOptions {
...
+  Off /**< Do not include geometric nonlinearities */
+};
+
+/**
+ * @brief Flag to enable shape diplacement (and its gradient) in a physics module
+ */
+enum class ShapeDisplacement
+{
+  On, /**< Allow shape displacement and its gradient */
+  Off /**< Don't allow displacement and its gradient */
```
jamiebramwell:
```suggestion
  Off /**< Don't allow shape displacement and its gradient */
```

src/serac/physics/state/finite_element_state.hpp
```diff
@@ -47,12 +47,84 @@ inline bool is_vector_valued(const GeneralCoefficient& coef)
...
+   * @param value The constant to set the finite element state to
+   * @return The modified finite element state
+   * @note This sets the true degrees of freedom and then broadcasts to the shared grid function entries. This means
+   * that if a different value is given on different processors, a shared DOF will be set to the owning processor value.
+   */
+  FiniteElementState& operator=(const double value)
+  {
+    FiniteElementVector::operator=(value);
+    return *this;
+  }
```
samuelpmishLLNL: I think these should be covered by the deleted 

```cpp
using FiniteElementVector::operator=;
```

jamiebramwell:
This should be ready for a final review.

src/serac/physics/solid_functional.hpp
```diff
@@ -324,24 +335,49 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
   {
     residual_->AddDomainIntegral(
         Dimension<dim>{},
-        DependsOn<0, 1, active_parameters + 2 ...>{},  // the magic number "+2" accounts for the fact that the
-                                                       // displacement and acceleration fields are always-on and come
-                                                       // first, so the `n`th parameter will actually be argument `n+2`
-        [this, material](auto /*x*/, auto& state, auto displacement, auto acceleration, auto... params) {
+        DependsOn<0, 1, 2,
+                  active_parameters + 3 ...>{},  // the magic number "+2" accounts for the fact that the
```
samuelpmishLLNL:
```suggestion
                  active_parameters + 3 ...>{},  // the magic number "+3" accounts for the fact that the
```

src/serac/physics/solid_functional.hpp
```diff
@@ -324,24 +335,49 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
+                                                 // displacement and acceleration fields are always-on and come
+                                                 // first, so the `n`th parameter will actually be argument `n+2`
```
samuelpmishLLNL:
```suggestion
                                                 // first, so the `n`th parameter will actually be argument `n+3`
```

src/serac/physics/solid_functional.hpp
```diff
@@ -324,24 +335,49 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
           }
 
+          // Note that the jacobian needs the fixup for the shape displacement.
+          // The material returns Kirchoff stress, which contains det(du / dX'). We want
+          // The final Jacobian to be det(du / dX), so we compute
+          // det(du / dX) = det(du / dX') * det(dX' / dX) = det(du / dX') * det(I + dp / dX)
+          auto flux = dot(stress, transpose(inv(deformation_grad))) * det(I + dp_dX);
+
           // This transpose on the stress in the following line is a
           // hack to fix a bug in the resdual operator. The stress
```
samuelpmishLLNL:
```suggestion
          // hack to fix a bug in the residual operator. The stress
```

src/serac/physics/solid_functional.hpp
```diff
@@ -405,11 +441,16 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
+        [body_force, this](auto x, auto /* displacement */, auto /* acceleration */, auto shape, auto... /* params */) {
+          auto dp_dX = get<DERIVATIVE>(shape);
+          // auto p     = get<VALUE>(shape);
+
           // note: this assumes that the body force function is defined
           // per unit volume in the reference configuration
-          return serac::tuple{body_force(x, time_), zero{}};
+
+          // TODO should we add the shape parameter to x?
+          return serac::tuple{body_force(x, time_) * det(dp_dX + I), zero{}};
```
samuelpmishLLNL: > TODO should we add the shape parameter to x?

I think so, and maybe we should be adding the displacement as well. We should decide which configuration the physical coordinates correspond to for the body force terms.

src/serac/physics/tests/serac_solid_functional_finite_diff.cpp
```diff
@@ -156,6 +157,126 @@ TEST(SolidFunctionalFiniteDiff, FiniteDifference)
...
+    }
+
+    // Compute numerical value of sensitivity of qoi with respect to bulk modulus
+    // by taking the inner product between adjoint load and displacement sensitivity
+    double dqoi_dshape = adjoint_load_form(ddisp_dshape);
+
+    // See if these are similar
+    SLIC_INFO(axom::fmt::format("dqoi_dshape: {}", dqoi_dshape));
+    SLIC_INFO(axom::fmt::format("sensitivity: {}", sensitivity(i)));
+    EXPECT_NEAR((sensitivity(i) - dqoi_dshape) / std::max(dqoi_dshape, 1.0e-2), 0.0, 8.0e-3);
```
samuelpmishLLNL: I think you mentioned earlier that these loose tolerances might have been related to the missing transpose on the stress, or some other small bugs in the implementation. Out of curiosity, how close are the relative errors in the current implementation?

src/serac/physics/tests/serac_solid_functional_shape.cpp
```diff
@@ -0,0 +1,184 @@
...
+    solid_solver_no_shape.advanceTimestep(dt);
+
+    pure_displacement = solid_solver_no_shape.displacement().gridFunction();
+    visit_dc.SetCycle(1);
+    visit_dc.Save();
+  }
+
+  for (int i = 0; i < shape_displacement.Size(); ++i) {
+    // Check the final displacement norm
+    EXPECT_NEAR(shape_displacement[i], pure_displacement[i], 1.0e-11);
```
samuelpmishLLNL, line 164: 👍

samuelpmishLLNL: This is looking better. I would have expected better agreement between these two problems, but we can investigate that later.

samuelpmishLLNL (approved):

src/serac/physics/solid_functional.hpp
```diff
@@ -130,28 +135,34 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
-    state_.push_back(displacement_);
+    residual_ = std::make_unique<Functional<test(trial, trial, trial, parameter_space...)>>(test_space, trial_spaces);
 
-    displacement_ = 0.0;
-    velocity_     = 0.0;
+    displacement_         = 0.0;
+    velocity_             = 0.0;
+    shape_displacement_   = 0.0;
+    adjoint_displacement_ = 0.0;
+    shape_displacement_   = 0.0;
```
btalamini: Is there a reason that `shape_displacement_` needs to be zeroed out both before and after `adjoint_displacement_`? I'm guessing no, but if so, a comment is in order.

src/serac/physics/solid_functional.hpp
```diff
@@ -130,28 +135,34 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
 
+    if (calc_shape_ == ShapeDisplacement::On) {
+      state_.push_back(shape_displacement_);
+    }
+
     // Create a pack of the primal field and parameter finite element spaces
-    mfem::ParFiniteElementSpace*                                             test_space = &displacement_.space();
-    std::array<mfem::ParFiniteElementSpace*, sizeof...(parameter_space) + 2> trial_spaces;
+    mfem::ParFiniteElementSpace* test_space = &displacement_.space();
+
```
btalamini, line 147: The "3" magic number appears here, too. Maybe it's worth pulling this into a named constant?

src/serac/physics/solid_functional.hpp
```diff
@@ -324,24 +335,49 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
+                                                 // first, so the `n`th parameter will actually be argument `n+2`
+        [this, material](auto /*x*/, auto& state, auto displacement, auto acceleration, auto shape, auto... params) {
           auto du_dX   = get<DERIVATIVE>(displacement);
           auto d2u_dt2 = get<VALUE>(acceleration);
-          auto stress  = material(state, du_dX, params...);
+          auto dp_dX   = get<DERIVATIVE>(shape);
+
+          // Compute the displacement gradient with respect to the shape-adjusted coordinate.
+
+          // Note that the current configuration x' = X + u + p, where X is the original reference
```
btalamini, line 352: We talked about this in person, so this comment is for record keeping: this is getting pretty involved with all of the change-of-configuration mappings. I support the suggestion to move to Cauchy or Piola as our expectation from the constitutive.

src/serac/physics/solid_functional.hpp
btalamini: +1 for adding the shape displacement to the coordinates once and for all. 

additional point: I vote to keep the reference coordinates as the default argument for the loading functions. Dead loading is both common and simpler (since it doesn't entail geometric nonlinearity). We *could* make an overload for a spatial version, but I suggest waiting until a user request for it comes in.

btalamini (approved):

src/serac/physics/tests/serac_solid_functional_finite_diff.cpp
```diff
@@ -156,6 +157,126 @@ TEST(SolidFunctionalFiniteDiff, FiniteDifference)
...
+      DirectSolverOptions{},
+      NonlinearSolverOptions{.rel_tol = 1.0e-6, .abs_tol = 1.0e-12, .max_iter = 10, .print_level = 1}};
+
+  // Construct a functional-based solid solver
+  SolidFunctional<p, dim> solid_solver(options, GeometricNonlinearities::On, "solid_functional");
+
+  solid_mechanics::NeoHookean mat{1.0, 1.0, 1.0};
+  solid_solver.setMaterial(mat);
+
+  solid_solver.shapeDisplacement() = shape_displacement_value;
```
kswartz92: At first glance I was confused why this was set with a `double`. Is this a rigid body translation? Also, perhaps it would be more consistent with other methods in solid functional to have a `setShapeDisplacement()` method?

src/serac/physics/state/finite_element_dual.hpp
```diff
@@ -56,9 +83,44 @@ class FiniteElementDual : public FiniteElementVector {
...
+   * @brief Fill a user-provided linear form based on the underlying true vector
+   *
+   * This distributes true vector dofs to the finite element (local) dofs by multiplying the true dofs
+   * by the restriction transpose operator.
+   *
+   * @see <a href="https://mfem.org/pri-dual-vec/">MFEM documentation</a> for details
+   *
+   * @param linear_form The linear form used to initialize the underlying true vector.
+   */
+  void fillLinearForm(mfem::ParLinearForm& linear_form) const
```
kswartz92, line 70: I think it's a great addition and much more clear to have the `fillLinearForm` and `setFromLinearForm` methods for `FiniteElementDual`. Are the `gridfunction` version of these methods still necessary? It seems confusing to a user to support both methods

jamiebramwell: Agreed. I'm hesitant to delete them though because people tend to use grid functions for everything. If LiDO is OK with me deleting, I'm OK with it as it does make more mathematical sense. Any thoughts @sethwatts ?

sethwatts: I would go farther and actually prefer that you delete the `GridFunction` filler due to the mathematical inconsistency.

jamiebramwell: Question: Would you ever want to set a dual vector from a coefficient?

sethwatts: I'm not coming up with any examples where we'd want to do that. @kswartz92 can you think of any? The closest I can think of is of a load vector, but I think the integration of a traction field into a load linear form is something we wouldn't want to skip.

kswartz92: I don't think I've ever set a dual vector from a coefficient

src/serac/physics/solid_functional.hpp
jamiebramwell: Nope. This is a typo.

src/serac/physics/solid_functional.hpp
jamiebramwell: I put this down as an issue. I agree we should move to either Cauchy or Piola in a follow on PR.

src/serac/physics/tests/serac_solid_functional_finite_diff.cpp
jamiebramwell: Yes - this is using the `mfem::Vector::operator=(double)` method. This brings up an important question though - would you rather have a set method and a const getter or direct access like this? @sethwatts @kswartz92 ?

src/serac/physics/tests/serac_solid_functional_finite_diff.cpp
samuelpmishLLNL: I ran these tests locally, and the the relative errors tended to be on the order of 1.0e-5 or 1.0e-6

src/serac/physics/tests/serac_solid_functional_finite_diff.cpp
jamiebramwell: Any thoughts on this @kswartz92 @sethwatts ? I believe this is the last outstanding issue in this PR.

sethwatts: I lean toward consistency with other methods (i.e., getter/setter) I will defer to @kswartz92 on this one.

kswartz92: I don't have a strong preference either way, but it seems more consistent with `SolidFunctional` and `FiniteElementState` methods to have a getter/setter. I can't think of any other places in serac where I've noticed something is set by grabbing a reference like this

PR #671, opened 2022-04-11
Fix the essential boundary condition application in parallel runs

### LINKED ISSUE #668: qs_attribute_solve_parallel test fails (sometimes)
When running `./ats.sh`, qs_attribute_solve_parallel takes a long time to converge (about 3 minutes). According to the logs, it tries 5000 iterations before it gives up. The tolerances that fail seem to switch between two values:

> ERROR: Test value 'displacement/l1norms[0]' out of tolerance: baseline value=0.0996687535543166, test value=1.95006410505963
> ERROR: Test value 'displacement/l2norms[0]' out of tolerance: baseline value=-0.216814893270359, test value=3.0725532465425
> ERROR: Test value 'displacement/linfnorms[0]' out of tolerance: baseline value=20.3046995090693, test value=13.7850688348039
> ERROR: Test value 'displacement/avgs[0]' out of tolerance: baseline value=0.29714063702572, test value=-0.407641749981539
> ERROR: Test value 'displacement/mins[0]' out of tolerance: baseline value=-6.88522152649478, test value=-18.488253015702
> ERROR: Test value 'displacement/maxs[0]' out of tolerance: baseline value=24.5078855697869, test value=4.95114675337821

Runs totally fine serially, though.

jamiebramwell:
This resolves an issue where we were applying boundary conditions using true DOF indices instead of local (grid function) indices. I'm honestly not sure why it hasn't appeared before now. 

Note that this does change the `BoundaryCondition` interface slightly as I explicitly denote whether one should use true or local degrees of freedom.

Resolves https://github.com/LLNL/serac/issues/668


jamiebramwell:
This definitely has a bug in parallel. I'm investigating...

jamiebramwell:
Found the bug! This will also fix a parallel issue @kswartz92 was seeing downstream.

chapman39 (approved):

src/serac/physics/boundary_conditions/boundary_condition.hpp
```diff
@@ -132,16 +132,29 @@ class BoundaryCondition {
...
+  void setTrueDofs(const mfem::Array<int> true_dofs);
+
+  /**
+   * @brief "Manually" set the DOF indices without specifying the field to which they apply
+   * @param[in] local_dofs The local (finite element/grid function) indices of the DOFs constrained by the boundary
+   * condition
+   *
+   * @note This will set both the true and local internal dof index arrays.
+   */
+  void setLocalDofs(const mfem::Array<int> local_dofs);
```
white238, line 150: Should we have descriptions here what the difference between a truedof, localdof, and a dof is?

white238 (approved):

src/serac/physics/boundary_conditions/boundary_condition.hpp
jamiebramwell: I really need to write a sphinx page in our docs about this as it's a fundamental concept in MFEM. I'll add an issue.

white238: That would be helpful as well, but a one sentence (if it can be that short) description at this level since they are all sitting here, would be useful as well.

chapman39 (approved):

Issue #667, opened 2022-04-06
Investigate why CTAD fails with clang in the parameterized neo-Hookean material model


jamiebramwell:
if something like
```cpp
return MaterialResponse{.density = density_, .stress = stress};
```
doesn't work, I don't think it should hold up this PR. We can just put a `// TODO` on it and make an issue, to revisit later.

_Originally posted by @samuelpmishLLNL in https://github.com/LLNL/serac/pull/658#discussion_r844430152_

jamiebramwell:
Note: this also fails in the non-parameterized version.

PR #674, opened 2022-04-14
resolve issue #667

samuelpmishLLNL:
see: https://github.com/LLNL/serac/issues/667

clang was failing to compile the following code:

```cpp
return MaterialResponse{.density = density_, .stress = stress};
```

this feature using `.density = ...` in the constructor ("[designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization)") isn't technically supported until C++20 but some compilers allow it anyway. In this case, mixing that with CTAD led to a compilation error.


adayton1 (approved):
LGTM

white238 (approved):

jamiebramwell:
Thanks for looking at this @samuelpmishLLNL !

PR #675, opened 2022-04-15
Use correct CMake MPI variables for Adiak

white238:
I'm not sure how this worked at all to be honest but it does build in most specs.  Maybe CMake is auto-finding MPI in those cases... Life is a mystery.

I have pushed this to Spack proper and hacked a blocked TPL build for Smith.  No need to rebuild things here.


adayton1 (approved):
Thanks for the fix!

jamiebramwell:
Thank you @white238 !!!!!!!!!

jamiebramwell (approved):

PR #681, opened 2022-04-22
Feature/talamini/thermal solid functional

btalamini:
Create a prototype of a material for functional-based thermomechanics.

The implementation is kept inside the file for its unit test suite, so it isn't affecting any code path yet.


src/serac/physics/materials/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause)
 
+add_subdirectory(tests)
```
jamiebramwell, line 7: 👍

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,303 @@
...
+   * @param[out] q0 Piola heat flux
+   */
+  template <typename T1, typename T2, typename T3>
+  auto calculate_constitutive_outputs(const tensor<T1, 3, 3>& grad_u,
+                                      T2 theta,
+                                      const tensor<T3, 3>& grad_theta,
+                                      State& /*state*/,
+                                      const tensor<double, 3, 3>& grad_u_old,
+                                      double /*theta_old*/,
+                                      double dt)
```
samuelpmishLLNL: @jamiebramwell I was thinking it would be good to change the other materials to follow this pattern, where instead of having the entire argument type be an arbitrary type, the arbitrary-ness is only in what is stored in the tensor. e.g.
```cpp
const tensor<T1, 3, 3>& grad_u
```
instead of
```cpp
const T1 & grad_u
```
This way, the function signature reveals more about the shapes of each argument.

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,303 @@
+#include <iostream>
```
jamiebramwell, line 13: This needs the copyright banner.

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,303 @@
...
+                                     {-0.51876824611259 ,  0.728725283562492, -0.44703351990878  },
+                                     { 0.058192140263313,  0.551784758906877,  0.831953877717782 }}};
+
+/**
+ * @brief Compute Green's strain from the displacement gradient
+ */
+template <typename T>
+auto green_strain(const tensor<T, 3, 3>& grad_u)
+{
+  return 0.5*(grad_u + transpose(grad_u) + transpose(grad_u)*grad_u);
```
jamiebramwell: @samuelpmishLLNL should this be `dot(...)` instead of `*`?

src/serac/physics/materials/tests/thermomechanical_material.cpp
jamiebramwell: I agree - this is nicer.

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,303 @@
...
+// An arbitrary rotation tensor for frame indifference tests
+static const tensor<double, 3, 3> Q{{{ 0.852932108456964,  0.405577416697288, -0.328654495524269 },
+                                     {-0.51876824611259 ,  0.728725283562492, -0.44703351990878  },
+                                     { 0.058192140263313,  0.551784758906877,  0.831953877717782 }}};
+
+/**
+ * @brief Compute Green's strain from the displacement gradient
+ */
+template <typename T>
+auto green_strain(const tensor<T, 3, 3>& grad_u)
```
jamiebramwell: We try to use `camelCase` for function names. See [here](https://serac.readthedocs.io/en/latest/sphinx/dev_guide/style_guide.html).

src/serac/physics/thermal_solid_functional.hpp
```diff
@@ -0,0 +1,121 @@
...
+   * @param geom_nonlin Flag to include geometric nonlinearities
+   * @param keep_deformation Flag to keep the deformation in the underlying mesh post-destruction
+   * @param name An optional name for the physics module instance
+   */
+  ThermalSolidFunctional(const typename ThermalConductionFunctional<order, dim>::SolverOptions& thermal_options,
+                         const typename SolidFunctional<order, dim>::SolverOptions& solid_options,
+                         GeometricNonlinearities geom_nonlin = GeometricNonlinearities::On,
+                         FinalMeshOption keep_deformation = FinalMeshOption::Deformed,
+                         const std::string& name = "")
+      : BasePhysics(2, order),
```
jamiebramwell: This should be `3` indicating 3 state fields (`displacement`, `velocity`, and `temperature`). However, I'm honestly not sure we use it anymore so I should probably just get rid of this argument.

src/serac/physics/tests/serac_thermal_solid_functional.cpp
```diff
@@ -0,0 +1,93 @@
...
+                                                         .prec        = HypreBoomerAMGPrec{}};
+
+  const NonlinearSolverOptions default_nonlinear_options = {
+    .rel_tol = 1.0e-4, .abs_tol = 1.0e-8, .max_iter = 10, .print_level = 1};
+
+  const typename SolidFunctional<p, dim>::SolverOptions solid_options = {default_linear_options,
+    default_nonlinear_options};
+  
+  // Construct a functional-based thermal-solid solver
+  ThermalSolidFunctional<p, dim> thermal_solid_solver(thermal_options,
```
jamiebramwell: I would add a note saying the material call still needs to be implemented before this module can be used.

jamiebramwell (commented):
Looks good! A few minor comments.

src/serac/physics/materials/tests/thermomechanical_material.cpp
btalamini: I saw an `operator*(...)` overload on line 931 of `tensor.hpp` that appears to make them do the same thing here. Should I not use that?

src/serac/physics/materials/tests/thermomechanical_material.cpp
btalamini: Fixing that. Does this apply to the methods in the material model struct as well?

src/serac/physics/materials/tests/thermomechanical_material.cpp
samuelpmishLLNL: Right now, `operator*` is used for both scalar multiplication of tensors and also a shortcut for `dot(...)` (i.e. single contractions / regular matrix multiplication), so they should be equivalent. 

We have discussed the possibility of making `operator*` be reserved for only scalar multiplication, and using free functions for other kinds of products (e.g. `dot(A, B)` for single contraction `double_dot(A, B)` for double contraction, `inner(A, B)` for "contract all", and `cross(A, B)` for the cross product).

btalamini: It seems like using a bare `*` for matrix multiplication is **almost** deprecated. I'll switch to `dot(A, B)`. It's still readable, and there's no added inertia to restricting `*` to scaling later.

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+
+/// @brief Green-Saint Venant isotropic thermoelastic model
+struct   ThermoelasticMaterial {
+  double E;          ///< Young's modulus
+  double nu;         ///< Poisson's ratio
+  double C;          ///< volumetric heat capacity
+  double alpha;      ///< thermal expansion coefficient
+  double theta_ref;  ///< datum temperature
+  double k;          ///< thermal conductivity
```
white238, line 46: 👍 comments!

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+  struct State {/* this material has no internal variables */};
+
+  /**
+   * @brief Evalute constitutive variables for thermomechanics
```
white238:
```suggestion
   * @brief Evaluate constitutive variables for thermomechanics
```

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,315 @@
...
+  ThermoelasticMaterial::State state{};
+  tensor<double, 3, 3> displacement_grad_old{};
+  double dt = 1.0;
+  auto generalized_fluxes = material.calculateThermalConstitutiveOutputs(
+      displacement_grad, temperature, temperature_grad, state,
+      displacement_grad_old, temperature_old, dt);
+  auto [heat_capacity, source, heat_flux] = generalized_fluxes;
+  // "inner" didn't work for me, but "dot" does
+  // ask sam and jamie about this
+  auto dissipation = -dot(heat_flux, temperature_grad)/temperature;
```
white238: Was this answered/asked?

btalamini: I talked to @samuelpmishLLNL about it. `inner` is intended to work here eventually, but this is not implemented yet. Should I do something to record this elsewhere?

src/serac/physics/materials/tests/thermomechanical_material.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+} // namespace serac
+
+
+int main (int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+
+  int result = RUN_ALL_TESTS();
+
```
white238, line 302: Does this touch any code that may output via SLIC?  If it does you need to initialize a SLIC logger. Example:

```
  axom::slic::SimpleLogger logger;  // create & initialize test logger, finalized when
                                    // exiting main scope
```

src/serac/physics/thermal_solid_functional.hpp
```diff
@@ -0,0 +1,125 @@
...
+ */
+template <int order, int dim>
+class ThermalSolidFunctional : public BasePhysics {
+public:
+  
+  /**
+   * @brief Construct a new coupled Thermal-Solid Functional object
+   *
+   * @param thermal_options The options for the linear, nonlinear, and ODE solves of the thermal operator
+   * @param solid)options The options for the linear, nonlinear, and ODE solves of the thermal operator
```
white238:
```suggestion
   * @param solid_options The options for the linear, nonlinear, and ODE solves of the thermal operator
```

src/serac/physics/thermal_solid_functional.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+  void advanceTimestep(double& dt) override
+  {
+    if (coupling_ == serac::CouplingScheme::OperatorSplit) {
+      double initial_dt = dt;
+      thermal_functional_.advanceTimestep(dt);
+      solid_functional_.advanceTimestep(dt);
+      SLIC_ERROR_ROOT_IF(std::abs(dt - initial_dt) > 1.0e-6,
+                         "Operator split coupled solvers cannot adaptively change the timestep");
+  } else {
```
white238: This file needs style ran over it. In your build directory type `make style` and clangformat will fix everything for you.

src/serac/physics/materials/tests/thermomechanical_material.cpp
white238: Add a TODO and move on imo.  Something like:

```
TODO: change to inner() when implemented
```

samuelpmishLLNL: Brandon and I talked about this in a webex last week. The takeaway was: `inner` _should_ work here too, but we haven't implemented the overloads for the different tensor dimensions yet. 

For some context, those gcc 11.x errors we saw a few months earlier seemed to be caused by a compiler bug that was handling variadic lists of integers incorrectly. Unfortunately, that's the C++ tool that makes it easy to write functions like `inner(A, B)` work for any shaped tensor in 1 function definition, so if we want serac to support gcc 11, then we have to manually write out separate function overloads for the cases 1D tensors, 2D tensors, 3D tensors, etc... -- Oof!

src/serac/physics/materials/tests/thermomechanical_material.cpp
btalamini: I don't know the answer to this. The contents of this file are not used anywhere else, but the includes could have such output.

src/serac/physics/thermal_solid_functional.hpp
btalamini: Ran it and committed. Leads me to a team style question: most of the results are good, but it butchered the displacement gradient initializations. They were aligned so that they looked like 3x3 matrices, now they're each crammed on a single line that runs well over 100 columns. Are exceptions tolerated in some way?

white238: Yes you can toggle it off like the following:

```
// clang-format off
... // code you dont want autoformatted
// clang-format on
```

src/serac/physics/materials/tests/thermomechanical_material.cpp
btalamini: After our discussion in the weekly meeting, I added this line in.

src/serac/physics/materials/tests/thermomechanical_material.cpp
jamiebramwell: Yes - classes are Pascal and methods are camel.

jamiebramwell (approved):

white238 (approved):
Thanks @btalamini !

Issue #689, opened 2022-05-06
Implement useful tensor operations


btalamini:
There are some core tensor operations that are needed, as well as some that would be nice to have.

In my opinion, these are the needs. They're all for 2D and 3D 2nd order tensors:
- Spectral decomposition for symmetric matrices
- Matrix exponential for symmetric matrices and matrix log function for SPD matrices
- Matrix square root for symmetric SPD matrices
- Polar decomposition for real matrices with positive determinant

Once the spectral decomposition is in place, we could use it to write convenience functions for the others.

Nice to have:
- Extend the linear solve function to work on matrices of right hand sides (ie `A * x = B`, where `B` is a matrix of RHS vectors). This would make it possible to avoid most calls to the matrix inverse.
- A representation of symmetric tensors, at least for 2nd order tensors. This would automatically guard some algorithms that only make sense on symmetric matrices and would save a little storage.
- Some means of invoking linear algebra libraries on 4th order tensors, especially if they have at least major symmetry. Computing eigenvalues of elastic tensors is the main use case I'm thinking about. Being able to cast the 4-th order tensors as 9x9 matrices (or 6x6 and symmetric 6x6 when there are major/minor symmetries) is handy.
- More flexible tensor contractions. It's hard to give all of the possible contractions meaningful names, so it would be nice to be able to express things in index notation. Numpy's `einsum(...)` function is one possible model (ie, `C_ijml, F_km -> B_ijkl`)
- Convenience function for ranges, like Python's `a = range(5)`. Sam posted some ideas [here](https://godbolt.org/z/3eda4PzsT)

btalamini:
Fixed by #1146

PR #691, opened 2022-05-07
Feature/talamini/tensor linear solves with matrix rhs

btalamini:
- Refactor to extract the components of the linear solve into resuable functions that are likely of wider utility: (1) LU factorization, (2) upper triangular matrix solve, (3) lower triangular matrix solve
- Re-implement `linear_solve(...)` function in terms of the new utility functions. As a benefit, it can now take a matrix RHS (instead of only a single vector). This is useful for avoiding explicit inverse matrices and their associated numerical problems. For example, if you're trying to convert the Cauchy stress to the first Piola stress, instead of 
```P = det(F)*sigma*transpose(inv(F))```, you can solve `PT = linear_solve(F, det(F)*sigma)` and get the answer by taking the transpose of `PT`. These kinds of operations also pop up frequently in adjoint methods.
- Re-implement the matrix inversion `inv(tensor<T, n, n> A)` using the LU factorization
- Creates unit tests of the linear solves, inverse, and dual number derivatives of them

This also fixes a trivial bug that prevented the existing `linear_solve` from compiling. I think it wasn't noticed because there was no code or test instantiating it yet.

### CODE REVIEW

File: src/serac/numerics/functional/tensor.hpp [hunk 1]
```diff
@@ -1170,58 +1172,147 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+ * L the lower triangular factor
+ * U the upper triangular factor
+ * P an array indicating the row permutations due to pivoting.
+ * Row i of A was exchanged with row P[i].
  */
 template <typename T, int n>
-SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const tensor<T, n> b)
+SERAC_HOST_DEVICE constexpr auto lu(const tensor<T, n, n>& A)
 {
+  // BT question for Sam Mish: Why not use the standard library abs?
```
samuelpmishLLNL: I'm not sure-- if it had `auto` type in its argument it would make more sense. It's always annoying to manage the different `abs` implementations from `std::`, C, CUDA, and now our AD versions. If all the tests compile/pass without defining this `abs` then we should take it out.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,58 +1172,147 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
   constexpr auto abs  = [](double x) { return (x < 0) ? -x : x; };
   constexpr auto swap = [](auto& x, auto& y) {
     auto tmp = x;
     x        = y;
     y        = tmp;
   };
 
-  tensor<double, n> x{};
+  auto           U = A;
+  auto           L = DenseIdentity<n>();
```
samuelpmishLLNL: `L` will be of type `tensor< double, n, n >` from this, which may not be what you want, given that the original matrix has type `tensor< T, n, n >`

btalamini: Oh, ok - not good. I could use `make_tensor()` to make an identity with type `T`.  Is that a good fix?

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,58 +1172,147 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+  tensor<int, n> P(make_tensor<n>([](auto i) { return i; }));
```
samuelpmishLLNL, line 1217: this is another thing that might be useful to add to the tensor class. C++ has `std::iota`, but that can't be used for declaration + initialization simultaneously.

maybe we could define something like one of these options: https://godbolt.org/z/3eda4PzsT

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,58 +1172,147 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
-    // zero entries below in this column
+  for (int i = 0; i < n; i++) {
+    // zero entries below in this column in U
+    // and fill in L entries
     for (int j = i + 1; j < n; j++) {
-      double c = -A[j][i] / A[i][i];
-      A[j] += c * A[i];
-      b[j] += c * b[i];
-      A[j][i] = 0;
+      double c = U[j][i] / U[i][i];
```
samuelpmishLLNL: using `double` explicitly might make this do the wrong thing when `T` is a dual type

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,58 +1172,147 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+}
+
+/**
+ * @brief Solves Ax = b for x using Gaussian elimination with partial pivoting
+ * @param[in] A The coefficient matrix A
+ * @param[in] b The righthand side vector b
+ * @return x The solution vector
+ */
+template <typename T, int n>
+SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(const tensor<T, n, n> A, const tensor<T, n>& b)
```
samuelpmishLLNL: we should probably allow for `A` and `b` that hold separate types too, e.g.

```cpp
template <typename S, typename T, int n>
SERAC_HOST_DEVICE constexpr auto linear_solve(const tensor<S, n, n> A, const tensor<T, n>& b) {
   ...
}
```

so that if only one argument contains dual number information, then it will still work

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,58 +1172,147 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+{
+  auto const [P, L, U] = lu(A);
+
+  // Forward substitution
+  // solve Ly = b
+  tensor<T, n> y{};
```
samuelpmishLLNL: re: https://github.com/LLNL/serac/pull/691/files#r868201191

I think `x` and `y` would have type `tensor< decltype(T{} / S{}), n >` in that case

src/serac/numerics/functional/tensor.hpp
btalamini: I was thinking that, too. I wanted to talk you about how best to do it. For example, if `A` is a dual type and `b` isn't, I could promote `b` to dual and call the implemented function again. Slightly less efficient, but doesn't repeat code. Or I could handle each permutation manually.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,58 +1172,147 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+  // Back substitution
+  // Solve Ux = y
+  tensor<T, n, m> x{};
+  for (int i = n - 1; i >= 0; i--) {
+    auto c = y[i];
+    for (int j = i + 1; j < n; j++) {
+      c -= U[i][j] * x[j];
+    }
+    x[i] = c / U[i][i];
   }
```
samuelpmishLLNL: in theory, the different shapes of the rhs argument can be handled in a single implementation, doing something like
```cpp
template <typename S, typename T, int n, int ... m>
SERAC_HOST_DEVICE constexpr auto linear_solve(const tensor<S, n, n>& A, const tensor<T, n, m ... >& b)
{
  using output_type = decltype(T{} / S{});
  auto [P, L, U] = lu(A);

  // Forward substitution
  // solve Ly = b
  tensor<output_type, n, m ... > y{};
  ...

  // Back substitution
  // Solve Ux = y
  tensor<output_type, n, m ... > x{};
  ...
  
  return x;
}
```

we might want to try that and see if g++11 handles it correctly or not

src/serac/numerics/functional/tensor.hpp
btalamini: The dual type case shouldn't happen, if I've done things right, because there is a specific implementation to handle that case. But your point stands, this will do the wrong thing if T is another type like quad precision, say. Will `T c = U[j][i] / U[i][i];` suffice?

src/serac/numerics/functional/tensor.hpp
btalamini: I'd like to write a test that fails, and then fix it. Let me see if I can do that.

src/serac/numerics/functional/tensor.hpp
samuelpmishLLNL: we should be able to just write 1 implementation that handles all of the different cases (`A` holds type `S`, and `b` is of arbitrary shape and holds type `T`), with something like https://github.com/LLNL/serac/pull/691#discussion_r868207090

src/serac/numerics/functional/tensor.hpp
btalamini: Or I guess `auto` would be best. Your comment applies to line 1200 as well.

btalamini: I wrote a test in https://github.com/LLNL/serac/pull/691/commits/b0570fe2cd38e112980fdcff2d14622bd8740021 that verifies that the LU factorization code fails to compile when instantiated with dual numbers, and then, with your help, we modified the implementation to handle this case. The test now passes. It checks the value and the gradient produced by LU.

src/serac/numerics/functional/tensor.hpp
btalamini: Question retracted. I'm leaving the lambda. I didn't think of the insufficient generality of the return type of `std::abs`

src/serac/numerics/functional/tensor.hpp
btalamini: Fixed this in b0570fe2cd38e112980fdcff2d14622bd8740021 (with your coaching). I think the right thing to do long-term is to parameterize the identity matrix data type, so that you can write `L = Identity<T, n>();` here. I made a note of this in the comments. Since that's a breaking change, I'll defer that for now.

src/serac/numerics/functional/tensor.hpp
btalamini: I agree this would be useful. I added that suggestion and your link to Issue #689.

src/serac/numerics/functional/tensor.hpp
btalamini: I actually don't want the generic `linear_solve(...)` to handle dual numbers. I wrote a specific implementation for dual numbers that avoids accumulating the derivatives through the factorization (similar to your custom derivative for `inv()`). This will be more efficient (at least for large enough matrices). It's also generic, in the sense that it doesn't depend on the underlying algorithm used for the solve. If we later implement an iterative solver option, this custom derivative will surely be more efficient than the straightforward approach. The problem I see with what I have is that it only works when *both* `A` and `b` hold dual numbers. When I try to use it to handle the case of only one geing dual, I run into compilation problems because my function templates are ambiguous. I see the problem, but I don't know how to solve it. Could you take a look?

src/serac/numerics/functional/tensor.hpp
btalamini: This has changed due to refactoring (to extract the upper/lower triangular solves as stand-alone functions, and to handle the multiple/single rhs case using your variadic template suggestion). It's `auto y{};` now. See my comment [above](https://github.com/LLNL/serac/pull/691#discussion_r870636670) though about not wanting this part of the code path to handle the dual number case. I think I want the underlying number type of `x`, `y`, `A`, and `b` to be the same.

src/serac/numerics/functional/tensor.hpp
btalamini: I changed it and the CI didn't choke. Looks like it works.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1185,43 +1200,187 @@ SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const t
...
+      U[j][i] = T{};
     }
   }
 
-  // Solve equation Ax=b for an upper triangular matrix A
-  for (int i = n - 1; i >= 0; i--) {
-    x[i] = b[i] / A[i][i];
-    for (int j = i - 1; j >= 0; j--) {
-      b[j] -= A[j][i] * x[i];
+  return LuFactorization<T, n>{.P = P, .L = L, .U = U};
```
samuelpmishLLNL: just a C++ note: the aggregate (brace) initialization doesn't require the designators (i.e. `.P = ..., .L = ...`). We use that syntax in a couple places in serac because there are some arguments that are defaulted. So, this line could just be written 
```cpp
return LuFactorization<T, n>{P, L, U};
```
or (I think) even
```cpp
return {P, L, U};
```
since you declared the full return type in the function declaration.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,13 +1172,26 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+ *
+ * The entries of P mean row i of the matrix was exchanged with row P[i].
  */
 template <typename T, int n>
-SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const tensor<T, n> b)
+struct LuFactorization {
+  tensor<int, n>  P;  ///< Row permutation indices due to partial pivoting
+  tensor<T, n, n> L;  ///< Lower triangular factor
+  tensor<T, n, n> U;  ///< Upper triangular factor
+};
```
samuelpmishLLNL: does `L` or `U` have 1s on the diagonal? If so, that would be good to note here

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1185,43 +1200,187 @@ SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const t
...
+ *
+ * U must be upper triangular. This is not checked, so
+ * failure to obey this will produce meaningless results.
+ *
+ * @param[in] U An upper triangular matrix
+ * @param[in] y The right hand side
+ * @return x the solution vector
+ */
+template <typename T, int n, int... m>
+SERAC_HOST_DEVICE constexpr auto solve_upper_triangular(const tensor<T, n, n> U, const tensor<T, n, m...> y)
```
samuelpmishLLNL:
```suggestion
SERAC_HOST_DEVICE constexpr auto solve_upper_triangular(const tensor<T, n, n> & U, const tensor<T, n, m...> & y)
```

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1185,43 +1200,187 @@ SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const t
...
   }
+  return y;
+}
 
+/**
+ * @overload
+ * @note For the case when no permutation of the rows is needed.
+ */
+template <typename T, int n, int... m>
+SERAC_HOST_DEVICE constexpr auto solve_lower_triangular(const tensor<T, n, n> L, const tensor<T, n, m...> b)
```
samuelpmishLLNL:
```suggestion
SERAC_HOST_DEVICE constexpr auto solve_lower_triangular(const tensor<T, n, n> & L, const tensor<T, n, m...> & b)
```

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1185,43 +1200,187 @@ SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const t
...
+ * meaningless results.
+ *
+ * @param[in] L A lower triangular matrix
+ * @param[in] b The right hand side
+ * @param[in] P A list of indices to index into b in a permuted fashion.
+ *
+ * @return y the solution vector
+ */
+template <typename T, int n, int... m>
+SERAC_HOST_DEVICE constexpr auto solve_lower_triangular(const tensor<T, n, n> L, const tensor<T, n, m...> b,
```
samuelpmishLLNL:
```suggestion
SERAC_HOST_DEVICE constexpr auto solve_lower_triangular(const tensor<T, n, n> & L, const tensor<T, n, m...> & b,
```

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,13 +1172,26 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+
+/**
+ * @brief Compute LU factorization of a matrix with partial pivoting
+ *
+ * @param[in] A The matrix to factorize
+ * @return An LuFactorization object
+ * @see LuFactorization
+ */
+template <typename T, int n>
+SERAC_HOST_DEVICE constexpr LuFactorization<T, n> lu(tensor<T, n, n> A)
```
samuelpmishLLNL:
```suggestion
SERAC_HOST_DEVICE constexpr LuFactorization<T, n> lu(const tensor<T, n, n> & A)
```

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1185,43 +1200,187 @@ SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const t
...
+
+  // Strip off derivatives, if any, and compute only x (ie no derivative)
+  auto lu_factors = lu(get_value(A));
+  auto x          = linear_solve(lu_factors, get_value(b));
+
+  // Compute directional derivative of x.
+  // If both b and A are not dual, the zero type
+  // makes these no-ops.
+  auto r  = get_gradient(b) - dot(get_gradient(A), x);
+  auto dx = linear_solve(lu_factors, r);
```
samuelpmishLLNL, line 1342: as it is right now, this line `auto dx = linear_solve(lu_factors, r);` may not be a no-op, even if `r` is of type `serac::zero` since there is no overload where `b` is a `serac::zero`. I think we can just add this function somewhere in the file to fix that:

```cpp
template <typename T, int n>
SERAC_HOST_DEVICE constexpr auto linear_solve(const tensor<T, n, n> &, zero) {
  return zero{};
}
```

note: this code was written in github, it may not compile

src/serac/numerics/functional/tests/tensor_unit_tests.cpp
```diff
@@ -152,3 +163,160 @@ TEST(tensor, implicit_conversion)
...
+  EXPECT_LT(squared_norm(dot(A, get_value(invA)) - Identity<4>()), tolerance);
+}
+
+template <int n>
+void checkLUDecomposition(const tensor<double, n, n>& A)
+{
+  auto lu_factorization = lu(A);
+  auto P                = lu_factorization.P;
+  auto L                = lu_factorization.L;
+  auto U                = lu_factorization.U;
```
samuelpmishLLNL:
```suggestion
  auto [P, L, U] = lu(A);
```

src/serac/numerics/functional/tensor.hpp
btalamini: I did put an overload for zero rhs (line 1370). It differs from your suggestion only in that it has `const zero` for the second argument instead of `zero`. Will that not suffice?

src/serac/numerics/functional/tests/tensor_unit_tests.cpp
```diff
@@ -152,3 +163,160 @@ TEST(tensor, implicit_conversion)
...
+  const tensor<double, 3> b{{-1, 2, 3}};
+
+  // central difference (2nd order accurate)
+  const double h     = 1e-6;
+  auto         dx_FD = (linear_solve(v + h * g, b) - linear_solve(v - h * g, b)) / (2 * h);
+
+  auto x = linear_solve(A, b);
+
+  EXPECT_LT(squared_norm(dx_FD - get_gradient(x)), tolerance);
+}
```
samuelpmishLLNL, line 327: one other test you might add is one where you try the LU and linear solve routines in a `constexpr` context, something like

```cpp
constexpr tensor<double, 3, 3> A{{{2, 1, -1}, {-3, -1, 2}, {-2, 1, 2}}};
constexpr tensor<double, 3>    b{{-1, 2, 3}};
constexpr auto x = linear_solve(A, b);
constexpr auto PLU = lu(A);
```

This will help verify that the constexpr evaluation works as intended too. Another benefit is that since undefined behavior (e.g. out-of-bounds array accesses, reading from uninitialized memory, etc.) is forbidden in a `constexpr` context, the compiler will locate any problems like that and let you know the exact line of code where it occurs.

src/serac/numerics/functional/tensor.hpp
samuelpmishLLNL: oops, I missed that! either one should be good

src/serac/numerics/functional/tensor.hpp
btalamini: Indeed the second one works. Thanks for the suggestion. Incorportated in https://github.com/LLNL/serac/pull/691/commits/8cd4d0423dd683767e4547bee9b5fad4017c49ac.

src/serac/numerics/functional/tensor.hpp
btalamini: Agreed. I added this comment. I also noted it in the Doxygen block for the function definition of `lu()`.

src/serac/numerics/functional/tensor.hpp
btalamini: Update: Sam and I merged the dual and non-dual number implementations while still using the analytic derivative approach that avoids using AD naively on the Gaussian elimination.

src/serac/numerics/functional/tensor.hpp
btalamini: Oops, I just overlooked this. I meant to go back and mark these arguments as references when I was finishing up. I fixed this and the others.

src/serac/numerics/functional/tensor.hpp
btalamini: See above.

src/serac/numerics/functional/tensor.hpp
btalamini: See above

src/serac/numerics/functional/tests/tensor_unit_tests.cpp
btalamini: Good idea. I added it (see line 259). This led me to add missing `constexpr` specifiers to a couple functions in `tensor.hpp` and `dual.hpp`.

samuelpmishLLNL (approved):

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1185,43 +1202,188 @@ SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const t
     y        = tmp;
   };
 
-  tensor<double, n> x{};
+  auto U = A;
+  // initialize L to Identity
+  auto L = tensor<T, n, n>{};
+  // This handles the case if T is a dual number
+  // TODO: consider making a dense identity that is templated on type
+  // BT 05/09/2022
```
jamiebramwell: I don't like leaving stamps in code as this is handled by git, but I defer to @white238 's judgement.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1170,13 +1172,28 @@ SERAC_HOST_DEVICE bool is_symmetric_and_positive_definite(tensor<double, 3, 3> A
...
+ *
+ * The convention followed is to place ones on the diagonal of the lower
+ * triangular factor.
+ * @param[in] A The matrix to factorize
+ * @return An LuFactorization object
+ * @see LuFactorization
  */
 template <typename T, int n>
-SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const tensor<T, n> b)
+SERAC_HOST_DEVICE constexpr LuFactorization<T, n> lu(const tensor<T, n, n>& A)
```
jamiebramwell: I would prefer `LU_factorize` or `factorize_LU` or something to be more explicit.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1269,54 +1431,10 @@ SERAC_HOST_DEVICE constexpr tensor<double, 3, 3> inv(const tensor<double, 3, 3>&
...
-    B[i] = B[i] / A[i][i];
-    for (int j = i - 1; j >= 0; j--) {
-      if (A[j][i] != 0.0) {
-        B[j] -= A[j][i] * B[i];
-      }
-    }
-  }
-
-  return B;
+  auto I = DenseIdentity<n>();
```
jamiebramwell, line 1435: 👍

jamiebramwell (approved):
Looks great! A few minor quibbles that I'm OK with either way. This should be really useful!

src/serac/numerics/functional/tensor.hpp
white238: I don't personally put my initials on TODO's normally, but I know others prefer to.  Please just do something like:

```
TODO - BT: some comment
```

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1185,43 +1202,188 @@ SERAC_HOST_DEVICE constexpr tensor<T, n> linear_solve(tensor<T, n, n> A, const t
...
+    }
+    x[i] = c / U[i][i];
+  }
   return x;
 }
 
+/**
+ * @brief Solves Ax = b for x using Gaussian elimination with partial pivoting
+ * @param[in] A The coefficient matrix A
+ * @param[in] b The righthand side vector b
```
white238, line 1321:
```suggestion
 * @param[in] b The right hand side vector b
```

src/serac/numerics/functional/tensor.hpp
btalamini: I'm happy to follow whatever convention there is. Removing the timestamp and formatting as in Chris's comment.

src/serac/numerics/functional/tensor.hpp
btalamini: I generally prefer verbing function names as well, with the exception of well understood math operations (eg `exp(A)` instead of `exponentiate(A)`). The LU factorization is a border case in my mind: `lu(A)` is fairly obvious in the context of tensors, but it isn't universal notation. If this thing were sitting in a more obvious context, (say `linear_algebra::lu(A)`) I'd argue to keep what I have, but it's just in the `serac` namespace. I've changed to `factorize_lu` - does this work?

src/serac/numerics/functional/tensor.hpp
btalamini: I agree with this style, but "righthand" and "lefthand" are used throughout `tensor.hpp`, and I think consistency is more important than the whitespace. I would prefer to leave it as is.

PR #693, opened 2022-05-12
Suggest new projection/install tree for spack to avoid collisions

adrienbernede:
I just ran into this:

```
==> Error: Install prefix collision for blt-0.4.1-bnajcsvekqqmr7pub7qx4iwshjni2iib
Prefix directory /Users/adrien/Work/Repos/serac_libs/apple-clang-13.1.6/blt-0.4.1 already used by another installed spec.
[ERROR: failure of spack install/dev-build]
```

Which confirms my opinion that the install tree should be more strict.

Note: with spack@0.18.0 adding full hash everywhere and in particular hashing the packages.py, it will be easier to have collisions.

- [ ] In CI this may very well cause a lot of new installations, not sure about that though -> Requires to run the full_workflow on Gitlab.
- [ ] See if what would be the consequence regarding the already installed dependencies if we need to update them.


scripts/spack/configs/config.yaml
```diff
@@ -19,7 +19,7 @@ config:
   install_tree:
     root: $spack/..
     projections:
-     all: "${COMPILERNAME}-${COMPILERVER}/${PACKAGE}-${VERSION}"
+      all: "${ARCHITECTURE}/${COMPILERNAME}-${COMPILERVER}/${PACKAGE}-${VERSION}-${HASH}"
```
white238: `ARCHITECTURE` is handled by the build scripts and this will duplicate that

adrienbernede: Removed

white238 (approved):

samuelpmishLLNL (approved):

jamiebramwell (approved):

adrienbernede:
I'll take those approvals as a reminder of the check I want to do before merging. ;)

white238:
> I'll take those approvals as a reminder of the check I want to do before merging. ;)

I actually just tested it on both our rz configs and there was no difference other than the hash's now slowing up in the host-configs.  Which like makes me slightly sad due to clutter, but I suppose I can live with it. It will be rough.

white238:
Feel free to merge.  I am going to do a rebuild of the TPLs very soon.

adrienbernede:
OK then ;)

PR #699, opened 2022-05-13
Add +asan variant to MFEM and Sundials Spack packages vendored into Serac

goxberry:
This pull request adds an `+asan` variant to the MFEM and Sundials Spack packages vendored into Serac, as part of upstreaming some Spack package tweaks from LiDO to Serac.

### DIFF
--- a/scripts/spack/packages/mfem/package.py
+++ b/scripts/spack/packages/mfem/package.py
@@ -164,45 +164,17 @@ class Mfem(Package, CudaPackage, ROCmPackage):
     variant('miniapps', default=False,
             description='Build and install miniapps')
 
-    # SERAC EDIT BEGIN - add ASan variant for memory debugging
-    # Custom variant not yet in upstream Spack; see
-    # https://github.com/spack/spack/pull/30210 for inspiration. Eventually,
-    # this variant should be replaced with a feature in upstream Spack that
-    # enables targeted use of ASan flags. The motivation for this variant is
-    # to inject ASan flags in this package, but not in its dependents;
-    # injecting ASan flags into specs will propagate them to MFEM's
-    # dependencts, which can result in undesirable behavior (e.g., building
-    # CMake with ASan flags).
+    # SERAC EDIT BEGIN - add AddressSanitizer variant for memory debugging
     variant('asan', default=False, description='Add Address Sanitizer flags')
 
-    # ASan is only supported by GCC and (some) LLVM-derived
-    # compilers. There's no convenient spec syntax for negating sets
-    # of compilers -- in this case, the conflicts arise with compilers
-    # that aren't gcc, clang, or apple-clang.
-    #
-    # The preferred approach taken by upstream Spack as of upstream
-    # commit 24c01d5 is to raise an exception within a package stage
-    # (e.g., xios does so in its install stage, pfunit does so in its
-    # setup_build_environment stage, wrf does so in its configure
-    # stage, elemental does so in its cmake_args stage).
-    #
-    # The trouble with this approach in isolation is that the
-    # concretizer can't detect those conflicts, so the exception is
-    # raised after building all of a package's dependents. Some of the
-    # more likely conflicts are listed here to enable
-    # concretization-time conflict detection; the list of compilers in
-    # the loop is every compiler listed in the spack.compilers package
-    # (https://spack.readthedocs.io/en/latest/spack.compilers.html)
-    # except gcc, clang, and apple-clang, to err on the conservative side.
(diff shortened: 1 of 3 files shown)


scripts/spack/packages/mfem/package.py
```diff
@@ -164,6 +164,55 @@ class Mfem(Package, CudaPackage, ROCmPackage):
     variant('miniapps', default=False,
             description='Build and install miniapps')
 
+    # Custom variant not yet in upstream Spack; see
```
white238: Any custom changes on our end need to be marked like this:

```
// SERAC EDIT START - <reason>
whatever change
// SERAC EDIT END
```
This makes it possible for us in the future to know whether changes in their package need to be moved forward when we update.

scripts/spack/packages/mfem/package.py
```diff
@@ -164,6 +164,55 @@ class Mfem(Package, CudaPackage, ROCmPackage):
...
+    # https://github.com/spack/spack/pull/30210 for inspiration. Eventually,
+    # this variant should be replaced with a feature in upstream Spack that
+    # enables targeted use of ASan flags. The motivation for this variant is
+    # to inject ASan flags in this package, but not in its dependents;
+    # injecting ASan flags into specs will propagate them to MFEM's
+    # dependencts, which can result in undesirable behavior (e.g., building
+    # CMake with ASan flags).
+    variant('asan', default=False, description='Add Address Sanitizer flags')
+
+    # ASan is only supported by GCC and (some) LLVM-derived
```
tepperly: How about:
```python
# ASan is the abbreviated name for memory error checking capabilities in modern compilers such 
# as GCC and some LLVM-derived compilers. When compiled with ASan enabled, the generated 
# code can detect numerous memory errors such as use after free, heap overflows, stack overflows,
# use outside of scope, and memory leaks.
```

scripts/spack/packages/mfem/package.py
goxberry: From looking at [other parts of `scripts/spack/packages/mfem/package.py`](https://github.com/LLNL/serac/blob/affb5793a6421abdde9fd29e5892e29c96df84bd/scripts/spack/packages/mfem/package.py#L52), would

```python
# SERAC EDIT BEGIN - reason
pass # <--- changes would go here
# SERAC EDIT END
```

work also?

white238: Yes please use the proper comment syntax given that this is a python file. I just quickly typed that up w/o thinking.

scripts/spack/packages/mfem/package.py
goxberry: Do you want that paragraph to replace just the first line or the entire block comment?

scripts/spack/packages/mfem/package.py
goxberry: Added to MFEM package in [`f4ae4e62`](https://github.com/LLNL/serac/pull/699/commits/f4ae4e62b5bb0e944ab92d352f4fedde82f39e14) and to Sundials package in [`008b116f`](https://github.com/LLNL/serac/pull/699/commits/008b116f0ff7894596f10747082a0e77febfbbf1).

scripts/spack/packages/mfem/package.py
white238: I would prefer a small comment. I personally find large comments to be distracting.

goxberry: Made comments briefer in [`88a20795`](https://github.com/LLNL/serac/pull/699/commits/88a207950b448cdb2e3feb81d359e2a4218186d6).

white238 (approved):

white238:
Should we forward the variants down the chain from the serac package?

goxberry:
I don't have a strong opinion either way. Happy to make whatever changes the project would like to that effect.

white238:
Is there a reason you can think of why you would not want it passed down the chain?  We already do this for the debug variant due to it being useful.

jamiebramwell (approved):

goxberry:
I can't think of a particularly compelling reason other than maybe the 2x instrumentation overhead for those libraries. In this case, the variant forwarding is limited, so it's not a build issue in the way that injecting environment variables into specs spreads throughout the DAG.

white238:
I say just merge... if its a problem in the future it is very easy to add.

goxberry:
Added `+asan` variant propagation in [`176bebf5`](https://github.com/LLNL/serac/pull/699/commits/176bebf5f023bfbcbc8243fd0ccae8f5e84f3f64).

PR #703, opened 2022-05-23
Refactor the `FiniteElementVector` class 

jamiebramwell:
This work will hopefully remove the need for physics module developers to manually sync the true vector (`HypreParVector`) and local vector (`ParGridFunction`) views of the finite element state data.

This refactor includes the following:
1. Makes `FiniteElementVector` inherit from `mfem::HypreParVector`. This is because the `FiniteElementState` is now semantically only a true vector and not a grid function.
2. Removes the need for `MaybeOwningPointer` in the `StateManager` class as the `FiniteElementState` objects no longer have `ParGridFunction` members that can either be managed by Sidre or not.
3. Removes optionally outputing formats other than sidre. This significantly cleans the interface as we require the sidre dependency anyway.
4. Fixes a memory bug in the `ThermalSolid` module and adds a missing cmake dependency in boundary condition.


samuelpmishLLNL:
Thanks for looking into this Jamie! I think this is an improvement already, but I have a few conceptual questions:

- do we still need the `FiniteElementVector::coefficient()` and `FiniteElementVector::vectorCoefficient()` member functions? grepping through, I only saw `coefficient()` used to set up `TransformedScalarCoefficient`-- an old hack to introduce coupling between physics. I didn't find any places where `vectorCoefficient()` is used.

- it seems like `.vector()` is used everywhere, does it make sense to just make `FiniteElementVector` itself an `mfem::Vector` with some additional functionality (projection operations, what else)?

- could we discuss the memory ownership of the different member variables (a lot use of this `MaybeOwningPointer`) on runs and restarts? I don't understand what's going on here very well.

samuelpmishLLNL:
One other thought: mfem uses the term "project" everywhere, but neglects to clarify what it actually means (least squares projection is what I originally imagined). When we eventually document this class on readthedocs, we should include some images that clarify that for a given mesh,
basis_functions_1D

when mfem says "project" they mean the following (sample a function at the nodes): 
sample_function

not to be confused with using a least square projection:
fit_to_function

jamiebramwell:
1) Probably not. I'll remove this.
2) Yes - this is very much a WIP as I work through the problems. I'm in the process of implementing all the functions needed to remove the `.vector()` calls. I'm playing around with the idea of making `FiniteElementVector` a vector itself to making the interface with pure MFEM objects cleaner, but I need to write a bunch of overloads that implement the correct synchronization calls.
3) Yeah, I should give a talk on this at a meeting in the very near future. Although Josh did write some nice docs on the issue here: https://serac.readthedocs.io/en/latest/sphinx/dev_guide/state_manager.html

src/serac/infrastructure/tests/serac_error_handling.cpp
```diff
@@ -112,18 +112,7 @@ TEST(SeracErrorHandling, BcRetrieveVecCoef)
...
-  axom::sidre::DataStore datastore;
-  serac::StateManager::initialize(datastore, "invalid_output_type");
-  serac::StateManager::setMesh(mesh::refineAndDistribute(buildDiskMesh(1000)));
-  ThermalConduction physics(1, ThermalConduction::defaultQuasistaticOptions());
-  // Try a definitely wrong number to ensure that an invalid output type is detected
-  EXPECT_THROW(physics.initializeOutput(static_cast<OutputType>(-7), ""), SlicErrorException);
-}
-
-TEST(SeracErrorHandling, InvalidCmdlineArg)
+TEST(serac_error_handling, invalid_cmdline_arg)
```
samuelpmishLLNL: I think we're trying to avoid `snake_case` test names now

edit: see https://github.com/LLNL/serac/pull/715

src/serac/physics/solid.cpp
```diff
@@ -516,27 +504,24 @@ const FiniteElementState& Solid::solveAdjoint(FiniteElementDual& adjoint_load,
...
 
-  adjoint_load.initializeTrueVec();
-
   // note: The assignment operator must be called after the copy constructor because
   // the copy constructor only sets the partitioning, it does not copy the actual vector
   // values
-  mfem::HypreParVector adjoint_load_vector(adjoint_load.trueVec());
-  adjoint_load_vector = adjoint_load.trueVec();
+  mfem::HypreParVector adjoint_load_vector(adjoint_load);
+  adjoint_load_vector = adjoint_load;
```
samuelpmishLLNL: was this redundant initialization pattern addressed by https://github.com/mfem/mfem/pull/2639 ?

samuelpmishLLNL:
the `.gridFunction(...)` member function gets used like
```cpp
displacement_.gridFunction(displacement_grid_function_);
```
where it's not really clear if the `mfem::GridFunction` is setting new values for the displacement field or vice versa.

Also, the function `gridFunction()` seems to just be an alias for `distributeSharedDofs()`, so would it make sense to get rid of one or the other?

samuelpmishLLNL:
I think that the name of these functions might be outdated:

```cpp
void FiniteElementDual::initializeTrueVec(const mfem::ParGridFunction& grid_function) { grid_function.ParallelAssemble(*this); }
```

```cpp
void FiniteElementState::initializeTrueVec(const mfem::ParGridFunction& grid_function) { grid_function.GetTrueDofs(*this); }
```

since it reads like a setter for a member variable, but the true vector is basically the entire FiniteElementVector itself. Would 

```cpp
void FiniteElementDual::operator=(const mfem::ParGridFunction& grid_function) { grid_function.ParallelAssemble(*this); }
void FiniteElementState::operator=(const mfem::ParGridFunction& grid_function) { grid_function.GetTrueDofs(*this); }
```

be more appropriate (or `SetFromGridFunction`, etc)

src/serac/physics/thermal_solid.cpp
```diff
@@ -13,15 +13,40 @@ namespace serac {
...
-      therm_solver_(order, therm_options, name, pmesh),
+      // Note that the solid solver must be constructed before the thermal solver as it mutates the mesh node grid
+      // function
       solid_solver_(order, solid_options, GeometricNonlinearities::On, FinalMeshOption::Deformed, name, pmesh),
+      therm_solver_(order, therm_options, name, pmesh),
       temperature_(therm_solver_.temperature()),
       velocity_(solid_solver_.velocity()),
       displacement_(solid_solver_.displacement())
 {
+  std::cout << "DURING CONSTRUCTION" << std::endl;
```
white238: debug messages.

jamiebramwell:
> the `.gridFunction(...)` member function gets used like
> 
> ```c++
> displacement_.gridFunction(displacement_grid_function_);
> ```
> 
> where it's not really clear if the `mfem::GridFunction` is setting new values for the displacement field or vice versa.
> 
> Also, the function `gridFunction()` seems to just be an alias for `distributeSharedDofs()`, so would it make sense to get rid of one or the other?

Does `void fillGridFunction(mfem::ParGridFunction& grid_func)` and `mfem::ParGridFunction& computeGridFunction()` make sense?

jamiebramwell:
> 
> Does `void fillGridFunction(mfem::ParGridFunction& grid_func)` and `mfem::ParGridFunction& computeGridFunction()` make sense?

On second thought, maybe  `void fillGridFunction(mfem::ParGridFunction& grid_func)` and `mfem::ParGridFunction& gridFunction()`  ?

jamiebramwell:
@samuelpmishLLNL , let me know what you think of this latest commit. I think it makes it more readable per your suggestions.

samuelpmishLLNL:
> @samuelpmishLLNL , let me know what you think of this latest commit. I think it makes it more readable per your suggestions.

This is looking a lot clearer!

One last hair to split: I don't think we need both of `FiniteElementState::operator=()` and `FiniteElementState::setFromGridFunction()`, since they are both aliases of the same thing. I'd stick with the latter (named function) for clarity.

jamiebramwell:
> > @samuelpmishLLNL , let me know what you think of this latest commit. I think it makes it more readable per your suggestions.
> 
> This is looking a lot clearer!
> 
> One last hair to split: I don't think we need both of `FiniteElementState::operator=()` and `FiniteElementState::setFromGridFunction()`, since they are both aliases of the same thing. I'd stick with the latter (named function) for clarity.

Yeah, I left the helper function to avoid having to explicitly write constructors for the derived classes. The question boils down to would it be better to have this redundant function, or have explicit constructors in the derived classes? Any thoughts @samuelpmishLLNL @btalamini @white238 ?

jamiebramwell:
To be specific, this is due to the error:

```
error: implicitly-declared ‘serac::FiniteElementState::FiniteElementState(const serac::FiniteElementState&)’ is deprecated [-Werror=deprecated-copy]
```

jamiebramwell:
> 

...I need to actually read comments. I was thinking you were advocating for the operator option. 🤦‍♀️

jamiebramwell:
OK, I think this is good to go now!

white238:
> > > @samuelpmishLLNL , let me know what you think of this latest commit. I think it makes it more readable per your suggestions.
> > 
> > 
> > This is looking a lot clearer!
> > One last hair to split: I don't think we need both of `FiniteElementState::operator=()` and `FiniteElementState::setFromGridFunction()`, since they are both aliases of the same thing. I'd stick with the latter (named function) for clarity.
> 
> Yeah, I left the helper function to avoid having to explicitly write constructors for the derived classes. The question boils down to would it be better to have this redundant function, or have explicit constructors in the derived classes? Any thoughts @samuelpmishLLNL @btalamini @white238 ?

I agree with @samuelpmishLLNL . We should have one way of doing things.  Also overloading `operator=` to do implicit conversions to me makes the code harder to read.

```
serac::FiniteElementState state;
mfem:SomeRandomClass foo;
state = foo;  //hard to determine what is going on due to conciseness
```

src/serac/numerics/expr_template_impl.hpp
```diff
@@ -76,11 +76,15 @@ using enable_if_mfem_vec = std::enable_if_t<std::is_base_of_v<mfem::Vector, std:
 template <typename vec>
 auto index(vec&& v, const int idx)
 {
-  if constexpr (std::is_same_v<std::decay_t<vec>, mfem::HypreParVector>) {
+  constexpr bool vec_is_a_HypreParVector = std::is_same_v<std::decay_t<vec>, mfem::HypreParVector> ||
```
white238, line 79: 👍 +1

src/serac/numerics/expr_template_impl.hpp
```diff
@@ -76,11 +76,15 @@ using enable_if_mfem_vec = std::enable_if_t<std::is_base_of_v<mfem::Vector, std:
...
+                                           std::is_base_of_v<mfem::HypreParVector, std::decay_t<vec>>;
+
+  if constexpr (vec_is_a_HypreParVector) {
     return static_cast<const double*>(v)[idx];
   }
 
-  if constexpr (!std::is_same_v<std::decay_t<vec>, mfem::HypreParVector>) {
+  constexpr bool vec_isnt_a_HypreParVector = !vec_is_a_HypreParVector;
```
white238: Did this not work w/o pulling it out? I prefer leaving boolean values positive then use the `!` for readability.

samuelpmishLLNL: Either one should work.

I think this was one of my changes-- the motivation was that expressions like `std::is_same_v<std::decay_t<vec>, mfem::HypreParVector>`  are a little heavy on the type-trait stuff, so giving it a plain name was preferable. The thing I like about
```cpp
if (vec_isnt_a_HypreParVector) { ... }
```
is that it just reads like an english sentence.

---------
I wish C++ supported boolean operators on types, so we could just write
```cpp
if (vec != HypreParVector) { ... }
```

samuelpmishLLNL (approved):

jamiebramwell:
> I think that the name of these functions might be outdated:
> 
> ```c++
> void FiniteElementDual::initializeTrueVec(const mfem::ParGridFunction& grid_function) { grid_function.ParallelAssemble(*this); }
> ```
> 
> ```c++
> void FiniteElementState::initializeTrueVec(const mfem::ParGridFunction& grid_function) { grid_function.GetTrueDofs(*this); }
> ```
> 
> since it reads like a setter for a member variable, but the true vector is basically the entire FiniteElementVector itself. Would
> 
> ```c++
> void FiniteElementDual::operator=(const mfem::ParGridFunction& grid_function) { grid_function.ParallelAssemble(*this); }
> void FiniteElementState::operator=(const mfem::ParGridFunction& grid_function) { grid_function.GetTrueDofs(*this); }
> ```
> 
> be more appropriate (or `SetFromGridFunction`, etc)

As Axom doesn't allow negative bools, I say we use `!vec_is_a_hypre_par_vector`.

src/serac/physics/coefficients/sensitivity_coefficients.hpp
```diff
@@ -58,12 +58,16 @@ class ShearSensitivityCoefficient : public mfem::Coefficient {
    */
   FiniteElementState& displacement_;
 
+  mfem::ParGridFunction displacement_grid_function_;
```
white238: Need doxygen

jamiebramwell: Even better! Deleting this!

jamiebramwell:
@btalamini could you take a look at this? Thanks!

btalamini:
I'm taking a look now. I don't know if I have enough context to give a very useful review, but I'll try.

jamiebramwell:
> I'm taking a look now. I don't know if I have enough context to give a very useful review, but I'll try.

Let me know if you want me to walk you through the changes.

src/serac/physics/solid.cpp
```diff
@@ -47,21 +46,17 @@ Solid::Solid(int order, const SolverOptions& options, GeometricNonlinearities ge
...
-    displacement_.trueVec()         = 0.0;
-    velocity_.trueVec()             = 0.0;
-    adjoint_displacement_.trueVec() = 0.0;
-  } else {
-    displacement_.initializeTrueVec();
-    velocity_.initializeTrueVec();
-    adjoint_displacement_.initializeTrueVec();
+    displacement_         = 0.0;
+    velocity_             = 0.0;
+    adjoint_displacement_ = 0.0;
```
btalamini, line 59: This is much better for physics module authors! :)

src/serac/physics/thermal_conduction.cpp
```diff
@@ -56,6 +56,8 @@ ThermalConduction::ThermalConduction(int order, const SolverOptions& options, co
   // Default to constant value of 1.0 for density and specific heat capacity
   cp_  = std::make_unique<mfem::ConstantCoefficient>(1.0);
```
btalamini, line 58: Is the purpose of these default values to make these parameters essentially optional (since quasi-static problems don't need them)? If so, I suggest putting this in the comment. The current comment repeats what the code already says and doesn't explain intent.

src/serac/physics/thermal_conduction.cpp
```diff
@@ -56,6 +56,8 @@ ThermalConduction::ThermalConduction(int order, const SolverOptions& options, co
...
   rho_ = std::make_unique<mfem::ConstantCoefficient>(1.0);
+
+  temperature_ = 2.0;
```
btalamini: This is a mysterious default value. A comment is probably in order here.

btalamini (approved):
Looks good, the clarity of the physics modules is definitely improved.

src/serac/physics/thermal_conduction.cpp
jamiebramwell: Yep. I'll add a comment.

src/serac/physics/thermal_conduction.cpp
jamiebramwell: This is a mystery to me as well! I bet it's leftover from debugging.

PR #706, opened 2022-05-25
performance refactor: batched kernel implementations

samuelpmishLLNL:
The main point of this PR is to change the implementation of the core finite element kernels from (pseudocode, using heat equation as an example):

```cpp
for (auto element : mesh) {

  // load the nodal temperatures for this element
  auto element_temperatures = temperatures[element];

  // integration loop
  for (auto quadrature_point : integration_rule) {

    // load the isoparametric coordinates, physical coordinates, and jacobian 
    // associated with this quadrature point
    auto [xi, x, J] = integration_point_data[quadrature_point];

    // compute the values of temperature, temperature gradient at this quadrature point
    auto [u, du_dx] = trial_space::interpolate(element_temperatures, xi, J);

    // evaluate the user's constitutive model at this quadrature point
    auto [source, flux] = qf(x, u, du_dx);

    // integrate the source and flux terms against shape functions and their derivatives (respectively)
    // to get the element residual contribution from this quadrature point
    element_residual += test_space::integrate(source, flux, xi, J);

  }

}
```

to 

```cpp
for (auto element : mesh) {

  // load the nodal temperatures for this element
  auto element_temperatures = temperatures[element];

  // load all the physical coordinates and jacobians
  // for the quadrature points associated with this element
  auto [x, J] = integration_point_data[element];

  // compute the values of temperature and temperature gradient at each quadrature point
  auto [u, du_dx] = trial_space::batch_interpolate(element_temperatures, xi, J);

  // evaluate the user's constitutive model at this quadrature point
  auto [sources, fluxes] = batch_apply(qf, x, u, du_dx);

  // integrate the source and flux terms against shape functions and their derivatives (respectively)
  // to get the complete element residual from all quadrature points
  element_residual += test_space::batch_integrate(sources, fluxes, xi, J);

}
```

that is, to implement functions that perform batched interpolation and integration (acting on all of the quadrature points at once, rather than one at a time). 

The main reason for this change is that computing these quantities in a batched way admits a more efficient algorithm that exploits the tensor product structure of the shape functions and quadrature rules to reduce the total amount of computation performed. To be clear: this more efficient algorithm is not an innovation by serac -- our initial kernel implementation was naive, so its performance was 

To make this change possible, this PR requires that specializations of the `finite_element` class template implement two functions:

- `finite_element< geometry, function_space >::interpolate(element_values, jacobians, quadrature_rule)` that takes in the values associated with an element, and returns an array of the interpolated values and derivatives for each quadrature point. ("E->Q")

- `finite_element< geometry, function_space >::integrate(sources_and_fluxes, jacobians, quadrature_rule, element_residual)` that takes in an array of the source and flux terms at each quadrature point, and numerically integrates them against the shape functions and shape function derivatives (respectively), adding the computed residuals to the last argument:

```math
\begin{align*}
R_I \coloneqq \int \phi_I \; s + \phi_{I,i} \; f_i \; \text{d}x
\end{align*}
```

- The kernel for computing element gradients reuses `finite_element< ... >::integrate` by observing that the boxed expression below has the same form as the residual expression above:

```math
\begin{align*}
R_I &\coloneqq \int \phi_I \; s + \phi_{I,i} \; f_i \; \text{d}x \\ \\
K_{IJ} &\coloneqq \frac{\partial R_I}{\partial U_J} \\ \\
&= \int \phi_I \; \frac{\partial s}{\partial U_J}  + \phi_{I,i} \; \frac{\partial f_i}{\partial U_J} \;  \text{d}x \\ \\
&= \int \phi_I \;\big(
\frac{\partial s}{\partial u} \frac{\partial u}{\partial U_J} + 
\frac{\partial s}{\partial u_{,k}} \frac{\partial u_{,k}}{\partial U_J}
\big) + \phi_{I,i} \; \big(
\frac{\partial f_i}{\partial u} \frac{\partial u}{\partial U_J} + 
\frac{\partial f_i}{\partial u_{,k}} \frac{\partial u_{,k}}{\partial U_J}
\big)\; \text{d}x \\ \\
&= \int \phi_I \;\big(
\frac{\partial s}{\partial u} \phi_J + 
\frac{\partial s}{\partial u_{,k}} \phi_{J,k}
\big) + \phi_{I,i} \; \big(
\frac{\partial f_i}{\partial u} \phi_J + 
\frac{\partial f_i}{\partial u_{,k}} \phi_{J,k}
\big) \; \text{d}x \\ \\
\Aboxed{K_{IJ} &= \int \phi_I \; \big(
d^{00} \; \phi_J + 
d^{01}_k \; \phi_{J,k}
\big) + \phi_{I,i} \; \big(
d^{10}_i \; \phi_J + 
d^{11}_{ik} \; \phi_{J,k}
\big) \;  \text{d}x}
\end{align*}
```
with 
```math
\begin{align*}
s = d^{00} \; \phi_J + d^{01}_k \phi_{J,k}, \qquad f_i = d^{10}_i \; \phi_J + d^{11}_{ik} \; \phi_{J,k}
\end{align*}
```


cmake/thirdparty/SetupSeracThirdParty.cmake
```diff
@@ -419,35 +422,39 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
     #------------------------------------------------------------------------------
     # Remove exported OpenMP flags because they are not language agnostic
     #------------------------------------------------------------------------------
-    set(_props)
-    if( ${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.13.0" )
-        list(APPEND _props INTERFACE_LINK_OPTIONS)
-    endif()
-    list(APPEND _props INTERFACE_COMPILE_OPTIONS)
+    if (NOT SERAC_ENABLE_CODEVELOP)
```
white238, line 425: Not connected to this PR but ....

Sigh this section of code makes me so sad... I am hopefully revamping all the BLT targets to get rid of this.

src/serac/physics/tests/solid.cpp
```diff
@@ -322,11 +322,11 @@ void functional_parameterized_solid_test(double expected_disp_norm)
 
   // add some nonexistent body forces / tractions to check that
   // these parameterized versions compile and run without error
-  solid_solver.addBodyForce(DependsOn<0>{}, [](const auto& x, double /*t*/, auto /* bulk */) { return x * 0.0; });
+  // solid_solver.addBodyForce(DependsOn<0>{}, [](const auto& x, double /*t*/, auto /* bulk */) { return x * 0.0; });
```
white238: Is this planning on being reenabled or is it dead?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -498,25 +496,25 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     residual_->AddBoundaryIntegral(
         Dimension<dim - 1>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
         [this, traction_function](auto x, auto n, auto, auto, auto shape, auto... params) {
-          auto dp_dX = get<DERIVATIVE>(shape);
-          auto p     = get<VALUE>(shape);
+          auto p = get<VALUE>(shape);
 
-          auto def_grad           = I + dp_dX;
-          auto inv_trans_def_grad = inv(transpose(def_grad));
+          // auto dp_dX = get<DERIVATIVE>(shape);
```
white238: dead code?

src/serac/numerics/functional/tuple_arithmetic.hpp
```diff
@@ -202,6 +202,30 @@ SERAC_HOST_DEVICE auto promote_to_dual_when(const T& x)
   }
 }
 
+/**
+ * @tparam dualify specify whether or not the input should be made into its dual type
+ * @tparam T the type of the values passed in
+ * @tparam n how many values were passed in
+ *
+ * @brief a function that optionally (decided at compile time) converts a list of values to their dual types
```
white238: I kind of like the brief at the top. Is this the normal way we do it in the code?

src/serac/numerics/functional/tuple_arithmetic.hpp
```diff
@@ -202,6 +202,30 @@ SERAC_HOST_DEVICE auto promote_to_dual_when(const T& x)
...
+ * @param x the values to be promoted
+ */
```
white238, line 212: Missing return description

src/serac/numerics/functional/tuple.hpp
```diff
@@ -260,6 +260,46 @@ SERAC_HOST_DEVICE constexpr const auto& get(const tuple<T...>& values)
   }
 }
 
+/**
+ * @brief a function intended to be used for extracting the ith type from a tuple
+ * get<i>(my_tuple) returns a reference, so it requires additional postprocessing to
+ * extract the underlying type (without a reference)
```
white238:
```suggestion
 * @brief a function intended to be used for extracting the ith type from a tuple.
 * As opposed to `get<i>(my_tuple)` which returns a reference, so it requires additional
 * postprocessing to extract the underlying type (without a reference).
```

If this is true then i feel this is more understandable.

src/serac/infrastructure/profiling.cpp
```diff
@@ -20,12 +20,6 @@ std::optional<cali::ConfigManager> mgr;
 }  // namespace
 #endif
 
-/**
- * @brief initialization for the profiling libraries (currently adiak and caliper, when enabled)
- *
- * @param comm the mpi communicator
- * @param options string of options passed to caliper, when enabled
- */
 void initialize([[maybe_unused]] MPI_Comm comm, [[maybe_unused]] std::string options)
```
samuelpmishLLNL, line 23: I'm not sure what's going on here. My doxygen_docs target tells me that `comm` is documented twice (once in the header, once in the .cpp file), so I deleted this redundant comment block. Now, my local doxygen build says this is correct, but the CI job complains that this function is not documented 😩

src/serac/numerics/functional/tests/functional_basic_elasticity.cpp
```diff
@@ -0,0 +1,166 @@
...
+      },
+      *mesh);
+
+  check_gradient(residual, U);
+
+  delete test_fes;
+  delete trial_fes;
+}
+
+TEST(basic, weird_mixed_test_2D) { weird_mixed_test<1, 2>(); }
```
white238: I do like weird tests.. lol.

src/serac/numerics/functional/quadrature.hpp
```diff
@@ -33,7 +33,7 @@ struct QuadratureRule {
   tensor<double, n, dim> points;
 
   /// @brief Returns the number of points in the rule
-  constexpr std::size_t size() const { return n; }
+  SERAC_HOST_DEVICE constexpr size_t size() const { return n; }
```
white238: What is this if not `std::size_t`?

src/serac/physics/tests/solid.cpp
samuelpmishLLNL: these should be uncommented, thanks for reminding me!

src/serac/numerics/functional/tuple_arithmetic.hpp
samuelpmishLLNL: I've moved all the `@briefs` I could find to the top (except for the `@file` comments, where we seem to be consistently putting the `@brief` lower down)

src/serac/numerics/functional/tuple_arithmetic.hpp
samuelpmishLLNL: Are the `@return` entries required? I usually try to describe in words what is returned in the `@brief`s, and `doxygen` isn't warning about missing `@return` statements.

src/serac/numerics/functional/tuple.hpp
samuelpmishLLNL: I'll take a stab at improving the wording. Without punctuation on the first line, it does look like a confusing run-on sentence, which was not intended.

src/serac/numerics/functional/quadrature.hpp
samuelpmishLLNL: I think `size_t` is `std::size_t`'s ancestor from C. 

There are definitely a lot of integer type inconsistencies in serac that we need to get figured out eventually. I made an mfem issue to hopefully look into fixing a big part of the problem:
https://github.com/mfem/mfem/issues/3229

btalamini:
https://github.com/LLNL/serac/blob/e9bcc81719b322201e85f451b87981c6de112687/src/serac/numerics/functional/boundary_integral.hpp#L31-L36

It seems that the doxygen block doesn't align with the class template parameters. I realize this is not code you modified in this PR, but I just caught it now.

samuelpmishLLNL:
> https://github.com/LLNL/serac/blob/e9bcc81719b322201e85f451b87981c6de112687/src/serac/numerics/functional/boundary_integral.hpp#L31-L36
> 
> It seems that the doxygen block doesn't align with the class template parameters. I realize this is not code you modified in this PR, but I just caught it now.

Interesting, usually doxygen warns if the descriptions don't agree with the argument names. I wonder if it doesn't check consistency of template parameters?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -491,32 +489,19 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
-          // Needed to be able to switch between dual and double square root functions
-          using std::sqrt;
-
-          // Compute the updated area contribution using the Piola transformation
-          // dA = det(F) * norm(F^-T n_0)
-          auto area_correction = det(def_grad) * norm(dot(inv_trans_def_grad, n));
-
-          return -1.0 * traction_function(x + p, shape_normal, ode_time_point_, params...) * area_correction;
+          auto p = get<VALUE>(shape);
+          return -1.0 * traction_function(x + p, n, ode_time_point_, params...);
```
jamiebramwell, line 507: Does this mean we can still calculate tractions on a shape-displaced mesh, but they are integrated in the reference (unperturbed, X) configuration? If so, we should put in a note for that or maybe emit a warning if `p` is nonzero on the boundary.

src/serac/physics/tests/pch.hpp
```diff
@@ -0,0 +1,6 @@
+
```
jamiebramwell: Where is this used?

src/serac/numerics/functional/tests/performance/J2_material_with_AD.cpp
```diff
@@ -1,218 +0,0 @@
-// Copyright (c) 2019-2022, Lawrence Livermore National Security, LLC and
```
jamiebramwell, line 1: Is there any reason to save these performance tests? Ideally we would have a benchmarking test repo for stuff like this. Thoughts @white238 ?

src/serac/numerics/functional/tests/functional_basic_elasticity.cpp
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) 2019-2022, Lawrence Livermore National Security, LLC and
```
jamiebramwell: Maybe rename this file `functional_basic_vector_h1.cpp` or something like that? It really isn't testing elasticity.

src/serac/numerics/functional/tests/functional_basic_elasticity.cpp
jamiebramwell: I can verify these are in fact weird. 😆

src/serac/numerics/functional/tests/functional_basic_thermal.cpp
```diff
@@ -20,59 +20,27 @@
 #include "serac/numerics/functional/functional.hpp"
```
jamiebramwell: `functional_basic_scalar_h1`?

src/serac/numerics/functional/tests/functional_material_state_test.cpp
```diff
@@ -1,333 +0,0 @@
-// Copyright (c) 2019-2022, Lawrence Livermore National Security, LLC and
```
jamiebramwell, line 1: Why are these tests deleted?

src/serac/numerics/functional/functional.hpp
```diff
@@ -128,6 +128,7 @@ mfem::ParFiniteElementSpace* generateParFiniteElementSpace(mfem::ParMesh* mesh)
       break;
   }
 
+  // note: this leaks memory: `fec` is never destroyed, but how to fix?
```
jamiebramwell, line 131: I don't really think there's a good way around this without a wrapper like `FiniteElementState` that manages the collection and space memory.

src/serac/physics/solid_mechanics.hpp
samuelpmishLLNL: Good point, I'm not sure what the best thing is to do for this function. I did make a note in the doxygen that it doesn't really work with shape sensitivity.

src/serac/numerics/functional/tests/functional_material_state_test.cpp
samuelpmishLLNL: These were the tests for Josh's old version of the quadrature data container, which isn't used any more.

src/serac/physics/tests/pch.hpp
samuelpmishLLNL: I'll delete it. I still want to move to using precompiled headers, but it seems like our version of CMake on the LC machines is too old.

src/serac/numerics/functional/functional.hpp
```diff
@@ -247,8 +248,10 @@ class Functional<test(trials...), exec> {
     for (uint32_t i = 0; i < num_trial_spaces; i++) {
       auto ndof_per_trial_element =
           static_cast<size_t>(trial_space_[i]->GetFE(0)->GetDof() * trial_space_[i]->GetVDim());
-      element_gradients_[i] = ExecArray<double, 3, exec>(num_elements, ndof_per_test_element, ndof_per_trial_element);
-      bdr_element_gradients_[i] = allocateMemoryForBdrElementGradients<double, exec>(*trial_space_[i], *test_space_);
+      // element_gradients_[i] = ExecArray<double, 3, exec>(num_elements, ndof_per_test_element,
```
jamiebramwell: dead code

src/serac/numerics/functional/finite_element.hpp
```diff
@@ -125,6 +266,95 @@ struct QOI {
...
+
+  for (int k = 0; k < q; k++) {
+    tensor<double, dim, dim> J;
+    for (int row = 0; row < dim; row++) {
+      for (int col = 0; col < dim; col++) {
+        J[row][col] = jacobians(col, row, k);
+      }
+    }
+
+    if constexpr (f == Family::H1 || f == Family::L2) {
```
jamiebramwell, line 294: I would add a comment that these are the Piola transforms. Yes its obvious, but I figure it can't hurt to overdocument.

src/serac/numerics/functional/domain_integral.hpp
```diff
@@ -79,7 +79,7 @@ class DomainIntegral {
     if constexpr (exec == ExecutionSpace::CPU) {
       KernelConfig<Q, geometry, test, trials...> eval_config;
 
-      evaluation_ = EvaluationKernel{eval_config, J, X, num_elements, qf, qdata};
+      evaluation_ = EvaluationKernel{DerivativeWRT<-1>{}, eval_config, J, X, num_elements, qf, qdata};
```
jamiebramwell: What does `DerivativeWRT<-1>` indicate?

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -89,135 +125,31 @@ struct KernelConfig {
...
-  std::shared_ptr<QuadratureData<qpt_data_type> > data_;          ///< (optional) user-provided quadrature data
-};
-
 /**
  * @overload
  * @note evaluation kernel that also calculates derivative w.r.t. `I`th trial space
  */
 template <int I, int Q, Geometry geom, typename test, typename... trials, typename derivatives_type, typename lambda,
-          typename qpt_data_type>
+          typename qpt_data_type, int... j>
```
jamiebramwell: what is this new parameter `j`?

src/serac/numerics/functional/domain_integral.hpp
jamiebramwell: I'm confused. How is the `j` parameter in the template instantiation deduced?

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -230,6 +162,14 @@ struct EvaluationKernel<DerivativeWRT<I>, KernelConfig<Q, geom, test, trials...>
   {
   }
 
+  /// @overload
+  EvaluationKernel(DerivativeWRT<-1>, KernelConfig<Q, geom, test, trials...>, const mfem::Vector& J,
```
jamiebramwell: I'm concerned about possible conflicts with `j` and `J`.

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -316,17 +246,24 @@ struct EvaluationKernel<DerivativeWRT<I>, KernelConfig<Q, geom, test, trials...>
   std::shared_ptr<QuadratureData<qpt_data_type> > data_;     ///< (optional) user-provided quadrature data
 };
 
-template <int Q, Geometry geom, typename test, typename... trials, typename lambda, typename qpt_data_type>
-EvaluationKernel(KernelConfig<Q, geom, test, trials...>, const mfem::Vector&, const mfem::Vector&, int, lambda,
-                 std::shared_ptr<QuadratureData<qpt_data_type> >)
-    -> EvaluationKernel<void, KernelConfig<Q, geom, test, trials...>, void, lambda, qpt_data_type>;
+// template <int Q, Geometry geom, typename test, typename... trials, typename lambda, typename qpt_data_type>
```
jamiebramwell: dead code

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -316,17 +246,24 @@ struct EvaluationKernel<DerivativeWRT<I>, KernelConfig<Q, geom, test, trials...>
...
+// EvaluationKernel(KernelConfig<Q, geom, test, trials...>, const mfem::Vector&, const mfem::Vector&, int, lambda,
+//                 std::shared_ptr<QuadratureData<qpt_data_type> >)
+//    -> EvaluationKernel<void, KernelConfig<Q, geom, test, trials...>, void, lambda, qpt_data_type,
+//                        std::make_integer_sequence<int, sizeof...(trials)> >;
 
 template <int i, int Q, Geometry geom, typename test, typename... trials, typename derivatives_type, typename lambda,
```
jamiebramwell: I there a reason this is lower case `i` instead of `I`? I think `I` would be better for consistency.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -87,21 +87,22 @@ struct QFunctionArgument;
...
 };
 
 /// @overload
 template <int p, int c, int dim>
 struct QFunctionArgument<H1<p, c>, Dimension<dim>> {
-  using type = serac::tuple<tensor<double, c>, serac::zero>;  ///< what will be passed to the q-function
+  using type = serac::tuple<tensor<double, c>, tensor<double, c, dim>>;  ///< what will be passed to the q-function
 };
 
 template <int i, int dim, typename... trials, typename lambda>
```
jamiebramwell, line 99: `I` for consistency

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -112,6 +113,24 @@ template <int Q, Geometry g, typename test, typename... trials>
 struct KernelConfig {
 };
 
+template <typename lambda, int dim, int n, typename... T, int... I>
```
jamiebramwell, line 116: Isn't `I` what we call `j` in the boundary kernels?

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -22,11 +22,34 @@ struct finite_element<Geometry::Hexahedron, H1<p, c> > {
...
   static constexpr int  components = c;
   static constexpr int  dim        = 3;
+  static constexpr int  n          = (p + 1);
   static constexpr int  ndof       = (p + 1) * (p + 1) * (p + 1);
+  static constexpr int  order      = p;
 
+  static constexpr int VALUE = 0, GRADIENT = 1;
+  static constexpr int SOURCE = 0, FLUX = 1;
+
+  // TODO: remove this
```
jamiebramwell: Why can this not be removed now?

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -73,5 +96,385 @@ struct finite_element<Geometry::Hexahedron, H1<p, c> > {
     return dN;
     // clang-format on
   }
+
```
jamiebramwell, line 95: I know it's boiler plate for every element type, but it would be good to write some prose on the meaning of `B` and `G`

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -73,5 +96,385 @@ struct finite_element<Geometry::Hexahedron, H1<p, c> > {
...
+    static constexpr auto G_ = [=]() {
+      tensor<double, q, n> G{};
+      for (int i = 0; i < q; i++) {
+        G[i] = GaussLobattoInterpolationDerivative<n>(points1D[i]);
+      }
+      return G;
+    }();
+
+    __shared__ tensor<double, q, n> B;
```
jamiebramwell, line 334: @white238 , ideally these would be dynamically allocated. Do you know if umpire (or some other tool) can do this with a single CPU/CUDA implementation?

src/serac/numerics/functional/detail/hexahedron_Hcurl.inl
```diff
@@ -166,5 +248,612 @@ struct finite_element<Geometry::Hexahedron, Hcurl<p>> {
...
+
+    tensor<tuple<source_t, flux_t>, q * q * q> output;
+
+    for (int qz = 0; qz < q; qz++) {
+      for (int qy = 0; qy < q; qy++) {
+        for (int qx = 0; qx < q; qx++) {
+          tensor<double, 3> phi_j{};
+          tensor<double, 3> curl_phi_j{};
+
+          switch (dir) {
```
jamiebramwell, line 327: Yes. This is reminding me why I hate Hcurl!

samuelpmishLLNL: There's a different dof numbering that cleans this up considerably, but that's not the one mfem uses.

src/serac/numerics/functional/domain_integral.hpp
samuelpmishLLNL: > What does `DerivativeWRT<-1>` indicate?

-1 is used to indicate no differentiation. Before this, there were two separate kernel implementations (Evaluate w/ and w/o AD) that were almost identical. This combines their implementations into 1, and optionally turns off AD with the -1 value. I'm on the fence if this change is an improvement or not.

> I'm confused. How is the `j` parameter in the template instantiation deduced?

in the class template argument deduction guide, it gets set from the number of trial spaces (at the end): 

https://github.com/LLNL/serac/blob/e9bcc81719b322201e85f451b87981c6de112687/src/serac/numerics/functional/domain_integral_kernels.hpp#L262-L266

src/serac/numerics/functional/domain_integral_kernels.hpp
samuelpmishLLNL: if we put a variadic pack of integers in the kernel type itself, then we can break tuples apart without needing to define extra helper functions like we did for stuff like:

```cpp
template <typename... S, typename... T, int... i>
SERAC_HOST_DEVICE constexpr auto minus_helper(const tuple<S...>& x, const tuple<T...>& y,
                                              std::integer_sequence<int, i...>) {
  return tuple{get<i>(x) - get<i>(y)...};
}

template <typename... S, typename... T>
SERAC_HOST_DEVICE constexpr auto operator-(const tuple<S...>& x, const tuple<T...>& y) {
  static_assert(sizeof...(S) == sizeof...(T));
  return minus_helper(x, y, std::make_integer_sequence<int, static_cast<int>(sizeof...(S))>());
}
```

src/serac/numerics/functional/domain_integral_kernels.hpp
samuelpmishLLNL: the more recent changes to these kernels try to use `I` and `J` to be consistent with the LaTeX explanation of the kernels at the top of the PR, but unfortunately there are some collisions (e.g. should `J` be the index or shorthand for the jacobians). I can make another pass over this to make some of these names more consistent.

src/serac/numerics/functional/domain_integral.hpp
white238: We shouldn't be using magic numbers. Possibly one of these?

```
constexpr int NO_DERIVATIVE = -1;
DerivativeWRT<NO_DERIVATIVE>
```

or:

```
using NoDerivative = DerivativeWRT<-1>;
```

src/serac/infrastructure/profiling.cpp
white238: Looks like this was caught by @adayton1 8 months ago, is this not working?

https://github.com/LLNL/serac/blob/a300c9c98c631d5503428aff9f67dbd2cdc0df90/scripts/azure-pipelines/docs_ignore_regexs.txt#L5-L6

src/serac/numerics/functional/domain_integral_kernels.hpp
samuelpmishLLNL: I've changed `j...` to `indices...` to avoid confusion with `J`.

src/serac/numerics/functional/detail/hexahedron_H1.inl
samuelpmishLLNL: I wonder why doxygen isn't producing warnings about these functions

src/serac/numerics/functional/finite_element.hpp
samuelpmishLLNL: I'm not familiar with the right technical terminology here. Is each conditional branch a separate kind of "Piola Transform", or does this type of transformation just generally have the name Piola associated with it?

src/serac/infrastructure/profiling.cpp
samuelpmishLLNL: Confusingly, I didn't even change this source file but now the CI job passed. It's worrying that our CI jobs don't seem to have deterministic output.

src/serac/numerics/functional/boundary_integral.hpp
```diff
@@ -85,28 +85,28 @@ class BoundaryIntegral {
       evaluation_ = EvaluationKernel{eval_config, J, X, N, num_elements, qf};
 
       for_constexpr<num_active_trial_spaces>([this, num_elements, quadrature_points_per_element, &J, &X, &N, &qf,
-                                              eval_config](auto i) {
+                                              eval_config](auto index) {
```
jamiebramwell, line 88: 👍

jamiebramwell (approved):
Great work @samuelpmishLLNL !

white238 (approved):

PR #709, opened 2022-06-01
Feature/talamini/parameterize thermomechanics

btalamini:
This is hashing out a design for a thermomechanics module that can also handle design-parameterized materials. There are a few decisions to make about the interface and how we want to handle both parameterized and non-parameterized materials. Please add other any others who should be involved in the discussion as additional reviewers.


src/serac/numerics/functional/functional.hpp
```diff
@@ -89,6 +89,41 @@ struct Index {
...
 
+/**
+ * @brief create an mfem::ParFiniteElementSpace from one of serac's
+ * tag types: H1, Hcurl, L2
+ *
+ * @tparam function_space a tag type containing the kind of function space and polynomial order
+ * @param mesh the mesh on which the space is defined
+ */
+template <typename function_space>
+mfem::ParFiniteElementSpace* generateParFiniteElementSpace(mfem::ParMesh* mesh)
```
jamiebramwell, line 100: 👍 This will be really handy!

src/serac/physics/tests/parameterized_thermomechanics_example.cpp
```diff
@@ -0,0 +1,244 @@
...
+    const auto            trEg = tr(Eg);
+
+    const auto S = 2.0 * G * dev(Eg) + K * (trEg - 3.0 * alpha * (theta - theta_ref)) * I;
+    const auto P = dot(F, S);
+    return dot(P, transpose(F));
+  }
+};
+
+/// @brief Constructs an MFEM mesh of a hollow cylinder restricted to the first orthant
+mfem::Mesh build_hollow_quarter_cylinder(std::size_t radial_divisions, std::size_t angular_divisions,
```
jamiebramwell: I would put this with the other mesh utilities in `src`.

src/serac/physics/tests/parameterized_thermomechanics_example.cpp
```diff
@@ -0,0 +1,244 @@
...
+  std::cout << "exact area of the top surface: " << exact_area << std::endl;
+
+  std::cout << "average vertical displacement: " << qoi(simulation.displacement()) / area(simulation.displacement())
+            << std::endl;
+
+  double deltaT = 1.0;
+  std::cout << "expected average vertical displacement: " << alpha0 * deltaT * height << std::endl;
+
+  serac::FiniteElementDual adjoint_load(mesh, simulation.displacement().space(), "adjoint_load");
+  auto                     dqoi_du = get<1>(qoi(DifferentiateWRT<0>{}, simulation.displacement()));
```
jamiebramwell, line 172: I know `VALUE` and `DERIVATIVE` live in the solid module, but maybe they should be somewhere more accessible so we could use them here?

jamiebramwell: Additionally, using defined `constexpr` values for the parameter indices might make this more readable.

src/serac/physics/materials/parameterized_solid_functional_material.hpp
```diff
@@ -40,8 +40,8 @@ struct ParameterizedLinearIsotropic {
   SERAC_HOST_DEVICE auto operator()(const DispGradType& du_dX, const BulkType& DeltaK, const ShearType& DeltaG) const
   {
     constexpr auto I       = Identity<dim>();
-    auto           K       = K0 + DeltaK;
-    auto           G       = G0 + DeltaG;
+    auto           K       = K0 + get<0>(DeltaK);
```
jamiebramwell, line 43: Why is this needed? Wouldn't this break the AD?

jamiebramwell: (If I'm reading it right, it's throwing away the derivative part of the dual number)

src/serac/physics/thermal_mechanics_functional.hpp
```diff
@@ -0,0 +1,350 @@
...
+  }
+
+  /**
+   * @brief This is an adaptor class that makes a thermomechanical material usable by
+   * the thermal module, by discarding the solid-mechanics-specific information
+   *
+   * @tparam ThermalMechanicalMaterial the material model being wrapped
+   */
+  template <typename ThermalMechanicalMaterial>
+  struct ThermalMaterialInterface {
```
jamiebramwell, line 129: Should this be in the materials directory? I can see it going either here or there.

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
```diff
@@ -0,0 +1,397 @@
...
+  auto zeroVector = [](const mfem::Vector&, mfem::Vector& u) { u = 0.0; };
+
+  // Set the initial displcament and boundary condition
+  thermal_solid_solver.setDisplacementBCs(ess_bdr, zeroVector);
+  thermal_solid_solver.setDisplacement(zeroVector);
+
+  // Finalize the data structures
+  thermal_solid_solver.completeSetup();
+
+  // thermal_solid_solver.initializeOutput(serac::OutputType::VisIt, "thermal_mechanics_without_input_file_output");
```
jamiebramwell: Dead code

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
```diff
@@ -0,0 +1,397 @@
...
+  double dt = 1.0;
+  thermal_solid_solver.advanceTimestep(dt);
+
+  // thermal_solid_solver.outputState();
+
+  // Check the final displacement norm
+  EXPECT_NEAR(expected_norm, norm(thermal_solid_solver.displacement()), 1.0e-4);
+}
+
+// TODO: investigate this failing test
```
jamiebramwell, line 190: Is this still failing?

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
```diff
@@ -0,0 +1,397 @@
...
+  constexpr int p = 2;
+  // this is the small strain solution, which works with a loose enought tolerance
+  // TODO work out the finite deformation solution
+  double alpha       = 1e-3;
+  double L           = 8;
+  double delta_theta = 1.0;
+  serac::functional_test_shrinking_3D<p>(std::sqrt(L * L * L / 3.0) * alpha * delta_theta);
+}
+
+TEST(Thermomechanics, parameterized)
```
jamiebramwell, line 355: I'm not sure what this stub is a placeholder for.

src/serac/physics/tests/parameterized_thermomechanics_example.cpp
samuelpmishLLNL: `VALUE` and `DERIVATIVE` describe the meaning of the entries of the argument tuples that get passed to the q-function, whereas DifferentiateWRT is telling Functional which `FiniteElementState` to differentiate. 

We could have some constexpr integers that are used to describe the argument indices., but they'll be specific to an individual QoI, as each QoI Functional might depend on different things (or the same set of things, but in a different order).

src/serac/physics/materials/parameterized_solid_functional_material.hpp
samuelpmishLLNL: This is a little bit confusing. `get<>()` is for tuples, `get_gradient()` and `get_value()` are for `dual`. 

The highlighted code change is related to how `SolidFunctional` previously used to strip out the derivatives of parameters before passing them to the material q-functions. An example of this from the addMaterial function (before):

```cpp
        [this, material](auto /*x*/, auto& state, auto displacement, auto acceleration, auto... params) {
          auto du_dX   = get<DERIVATIVE>(displacement);
          auto d2u_dt2 = get<VALUE>(acceleration);
          auto stress  = material(state, du_dX, serac::get<VALUE>(params)...); // <----------
          ....
```

The problem was that in order to reuse the single-physics solvers (thermal and mechanics) in the thermomechanics module, they each need to know about the other primal field as a "parameter" (displacement and temperature, respectively). But, they also need the gradients of those primal fields to work, so we changed `SolidFunctional` to do this instead:

```cpp
        [this, material](auto /*x*/, auto& state, auto displacement, auto acceleration, auto... params) {
          auto du_dX   = get<DERIVATIVE>(displacement);
          auto d2u_dt2 = get<VALUE>(acceleration);
          auto stress  = material(state, du_dX, params...); // <----------
          ....
```

But that means that the previous materials that only use the value of the parameters now have to index into the tuple from the q-function argument explicitly to pull out the value.

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
samuelpmishLLNL: this is that "failing" test, which is currently commented out rather than completely deleted

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
samuelpmishLLNL: it is failing because there was a placeholder analytic solution that isn't quite what we want. In general though we don't (or at least I don't) know a good way to check for the "right" answer. The actual analytic solution to this problem will never be found by the thermomechanics module as-written since the operator-split approach doesn't actually enforce that mechanical and thermal equilibrium are simultaneously satisfied.

One possibility might be to phrase this as a time-dependent problem (applying the deformation gradually), and then it should approximate the "right" answer in the limit, as you take enough time-steps. That would make the test much more expensive though, I'm not sure what the right thing is to do with this test.

samuelpmishLLNL (approved):

src/serac/physics/thermal_mechanics_functional.hpp
btalamini: It's glue code between the materials and the physics module, which does make it tricky to assign a home. I was also conflicted, but I prefer the current arrangement because the glue code is a consequence of the operator split solver algorithm. Putting it with the materials will make material model developers confront the solver algorithm details. I'd prefer to let them focus on the physics and ignore how the solver will use their model.

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
btalamini: Sam and I worked on these together, and I voted to disable this test, for the reasons Sam gave above. I have a suggestion to replace it: we could manufacture a solution for the coupled problem, set the discrete temperature and displacements fields to that solution, and check that both residuals are zero. It's not a foolproof test, but it's not trivial, either. This will be easier to do once the patch test/manufactured solution PR comes through.

btalamini: Sam and I worked on these together, and I voted to disable this test, for the reasons Sam gave above. I have a suggestion to replace it: we could manufacture a solution for the coupled problem, set the discrete temperature and displacements fields to that solution, and check that both residuals are zero. It's not a foolproof test, but it's not trivial, either. This will be easier to do once the patch test/manufactured solution PR comes through.

samuelpmishLLNL: Looking over the other two tests: they pass, but don't really do anything (the solutions are practically trivial). I wonder: is it worth stealing some of the work from the patch-test branch to make them into more meaningful tests, or should we just try to get this PR in first and fix them in the patch-test PR?

btalamini: I'd prefer the latter so that we can see if patch testing plain mechanics/heat transfer reveals any issues, and also so that I can participate on the thermomechanics testing.

samuelpmishLLNL:
Also: I can't seem to remove the DRAFT tag, since it isn't my PR. @btalamini are you able to do that?

src/serac/physics/materials/parameterized_solid_functional_material.hpp
jamiebramwell: Ahhh, I missed that this got moved.

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
jamiebramwell: Agreed. This can wait for the patch test PR. I'd remove it here to avoid confusion.

src/serac/physics/tests/serac_thermal_mechanics_functional.cpp
jamiebramwell: Yeah, I would completely delete it in favor of a github issue.

jamiebramwell (approved):

PR #717, opened 2022-06-07
Update time in certain physics types

chapman39:


chapman39:
- adds a setCycle function to base physics
- updates time for certain physics types that weren't updating time in advanceTimestep
- @jamiebramwell [besides `t` not being set from inlet in `serac.cpp`](https://github.com/LLNL/serac/pull/717/files#diff-d9fef39355c0c509f4da9ffdabd95167b3b4cbe7f3955deba262e082ad5b9c29R207), I'm not sure if you'd want me to keep this as well as the `isRestart` stuff in `solid.cpp`

chapman39:
@white238 I tried looking for a way to get the time from inlet with something like `inlet["time"]` ([for this line we were talking about](https://github.com/LLNL/serac/pull/717/files#diff-d9fef39355c0c509f4da9ffdabd95167b3b4cbe7f3955deba262e082ad5b9c29R207)) and haven't gotten much results. I'm not sure if there's some way to view the entire dictionary of inlet, but I think that would be useful.

white238:
> @white238 I tried looking for a way to get the time from inlet with something like `inlet["time"]` ([for this line we were talking about](https://github.com/LLNL/serac/pull/717/files#diff-d9fef39355c0c509f4da9ffdabd95167b3b4cbe7f3955deba262e082ad5b9c29R207)) and haven't gotten much results. I'm not sure if there's some way to view the entire dictionary of inlet, but I think that would be useful.

Inlet only has access (or really knows how to access) part of Sidre.  When getting the time during a restart you should be able to access it via the `MFEMSidreDatacollection`

chapman39:
So I am now getting the time from `StateManager::load`. I changed its return value... let me know if you'd like a different approach.

In any case, it looks like time is one `dt` too far. I could subtract `t` by `dt`, which would resolve the issue, but I don't think this is going to always work since @jamiebramwell you mentioned that each cycle may have a different `dt` or something. Maybe something can be changed in the time step loop (line 222 of serac.cpp).

EDIT:
This issue has been resolved.

chapman39:
I changed `ti` to `cycle` in `serac.cpp` since that makes more sense to me.

white238:
@kennyweiss do you have any idea why this works when we never call the following function or maybe why it's not necessary?

https://github.com/LLNL/axom/blob/2ad90f2e4ec001241f41102992575bf07a6bee30/src/axom/sidre/core/MFEMSidreDataCollection.hpp#L474-L476

I just ran into this function and it seems like it should be a necessary thing. We do call `UpdateStateFromDS`

kennyweiss:
> @kennyweiss do you have any idea why this works when we never call the following function or maybe why it's not necessary?
> 
> https://github.com/LLNL/axom/blob/2ad90f2e4ec001241f41102992575bf07a6bee30/src/axom/sidre/core/MFEMSidreDataCollection.hpp#L474-L476
> 
> I just ran into this function and it seems like it should be a necessary thing. We do call `UpdateStateFromDS`

I think I need more context to disambiguate the "this" in your sentence. (i.e. what is working?)

The direct answer to your question might be that you're pushing the cycle/time to the datastore during `save` here: https://sourcegraph.com/github.com/LLNL/serac/-/blob/src/serac/physics/state/state_manager.cpp?L149-151

#### Background that might be helpful:
The `MFEMSidreDataCollection` has a `time,`  `dt` and `cycle` variable (by value) (from the `mfem::DataCollection` base class). The underlying `sidre::DataStore` has its own copy of these (also by value). It looks like (each of?) your `BasePhysics` have their own (by value) copy, so to synchronize, you need to push/pull those values to multiple places. 

The one that matters for checkpoint/restart is the one in the datastore. The one that matters in the code is likely the one in (each of) the physics class(es).

If the "Physics" don't have a notion of time/cycle that is independent of the overall simulation, it might be worth considering letting the state manage the cycle and time and having the physics access it from there.

white238:
I suppose I could be more clear..

I was confused how the time variables were getting to the datastore so that the correct values were there when restart, which they are.  I tracked it down.  `MFEMSidreDataCollection::Save` calls `MFEMSidreDataCollection::PrepareToSave` which then calls `MFEMSidreDataCollection::UpdateStateToDS`.

brain kicked in eventually.

Edit:  This does actually bring up a good point in your thorough comment.  We are setting time and cycle but not dt (via `MFEMSidreDataCollection::SetTimeStep`).  @jamiebramwell, does the `dt` every change from the input file value?

white238 (dismissed):

jamiebramwell:
> 

Yes - `dt` could change from the input deck value.

jamiebramwell:
> Edit: This does actually bring up a good point in your thorough comment. We are setting time and cycle but not dt (via `MFEMSidreDataCollection::SetTimeStep`). @jamiebramwell, does the `dt` every change from the input file value?

Yes - `dt` can change from the input deck value.

jamiebramwell (approved):

white238 (approved):

PR #718, opened 2022-06-08
Use camp tuple and methods

adayton1:
* Replaces the following in serac with camp implementations that are fully variadic:
** tuple
** tuple_size
** tuple_element
** several methods relating to tuples


adayton1:
It appears that the issue I was running into comes from using camp::make_tuple instead of serac::make_tuple. Rather than holding this merge request up, I decided to continue using serac::make_tuple and determine how to replace it in a follow on pull request.

adayton1:
I think the failures are from using public packages without the patch I have for camp. @white238, do I need to run the build tpls script without any arguments to install new public packages?

src/serac/numerics/functional/tuple.hpp
```diff
@@ -13,216 +13,32 @@
...
-    return values.v5;
-  }
-  if constexpr (i == 6) {
-    return values.v6;
-  }
-  if constexpr (i == 7) {
-    return values.v7;
-  }
+  return camp::get<i>(values);
 }
```
samuelpmishLLNL, line 50: What is the motivation for wrapping `get` in the serac namespace when the other parts of the tuple ecosystem use the camp namespace?

src/serac/numerics/functional/tuple.hpp
```diff
@@ -566,116 +358,9 @@ SERAC_HOST_DEVICE constexpr auto operator*(const double a, const tuple<T...>& x)
...
- * @tparam T the types stored in the tuple
- * @param out the ostream to write the output to
- * @param A the tuple of values
- * @brief print a tuple of values
- */
-template <typename... T>
-auto& operator<<(std::ostream& out, const serac::tuple<T...>& A)
-{
-  return print_helper(out, A, std::make_integer_sequence<size_t, sizeof...(T)>());
-}
```
samuelpmishLLNL, line 600: does `camp::tuple` support `operator<<` already, or are we getting rid of this feature?

src/serac/numerics/functional/tuple.hpp
adayton1: The camp::get signature is more generic, so I was worried there might be conflicts with the serac overload of get that takes a variant if we just used the camp version.

```
// serac overload
template <int i, typename... T>
SERAC_HOST_DEVICE constexpr auto& get(camp::tuple<T...>& values);

// camp overload
template <camp::idx_t index, class Tuple>
CAMP_HOST_DEVICE constexpr auto& get(const Tuple& t) noexcept
```

I would like to revamp camp's get functions in a future PR.

src/serac/numerics/functional/tuple.hpp
adayton1: All the functions that I removed already have support in camp.

src/serac/numerics/functional/tuple.hpp
```diff
@@ -13,216 +13,32 @@
...
-tuple(T...) -> tuple<T...>;
-
 /**
  * @brief helper function for combining a list of values into a tuple
  * @tparam T types of the values to be tuple-d
  * @param args the actual values to be put into a tuple
  */
 template <typename... T>
-SERAC_HOST_DEVICE tuple<T...> make_tuple(const T&... args)
+SERAC_HOST_DEVICE camp::tuple<T...> make_tuple(const T&... args)
```
white238, line 31: Can you add a comment to these as why they exist here instead of in camp?

adayton1: Done.

adayton1:
I ran the build_tpls script. @white238, do I need to run something else to update the host configs?

samuelpmishLLNL (approved):

jamiebramwell (approved):
@adayton1 Looks great! Thanks!!!

adayton1:
nvcc is crashing when trying to compile the following code. A compiler bug has been submitted through LC.

```
#include "camp/tuple.hpp"
#include <utility>

template <typename... S, typename... T, int... i>
CAMP_HOST_DEVICE constexpr auto plus_helper(const camp::tuple<S...>& x,
                                            const camp::tuple<T...>& y,
                                            std::integer_sequence<int, i...>)
{
     return camp::tuple{camp::get<i>(x) + camp::get<i>(y)...};
}

template <typename... S, typename... T>
CAMP_HOST_DEVICE constexpr auto operator+(const camp::tuple<S...>& x,
                                          const camp::tuple<T...>& y)
{
   static_assert(sizeof...(S) == sizeof...(T));
   return plus_helper(x, y, std::make_integer_sequence<int, static_cast<int>(sizeof...(S))>());
}

int main(int argc, char* argv[]) {
   camp::tuple a{0};
   camp::tuple b{1};
   camp::tuple c = a + b;
   (void) c;
   return 0;
}
```

adayton1:
I have another data point. I tried using camp functionality instead of std::integer_sequence and std::make_integer_sequence, but nvcc still crashes.

```
#include "camp/tuple.hpp"
#include <utility>

template <typename... S, typename... T, camp::idx_t... i>
CAMP_HOST_DEVICE constexpr auto plus_helper(const camp::tuple<S...>& x,
                                            const camp::tuple<T...>& y,
                                            camp::idx_seq<i...>)
{
     return camp::tuple{camp::get<i>(x) + camp::get<i>(y)...};
}

template <typename... S, typename... T>
CAMP_HOST_DEVICE constexpr auto operator+(const camp::tuple<S...>& x,
                                          const camp::tuple<T...>& y)
{
   static_assert(sizeof...(S) == sizeof...(T));
   return plus_helper(x, y, camp::make_idx_seq_t<sizeof...(S)>{});
}

int main(int argc, char* argv[]) {
   camp::tuple a{0};
   camp::tuple b{1};
   camp::tuple c = a + b;
   (void) c;
   return 0;
}
```

white238:
~~Data point from @trws, removing all CTAD from `camp::tuple` should allow this to compile.~~ Correction below

trws:
Small clarification, the CTAD is mostly fine, it's just in copy deduction that we have problems.  In this example it's specifically the `camp::tuple c = a + b;` line.  Changing that out for auto should let it work.

white238:
> Small clarification, the CTAD is mostly fine, it's just in copy deduction that we have problems. In this example it's specifically the `camp::tuple c = a + b;` line. Changing that out for auto should let it work.

Apparently my reading comprehension is low today. Thanks for the correction @trws !

white238:
Hi Chris,

Just heard back from the Nvidia guy at the lab. The fix for the CAMP tuple bug is slated for the CUDA 12.1 release.

Thanks,
Alan Dayton

jamiebramwell:
Closing as significant refactoring is happening. This PR will be used as reference once the refactoring is complete.

PR #739, opened 2022-07-01
Clean up the boundary condition projection methods

jamiebramwell:
This cleans the boundary condition projection functions (particularly removing the duplicate and broken `projectBdr` methods) and disables a the first order system ODE test as it is currently broken in develop. This issue is tracked in https://github.com/LLNL/serac/issues/741 .

Specifically, this PR

- Removes the state references in the boundary condition object. Only a finite element space is needed to correctly compute the degree of freedom lists from the given attributes and optional vector components.
- Streamlines the application of essential boundary conditions by combining the matrix elimination and RHS modification calls.
- Ensures consistent use of the time parameter while setting essential degrees of freedom.
- Removes post-construction setting of DOFs for boundary conditions.
- Increases the wait time for files on LC for integration tests.


samuelpmishLLNL (dismissed):

src/serac/physics/boundary_conditions/boundary_condition.hpp
```diff
@@ -178,47 +178,27 @@ class BoundaryCondition {
 
   /**
    * @brief Projects the boundary condition over a field
+   * @param[in] time The time at which to project the boundary condition
    * @param[inout] state The field to project over
    */
```
samuelpmishLLNL: These doxygen comments are almost tautological-- can we improve the explanation of what these functions do? 

I'm not too familiar with these functions, is the following explanation accurate:

```cpp
  /**
   * @brief for each dof that is associated with this boundary condition, overwrite 
   * its value in `state` by evaluating the provided function at the node associated with that dof
   * @param[inout] state The vector of dofs to modify
   * @param[in] time The time at which to evaluate the function
   */
```

src/serac/physics/boundary_conditions/boundary_condition.hpp
```diff
@@ -178,47 +178,27 @@ class BoundaryCondition {
...
   /**
    * @brief Projects the boundary condition over boundary to a DoF vector
    * @param[in] dof_values The discrete dof values to project
-   * @param[in] time The time for the coefficient, used for time-varying coefficients
+   * @param[in] time The time at which to project the boundary condition
    * @pre A corresponding field (FiniteElementState) has been associated
    * with the calling object via BoundaryCondition::setTrueDofs(FiniteElementState&)
    */
-  void projectBdrToDofs(mfem::Vector& dof_values, const double time) const;
+  void projectToDofs(mfem::Vector& dof_values, const double time) const;
```
samuelpmishLLNL: this seems equivalent to `project(FiniteElementState& state, const double time)` above, is there a reason to have both?

samuelpmishLLNL:
As I'm working in the solid mechanics module, I'm noticing that these boundary condition classes are sort of hard to use. 

One of the confusing aspects is that `time` is handled implicitly in some functions and explicitly in others, and the functions where `time` is passed in look like they actually mutate the internal state of the boundary condition object:
```cpp
  auto f = [](mfem::Vector x, double time) { ... };
  BoundaryCondition bc(f, ...);
  FiniteElementState u = ... ;

  // for (i in dofs(bc)) { u_i := f(x_i, t = 1.0); }
  bc.setTime(1.0);
  bc.project(u);

  // for (i in dofs(bc)) { u_i := f(x_i, t = 2.0); }
  bc.project(u, 2.0);

  // at which time is f(x, t) evaluated here?
  bc.project(u);
```

Also we have a lot of equivalent ways to do the same thing
`BoundaryCondition::setTime(double) + BoundaryCondition::project(FiniteElementState)`
`BoundaryCondition::project(FiniteElementState, double /* time */)`
`FiniteElementVector::project(mfem::Coefficient&)`
`FiniteElementVector::projectBdr(mfem::Coefficient&, const mfem::Array<int>&)`

Grammatically, it's weird to have the verb "project" used to mean the same thing in both directions, i.e. 
we're defining `A.project(B)` to be semantically equivalent to `B.project(A)`. 

Additionally, the way that these projections specify which dofs should be modified is inconsistent. Some specify this with explicit  attribute lists, others use explicit dof lists, or implicit dof lists, or implicit dof lists with implicit references to vectors. This is compounded by the earlier ambiguity about what time is being used to evaluate the provided functions.

src/serac/physics/boundary_conditions/boundary_condition.hpp
```diff
@@ -133,140 +133,67 @@ class BoundaryCondition {
...
-   * @param[in] k_mat_post_elim A stiffness (system) matrix post-elimination
+   * @param[inout] k_mat A stiffness (system) matrix post-elimination. The rows and cols of the essential dofs will be
+   * eliminated.
    * @param[out] rhs The RHS vector for the system
    * @param[inout] state The state from which the solution DOF values are extracted and used to eliminate
-   * @param[in] time Simulation time, used for time-varying boundary coefficients
-   * @pre BoundaryCondition::eliminateFrom has been called
+   * @pre The input state solution must contain the correct essential DOFs. This can be done by calling the @a
+   * BoundaryCondition::setDofs method.
    */
```
samuelpmishLLNL: Some ambiguities in this description:
- what does "elimination" mean here? This word typically means row-reduction in numerical linear algebra, but that's not what it means here.
- what is a "system" matrix?
- is `k_mat` required to be "post-elimination", or is this function the one to do the eliminating?
- is `state` an `inout` or just an input?

I'm not sure how mfem actually modifies the coefficient matrix (does it leave a 1 on the diagonal, or something else?), so the pseudocode below may be slightly inaccurate: 

```cpp
/**
 * @brief modify the system of equations, A x = b, by replacing equations that correspond to 
 * essential boundary conditions with ones that prescribe the desired values. To preserve symmetry,
 * the off-diagonal entries of associated columns of A are also zeroed out, and b is modified accordingly. 
 * This function is equivalent to the MATLAB code:
 * 
 *     A[essential_dofs] = eye(n)[essential_dofs];
 *     b -= A[:, essential_dofs] * x[essential_dofs];
 *     A[:, essential_dofs] = eye(n)[:, essential_dofs];
 * 
 * where `essential_dofs` are the rows corresponding to this essential boundary condition
 * 
 * @param[inout] A the coefficient matrix of the system of equations
 * @param[in] x the vector containing the prescribed values at the essential boundary condition indices 
 * @param[inout] b the righthand side
 *
 * @note `BoundaryCondition::setDofs` can be used to set the essential dofs of `x`.
 */

Comment (samuelpmishLLNL) [reply src/serac/physics/boundary_conditions/boundary_condition.hpp]: Also: does this modify the essential dofs of `b`?

File: src/serac/physics/boundary_conditions/boundary_condition.hpp [hunk 4]
```diff
@@ -133,140 +133,67 @@ class BoundaryCondition {
...
+  /**
+   * @brief "Manually" set the DOF indices without specifying the field to which they apply
+   * @param[in] true_dofs The true vector indices of the DOFs constrained by the boundary condition
    *
-   * Used for time-dependent coefficients
+   * @note This will set both the true and local internal dof index arrays.
+   * @note True and local dofs are described in the <a href="https://mfem.org/pri-dual-vec/">MFEM documentation</a>
    */
-  void setTime(const double time);
+  void setTrueDofList(const mfem::Array<int> true_dofs);
```
samuelpmishLLNL:
```suggestion
  void setTrueDofList(const mfem::Array<int> & true_dofs);
```

src/serac/physics/boundary_conditions/boundary_condition.hpp
jamiebramwell: Yes - it modifies the essential entries of `b`. I have modified the docs to reflect this.

samuelpmishLLNL (approved):

jamiebramwell:
@btalamini , I added a comment about the essential dofs as we discussed yesterday.

src/serac/infrastructure/tests/serac_error_handling.cpp
```diff
@@ -54,21 +54,17 @@ TEST(SeracErrorHandling, EquationSolverKinsolNotAvailable)
...
-{
-  auto mesh      = mesh::refineAndDistribute(buildDiskMesh(10));
-  int  num_attrs = mesh->bdr_attributes.Max();
-
-  auto              coef = std::make_shared<mfem::ConstantCoefficient>();
-  BoundaryCondition bc(coef, 0, std::set<int>{1}, num_attrs);
-  EXPECT_THROW(bc.project(), SlicErrorException);
-}
-
 TEST(SeracErrorHandling, BcOneComponentVectorCoef)
```
btalamini, line 57: FWIW, I'm a fan of very descriptive test names, favoring clarity to brevity. I prefer sentence-length names that describe exactly what's being tested, so that when it fails, anyone can understand what went wrong. It's hard to know what this test is actually testing as it stands. You have to read the test logic to know. I'm not requesting you to change this now, just making a remark while it's topical.

src/serac/physics/tests/serac_newmark_test.cpp
```diff
@@ -211,6 +211,11 @@ TEST_F(NewmarkBetaTest, EquilbriumLua)
   solid_solver->outputState();
 }
 
+/*
+
+TODO This is currently disabled as this test occasionally fails when run in parallel but only following another test.
+This is tracked in https://github.com/LLNL/serac/issues/741 .
```
btalamini, line 217: Glad you made an issue!

btalamini (approved):
Looks good.

PR #740, opened 2022-07-05
update solid mechanics module

samuelpmishLLNL:
Notable changes:

1. Add preliminary support for materials with internal variables</summary>

  To simulate a material that has internal variables, the general steps are:

```cpp
// create the solid mechanics module
SolidFunctional<p, dim> solid_solver(...);

// create a material model with the desired parameters
solid_mechanics::J2 mat{
  1000, // Young's modulus
  0.25, // Poisson's ratio
  1.0,  // isotropic hardening constant
  2.0,  // kinematic hardening constant
  5.0,  // yield stress
  1.0   // mass density
};

// create an instance of the initial state
solid_mechanics::J2::State initial_state{};

// initialize a state object for each quadrature point
auto state = solid_solver.createQuadratureDataBuffer(initial_state);

// register the material with the solid mechanics module, and give it the buffer of quadrature state
solid_solver.setMaterial(mat, state);
```

There is still more work to be done here, including adding more ways to create the quadrature data buffer (loading binary data from a restart, or using a lambda function to initialize spatially-dependent quadrature data).

Materials that do not have internal variables still just call 

```cpp
solid_solver.setMaterial(mat);
```

-------------------------
  
2. Simplify material interface requirements for solid mechanics</summary>

the new interface:

```cpp
struct LinearIsotropic {
  // each material must define a type called "State" to hold the internal variables
  // if a material doesn't need internal variables, then you can just set this to "Empty"
  using State = Empty;

  // new operator() function signature:
  // first argument is an instance of State, as above
  // then you get the displacement gradient
  // the spatial position and displacement vector are no longer passed
  //
  // additionally, now you only return the Cauchy stress, rather than a struct
  template <typename DispGradType>
  SERAC_HOST_DEVICE auto operator()(State & /* state */, const DispGradType& du_dX) const
  {
    auto I      = Identity<dim>();
    auto lambda = K - (2.0 / dim) * G;
    auto epsilon = 0.5 * (transpose(du_dX) + du_dX);
    return lambda * tr(epsilon) * I + 2.0 * G * epsilon;
  }

  // materials are required to define a member variable named "density"
  double density;   
  double K;
  double G;
};
```

-------------------------

3. Add a J2 material model with linear isotropic and kinematic hardening</summary>

  tested against an analytic solution from [here](https://github.com/LLNL/serac/files/9049523/j2_verification.pdf):
  J2_verification

  mathematica notebook with analytic solution and plotting available [here](https://github.com/LLNL/serac/files/9049652/J2_verification.nb.zip)

Some interns asked about what the isotropic hardening ($H_i$) and kinematic hardening ($H_k$) parameters of this model do, so I made some animations to show:

<details>
<summary> animations </summary>

J2_neither
J2_isotropic
J2_kinematic
J2_both
</details>

-------------------------

4. Improvements to the Solid Mechanics module</summary>

  - using a single `Functional` object to represent the residual calculation
  - managing material state buffers so that quadrature data is reset on each residual evaluation, and then finally updated after finding a converged solution
  - simplified previous `QuadratureData` implementation
  
-------------------------

5. Miscellaneous small changes</summary>

  - adding another tool for general material verification, `single_quadrature_point_test()`
  - adding a convenience function `diag()` for creating diagonal matrices from a vector of the diagonal entries
  - added a javascript tool on our readthedocs page to do conversions between elastic properties
  - added animations for J2 hardening above to readthedocs page


src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -96,6 +96,88 @@ void functional_solid_test_static(double expected_disp_norm)
...
+    0.25, // Poisson's ratio
+    1.0,  // isotropic hardening constant
+    2.0,  // kinematic hardening constant
+    5.0,  // yield stress
+    1.0   // mass density
+  };
+
+  solid_mechanics::J2::State initial_state{};
+
+  auto state = solid_solver.createQuadratureDataBuffer(initial_state);
```
jamiebramwell, line 132: Can this be combined with the `setMaterial` call?

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -96,6 +96,88 @@ void functional_solid_test_static(double expected_disp_norm)
...
+  solid_solver.setMaterial(mat, state);
+
+  // Define the function for the initial displacement and boundary condition
+  auto bc = [](const mfem::Vector&, mfem::Vector& bc_vec) -> void { bc_vec = 0.0; };
+
+  // Set the initial displacement and boundary condition
+  solid_solver.setDisplacementBCs(ess_bdr, bc);
+  solid_solver.setDisplacement(bc);
+
+  solid_solver.setPiolaTraction([](auto x, auto /*n*/, auto t){
```
jamiebramwell: `setReferenceTraction` maybe? I'm OK with either as long as it's documented.

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -366,21 +441,23 @@ void functional_parameterized_solid_test(double expected_disp_norm)
...
-
-TEST(SolidFunctional, 2DLinearDynamic) { functional_solid_test_dynamic<1, 2>(1.52116682); }
-TEST(SolidFunctional, 2DQuadDynamic) { functional_solid_test_dynamic<2, 2>(1.52777214); }
-
-TEST(SolidFunctional, 3DLinearDynamic) { functional_solid_test_dynamic<1, 3>(1.520679017); }
-TEST(SolidFunctional, 3DQuadDynamic) { functional_solid_test_dynamic<2, 3>(1.527009514); }
-
-TEST(SolidFunctional, 2DLinearPressure) { functional_solid_test_boundary<1, 2>(0.065326222, TestType::Pressure); }
-TEST(SolidFunctional, 2DLinearTraction) { functional_solid_test_boundary<1, 2>(0.126593590, TestType::Traction); }
+//TEST(SolidFunctional, 2DLinearStatic) { functional_solid_test_static<1, 2>(1.511052595); }
```
jamiebramwell: Is there a reason these are disabled? Or is it just leftover from debugging?

src/serac/physics/materials/parameterized_solid_functional_material.hpp
```diff
@@ -16,34 +16,19 @@
 #include "serac/physics/materials/solid_functional_material.hpp"
 
 /// SolidFunctional helper data types
-namespace serac::solid_util {
+namespace serac::solid_mechanics {
```
jamiebramwell, line 19: 👍

src/serac/physics/materials/parameterized_solid_functional_material.hpp
```diff
@@ -75,15 +56,9 @@ class ParameterizedLinearIsotropicSolid {
...
-private:
-  /// Density
-  double density_;
-
-  /// Bulk modulus
-  double bulk_modulus_offset_;
-
-  /// Shear modulus
-  double shear_modulus_offset_;
+  double density;
```
jamiebramwell: Do we want these to be user-modifiable after construction?

src/serac/physics/solid_functional.hpp
```diff
@@ -22,6 +22,77 @@
 #include "serac/physics/solid.hpp"
 #include "serac/physics/materials/functional_material_utils.hpp"
 
+
+// DELETE BELOW
```
jamiebramwell: Should this be deleted?

src/serac/physics/solid_functional.hpp
```diff
@@ -184,6 +279,29 @@ class SolidFunctional : public BasePhysics {
...
 
+  template < typename T >
+  QuadratureData<T> createQuadratureDataBuffer(T initial_state) {
+    //const mfem::FiniteElement&   el = *(displacement_.space().GetFE(0));
+    //const mfem::IntegrationRule& ir = mfem::IntRules.Get(el.GetGeomType(), el.GetOrder() * 2);
+    //constexpr auto flags = mfem::GeometricFactors::COORDINATES | mfem::GeometricFactors::JACOBIANS;
+    //auto geom = mesh_.GetGeometricFactors(ir, flags);
+
+    constexpr auto Q                                 = order + 1;
+    constexpr auto num_quadrature_points_per_element = (dim == 2) ? Q * Q : Q * Q * Q;
```
jamiebramwell: A note that this is assuming hexes or quads with tensor product integration rules would be nice. I know it is also required by other parts of the code, but it would help readers understand why this is hard coded.

src/serac/physics/solid_functional.hpp
```diff
@@ -264,55 +393,31 @@ class SolidFunctional : public BasePhysics {
...
-          auto source = zero{};
-
-          auto response = parameterized_material(x, u, du_dX, serac::get<0>(params)...);
-
-          auto flux = response.stress;
+        [this, material](auto /*x*/, auto & state, auto displacement, auto acceleration, auto ... params) {
+          auto a = get<VALUE>(acceleration);
+          auto du_dX = get<DERIVATIVE>(displacement);
+          auto body_force = material.density * a + 0.0 * du_dX[0];
+          auto stress = material(state, du_dX, serac::get<0>(params)...);
```
jamiebramwell:
```suggestion
          auto stress = material(state, du_dX, serac::get<VALUE>(params)...);
```

src/serac/physics/solid_functional.hpp
```diff
@@ -485,28 +534,36 @@ class SolidFunctional : public BasePhysics {
...
-    auto residual = std::make_unique<mfem_ext::StdFunctionOperator>(
+    return std::make_unique<mfem_ext::StdFunctionOperator>(
         displacement_.space().TrueVSize(),
 
         // residual function
         [this](const mfem::Vector& u, mfem::Vector& r) {
-          functional_call_args_[0] = u;
-
-          r = (*K_functional_)(functional_call_args_);
+          #if 0
```
jamiebramwell: Dead code.

src/serac/physics/tests/serac_solid_functional.cpp
samuelpmishLLNL: For the homogeneous case, that seems like a good shortcut. I think we still need ways to creating spatially-varying quadrature data and loading binary data buffers for restarted runs

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -111,69 +53,103 @@ class LinearIsotropicSolid {
...
-    double poisson_ratio = (3 * K - 2 * G) / (6 * K + 2 * G);
+    //
+    // see pg. 260, box 7.5,
+    // in "Computational Methods for Plasticity"
+    //
 
-    SLIC_ERROR_ROOT_IF(poisson_ratio < 0.0, "Poisson ratio must be positive in the neo-Hookean material model.");
-  }
+    auto new_strain = sym(du_dX);
+    auto dstrain = new_strain - state.old_strain;
```
btalamini: Instead of caching both `old_strain` and `el_strain` in `state`, you can save some memory by only storing the difference `state.pl_strain_tensor = old_strain - el_strain` (in other words, make the plastic strain tensor the state variable). Then the trial elastic strain on line 120 is just
```c++
auto el_strain = new_strain - state.pl_strain_tensor
```

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -111,69 +53,103 @@ class LinearIsotropicSolid {
...
+  double Hk;       ///< kinematic hardening constant
+  double sigma_y;  ///< yield stress
+  double density;  ///< mass density
+
+  /// @brief variables required to characterize the hysteresis response
+  struct State {
+    tensor<double, dim, dim> beta;           ///< back-stress tensor
+    tensor<double, dim, dim> old_strain;     ///< previous strain
+    tensor<double, dim, dim> el_strain;      ///< elastic strain
+    double                   pl_strain;      ///< plastic strain
```
btalamini: The state variable names could be more consistent. The name `el_strain` is used for a tensor, which is fine, but then I think the name `pl_strain` should also be used for the tensor-valued plastic strain (as in `pl_strain = strain - el_strain`). The scalar variable often goes by names that emphasize its scalar nature such as "equivalent plastic strain", "uniaxial equivalent plastic strain", "accumulated plastic strain", etc.

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -111,69 +53,103 @@ class LinearIsotropicSolid {
...
-  /// Bulk modulus in the stress free configuration
-  double bulk_modulus_;
+    } else {
+
+      state.el_strain = get_value(el_strain);
+
+    }
+
+    state.old_strain = get_value(new_strain);
+
```
btalamini: It looks like you never update several of the state variables (I think `q` and `plastic_strain_inc`). They're not needed for functionality, of course. Did you put them in the state variable set for output?

src/serac/numerics/functional/dual.hpp
```diff
@@ -246,6 +246,9 @@ SERAC_HOST_DEVICE auto log(dual<gradient_type> a)
   return dual<gradient_type>{log(a.value), a.gradient / a.value};
 }
 
+/** @brief implementation of the natural logarithm function for doubles */
+SERAC_HOST_DEVICE auto log(double a) { return std::log(a); }
```
btalamini: This is good, there were a few TODOs peppered through the code to force this lookup.

src/serac/physics/materials/tests/J2_material.cpp
```diff
@@ -0,0 +1,119 @@
...
+  double tmax = 2.0;
+  int num_steps = 64;
+
+  double G = 79000;
+  double K = 10 * G;
+
+  double E = 9 * K * G / (3 * K + G);
+  double nu = (3 * K - 2 * G) / (2 * (3 * K + G));
+
+  solid_mechanics::J2alt material{
```
btalamini: Is "J2alt" the right material name? Sorry if this is a stupid question.

src/serac/physics/solid_functional.hpp
```diff
@@ -105,22 +186,19 @@ class SolidFunctional : public BasePhysics {
...
+        trial_spaces[i + 2]         = &(parameter_states_[i].get().space());
         parameter_sensitivities_[i] = std::make_unique<FiniteElementDual>(mesh_, parameter_states_[i].get().space());
-        functional_call_args_.emplace_back(parameter_states_[i].get());
       }
     }
 
-    M_functional_ = std::make_unique<Functional<test(trial, parameter_space...)>>(&displacement_.space(), trial_spaces);
-
-    K_functional_ = std::make_unique<Functional<test(trial, parameter_space...)>>(&displacement_.space(), trial_spaces);
+    residual_ = std::make_unique<Functional<test(trial, trial, parameter_space...)>>(test_space, trial_spaces);
```
btalamini, line 141: 👍

src/serac/physics/solid_functional.hpp
btalamini: Seconding Jamie's point. A note on the fixed precision/accuracy of the quadrature would be helpful as well. 

Related: do you anticipate allowing the quadrature rule to be specified as a compile time option?

src/serac/physics/solid_functional.hpp
```diff
@@ -264,55 +344,31 @@ class SolidFunctional : public BasePhysics {
...
+          auto du_dX = get<DERIVATIVE>(displacement);
+          auto body_force = material.density * a + 0.0 * du_dX[0];
+          auto stress = material(state, du_dX, serac::get<0>(params)...);
 
           if (geom_nonlin_ == GeometricNonlinearities::On) {
-            auto deformation_grad = du_dX + I_;
-            flux                  = flux * inv(transpose(deformation_grad));
+            auto F = I + du_dX;
+            body_force = body_force * det(F);
+            stress = dot(stress, inv(transpose(F)));
```
btalamini: Naive question (though maybe directed at @jamiebramwell ): If I'm reading this right, if the user specifies that the problem is geometrically nonlinear, the stress is then transformed from the Kirchhoff stress to the Piola stress. Why not use the natural pairing of stress and integration domain either by (a) requiring the Piola stress as the output from the material models, and leaving the residual calculation as is, or (b) leave the stress output convention alone, but compute the shape function gradients with the current coordinates in the residual?

src/serac/physics/solid_functional.hpp
```diff
@@ -525,59 +484,138 @@ class SolidFunctional : public BasePhysics {
...
+    }
+
+    // if our simulation involves materials that have internal state 
+    if (residual_->material_state_buffers_.size()) {
+
+      // after finding displacements that satisfy equilibrium, 
+      // compute the residual one more time, this time enabling
+      // the material state buffers to be updated
+      residual_->update_qdata = true;
+      (*residual_)(displacement_, zero_, parameter_states_[parameter_indices] ...);
```
btalamini: Could we take advantage of this additional call to take care of other business?

- If adjoint sensitivities are requested from material parameters, it would be efficient to promote them to dual numbers in the material operator at this call site only.  This will avoid unnecessary operations during the equilibrium search. (Furthermore, materials with iterative procedures can detect when the parameters are dual and do the adjoint computation by hand at the solution, rather than accumulating through the iteration).
- If we can turn off the zeroing out of the residual terms corresponding to essential BCs, then those slots will collect the reactions, for use in output and QoI.

scripts/spack/configs/linux_ubuntu_20/packages.yaml
```diff
@@ -79,8 +80,8 @@ packages:
   cuda:
     buildable: false
     externals:
-    - spec: cuda@11.2.0
-      prefix: /usr/local/cuda-11.2
+    - spec: cuda@11.7.0
```
white238, line 83: @jamiebramwell do you have 11.7 installed on your machine? Was this moved forward for a reason (bug in 11.2, etc)?

jamiebramwell: I don't have cuda installed on my machine at all. I believe this was for @samuelpmishLLNL 's local build.

white238: k.. at some point in the future we may have to make these user-specific.

samuelpmishLLNL: I wish these tools would just identify the version number automatically when you point it at a directory, it's just a one-liner to deduce this information:

```sh
sam@provolone:~$ /usr/local/cuda/bin/nvcc --version | grep -o -P "release.{0,5}" | grep -o "....$"
11.7
```

instead we're requiring users to either:
- have incredibly specific versions of certain libraries
- manually dive into these config files to build their dependencies

both of which are unpleasant

samuelpmishLLNL: that being said, I didn't intend to include this file in the PR, I can revert it back to 11.2 (although nvcc 11.7 does include a number of fixes for compiler bugs we submitted to NVIDIA)

src/serac/physics/solid_functional.hpp
samuelpmishLLNL: > If we can turn off the zeroing out of the residual terms corresponding to essential BCs, then those slots will collect the reactions, for use in output and QoI.

👍 

`Functional` doesn't know anything about essential b.c.s, so the output of this call will have the nodal reaction forces. Instead of discarding the output from the highlighted code, we're storing it now and there's a getter `SolidFunctional::nodalForces()` to access it.

> If adjoint sensitivities are requested from material parameters, it would be efficient to promote them to dual numbers in the material operator at this call site only. This will avoid unnecessary operations during the equilibrium search. (Furthermore, materials with iterative procedures can detect when the parameters are dual and do the adjoint computation by hand at the solution, rather than accumulating through the iteration).

For reference, the derivatives w.r.t. parameter fields aren't calculated during the equilibrium solve, they're lazily-evaluated when the user calls "computeSensitivity()`.

src/serac/physics/materials/parameterized_solid_functional_material.hpp
samuelpmishLLNL: In this case, it doesn't bother me. We make a copy of the material object when calling `SolidFunctional::addMaterial()`, so the user doesn't have a way to modify the values that are used the kernels afterwards.

Were there other specific concerns you had about the mutability of these member variables?

src/serac/physics/tests/serac_solid_functional.cpp
samuelpmishLLNL: it was just a prototyping thing, enabling fewer tests shaves off a few seconds of compilation time (and reduces redundant compilation error messages when making interface changes)

src/serac/physics/tests/serac_solid_functional.cpp
samuelpmishLLNL: I'm good with either, I just want the configuration to be in the name somewhere so you don't see code like

```cpp
simluation.setTraction(some_function, true);
```

and have to think about what the boolean means

src/serac/physics/tests/serac_solid_functional.cpp
samuelpmishLLNL: after our discussion on Thursday (deciding to use shared_ptr for quadrature data), it won't be possible to combine these calls any more

src/serac/physics/tests/serac_solid_functional.cpp
samuelpmishLLNL: Also, I wonder if "set" is the right word to use here. To me "set" implies assignment `r := f`, but what it's really doing is addition `r += f` (adding another surface integral to be evaluated)

btalamini: > I'm good with either, I just want the configuration to be in the name somewhere so you don't see code like
> 
> ```c++
> simluation.setTraction(some_function, true);
> ```
> 
> and have to think about what the boolean means


Yeah, would be nice if C++ had keyword arguments like Python:
```python
solid_solver.setTraction(some_function, reference_configuration=true)
```
which is self-documenting.

btalamini: > Also, I wonder if "set" is the right word to use here. To me "set" implies assignment `r := f`, but what it's really doing is addition `r += f` (adding another surface integral to be evaluated)

How about "apply"? "add" could work, too, but I slightly prefer the former. Whichever we choose, it should also be used for the body force for uniformity.

btalamini (approved):
Looks correct. Small optimization possible: `normalize(eta)` is recomputed several times, you could store it in a variable and reuse.

src/serac/physics/materials/parameterized_solid_functional_material.hpp
jamiebramwell: Not really - As long as we copy them into the module (and document that so people don't try to change it on-the-fly), that seems good to me.

src/serac/physics/solid_functional.hpp
jamiebramwell: I think I like the option of (b) better. When we were thinking of material interfaces, it came up that most of the current codes use Kirchoff (or really Cauchy as I have in the shape optimization branch) stress. That's why we went this way. However the old MFEM implementation used Piola, so we're kind of in an in-between state. I agree one or the other (my vote is for Cauchy) is preferable.

jamiebramwell:
Built docs here: https://serac.readthedocs.io/en/mish2-update_solid_mechanics_module/

scripts/spack/configs/linux_ubuntu_20/packages.yaml
jamiebramwell: For now, I think this could either be a new config directory (`linux_ubuntu_20_cuda`?) or just removed from this PR.

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -96,6 +86,94 @@ void functional_solid_test_static(double expected_disp_norm)
...
+  };
+
+  solid_mechanics::J2::State initial_state{};
+
+  auto state = solid_solver.createQuadratureDataBuffer(initial_state);
+
+  solid_solver.setMaterial(mat, state);
+
+  // it is confusing to me to just pull out magic numbers
+  // for the element and boundary attributes
```
jamiebramwell: I believe Jonathan added some lambda-based mesh marking infrastructure a while ago. Would that be preferable? At some point we do have to interact with externally generated meshes, and tags like this is how I typically see it done.

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -96,6 +86,94 @@ void functional_solid_test_static(double expected_disp_norm)
...
+
+  // prescribe zero displacement at the supported end of the beam,
+  std::set<int> support           = {1};
+  auto          zero_displacement = [](const mfem::Vector&, mfem::Vector& u) -> void { u = 0.0; };
+  solid_solver.setDisplacementBCs(support, zero_displacement);
+
```
jamiebramwell, line 140: Should these be two separate tests?

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -366,21 +381,24 @@ void functional_parameterized_solid_test(double expected_disp_norm)
   EXPECT_NEAR(expected_disp_norm, norm(solid_solver.displacement()), 1.0e-6);
 }
 
-TEST(SolidFunctional, 2DLinearStatic) { functional_solid_test_static<1, 2>(1.511052595); }
+// TEST(SolidFunctional, 2DLinearStatic) { functional_solid_test_static<1, 2>(1.511052595); }
 TEST(SolidFunctional, 2DQuadStatic) { functional_solid_test_static<2, 2>(2.18604855); }
```
jamiebramwell: Should these still be disabled?

src/serac/physics/tests/liquid_crystal_elastomer_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+                             transition_temperature, Nb2);
+
+  LiquidCrystalElastomer::State initial_state{};
+
+  auto qdata = solid_solver.createQuadratureDataBuffer(initial_state);
+
+  solid_solver.setMaterial(mat, qdata);
+
+  // prescribe zero displacement at the supported end of the beam
+#if 0
```
jamiebramwell: dead code

src/serac/physics/common.hpp
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2019-2021, Lawrence Livermore National Security, LLC and
```
jamiebramwell: 2022

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -113,9 +84,14 @@ SolverOptions defaultDynamicOptions()
  *  where \f$\mathbf{M}\f$ is a mass matrix, \f$\mathbf{K}\f$ is a stiffness matrix, \f$\mathbf{u}\f$ is the
  *  temperature degree of freedom vector, and \f$\mathbf{f}\f$ is a thermal load vector.
  */
-
-template <int order, int dim, typename... parameter_space>
-class ThermalConductionFunctional : public BasePhysics {
+template <int order, int dim, typename parameters = Parameters<>,
+          typename parameter_indices = std::make_integer_sequence<int, parameters::n>>
+class ThermalConductionFunctional;
```
jamiebramwell, line 89: I'll have to remember this pattern!

src/serac/numerics/functional/functional.hpp
```diff
@@ -25,6 +25,10 @@
 
 namespace serac {
 
+template <int i>
+struct DifferentiateWRT {
+};
```
jamiebramwell, line 30: There are a few of these encoded int types (`Dim`, `Index`, this one). Should we have a single index type, or is it better to have more explicit names even though the functionality is essentially duplicated? I think I lean towards the latter.

src/serac/physics/thermal_conduction_functional.hpp
```diff
@@ -594,9 +550,13 @@ class ThermalConductionFunctional : public BasePhysics {
   /// An auxilliary zero vector
   mfem::Vector zero_;
 
+  /// sam: is this the correct description of u_?
```
jamiebramwell: I believe this was tied to the ODE solve. I can go investigate.

src/serac/physics/solid_functional.hpp
```diff
@@ -105,22 +125,20 @@ class SolidFunctional : public BasePhysics {
     state_.push_back(adjoint_displacement_);
 
     // Create a pack of the primal field and parameter finite element spaces
-    std::array<mfem::ParFiniteElementSpace*, sizeof...(parameter_space) + 1> trial_spaces;
+    mfem::ParFiniteElementSpace*                                             test_space = &displacement_.space();
+    std::array<mfem::ParFiniteElementSpace*, sizeof...(parameter_space) + 2> trial_spaces;
     trial_spaces[0] = &displacement_.space();
-
-    functional_call_args_.emplace_back(displacement_);
+    trial_spaces[1] = &displacement_.space();  // the accelerations have the same trial space as displacement
```
jamiebramwell, line 131: velocity?

src/serac/physics/solid_functional.hpp
```diff
@@ -215,108 +283,44 @@ class SolidFunctional : public BasePhysics {
...
-
-          auto source = zero{};
-
-          auto response = parameterized_material(x, u, du_dX, serac::get<0>(params)...);
-
-          auto flux = response.stress;
+        [this, material](auto /*x*/, auto& state, auto displacement, auto acceleration, auto... params) {
+          auto a          = get<VALUE>(acceleration);
+          auto du_dX      = get<DERIVATIVE>(displacement);
+          auto body_force = material.density * a + 0.0 * du_dX[0];
```
jamiebramwell: Are these multiply by zero operations still needed? I remember at one point they were for type deduction reasons.

src/serac/physics/solid_functional.hpp
```diff
@@ -215,108 +283,44 @@ class SolidFunctional : public BasePhysics {
...
+          auto stress     = material(state, du_dX, serac::get<0>(params)...);
```
jamiebramwell: `serac::get<VALUE>`?

src/serac/physics/solid_functional.hpp
```diff
@@ -349,36 +353,23 @@ class SolidFunctional : public BasePhysics {
...
-          double geom_factor = (geom_nonlin_ == GeometricNonlinearities::On ? 1.0 : 0.0);
-
-          auto deformation_grad = du_dX + I_;
-
-          auto source = parameterized_body_force(x, time_, u, du_dX, serac::get<0>(params)...) *
-                        (1.0 + geom_factor * (det(deformation_grad) - 1.0));
-          return serac::tuple{source, flux};
+        [body_force, this](auto x, auto displacement, auto /* acceleration */, auto... /*params*/) {
+          auto du_dX    = get<DERIVATIVE>(displacement);
+          auto one_dual = (1.0 + 0.0 * du_dX[0][0]);
```
jamiebramwell: Why is this needed?

src/serac/physics/solid_functional.hpp
```diff
@@ -388,129 +379,41 @@ class SolidFunctional : public BasePhysics {
...
-    auto residual = std::make_unique<mfem_ext::StdFunctionOperator>(
+    return std::make_unique<mfem_ext::StdFunctionOperator>(
         displacement_.space().TrueVSize(),
 
         // residual function
         [this](const mfem::Vector& u, mfem::Vector& r) {
-          functional_call_args_[0] = u;
-
-          r = (*K_functional_)(functional_call_args_);
+          r = (*residual_)(u, zero_, parameter_states_[parameter_indices]...);
```
jamiebramwell, line 398: With this change, we can probably delete the array interface to functional I introduced.

src/serac/physics/solid_functional.hpp
```diff
@@ -524,56 +427,138 @@ class SolidFunctional : public BasePhysics {
...
+          [this](const mfem::Vector& d2u_dt2) -> mfem::Operator& {
+            add(1.0, u_, c0_, d2u_dt2, predicted_displacement_);
 
-            mfem::Vector K_arg(u_.Size());
-            add(1.0, u_, c0_, d2u_dt2, K_arg);
-            functional_call_args_[0] = K_arg;
+            // K := dR/du
+            auto K = serac::get<1>((*residual_)(differentiate_wrt(predicted_displacement_), d2u_dt2,
+                                                parameter_states_[parameter_indices]...));
+            std::unique_ptr<mfem::HypreParMatrix> k_mat(assemble(K));
```
jamiebramwell, line 448: `DERIVATIVE`

src/serac/physics/solid_functional.hpp
```diff
@@ -524,56 +427,138 @@ class SolidFunctional : public BasePhysics {
...
 
-            return *J_;
-          });
+      displacement_ += du_;
+
+      for (auto& bc : bcs_.essentials()) {
+        bc.setDofs(du_, time_);
+      }
+
+      // do I have to do this?
```
jamiebramwell: I believe the answer is no. This should already be set on line 468 and the underlying operator for the linear solver (i.e. the Jacobian) is set in the during the nonlinear solver `Mult` call.

src/serac/physics/solid_functional.hpp
```diff
@@ -606,11 +591,12 @@ class SolidFunctional : public BasePhysics {
     mfem::HypreParVector adjoint_essential(adjoint_load);
     adjoint_essential = 0.0;
 
-    functional_call_args_[0] = displacement_;
-
-    auto [r, drdu] = (*K_functional_)(functional_call_args_, Index<0>{});
-    auto jacobian  = assemble(drdu);
-    auto J_T       = std::unique_ptr<mfem::HypreParMatrix>(jacobian->Transpose());
+    // sam: is this the right thing to be doing for dynamics simulations,
+    // or are we implicitly assuming this should only be used in quasistatic analyses?
```
jamiebramwell, line 580: No - this adjoint solve only works for quasi-statics. I have a task for next year to add a method for transient problems. We should probably put a check in here that this is only called for quasi-static simulations.

src/serac/physics/solid_functional.hpp
```diff
@@ -625,7 +611,7 @@ class SolidFunctional : public BasePhysics {
     lin_solver.Mult(adjoint_load_vector, adjoint_displacement_);
 
     // Reset the equation solver to use the full nonlinear residual operator
-    nonlin_solver_.SetOperator(*residual_);
+    nonlin_solver_.SetOperator(*residual_with_bcs_);
```
jamiebramwell: As stated above, I don't think this is actually needed.

src/serac/physics/state/state_manager.hpp
```diff
@@ -58,33 +58,33 @@ class StateManager {
...
-  }
+  ///**
+  // * @brief Factory method for creating a new QuadratureData object
+  // * @tparam T The type of the per-qpt data
+  // * @param[in] name The name of the quadrature data field
+  // * @param[in] p The order of the quadrature rule
+  // * @param[in] mesh_tag A string that uniquely identifies the mesh on which the quadrature data is to be defined
+  // * @see QuadratureData::QuadratureData
+  // */
+  // template <typename T>
```
jamiebramwell: Dead code.

src/serac/physics/state/state_manager.hpp
```diff
@@ -114,7 +114,7 @@ class StateManager {
   {
     datacolls_.clear();
     is_restart_ = false;
-    syncable_data_.clear();
+    // syncable_data_.clear();
```
jamiebramwell: dead code

src/serac/physics/state/state_manager.hpp
```diff
@@ -169,7 +169,7 @@ class StateManager {
    * @brief A set of @p QuadratureData<T> objects that need to be synchronized before saving to disk
    * FIXME Need one set per datacoll since we can save at different times?
    */
-  static std::vector<std::unique_ptr<SyncableData>> syncable_data_;
+  // static std::vector<std::unique_ptr<SyncableData>> syncable_data_;
```
jamiebramwell: dead code

src/serac/physics/state/state_manager.cpp
```diff
@@ -139,9 +129,10 @@ void StateManager::save(const double t, const int cycle, const std::string& mesh
   std::string file_path = axom::utilities::filesystem::joinPath(datacoll.GetPrefixPath(), datacoll.GetCollectionName());
   SLIC_INFO_ROOT(axom::fmt::format("Saving data collection at time: {} to path: {}", t, file_path));
 
-  for (const auto& data : syncable_data_) {
-    data->sync();
-  }
+  // sam: ?
```
jamiebramwell: I believe this was used to update the underlying pure double arrays stored by sidre.

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -111,69 +53,89 @@ class LinearIsotropicSolid {
...
 
   /**
-   * @brief Material response call for a neo-Hookean solid
+   * @brief stress calculation for a NeoHookean material model
    *
-   * @tparam PositionType Spatial position type
-   * @tparam DisplacementType Displacement type
    * @tparam DispGradType Displacement gradient type
-   * @param displacement_grad displacement gradient with respect to the reference configuration (displacement_grad)
+   * @param du_dX displacement gradient with respect to the reference configuration (displacement_grad)
```
jamiebramwell, line 63: Update parameter name

src/serac/physics/materials/tests/J2_material.cpp
```diff
@@ -0,0 +1,111 @@
...
+ * a couple of the coefficents in the equations appear with the wrong sign,
+ * but the implementation below fixes those typos
+ */
+
+#include <iostream>
+
+#include "serac/physics/materials/solid_functional_material.hpp"
+#include "serac/physics/materials/material_verification_tools.hpp"
+
+using namespace serac;
```
jamiebramwell, line 27: Should this go in tensor.hpp?

src/serac/physics/materials/tests/J2_material.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  auto history = single_quadrature_point_test(tmax, num_steps, material, initial_state, du_dX);
+
+  for (auto [t, state, dudx, stress] : history) {
+    if (t > 0) {
+      double rel_error = norm(stress - analytic_soln(t)) / norm(stress);
+      std::cout << t << ": " << rel_error << std::endl;
+    }
+
+    // for plotting:
```
jamiebramwell: dead code

src/serac/numerics/quadrature_data.hpp
```diff
@@ -20,335 +20,91 @@
...
-  /**
-   * @brief A copy of the element_offsets member from mfem::QuadratureSpace
-   */
-  UnifiedArray<int> offsets_;
-  /**
-   * @brief The stride of the array
-   */
-  static constexpr int stride_ = detail::ceil(sizeof(T) / static_cast<double>(sizeof(double)));
+  T*     data;    ///< pointer to the buffer of quadrature data
+  size_t stride;  ///< how many quadrature points per element
```
jamiebramwell, line 70: `q_pts_per_element`?

src/serac/numerics/quadrature_data.hpp
```diff
@@ -20,335 +20,91 @@
...
+  Empty data;
+};
 
-template <typename T, typename Derived>
-SERAC_HOST_DEVICE T& QuadratureDataImpl<T, Derived>::operator()(const int element_idx, const int q_idx)
-{
-  const auto idx = asDerived().offsets()[static_cast<std::size_t>(element_idx)] + q_idx;
-  return asDerived().data()[static_cast<std::size_t>(idx)];
-}
+extern std::shared_ptr<QuadratureData<Nothing> > NoQData;
```
jamiebramwell, line 107: Why is this needed?

src/serac/numerics/functional/functional.hpp
```diff
@@ -25,6 +25,10 @@
 
 namespace serac {
 
+template <int i>
+struct DifferentiateWRT {
```
jamiebramwell, line 29: Do we still need `Index` on line 83?

src/serac/numerics/functional/functional.hpp
```diff
@@ -732,6 +709,10 @@ class Functional<test(trials...), exec> {
 
   /// @brief 3D array that stores each boundary element's gradient of the residual w.r.t. trial values
   ExecArray<double, 3, exec> bdr_element_gradients_[num_trial_spaces];
+
+public:  // TODO
+  /// @brief flag for denoting when a residual evaluation should update the material state buffers
```
jamiebramwell: I'm confused why this is a vector-of-vectors.

jamiebramwell: I can't find where this is used.

samuelpmishLLNL:
> https://serac.readthedocs.io/en/mish2-update_solid_mechanics_module/

The only updated parts are here:

https://serac.readthedocs.io/en/mish2-update_solid_mechanics_module/sphinx/theory_reference/solid.html#material-parameters

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: Good catch-- this was leftover from the initial material state implementationI did on this branch. Gone now.

src/serac/physics/tests/serac_solid_functional.cpp
samuelpmishLLNL: maybe we could adopt a convention for these tests where something like `{1}` is homogeneous essential b.c., `{2}` is non-homogenous essential b.c. `{3}` is natural, etc

It's just unpleasant to have to inspect the .mesh files every time to figure out what's going on.

It'd be nice if the mesh files had a string key to help with labeling sets of elements/boundary elements, but that's probably not going to happen.

src/serac/physics/solid_functional.hpp
samuelpmishLLNL: right now, the residual is just a function of $u, \ddot{u}$. We can add velocity later, if people need viscoelastic models

src/serac/physics/solid_functional.hpp
samuelpmishLLNL: This one still does something. The body force term here doesn't directly depend on the displacement, so it makes more sense to not have the ` + 0.0 * du_dX[0]` part initially. The problem is: if the geometric nonlinearity flag is turned on, then the body force is modified by multiplying by $\det(\boldsymbol{F})$ (making it dependent on the displacement). Since the geometric nonlinearity flag is a runtime choice, we can't write something like

```cpp
auto body_force = material.density * a;
auto stress     = material(state, du_dX, ...);
if (geom_nonlin_ == GeometricNonlinearities::On) {
  return {body_force * det(F), dot(stress, inv(transpose(F)))};
} else {
  return {body_force, stress};   
}
```
because the types of `body_force * det(F)` and `body_force` are different (and same with the stresses): the former tracks derivative information w.r.t. displacements, while the latter doesn't. C++ functions have to have a single return type, so you can't have a (runtime) conditional statement with branches that return different types.

src/serac/physics/solid_functional.hpp
samuelpmishLLNL: see https://github.com/LLNL/serac/pull/740#discussion_r931626982 re: conditional branches and returning only a single type from a function

(admittedly, this line of code is not elegant)

src/serac/physics/solid_functional.hpp
samuelpmishLLNL: I think I already did that

src/serac/physics/state/state_manager.hpp
samuelpmishLLNL: I wasn't sure what to do with this code, it wasn't compiling because I removed the old quadrature data logging. However, that capability didn't appear to be tested, or allow us to do restarts anyway.

src/serac/physics/materials/solid_functional_material.hpp
samuelpmishLLNL: I believe the parameter name in the doxygen comment already matches the function declaration below it

src/serac/physics/solid_functional.hpp
jamiebramwell: Per our discussion yesterday, the inertial force should not be modified as this is the reference density.

src/serac/physics/state/state_manager.hpp
jamiebramwell: If it's not working or testing, I say we delete it. We can always get it from the git history.

src/serac/numerics/quadrature_data.hpp
samuelpmishLLNL: It's used as a default value for `Functional::AddVolumeIntegral()`, so you can pass it q-functions that have no notion of internal variables (like body forces)

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: Good question, we have a number of these things floating around. It would be good to consolidate them, but I'm tempted to do this in a later PR.

src/serac/numerics/functional/functional.hpp
samuelpmishLLNL: oops, I responded to this in another comment https://github.com/LLNL/serac/pull/740#discussion_r933637162

src/serac/numerics/functional/functional.hpp
jamiebramwell: I'll add an issue on this.

jamiebramwell (approved):
Looks good to me! Great work!!

samuelpmishLLNL:
For reference, Jamie and I spoke on Friday about the tests in `serac_solid_functional` and decided to rebaseline many of them because :
1. the old "accepted" values were not properly verified in the first place
2. they also used a different interpretation of the body force (per unit volume in current vs. per unit volume in reference)
3. these tests are going to be imminently replaced and improved in favor of standardized patch tests with analytic / manufactured solutions.

PR #742, opened 2022-07-06
Adding diag operation for square rank 2 tensors

kswartz92:
This operation returns a matrix with the same diagonal entries as the input matrix, with zeros in the off-diagonal positions. It is needed for a linear, cubic solid material model.

### DIFF
--- a/src/serac/numerics/functional/tensor.hpp
+++ b/src/serac/numerics/functional/tensor.hpp
@@ -1065,15 +1065,15 @@ SERAC_HOST_DEVICE constexpr auto dev(const tensor<T, n, n>& A)
 }
 
 /**
- * @brief Returns a matrix (rank-2 tensor) containing the diagonal entries of the input matrix
+ * @brief Returns a square matrix (rank-2 tensor) containing the diagonal entries of the input square matrix
  * with zeros in the off-diagonal positions
- * @param[in] A The input matrix
+ * @param[in] A The input square matrix
  * This operation is used to compute a term in the constitutive response of a linear, cubic solid material
  */
 template <typename T, int n>
 SERAC_HOST_DEVICE constexpr auto diag(const tensor<T, n, n>& A)
 {
-  auto D = make_tensor<n, n>([](int, int) { return 0.0; });
+  auto D = make_tensor<n, n>([](int, int) {return 0.0;});
   for (int i = 0; i < n; i++) {
     D[i][i] = A[i][i];
   }
--- a/src/serac/numerics/functional/tensor.hpp
+++ b/src/serac/numerics/functional/tensor.hpp
@@ -1065,6 +1065,22 @@ SERAC_HOST_DEVICE constexpr auto dev(const tensor<T, n, n>& A)
 }
 
 /**
+ * @brief Returns a matrix (rank-2 tensor) containing the diagonal entries of the input matrix
+ * with zeros in the off-diagonal positions
+ * @param[in] A The input matrix
+ * This operation is used to compute a term in the constitutive response of a linear, cubic solid material
+ */
+template <typename T, int n>
+SERAC_HOST_DEVICE constexpr auto diag(const tensor<T, n, n>& A)
+{
+  auto D = make_tensor<n, n>([](int, int) { return 0.0; });
+  for (int i = 0; i < n; i++) {
+    D[i][i] = A[i][i];
+  }
+  return D;
+}
+
+/**
  * @brief Obtains the identity matrix of the specified dimension
  * @return I_dim
  */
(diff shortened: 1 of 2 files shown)


samuelpmishLLNL:
I like the idea of adding this feature, there is a related function in my solid mechanics update branch:
https://github.com/LLNL/serac/blob/bb0928abfff401463ba4d331acf00285c669d344/src/serac/physics/materials/tests/J2_material.cpp#L29-L35

One concern I have is that the function `diag` in MATLAB and numpy means either (depending on what the input type is)
`diag({a, b, c}) -> {{a, 0, 0}, {0, b, 0}, {0, 0, c}}` 
or
`diag({{a, e, f}, {g, b, h}, {i, j, c}}) -> {a, b, c}`

So I think it would be better to avoid conflicting with the existing convention by either:

- renaming the dense matrix -> diagonal matrix function in this PR to something else (maybe `DiagonalMatrix`)

 or 

- conform to the MATLAB-esque `diag` function convention, in which case the dense matrix -> diagonal matrix mapping in this PR would be equivalent to
```cpp
auto D = diag(diag(A));
```

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1064,6 +1064,22 @@ SERAC_HOST_DEVICE constexpr auto dev(const tensor<T, n, n>& A)
...
+/**
+ * @brief Returns a square matrix (rank-2 tensor) containing the diagonal entries of the input square matrix
+ * with zeros in the off-diagonal positions
+ * @param[in] A The input square matrix
+ * This operation is used to compute a term in the constitutive response of a linear, cubic solid material
+ */
+template <typename T, int n>
+SERAC_HOST_DEVICE constexpr auto diag(const tensor<T, n, n>& A)
+{
+  auto D = make_tensor<n, n>([](int, int) {return 0.0;});
```
samuelpmishLLNL: other possible implementations (that produce identical assembly for the 3x3 case): https://godbolt.org/z/c8nxGP897

personally, I prefer the implementation of `diag2` and `diag3` slightly

kswartz92: I'm okay with any of these implementations. I wasn't sure if the `tensor` constructor initialized values to 0, but it seems from looking at `diag2` that it does.

kswartz92:
> I like the idea of adding this feature, there is a related function in my solid mechanics update branch:
> 
> https://github.com/LLNL/serac/blob/bb0928abfff401463ba4d331acf00285c669d344/src/serac/physics/materials/tests/J2_material.cpp#L29-L35
> 
> One concern I have is that the function `diag` in MATLAB and numpy means either (depending on what the input type is) `diag({a, b, c}) -> {{a, 0, 0}, {0, b, 0}, {0, 0, c}}` or `diag({{a, e, f}, {g, b, h}, {i, j, c}}) -> {a, b, c}`
> 
> So I think it would be better to avoid conflicting with the existing convention by either:
> 
> * renaming the dense matrix -> diagonal matrix function in this PR to something else (maybe `DiagonalMatrix`)
> 
> or
> 
> * conform to the MATLAB-esque `diag` function convention, in which case the dense matrix -> diagonal matrix mapping in this PR would be equivalent to
> 
> ```c++
> auto D = diag(diag(A));
> ```

Yeah, I agree the naming could be confusing since it matches matlab/python names. I am on board with renaming it to something like `DiagonalMatrix`

btalamini:
For what it's worth, I agree with both of Sam's points: (1) This is a nice feature to have and we should get it in. (2) It shouldn't be named "diag" because of the more common usages Sam noted. I prefer the `diag2` implementation, but any of them are fine, really.

src/serac/numerics/functional/tensor.hpp
btalamini: The tensor constructor **doesn't** initialize values. It's the empty braces in `tensor<T, n, n> D{};` that ensure it is initialized to zero. I learned this the hard way. Just wanted to clear this up to save you some debugging later.

samuelpmishLLNL: `tensor` actually doesn't define any constructors, and since it just a plain-old-data struct then you get [zero-initialization](https://en.cppreference.com/w/cpp/language/zero_initialization) with the empty curly braces. Omitting the curly braces does leave the values uninitialized, like it does with other basic types in C++.

kswartz92: thanks for the clarification, definitely saved me some future debugging time!

samuelpmishLLNL (approved):

jamiebramwell (approved):
Looks great! Thanks @kswartz92 !

Issue #737, opened 2022-06-30
Add ability to rebaseline individual tests


white238:
Currently we only have the ability to rebaseline all the tests.  It would be nice to be able to list singular or comma-delimited tests to be rebaselined.

Something like:

```
./ats.sh --baseline=test1,test2
```

PR #743, opened 2022-07-13
Increase the wait for summary files in testing

jamiebramwell:
This increases the wait for summary files as it takes a while for these files to appear on LC during ats CI runs.

### DIFF
--- a/scripts/testing/tolerance_test.py
+++ b/scripts/testing/tolerance_test.py
@@ -35,7 +35,7 @@ def as_list(v):
 def ensure_file(path, wait_for_file=False):
     exists = os.path.exists(path)
     if wait_for_file:
-        for i in range(5):
+        for i in range(12):
             if exists:
                 break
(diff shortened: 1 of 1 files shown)


chapman39 (approved):
I think we should try figure out a better solution soon. When I get to https://github.com/LLNL/serac/issues/737 I could try and look into this issue again, so we don't have to keep increasing the sleep time.

white238:
I'll talk to LC about this but I've had this problem in the past.  LC, at the time, told me they guarantee files will be synced between nodes at most in 1 minute, usually its sooner.

I'm sure there are better answers to this and I'll ask around, but my current ideas require more work than this.  For example, wrapping serac with the tolerance test so it forces everything onto the same nodes. This could be complicated to get right.  Another idea is to add some logic into ATS directly that would allow you to run following checks on the same node as the previous test.

white238 (approved):

Issue #748, opened 2022-07-20
Enable Codevelop build for blueos


white238:
Currently getting a compile flag that is ill formed:

```
-Xcompiler= -Xcompiler=-fopenmp=libomp
```

the first `-Xcompiler=` causes the rest to be past to the host compiler, which cannot understand the second `-Xcompiler=`.

Also there are a lot of CMake warnings about our built zlib package.  For example:

```
CMake Warning at cmake/blt/cmake/BLTMacros.cmake:770 (add_executable):
  Cannot generate a safe runtime search path for target
  simple_conduction_with_input_file because files in some directories may
  conflict with libraries in implicit directories:

    runtime library [libz.so.1] in /usr/lib64 may be hidden by files in:
      /usr/WS2/smithdev/libs/serac/blueos_3_ppc64le_ib_p9/2022_06_01_21_11_45/clang-10.0.1/zlib-1.2.12-i6lqmmtb2os46rfflvu43zelz5sv2lcb/lib

  Some of these libraries may not be found correctly.
Call Stack (most recent call first):
  examples/simple_conduction/CMakeLists.txt:7 (blt_add_executable)
```

PR #751, opened 2022-07-25
Update Axom

### LINKED ISSUE #711: Spack is grabbing packages from other compiler specs
Look into this at a later date.

### LINKED ISSUE #752: Replace black/ whitelist to deny/ allowlist, respectively
Referring to this section here: https://github.com/LLNL/serac/blob/fbeabd9362f67fb6cc405ee3e4c137ca2187afb2/scripts/spack/packages/mfem/package.py#L172-L189

@tepperly mentioned in a LiDO merge request to update the phrasing of these variables, so I figured we should change them here, so I don't have to keep changing them after a LiDO tpl build.

white238:
Update Axom to include the following:

* CMake fix for exported targets coming from Axom
* `SLIC_*ROOT_*` macros now included with SLIC

Update Uberenv to include new fix for `--fresh` concretizer option which does not search for outside libraries in previous compiler specs. Fixes #711 

Update Spack due to `ldlibs` and `ldflags` not being used in `CachedCMakePackage`.  This has been pushed up to Spack proper and is waiting for review.

Remove white/black list terminology. Fixes #752.

Clean-up codevelop mfem includes.

Use system `zlib` on blueos systems. #748 (does not fully fix this but should address the multitude of zlib warnings)

### DIFF
--- a/cmake/thirdparty/SetupSeracThirdParty.cmake
+++ b/cmake/thirdparty/SetupSeracThirdParty.cmake
@@ -203,9 +203,12 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
 
         # Patch the mfem target with the correct include directories
         get_target_property(_mfem_includes mfem INCLUDE_DIRECTORIES)
-        target_include_directories(mfem SYSTEM INTERFACE $<BUILD_INTERFACE:${_mfem_includes}>)
+        target_include_directories(mfem SYSTEM INTERFACE ${_mfem_includes})
+        target_include_directories(mfem SYSTEM INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>)
         target_include_directories(mfem SYSTEM INTERFACE $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/mfem>)
-        
+
+        blt_print_target_properties(TARGET mfem)
+
         #### Restore previously stored data
         foreach(_tpl ${tpls_to_save})
             set(${_tpl}_DIR "${${_tpl}_DIR_SAVE}" CACHE PATH "" FORCE)
--- a/host-configs/rzgenie-toss_3_x86_64_ib-gcc@8.3.1.cmake
+++ b/host-configs/rzgenie-toss_3_x86_64_ib-gcc@8.3.1.cmake
@@ -11,11 +11,11 @@
 #------------------------------------------------------------------------------
 if(DEFINED ENV{SPACK_CC})
 
-  set(CMAKE_C_COMPILER "/usr/WS2/smithdev/libs/serac/toss_3_x86_64_ib/2022_06_01_21_10_59/spack/lib/spack/env/gcc/gcc" CACHE PATH "")
+  set(CMAKE_C_COMPILER "/usr/WS2/smithdev/libs/serac/toss_3_x86_64_ib/2022_07_22_20_32_34/spack/lib/spack/env/gcc/gcc" CACHE PATH "")
 
-  set(CMAKE_CXX_COMPILER "/usr/WS2/smithdev/libs/serac/toss_3_x86_64_ib/2022_06_01_21_10_59/spack/lib/spack/env/gcc/g++" CACHE PATH "")
+  set(CMAKE_CXX_COMPILER "/usr/WS2/smithdev/libs/serac/toss_3_x86_64_ib/2022_07_22_20_32_34/spack/lib/spack/env/gcc/g++" CACHE PATH "")
 
-  set(CMAKE_Fortran_COMPILER "/usr/WS2/smithdev/libs/serac/toss_3_x86_64_ib/2022_06_01_21_10_59/spack/lib/spack/env/gcc/gfortran" CACHE PATH "")
+  set(CMAKE_Fortran_COMPILER "/usr/WS2/smithdev/libs/serac/toss_3_x86_64_ib/2022_07_22_20_32_34/spack/lib/spack/env/gcc/gfortran" CACHE PATH "")
 
 else()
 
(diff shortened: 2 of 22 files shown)


cmake/thirdparty/SetupSeracThirdParty.cmake
```diff
@@ -203,9 +203,12 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
 
         # Patch the mfem target with the correct include directories
         get_target_property(_mfem_includes mfem INCLUDE_DIRECTORIES)
-        target_include_directories(mfem SYSTEM INTERFACE $<BUILD_INTERFACE:${_mfem_includes}>)
+        target_include_directories(mfem SYSTEM INTERFACE ${_mfem_includes})
```
white238, line 206: @chapman39 This was the cause of the weird include directories we were seeing friday.  I cannot explain how something that worked for 16 months just now started failing or how this was the cause.

For people that missed the fun, this was taking a list of include directories and making them system includes and inherited down the chain but magically started tacking on the `PROJECT_SOURCE_DIR` in a way that broke it.  For example:

```
/usr/workspace/white238/serac/repo<$BUILD_INTERFACE:/some;/list;/of;/includes>
```

when it should have just been this:

```
<$BUILD_INTERFACE:/some;/list;/of;/includes>
```

Note the extra path and missing the semi-colon between `repo` and `<$BUILD_INTERFACE`.

host-configs/rzgenie-toss_3_x86_64_ib-gcc@8.3.1.cmake
```diff
@@ -53,27 +53,27 @@ set(ENABLE_OPENMP ON CACHE BOOL "")
...
-set(AXOM_DIR "${TPL_ROOT}/axom-0.6.1.3-basyhw2gn6m2qbkglqvu6rv6o2ipsy2z" CACHE PATH "")
+set(AXOM_DIR "${TPL_ROOT}/axom-0.6.1.4-d5rbythhy7tuc2aihtzv5s52mbpynozy" CACHE PATH "")
 
 set(CAMP_DIR "${TPL_ROOT}/camp-2022.03.0-d5tbmmrjeftx6iopin4iab4b5vfk36s4" CACHE PATH "")
 
-set(CONDUIT_DIR "${TPL_ROOT}/conduit-0.8.3-uxzzvoai6ttvccjo4usntqtf2sg5sjew" CACHE PATH "")
+set(CONDUIT_DIR "${TPL_ROOT}/conduit-0.8.3-iwasyu6c2juyf6nkoppx3yl53g3wsuf2" CACHE PATH "")
 
-set(LUA_DIR "/usr/WS2/smithdev/libs/serac/toss_3_x86_64_ib/2022_06_01_21_10_59/clang-10.0.0/lua-5.3.5-xswudgamrkym66lww3pvvsl6g3fnkl4h" CACHE PATH "")
+set(LUA_DIR "${TPL_ROOT}/lua-5.3.5-e65dfjw7i24gddfjpgxpltcv5gx5m6sz" CACHE PATH "")
```
white238: Oh yea.. back to non-reuse.

chapman39:
Fixes #752 thanks

Edit: actually, it looks like this needs to be changed in serac, MFEM, and sundials ([sourcegraph](https://sourcegraph.com/search?q=context:global+repo:github.com/llnl/serac+blacklist&patternType=literal))

samuelpmishLLNL:
What is the significance of the (seemingly) random number/letter suffix on the individual libraries? e.g.

```cmake
set(AXOM_DIR "${TPL_ROOT}/axom-0.6.1.5-n2xnyhevwnspmjjf5ls5xaoac65ekdov" CACHE PATH "")
#                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I assumed initially that it was a commit hash or something, but the suffix in each .cmake file is different

white238:
> What is the significance of the (seemingly) random number/letter suffix on the individual libraries? e.g.
> 
> ```cmake
> set(AXOM_DIR "${TPL_ROOT}/axom-0.6.1.5-n2xnyhevwnspmjjf5ls5xaoac65ekdov" CACHE PATH "")
> #                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> I assumed initially that it was a commit hash or something, but the suffix in each .cmake file is different

It is a hash of the spack spec for that package.  I am not sure all the things that go into it but version, variants, compiler, dependencies, etc.

samuelpmishLLNL:
hmm, it seems horrible that a human has to go into each of these directories and manually copy and paste these unique identifiers everywhere. Would it be possible to remove those suffixes so we could deduplicate the common entries in all of these hostconfigs?

samuelpmishLLNL:
concretely, it seems like it would be nice to be able to define something like

```
# common_entries.cmake
set(AXOM_DIR "${TPL_ROOT}/axom-0.6.1.5" CACHE PATH "");
set(RAJA_DIR ... CACHE PATH "");
set(MFEM_DIR ... CACHE PATH "");
...
```

And then each specific host config could just have something like

```
# gcc@11.1.cmake
set(TPL_ROOT "/home/serac/serac_tpls/gcc-11.1.0" CACHE PATH "")

# only define the TPL structure once, rather than duplicating it in each .cmake file
include("path/to/common_entries.cmake")

# if we want to overwrite something defined in common_entries.cmake
set(AXOM_DIR "/some/other/path/to/axom/" CACHE PATH "" FORCE)
```

but appending a unique hash on each build of each library defeats this possibility

samuelpmishLLNL:
here's an example: https://gist.github.com/samuelpmishLLNL/5a8d356edde8fbe4c28a7cb1bc4c2599

white238:
> hmm, it seems horrible that a human has to go into each of these directories and manually copy and paste these unique identifiers everywhere. Would it be possible to remove those suffixes so we could deduplicate the common entries in all of these hostconfigs?

These files are auto-generated by our spack package.

samuelpmishLLNL:
> > hmm, it seems horrible that a human has to go into each of these directories and manually copy and paste these unique identifiers everywhere. Would it be possible to remove those suffixes so we could deduplicate the common entries in all of these hostconfigs?
> 
> These files are auto-generated by our spack package.

Ah, that makes way more sense!

samuelpmishLLNL (approved):

chapman39 (approved):

PR #753, opened 2022-07-26
Fix boundary condition bug

jamiebramwell:
This fixes a bug in the boundary condition class where the correct degrees of freedom were not being calculated properly for vector-valued functions.

### DIFF
--- a/src/serac/physics/tests/serac_solid_functional_boundary.cpp
+++ b/src/serac/physics/tests/serac_solid_functional_boundary.cpp
@@ -1,0 +1,117 @@
+// Copyright (c) 2019-2022, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include <fstream>
+
+#include "axom/slic/core/SimpleLogger.hpp"
+#include <gtest/gtest.h>
+#include "mfem.hpp"
+
+#include "serac/serac_config.hpp"
+#include "serac/mesh/mesh_utils.hpp"
+#include "serac/physics/state/state_manager.hpp"
+#include "serac/physics/solid_functional.hpp"
+#include "serac/physics/materials/solid_functional_material.hpp"
+#include "serac/physics/materials/parameterized_solid_functional_material.hpp"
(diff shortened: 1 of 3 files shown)


white238 (approved):

src/serac/physics/tests/serac_solid_functional_boundary.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  // Perform the quasi-static solve
+  double dt = 0.1;
+
+  solid_solver.advanceTimestep(dt);
+  solid_solver.outputState("paraview_output");
+
+  // Check the final displacement norm
+  EXPECT_NEAR(0.171720048, norm(solid_solver.displacement()), 1.0e-6);
+
```
samuelpmishLLNL: I don't want to be too much of a jerk here, but I said earlier that I will not approve any PRs that introduce tests like this. I understand this is a 1-line-of-code change PR, but we can't keep doing this.

Let's start doing basic patch tests or something-- linear elasticity is fine for testing this feature, so we have analytic solutions to compare the entire displacement solution to.

I just glanced at mfem and didn't see any obvious patch test meshes in their repo (?!) so I guess we have to write our own, but it's pretty easy to do.

jamiebramwell: I totally agree. I tried to be a bit better here by actually checking the exact dof values. I agree we need real patch tests.

white238: I created an issue so we can track this and start the conversation outside of a bugfix PR.

#754

samuelpmishLLNL (approved):

src/serac/physics/tests/serac_solid_functional_boundary.cpp
```diff
@@ -80,8 +84,18 @@ TEST(SolidFunctional, BoundaryCondition)
   solid_solver.advanceTimestep(dt);
   solid_solver.outputState("paraview_output");
 
-  // Check the final displacement norm
-  EXPECT_NEAR(0.171720048, norm(solid_solver.displacement()), 1.0e-6);
+  // compare approximate solution to the exact,
```
jamiebramwell, line 87: This is great! Thanks!!!

PR #762, opened 2022-08-11
Remove the mesh morphing from the `SolidFunctional` module

### LINKED ISSUE #760: Geometric nonlinearities flag causes invalid reference config
Currently the flag `GeometricNonlinearities::On` causes two distinct consequences in the solid mechanics module:

1. Finite deformation is accounted for in the simulation
2. When field output is dumped, the nodes of the mesh are moved to the current deformed stated

The Second one is enforced by changing the nodal coordinates of the mesh after a time step is complete. Then, at the start of the next timestep, the coordinates are moved back so that the reference configuration is again established. In between these operations, the reference configuration is left in an invalid state. Calls to, say, error norm calculations produce erroneous results because they are interpolated on the wrong mesh. There is no warning, the user is just expected to know this. 

Suggested change: Leave the coordinates of the mesh alone. If we want to plot on the deformed state, do this addition of the original coordinates and displacements in the output handling. Better yet, if Visit can deform the mesh with a given displacement field, drop this feature. 

If we want to do physics in an updated Lagrangian frame, this should be a separate option from  `GeometricNonlinearities`.

jamiebramwell:
To ensure consistency between physics modules, this PR removes the mesh mutation from the `SolidFunctional` module. This is OK for output as both ParaView and Sidre provide displace operators.

Resolves https://github.com/LLNL/serac/issues/760

### DIFF
--- a/src/serac/physics/solid_functional.hpp
+++ b/src/serac/physics/solid_functional.hpp
@@ -683,12 +683,6 @@ protected:
   /// @brief A flag denoting whether to compute geometric nonlinearities in the residual
   GeometricNonlinearities geom_nonlin_;
 
-  /// @brief Pointer to the reference mesh data
-  std::unique_ptr<mfem::ParGridFunction> reference_nodes_;
-
-  /// @brief Pointer to the deformed mesh data
-  std::unique_ptr<mfem::ParGridFunction> deformed_nodes_;
-
   /// @brief Coefficient containing the essential boundary values
   std::shared_ptr<mfem::VectorCoefficient> disp_bdr_coef_;
 
--- a/src/serac/physics/solid_functional.hpp
+++ b/src/serac/physics/solid_functional.hpp
@@ -99,7 +99,7 @@ public:
    */
   SolidFunctional(
       const SolverOptions& options, GeometricNonlinearities geom_nonlin = GeometricNonlinearities::On,
-      FinalMeshOption keep_deformation = FinalMeshOption::Deformed, const std::string& name = "",
+      const std::string&                                                                 name             = "",
       std::array<std::reference_wrapper<FiniteElementState>, sizeof...(parameter_space)> parameter_states = {})
       : BasePhysics(2, order, name),
         velocity_(StateManager::newState(FiniteElementState::Options{
@@ -114,8 +114,7 @@ public:
               nonlin_solver_, bcs_),
         c0_(0.0),
         c1_(0.0),
-        geom_nonlin_(geom_nonlin),
-        keep_deformation_(keep_deformation)
+        geom_nonlin_(geom_nonlin)
   {
     SLIC_ERROR_ROOT_IF(mesh_.Dimension() != dim,
                        axom::fmt::format("Compile time dimension and runtime mesh dimension mismatch"));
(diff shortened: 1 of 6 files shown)


samuelpmishLLNL:
Screen Shot 2022-08-11 at 9 16 14 AM 

👍 

----------

To be clear, now everything is always kept in the reference configuration?

samuelpmishLLNL:
Thanks for looking into this, I was confused by what was going on with all of these mesh transformations in this file, but I wasn't confident enough about what was going on to modify it in the previous solid mechanics update.

jamiebramwell:
> Screen Shot 2022-08-11 at 9 16 14 AM
> 
> 👍
> 
> To be clear, now everything is always kept in the reference configuration?

Yes - the mesh is now always in the reference configuration. I think in the future it might be nice to register a displacement field with the state manager so a physics developer can grab the deformed configuration if necessary, but we can develop that feature when it is needed.

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -375,12 +372,12 @@ void functional_parameterized_solid_test(double expected_disp_norm)
   EXPECT_NEAR(expected_disp_norm, norm(solid_solver.displacement()), 1.0e-6);
 }
 
-TEST(SolidFunctional, 2DLinearStatic) { functional_solid_test_static<1, 2>(1.5110743593501033); }
-TEST(SolidFunctional, 2DQuadStatic) { functional_solid_test_static<2, 2>(2.1864070695817928); }
-TEST(SolidFunctional, 2DQuadParameterizedStatic) { functional_parameterized_solid_test<2, 2>(2.1864070695817928); }
+TEST(SolidFunctional, 2DLinearStatic) { functional_solid_test_static<1, 2>(1.5110529858788075); }
```
jamiebramwell, line 375: Note - these norms changed because the integration occurs on the reference instead of the deformed configuration now. This is actually one of the side effects that inspired this PR.

samuelpmishLLNL (approved):

src/serac/physics/tests/serac_solid_functional.cpp
btalamini: Yes, it was patch test failures that caught this. Once that branch is merged, most of these tests will be deleted.

src/serac/physics/solid_functional.hpp
```diff
@@ -190,26 +181,7 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
-    auto mesh_fe_space = new mfem::ParFiniteElementSpace(displacement_.space(), &mesh_, mesh_fe_coll);
-    auto mesh_nodes    = new mfem::ParGridFunction(mesh_fe_space);
-    mesh_nodes->MakeOwner(mesh_fe_coll);
-
-    *mesh_nodes = *reference_nodes_;
-
-    // Set the mesh to the newly created nodes object and pass ownership
-    mesh_.NewNodes(*mesh_nodes, true);
-  }
+  ~SolidFunctional() {}
```
btalamini, line 184: The trivial destructor is an unexpected win. Even more evidence in favor of this change.

btalamini (approved):

PR #763, opened 2022-08-11
Clean up the direct solver option

jamiebramwell:
This cleans up the implementation of the SuperLU linear solver option. Instead of wrapping it as a nonlinear solver, it wraps it as a linear solver. This significantly cleans and streamlines the infrastructure in the `EquationSolver` class.


btalamini (approved):

jamiebramwell:
@samuelpmishLLNL , could you take a look at this? Thanks!

src/serac/numerics/equation_solver.hpp
```diff
@@ -116,49 +108,54 @@ class EquationSolver : public mfem::Solver {
                                                                const NonlinearSolverOptions& nonlin_options);
 
   /**
-   * @brief A wrapper class for combining a nonlinear solver with a SuperLU direct solver
+   * @brief A wrapper class for using the MFEM super LU solver with a HypreParMatrix
    */
-  class SuperLUNonlinearOperatorWrapper : public mfem::Operator {
+  class SuperLU : public mfem::Solver {
```
white238: I find this class name confusing. At first I thought it was using a class from the SuperLU library not from us. ``SuperLUSolver`` maybe? The previous was very verbose though...

jamiebramwell: Done.

white238 (approved):

PR #765, opened 2022-08-12
WIP: Add a sliding wall boundary condition

jamiebramwell:
This PR adds a sliding wall boundary condition. It works by rotating the frame of reference such that one of the normal of the sliding wall face is aligned with a coordinate axis. This allows us to use the MFEM-native component-wise displacement boundary condition.

It contains internal error checking to ensure that the indicated face has a consistent normal on all marked faces.

image


src/serac/numerics/functional/tensor.hpp
```diff
@@ -939,6 +939,25 @@ SERAC_HOST_DEVICE constexpr auto operator*(const tensor<S, m...>& A, const tenso
...
+template <typename S, typename T, int m>
+SERAC_HOST_DEVICE constexpr auto cross(const tensor<S, m>& A, const tensor<T, m>& B)
+{
+  static_assert(m == 3, "Cross products are only defined on vectors of dimension 3.");
+
+  tensor<decltype(S{} * T{}), m> cross_product{
+      {A[1] * B[2] - A[2] * B[1], A[2] * B[0] - A[0] * B[2], A[0] * B[1] - A[1] * B[0]}};
+
+  return cross_product;
+}
```
samuelpmishLLNL, line 959: Whoops, I don't know how we made it this far without a cross product!

If you want to enforce that dim == 3 (which we should), you can just put that directly in the function signature, and then you don't need the static assert,

```suggestion
template <typename S, typename T>
SERAC_HOST_DEVICE constexpr  tensor<decltype(S{} * T{}), 3> cross(const tensor<S, 3>& u, const tensor<T, 3>& v)
{
  return {u[1] * v[2] - u[2] * v[1], u[2] * v[0] - u[0] * v[2], u[0] * v[1] - u[1] * v[0]}};
}
```

It's also often useful to define a couple special cases of "2D" cross product:

```cpp
template <typename S, typename T>
SERAC_HOST_DEVICE constexpr  auto cross(const tensor<S, 2>& u, const tensor<T, 2>& v) {
  return u[0] * v[1] - u[1] * v[0]; // out-of-plane component, really just det({u, v})
}

template <typename T>
SERAC_HOST_DEVICE constexpr tensor<T, 2> cross(const tensor<T, 2>& u) {
  return {-u[1], u[0]}; // in-plane result of cross(vec3{0, 0, 1}, vec3{u[0], u[1], 0})
}
```

I'm writing this code in the browser, so it might have some mistakes

src/serac/physics/solid_functional.hpp
```diff
@@ -249,12 +249,174 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
+            for (int comp = 0; comp < normal.Size(); ++comp) {
+              normal(comp) = rank_normals[static_cast<std::size_t>(comp + (i * dim))];
+            }
+            attribute_normals.emplace(attr, normal);
+          }
+        }
+      }
+    }
+
+    // Loop over the computed attribute nomral vectors
```
samuelpmishLLNL:
```suggestion
    // Loop over the computed attribute normal vectors
```

src/serac/physics/solid_functional.hpp
```diff
@@ -608,6 +788,96 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
+      // Rotate the velocity
+      auto& velocity_gf = velocity_.gridFunction();
+      rotateGridFunction(*coordinate_transform_, velocity_gf);
+      velocity_.setFromGridFunction(velocity_gf);
+
+      // Rotate the adjoint displacement
+      auto& adjoint_disp = adjoint_displacement_.gridFunction();
+      rotateGridFunction(*coordinate_transform_, adjoint_disp);
+      adjoint_displacement_.setFromGridFunction(adjoint_disp);
+    }
```
samuelpmishLLNL, line 819: these rotate and undoRotate functions worry me for a few reasons:
- they seem pretty expensive (manually touching every single dof of many different fields)
- they're done frequently 
- the underlying implementation (rotateGridFunction) is CPU-specific

src/serac/physics/solid_functional.hpp
```diff
@@ -352,7 +521,14 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
     residual_->AddBoundaryIntegral(
         Dimension<dim - 1>{},
         [this, traction_function](auto x, auto n, auto, auto, auto... params) {
-          return -1.0 * traction_function(x, n, time_, params...);
+
+          auto traction_vector = traction_function(x, n, time_, params...);
+
+          if (coordinate_transform_) {
+            traction_vector = dot(*coordinate_transform_, traction_vector);
+          }
```
samuelpmishLLNL, line 529: this pattern doesn't work on a GPU, using the `this` pointer to get at members of `SolidFunctional` inside the q-function requires dereferencing a host pointer (it's also won't work to pass `time_` either).

src/serac/physics/solid_functional.hpp
```diff
@@ -249,12 +249,174 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
+            attribute_normals.emplace(attr, normal);
+          }
+        }
+      }
+    }
+
+    // Loop over the computed attribute nomral vectors
+    for (auto& normal : attribute_normals) {
+      SLIC_ERROR_ROOT_IF(dim != 2 && dim != 3, "Dimension must be 2 or 3 for sliding wall boundary conditions.");
+
```
samuelpmishLLNL, line 378: The code above mentions that it ensures the normals are all the same, and then we loop over multiple normals here. Do we need a loop, or is there just one normal direction?

jamiebramwell: The next step is adding multiple normals (i.e. multiple sliding walls).

src/serac/physics/solid_functional.hpp
jamiebramwell: Do you think this would be better in the residual calculation? The only problem is then modifying the stiffness would get complicated.

PR #769, opened 2022-08-16
WIP: Parameter buckets

jamiebramwell:
This is a first cut at adding "buckets" for different flavors of parameters. For example, you can specify material and load parameters separately enabling parameterized loads and materials to be used at the same time with out each class knowing about the other's parameter list. 

Feedback is welcome!!

### DIFF
--- a/src/serac/physics/common.hpp
+++ b/src/serac/physics/common.hpp
@@ -20,7 +20,17 @@ namespace serac {
  * user-controlled parameters (e.g. for design optimization)
  */
 template <typename... T>
-struct Parameters {
+struct MaterialParameters {
+  static constexpr int n = sizeof...(T);  ///< how many parameters were specified
+};
+
+template <typename... T>
+struct ShapeParameter {
+  static constexpr int n = sizeof...(T);  ///< how many parameters were specified
+};
+
+template <typename... T>
+struct LoadParameters {
   static constexpr int n = sizeof...(T);  ///< how many parameters were specified
 };
 
--- a/src/serac/physics/solid_functional.hpp
+++ b/src/serac/physics/solid_functional.hpp
@@ -67,14 +67,18 @@ const SolverOptions default_dynamic_options = {
  * @tparam order The order of the discretization of the displacement and velocity fields
  * @tparam dim The spatial dimension of the mesh
  */
-template <int order, int dim, typename parameters = Parameters<>,
-          typename parameter_indices = std::make_integer_sequence<int, parameters::n>>
+template <int order, int dim, typename mat_parameters = MaterialParameters<>,
+          typename load_parameters        = LoadParameters<>,
+          typename mat_parameter_indices  = std::make_integer_sequence<int, mat_parameters::n>,
+          typename load_parameter_indices = std::make_integer_sequence<int, load_parameters::n>>
 class SolidFunctional;
 
 /// @overload
-template <int order, int dim, typename... parameter_space, int... parameter_indices>
-class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_sequence<int, parameter_indices...>>
-    : public BasePhysics {
+template <int order, int dim, typename... mat_parameter_space, typename... load_parameter_space,
+          int... mat_parameter_indices, int... load_parameter_indices>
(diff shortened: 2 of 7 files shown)


samuelpmishLLNL:
Another possible design is one where the user specifies which parameters they care about for each calculation. For example, something like:

```cpp
template < typename ... parameters >
struct SolidMechanics {

    template < int ... which_parameters_to_use, typename material_model >
    void addMaterial(material_model material) {
        // ...
    }

    template < int ... which_parameters_to_use, typename function >
    void addTraction(function T) {
        // ...
    }

};

// dummy types for the sake of illustration
struct FiberOrientation{};
struct CoefThermalExpansion{};
struct Temperature{};
struct Pressure{};

int main() {

    SolidMechanics < FiberOrientation, CoefThermalExpansion, Temperature, Pressure > simulation;

    // index 0 corresponds to the fiber orientation parameter, so it get passed in to this q-function
    simulation.addMaterial<0>([=](auto displacement, auto fiber){
        auto du_dx = get<1>(displacement); 
        auto n = get<0>(fiber);
        return stress_anisotropic_material(du_dx, n);
    });

    // indices 1 and 2 correspond to thermal expansion coefficient and temperature parameters,
    // so those get passed in to this q-function
    simulation.addMaterial<1, 2>([=](auto displacement, auto CTE, auto temperature){
        auto alpha = get<0>(CTE);
        auto theta = get<0>(temperature);
        auto du_dx = get<1>(displacement); 
        return stress_for_thermomechanical_model(du_dx, alpha, theta);
    });

    // index 3 corresponds to the pressure parameter, so it get passed in to this boundary q-function
    simulation.addTraction<3>([=](auto x, auto n, auto pressure){
        auto p = get<0>(pressure);
        return p * n;
    });

}
```

This would give finer control over precisely which parameters matter for each calculation. The part I'm unsure about right now is that it would be clearer to write `addTraction< Pressure >(...)` instead of using its index.

It would also require some changes to the `DomainIntegral` and `BoundaryIntegral` types, but I don't think it would be too bad.

jamiebramwell:
> Another possible design is one where the user specifies which parameters they care about for each calculation. For example, something like:
> 
> ```c++
> template < typename ... parameters >
> struct SolidMechanics {
> 
>     template < int ... which_parameters_to_use, typename material_model >
>     void addMaterial(material_model material) {
>         // ...
>     }
> 
>     template < int ... which_parameters_to_use, typename function >
>     void addTraction(function T) {
>         // ...
>     }
> 
> };
> 
> // dummy types for the sake of illustration
> struct FiberOrientation{};
> struct CoefThermalExpansion{};
> struct Temperature{};
> struct Pressure{};
> 
> int main() {
> 
>     SolidMechanics < FiberOrientation, CoefThermalExpansion, Temperature, Pressure > simulation;
> 
>     // index 0 corresponds to the fiber orientation parameter, so it get passed in to this q-function
>     simulation.addMaterial<0>([=](auto displacement, auto fiber){
>         auto du_dx = get<1>(displacement); 
>         auto n = get<0>(fiber);
>         return stress_anisotropic_material(du_dx, n);
>     });
> 
>     // indices 1 and 2 correspond to thermal expansion coefficient and temperature parameters,
>     // so those get passed in to this q-function
>     simulation.addMaterial<1, 2>([=](auto displacement, auto CTE, auto temperature){
>         auto alpha = get<0>(CTE);
>         auto theta = get<0>(temperature);
>         auto du_dx = get<1>(displacement); 
>         return stress_for_thermomechanical_model(du_dx, alpha, theta);
>     });
> 
>     // index 3 corresponds to the pressure parameter, so it get passed in to this boundary q-function
>     simulation.addTraction<3>([=](auto x, auto n, auto pressure){
>         auto p = get<0>(pressure);
>         return p * n;
>     });
> 
> }
> ```
> 
> This would give finer control over precisely which parameters matter for each calculation. The part I'm unsure about right now is that it would be clearer to write `addTraction< Pressure >(...)` instead of using its index.
> 
> It would also require some changes to the `DomainIntegral` and `BoundaryIntegral` types, but I don't think it would be too bad.

I like this design a lot. We should discuss pros and cons with the larger group today in the meeting.

jamiebramwell:
Closed in favor of more fine-tuned parameter selection in https://github.com/LLNL/serac/pull/782

PR #770, opened 2022-08-17
Update build instructions for macOS

btalamini:
Use MacPorts.


src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+Activate the particular package versions with MacPorts:
 
 .. code-block:: bash
 
-  $ brew install gcc@8
-  $ brew install llvm@11
-  $ brew install mpich
+   $ sudo port select clang mp-clang-12
+   $ sudo port select mpi openmpi-clang-fortran
+   $ sudo port select gcc mp-gcc12
```
samuelpmishLLNL: Is it a requirement to use macports instead of brew, and if so, why?

btalamini: No, it's not a requirement. On [this line](https://github.com/LLNL/serac/blob/talamini/docs/update_macOS_build/src/docs/sphinx/quickstart.rst?plain=1#L279) I state that it's *one* way to get needed packages.
I'm not against having Homebrew instructions as well, but someone needs to work through that process and then maintain those docs. I'm willing to give you a hand if you want to try a Homebrew build.

samuelpmishLLNL:
a built copy of these documents is available [here](https://serac.readthedocs.io/en/talamini-docs-update_macos_build/sphinx/quickstart.html#building-serac-dependencies-on-macos)

samuelpmishLLNL:
I'm struggling to follow this guide. There are useful parts, but there enough missing steps that it's not clear what I should be doing, and I'm just at the point of trying to install and symlink the stuff from MacPorts.

```sh
sam@Sams-Air darwin_monterey % sudo port select mpi openmpi-clang-fortran
Selecting 'openmpi-clang-fortran' for 'mpi' failed: The specified version 'openmpi-clang-fortran' is not valid.
```
-----------

The packages.yaml also contains entries of libraries that were not instructed to be installed by MacPorts (`tar`, `perl`, `netlib-lapack`), where should I get these?

I'm dreading having to manually go into that `packages.yaml` file and manually edit the specific directories and versions of each of the 17 packages. Why do I have to do this?

-------------------

instead of 
>The operating_system field should be set to `bigsur` for macOS 11. Refer to the spack documentation for other versions of macOS.

we should inline some of the options users are likely to need, since spack's documentation is unintuitive to navigate:   
```
"mojave", "catalina", "bigsur", "monterey", "ventura"
```
These are the values scraped from spack's readthedocs, but I can't confirm if they work or not, because I can't get to that step yet.

Similarly, I think "aarch64" is the appropriate entry for "target" on the newer machines, but I couldn't find good information about this from the spack documentation or check that it works.

--------------

I've installed several gigabytes of compilers and tools I already have, since this guide is only written from the perspective of MacPorts (in my experience, `brew` is the more commonly-used tool). I was alarmed to see that `port install openmpi-clang12` went on to install 3 separate versions of LLVM, several versions of clang, etc.

-----------

This comment isn't directed at Brandon or the content of this PR, but: _man_ this TPL build system is infuriating. I've spent several hours today trying to follow these steps (and this is after the additional clarification provided by the PR) and have very little to show for it. I can't imagine anyone not being paid by LLNL actually jumping through all of these hoops.

btalamini:
> I'm struggling to follow this guide. There are useful parts, but there enough missing steps that it's not clear what I should be doing, and I'm just at the point of trying to install and symlink the stuff from MacPorts.
> 
> ```shell
> sam@Sams-Air darwin_monterey % sudo port select mpi openmpi-clang-fortran
> Selecting 'openmpi-clang-fortran' for 'mpi' failed: The specified version 'openmpi-clang-fortran' is not valid.
> ```

I'm in agreement that this deserves editing for clarity. There was essentially nothing about the Mac build before, so I did my reasonable best effort at capturing everything I had to do. Let's capture the parts that gave you trouble and add as necessary. I'm willing to help on Teams to work through the rough patches.

> 
> The packages.yaml also contains entries of libraries that were not instructed to be installed by MacPorts (`tar`, `perl`, `netlib-lapack`), where should I get these?
> 

On my system, getting the ports I documented got the other things as dependencies. Relying on this is probably not wise, since MacPorts could change their packages. On the other hand, filling in the others will require identifying the port names for everything, and as you note below, there's an awful lot to specify.

> I'm dreading having to manually go into that `packages.yaml` file and manually edit the specific directories and versions of each of the 17 packages. Why do I have to do this?
> 

Only the spack maintainers know the answer to this one. Seems like it should be possible to specify `/opt/local` as a global prefix, and then only to override the few that are different. May be worth opening an issue on spack.

> instead of
> 
> > The operating_system field should be set to `bigsur` for macOS 11. Refer to the spack documentation for other versions of macOS.
> 
> we should inline some of the options users are likely to need, since spack's documentation is unintuitive to navigate:
> 
> ```
> "mojave", "catalina", "bigsur", "monterey", "ventura"
> ```
> 
> These are the values scraped from spack's readthedocs, but I can't confirm if they work or not, because I can't get to that step yet.
> 
> Similarly, I think "aarch64" is the appropriate entry for "target" on the newer machines, but I couldn't find good information about this from the spack documentation or check that it works.
>

Agree with these, but couldn't test them, and didn't want to say more than I knew to be true.

> I've installed several gigabytes of compilers and tools I already have, since this guide is only written from the perspective of MacPorts (in my experience, `brew` is the more commonly-used tool). I was alarmed to see that `port install openmpi-clang12` went on to install 3 separate versions of LLVM, several versions of clang, etc.

I won't argue the relative merits of one Apple package manager versus another; people can use whatever they want. I just wrote down the process using the one *I* know how to use. Again, happy to assist with a Homebrew-based build (but I don't have the time to try a Homebrew build on my own just for the sake of the doc page).

goxberry:
> Similarly, I think "aarch64" is the appropriate entry for "target" on the newer machines

On an M1 Mac, I'd use `arch=m1`.

> but I couldn't find good information about this from the spack documentation or check that it works.

See https://spack.readthedocs.io/en/latest/basic_usage.html#support-for-specific-microarchitectures for supported values of `arch`.

goxberry:
> > I'm dreading having to manually go into that packages.yaml file and manually edit the specific directories and versions of each of the 17 packages. Why do I have to do this?


> Only the spack maintainers know the answer to this one. Seems like it should be possible to specify /opt/local as a global prefix, and then only to override the few that are different. May be worth opening an issue on spack.

Because that's usually not enough information. In addition to packages and versions, introspection also needs to be done on variants. This issue has come up before with Spack maintainers, which is why the [`spack external find`](https://spack.readthedocs.io/en/latest/build_settings.html#automatically-find-external-packages) command exists. It doesn't work for every package -- [logic must be added to a package to do the required introspection](https://spack.readthedocs.io/en/latest/packaging_guide.html#making-a-package-discoverable-with-spack-external-find) -- but this problem is known, and is being addressed for some of the packages at the ~~root~~ leaves of the dependency DAG (e.g., MPI implementations).

src/docs/sphinx/quickstart.rst
goxberry: OpenMPI tends to be better supported by macOS package managers, so if there are `brew`-based instructions, I'd lean towards installing OpenMPI. At one point, I believe an old version of OpenMPI was distributed with some of the macOS developer tools, which I think was the justification for those package managers' MPI preferences.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+.. code-block:: bash
+
+   $ python3 -m venv <virtual-env-name>
+
+Since spack will need ``clingo``, activate the python vitual environment and 
+install the package to meet this dependency:
+
+.. code-block:: bash
+
+   $ source <virtual-env-name>/bin/activate
```
goxberry: I would skip this activation step and tell users to run `<virtual-env-name>/bin/pip3 install clingo` to avoid having to worry about the state of environment variables.

Similarly, when invoking the `uberenv.py` script, I'd tell users to run `<virtual-env-name/bin/python3 ./scripts/uberenv/uberenv.py --prefix=../path/for/TPLs` etc., to avoid the environment variables state issues.

Unexpected environment state is a really common failure mode, which is why many package managers go out of their way to sandbox environment variables; it will save you a lot of headaches if you sidestep this issue entirely by advising users to use absolute paths instead of activating an environment.

samuelpmishLLNL:
> > > I'm dreading having to manually go into that packages.yaml file and manually edit the specific directories and versions of each of the 17 packages. Why do I have to do this?
> 
> > Only the spack maintainers know the answer to this one. Seems like it should be possible to specify /opt/local as a global prefix, and then only to override the few that are different. May be worth opening an issue on spack.
> 
> Because that's usually not enough information. In addition to packages and versions, introspection also needs to be done on variants. This issue has come up before with Spack maintainers, which is why the [`spack external find`](https://spack.readthedocs.io/en/latest/build_settings.html#automatically-find-external-packages) command exists. It doesn't work for every package -- [logic must be added to a package to do the required introspection](https://spack.readthedocs.io/en/latest/packaging_guide.html#making-a-package-discoverable-with-spack-external-find) -- but this problem is known, and is being addressed for some of the packages at the ~root~ leaves of the dependency DAG (e.g., MPI implementations).

It seems untenable that our build system requires a human to manually specify that `readline` is at exactly version 8.1.2.000. The argument that this sort of thing is necessary doesn't make any sense, since every other package manager I've used figures this sort of thing out automatically.

On a related note, how am I supposed to know how to format the library version in that packages.yaml file?
When I install, say, unzip with MacPorts, it tells me that it's installed version `unzip-6.0_4`. The packages.yaml file is formatted as `unzip@6.0`. Does the _4 matter? How am I supposed to know when to truncate the version string, and by how much? By contrast, entries like `readline@8.1.2.000` have totally different formatting. How many trailing zeroes are necessary? Why does `perl` have a "v" in its version prefix and other libraries don't? 

If we're expecting users to do this manually 17 times, we have to actually explain what is expected of them.

-----------

In contrast: the other day, I was looking for a tool to help me inspect what was taking up space in our test binaries. A quick google search led me to `bloaty`, which I had never heard of before. My experience with `bloaty` went something like this:

1. They have a github page. I clone their repository with `--recursive` since I saw it has a .gitmodule file in the repo.
2. The project contains a CMakeLists.txt file, so I just configure `cmake` and build it.
3. The dependencies and main executable build successfully in ~30 seconds.
4. `bloaty --help` helps me discover which flags to use in order to find the information I need.
5. Identify that serac's binaries are 95% debug symbols from TPLs installed by `spack`.

I went from having never heard of this tool, to building it (and all of its dependencies) without needing any documentation, to using the tool to solve the problem I had, and it all took a matter of minutes. In serac, it's taking me hours just to process one of many steps in a guide to build just the TPLs, let alone the project itself. And I'm pretty sure I've done this step wrong, because it's completely unclear what needs to actually happen.

Why is it the case that building `serac` is so painful, and yet `bloaty` builds so effortlessly?

goxberry:
> It seems untenable that our build system requires a human to manually specify that readline is at exactly version 8.1.2.000. The argument that this sort of thing is necessary doesn't make any sense, since every other package manager I've used figures this sort of thing out automatically.

Most package managers I've used don't try to integrate with software installed outside the package management system. re: version detection, that is typically the province of build configuration systems. But Spack is trying to leverage packages installed external to it, so it needs enough information to determine whether package prerequisites are satisfied by those external packages.

Attempting to use external packages is optional. You don't *need* to do it if you're willing to have Spack build more software.

jamiebramwell:
This thread is getting a bit heated. I would encourage revisiting our [code of conduct](https://github.com/LLNL/serac/blob/develop/CODE-OF-CONDUCT.md), particularly the statements about respectful discourse and empathy.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+   $ pip install clingo
+
+The python virtual environment must be active when invoking the ``uberenv.py`` script below.
+
+Configuring spack
+^^^^^^^^^^^^^^^^^
+
+In ``compilers.yaml``:
```
goxberry: It should be possible to automate this step. `spack compiler find` would be the command; if the compilers aren't found in the default search paths, you should be able to run `spack compiler find /opt/local` or `spack compiler find /opt/local/bin` to generate it.

Homebrew typically required supplying a path to `spack compiler find` for any of the "versioned" compilers e.g., `gcc@8`, so you'd have to give the install prefix up to the level above `bin`, e.g, `/opt/Homebrew/Cellar/gcc@8/8.5.0/` should work, I think.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+   packages:
+   all:
+      compiler: [clang, gcc]
+      providers:
+         blas: [netlib-lapack]
+         lapack: [netlib-lapack]
+         mpi: [openmpi]
+
+   mpi:
+      buildable: false
```
goxberry, line 369: On a macOS or personal machine, the `buildable: false` part isn't as necessary as it is on an HPC machine; if automating the construction of enough of `packages.yaml` isn't feasible, building it with Spack might be worth the time versus developer frustration tradeoff. I can also see the appeal of leaving it, for the sake of pattern matching and consistency.

src/docs/sphinx/quickstart.rst
goxberry: You may also need to add a `--scope` flag so that command creates/modifies that file in the right place.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+   openmpi:
+      externals:
+      - spec: openmpi@4.1.4
```
goxberry, line 372: In theory, you should be able to use `spack external find openmpi`, but you may have to add flags, like `-p /opt/local` and `--scope` with whatever scope you're putting the `packages.yaml` in -- `uberenv`'s configuration might dictate what the right scope is (e.g., `system`, `user`). See https://spack.readthedocs.io/en/latest/command_index.html#spack-external-find for details.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+         prefix: /opt/local
+   
+   netlib-lapack:
```
goxberry, line 375: Looks like this package doesn't have `spack external find` hooks, but `openblas` does, if using it in place of `netlib-lapack` is an option.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+      buildable: false
+      externals:
+      - spec: netlib-lapack@3.10.1
+         prefix: /opt/local
+   autoconf:
```
goxberry, line 380: Should be `spack external find`-able.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+      buildable: false
+      externals:
+      - spec: autoconf@2.71
+         prefix: /opt/local
+   automake:
```
goxberry, line 385: Also `spack external find`-able.

goxberry: Of the rest of these, it looks like `gettext`, `libtool`, `m4`, `perl`, `pkg-config`, and `tar` could be found with `spack external find` (and appropriate flags).

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+
+* cmake (CMake is present by default on current MacOS distributions, but is often out of date, hence we recommend installing this package.)
+* cppcheck
+* doxygen
+
+Then, append the following to ``packages.yaml``:
+
+.. code-block:: yaml
+
+   cmake:
```
goxberry: All of these are `spack external find`-able except `cppcheck`.

src/docs/sphinx/quickstart.rst
white238: I copy/pasted a known working meta-package (or whatever these are called).  There was no particular reason we set this to netlib-lapack.

white238:
@goxberry Thanks for the input! It's good to know that a lot of these could have been found with `spack external find`.  I am trying to move to environments as quickly as possible which will clean up a lot of these difficulties i think but as usual other things pop up the priority tree.

white238:
Please remember when reviewing these steps that this was a first cut at documenting what we had to do to get a working build at all.  The intention was to document these steps so we didn't lose found knowledge then iterate to improve the process.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+   $ sudo port select mpi openmpi-clang-fortran
+   $ sudo port select gcc mp-gcc12
 
-Build third-party libraries via Uberenv:
+This step tells MacPorts to make symbolic links in your path so that, for example, ``clang`` will invoke the clang12 from the MacPorts package and not the ``clang`` shipped by Apple. 
+It also sets up a set of symlinks so that Clang, GCC, and the MPI wrappers all work without you having to muck with environment variables to locate header files and libraries. 
+It may be possible to skip this step and give full paths to your compilers in ``compilers.yaml`` (instead of the symlinks ``/opt/local/bin/clang``, etc.), but we haven't tried this.
 
-``python3 scripts/uberenv/uberenv.py --spec=%apple-clang@11.0.1 --prefix=../path/to/install``
+While building ParMetis, spack invokes the MPI compiler wrapper with the ``mpic++`` command, but MacPorts does not create this particular synonym for OpenMPI.
```
white238:
```suggestion
While building ParMetis, Spack invokes the MPI compiler wrapper with the ``mpic++`` command, but MacPorts does not create this particular synonym for OpenMPI. 
```

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+I don't like to let software packages change my Python environment, particularly by adding or removing packages with ``pip``, as this can lead to version dependency conflicts.  
+I create a Python virtual environment in the serac root directory with the built-in ``venv`` module. 
+
+.. (I use the name "venv" for the virtual environment; We might consider adding `venv/` to the list of git ignores in the root directory).
+
+.. code-block:: bash
+
+   $ python3 -m venv <virtual-env-name>
+
+Since spack will need ``clingo``, activate the python vitual environment and
```
white238:
```suggestion
Since Spack will need ``clingo``, activate the python vitual environment and 
```

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+         cc: /opt/local/bin/clang
+         cxx: /opt/local/bin/clang++
+         f77: /opt/local/bin/gfortran
+         fc: /opt/local/bin/gfortran
+      spec: clang@12.0.1
+      target: x86_64
+
+NOTES: 
+
+* The `operating_system` field should be set to ``bigsur`` for macOS 11. Refer to the spack documentation for other versions of macOS.
```
white238:
```suggestion
* The ``operating_system`` field should be set to ``bigsur`` for macOS 11. Refer to the spack documentation for other versions of macOS.
```

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+Notes:
+
+* OpenGL is not supported on modern Macs, hence those packages are not specified as they are on other platforms. 
+  This means that the optional ``glvis`` spec can't be built.
+* The version specs should be set to the actual versions of the packages you have, which will not neccesarily be the same as the above.
+
+The above spack settings and MacPorts packages will cover the basic installation of Serac. 
+If you want to build the optional devtools, you should install the additional packages with MacPorts:
+
+* cmake (CMake is present by default on current MacOS distributions, but is often out of date, hence we recommend installing this package.)
```
white238, line 473: This is required by Serac build and probably shouldn't be in this list.  It is also listed in the original list of required packages.

white238:
```suggestion
```

CMake is required by Serac build and probably shouldn't be in this list.  It is also listed in the original list of required packages.

goxberry:
I've tried to add some additional comments where parts of config files can be automatically generated using Spack commands. All of the comments assume use of whatever the commit of the development branch was at the time I made those comments.

I agree with @samuelpmishLLNL's point that asking users to manually construct more than small snippets of these files -- like the first several lines of `packages.yaml` -- is probably too much, even though I can appreciate the time and effort that goes into writing up an install guide like the one in this PR. In addition, manually writing these files is likely to be a challenge to maintain for both users and Serac developers. While I personally don't have the time to work through the directions and figure out the right combinations of flags and arguments needed to automatically generate the configuration files, I think doing so has the potential to address some of @samuelpmishLLNL's concerns, and it may make the install guide easier to maintain in the process. I also think these commands would address @btalamini's point that users should be able to supply an install prefix so that Spack can detect these externals.

For the packages that can't be detected automatically, is dropping them from `packages.yaml` an option? The increase in TPL build times could be worth avoiding the hassle of trying to debug manually-created configuration files.

src/docs/sphinx/quickstart.rst
white238: Can you move this whole cmake section to the above packages.yaml section

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+
+The invocation of ``uberenv.py`` is slightly modified from the standard instructions above: 
+
+.. code-block:: bash
+
+   $ ./scripts/uberenv/uberenv.py --prefix=../path/for/TPLs --spec="clang@12.0.1 ^openmpi@4.1.4"
+
+Notice the caret with the MPI spec. 
+This is necessary to force spack to use the installed MPI. 
+Without this, current versions of spack ignore the ``packages.yaml`` file and try to build a version of MPI from source.
```
white238:
```suggestion
Without this, current versions of Spack ignore the ``packages.yaml`` file and try to build a version of MPI from source.
```

goxberry:
@white238 No problem! I agree; environments have the potential to clean up build instructions significantly.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+         prefix: /usr
+   zlib:
+      buildable: false
+      externals:
+      - spec: zlib@1.2.12
+         prefix: /opt/local
+
+Notes:
+
+* OpenGL is not supported on modern Macs, hence those packages are not specified as they are on other platforms.
```
goxberry: Depends on what you mean by "supported". macOS still supports them, but Spack doesn't use the correct locations when it attempts to look up headers and libraries in the `opengl` package on macOS 12.5 (and possibly earlier versions).

OpenGL has been deprecated since macOS 10.14, but the XCode Command Line Tools still install OpenGL 4.1 and GLU 1.3. On earlier versions of macOS, it used to be possible to specify the `opengl` package in `packages.yaml` with an install prefix. On newer versions of macOS (certainly macOS 12.5 on an M1 Mac), those paths have changed. The paths in the Spack `opengl` package could be updated using the path information in https://github.com/ziglang/zig/issues/2208#issuecomment-582040352, but it's probably not worth the hassle unless someone really wants to use `glvis` on macOS.

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
-Install required compilers and MPI via `homebrew <https://brew.sh/>`_:
+Activate the particular package versions with MacPorts:
 
 .. code-block:: bash
 
-  $ brew install gcc@8
-  $ brew install llvm@11
-  $ brew install mpich
+   $ sudo port select clang mp-clang-12
+   $ sudo port select mpi openmpi-clang-fortran
```
chapman39: should be

`sudo port select mpi openmpi-clang12-fortran`

instead of 

`sudo port select mpi openmpi-clang-fortran`

src/docs/sphinx/quickstart.rst
chapman39: it may be a good idea to specify which `compilers.yaml`/ `packages.yaml` when starting this segment, like `scripts/spack/configs/darwin/compilers.yaml`

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+* For ``<path/to/python/venv>``, specify the the virtual environment directory.
+
+Building dependencies
+^^^^^^^^^^^^^^^^^^^^^
+
+The invocation of ``uberenv.py`` is slightly modified from the standard instructions above: 
+
+.. code-block:: bash
+
+   $ ./scripts/uberenv/uberenv.py --prefix=../path/for/TPLs --spec="clang@12.0.1 ^openmpi@4.1.4"
```
chapman39: should be `./scripts/uberenv/uberenv.py --prefix=../path/for/TPLs --spec="%clang@12.0.1 ^openmpi@4.1.4"` (needed `%` by clang)

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,268 @@ must be specified using either:
...
+The python virtual environment must be active when invoking the ``uberenv.py`` script below.
+
+Configuring spack
+^^^^^^^^^^^^^^^^^
+
+In ``compilers.yaml``:
+
+.. code-block:: yaml
+
+   - compiler:
```
chapman39, line 330: needs `compilers:` a line above `- compiler:`

goxberry: In the spirit of @white238's comment re: capturing existing institutional knowledge, this omission may be an artifact of using an older version of Spack configuration schema, which has changed at least once. 

IIRC, there are some automated tools to migrate config files using older schemas to the current schema. [`spack config update`](https://spack.readthedocs.io/en/latest/command_index.html#spack-config-update) looks like the relevant command (based on the context of https://github.com/spack/spack/issues/18031), but the documentation for that command is...sparse.

btalamini: Thanks. Fixed.

src/docs/sphinx/quickstart.rst
btalamini: Good catch. Fixed.

src/docs/sphinx/quickstart.rst
btalamini: I did this, and added more useful information (if, like me, you don't like having unstaged changes dangling in git)

src/docs/sphinx/quickstart.rst
btalamini: Thanks! updated.

src/docs/sphinx/quickstart.rst
btalamini: Yeah, good catch. I was lazy about distinguishing Spack, the software, from `spack`, the executable. I changed all the instances.

btalamini:
> I'm struggling to follow this guide. There are useful parts, but there enough missing steps that it's not clear what I should be doing, and I'm just at the point of trying to install and symlink the stuff from MacPorts.
> 
> ```shell
> sam@Sams-Air darwin_monterey % sudo port select mpi openmpi-clang-fortran
> Selecting 'openmpi-clang-fortran' for 'mpi' failed: The specified version 'openmpi-clang-fortran' is not valid.
> ```

Sam, Alex caught the error here. The command had a typo in it. I'm sorry about that.

src/docs/sphinx/quickstart.rst
chapman39: might've just been a type-o, since the original `packages.yaml` file included the `compilers:` line.

i see what you mean about that command though. i tried using it for fun to see if it'd actually add the `compilers:` line itself. apparently these are the options (even though the documentation doesn't say): `compilers concretizer mirrors repos packages modules config upstreams bootstrap`

i ran
`spack config update configs`
...and got `==> No updates needed for "config" section.`

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,279 @@ must be specified using either:
...
+   openmpi:
+      externals:
+      - spec: openmpi@4.1.4
+         prefix: /opt/local
+   
+   netlib-lapack:
+      buildable: false
+      externals:
+      - spec: netlib-lapack@3.10.1
+         prefix: /opt/local
```
chapman39: the extra space behind `prefix` will cause `spack spec` to fail!
```suggestion
        prefix: /opt/local
```

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,279 @@ must be specified using either:
...
+   autoconf:
+      buildable: false
+      externals:
+      - spec: autoconf@2.71
+         prefix: /opt/local
+   automake:
+      buildable: false
+      externals:
+      - spec: automake@1.16.5
+         prefix: /opt/local
```
chapman39: i noticed more of the `prefix` sections throughout `packages.yaml` needs an extra space removed (so that it aligns with spec). after that i think everything else lgtm

src/docs/sphinx/quickstart.rst
btalamini: Thanks, I did this one. It will show up in the next push.

src/docs/sphinx/quickstart.rst
btalamini: done.

src/docs/sphinx/quickstart.rst
btalamini: Oo, that was a nasty one. I think there were actual TABS in there instead of spaces. Should be good now. Nice catch!

src/docs/sphinx/quickstart.rst
btalamini: Efficiency issues aside, I was unable to get Spack to successfully build MPI on my system, so the flag was necessary for me. Maybe it will succeed once we automate more of the dependencies.

src/docs/sphinx/quickstart.rst
btalamini: I can see why setting the environment state could be dangerous. But in this case, doesn't the virtual environment make the user environment *more* predictable? The user could have any setting for `PYTHONHOME` or `PYTHONPATH`, and those could change the behavior of `<virtual-env-name>/bin/pip3`. The virtual environment will clear the python environment variables and make sure that pip looks for and installs packages where we expect.

goxberry: > But in this case, doesn't the virtual environment make the user environment _more_ predictable?

It doesn't if they've already activated a virtual environment instantiated via `venv` or `virtualenv`, or some other software (e.g., `conda` from Anaconda's Python distro or `lmod`) is managing environment variables. "Other software managing the environment variables" tends to be the failure mode I saw at LLNL.

> The user could have any setting for `PYTHONHOME` or `PYTHONPATH`, and those could change the behavior of `<virtual-env-name>/bin/pip3`.

That's true. But in this scenario, those environment variables would also affect the instantiation of the virtual environment because one could manipulate those environment variables so that alternate versions of `venv` and `pip` take precedence over the "stock" versions. That pathology could be mitigated by instantiating the virtual environment via:

```console
% /absolute/path/to/python3 -E -m venv <virtual-env-name>
```
and provisioning the virtual environment via:

```console
% /absolute/path/to/<virtual-env-name>/bin/python3 -E -m pip install clingo
```

or, if you want, you could write a wrapper script that unsets `PYTHONHOME` and `PYTHONPATH` in that scope, or in a subshell. Python's `-s` and `-I` flags could also be used on sufficiently recent Pythons to provide further isolation, if that's what you're going for.

I usually saw `PYTHONPATH` used in DTED environments to kludge around incorrect `RPATH`s in shared libraries (e.g., due to a borked Silo install, or Cubit's odd `RPATH`-ing, at least prior to version 15.9). Typically, this environment variable didn't materially affect building something like `clingo` because the only paths in `PYTHONPATH` pointed to Silo libraries and maybe Cubit libraries. But again, as you correctly point out, it is possible to set that variable and/or `PYTHONHOME` in ways that will foul up script execution.

btalamini: It's true, there are really an infinite number of ways that a user's environment could be surprising. This is convincing me to say less. I'm just going to say that the script will try to install the clingo package if it's not already present. The user can decide if and how they want to isolate that.

samuelpmishLLNL (approved):

src/docs/sphinx/quickstart.rst
```diff
@@ -267,28 +267,263 @@ must be specified using either:
 
 ``python3 scripts/uberenv/uberenv.py --spack-config-dir=scripts/spack/configs/linux_ubuntu_18 --prefix=../path/to/install``
 
-Preparing OSX for Serac Installation
+Building Serac dependencies on MacOS
```
adrienbernede, line 270: May I suggest that the reader be warned about the existence of this section back in ``Building Serac's Dependencies via Spack/uberenv`` ?

This quick-start guide did not pass the lazy sequential reader test: as a lazy sequential reader, on my mac I run
``python3 scripts/uberenv/uberenv.py --prefix=../serac_libs``, then it fails, then I read more :)

adrienbernede (commented):
Simple suggestion

adrienbernede:
First of all, thank you @btalamini for this work. It is quite a tough mission to explain how to build Serac on MacOS, not to mention that there are now 2 possible architectures.

Earlier this year, I tried to install Serac on my M1 iMac, and failed miserably at clingo install. I was expecting the challenge to be within the `packages.yaml`, because I knew about the tough points mentioned above.

I can understand @samuelpmishLLNL frustration, although I wanted to share a quick answer to his comments:
- I agree that there should be a quick way to install _a_ usable target of Serac on any system, for a quick getting starting at running Serac.
- Serac being an HPC code, both fine grain configuration of the installation (and dependencies) as well as reproducibility are paramount.

What I am trying to say is that we describe here the workflow targeting the second point. For pure, super simple installation, maybe another technology (container?) should be considered / described.

Now for the PR review:
I am a lazy sequential reader. For a quick-start guide I am looking for actionable information for my specific context. I would like to suggest that, in a subsequent PR, this documentation has clearer partitions between say "LC", "Mac OS", "Windows" contexts and maybe sub-partitions "user quick install" and "developer controlled install".

Thanks for reading me :)

Note: admittedly, I haven’t been through the process described in this PR yet.

btalamini:
@white238, I'd like to get your approval / additional comments on this, since it's your bailiwick. Would you mind taking another look at your earliest convenience?

chapman39 (approved):

white238:
Approved.... sorry I am still catching up.

btalamini:
No worries. We can always amend the instructions later.

From: Chris White ***@***.***>
Reply-To: LLNL/serac ***@***.***>
Date: Monday, September 12, 2022 at 11:12 AM
To: LLNL/serac ***@***.***>
Cc: "Talamini, Brandon Louis" ***@***.***>, State change ***@***.***>
Subject: Re: [LLNL/serac] Update build instructions for macOS (PR #770)


Sorry I am still catching up.

Issue #773, opened 2022-08-20
Stress in residual operator contracted on wrong index


btalamini:
The functional-based solid mechanics operator uses a total Lagrangian setting, so it uses the first Piola-Kirchhoff stress in the residual calculation. However, in the kernel [here](https://github.com/LLNL/serac/blob/5aeeea93f029d3b3afcbe30e9ce39253c1ea77e8/src/serac/numerics/functional/integral_utilities.hpp#L523) the stress is contracted with the shape function derivatives on its first index, which is the wrong one. 

Needed: fix this in a dimensionally independent way, so both scalar and vector-valued primal field problems work.

jamiebramwell:
@samuelpmishLLNL , did we decide to leave it the way it is?

btalamini:
This bug was fixed in #779 (by transposing the Piola-Kirchhoff stress):

https://github.com/LLNL/serac/blob/7a6ef7a6e7e502608c21d2ef1fc0a43fedd3fef5/src/serac/physics/solid_mechanics.hpp#L393-L402

However, as you can see, the comments written there treat this fix as temporary. I think we're now taking the view that this is a design requirement: the rightmost index of the flux needs to be the one that contracts with the test function gradient index. We can close the issue, but we should update the comments and the documentation.

jamiebramwell:
That's fair. I'll leave this open as an issue to better update the documentation.

Issue #754, opened 2022-07-26
Move to patch testing


white238:
@samuelpmishLLNL pointed out there is a better way to handle some of the unit tests we are writing.  We should create a few examples and document how to do these so we can move forward in future tests in a more robust manner.

Someone needs to further flush this out who understands it.

samuelpmishLLNL:
Here's a review of what we're trying to test, some of the weird practices taking place in serac's test suite, and some suggestions on how to improve.

In a nutshell, we have some equations that come from the physics, $r_1(x,y) = 0, r_2(x,y) = 0$, that we'd like to solve,

<img width="550" alt="exact_solution" src="https://user-images.githubusercontent.com/61714427/181161349-f8aa07f0-10b0-4224-855f-eb1c2bd1418c.png">

We find an approximate solution by making an initial guess and iterating until the norm of $r$ is smaller than some tolerance (i.e. until the point gets inside the red disk around the solution,  tolerance $\approx$ radius), and then stop.

<img width="550" alt="iterate" src="https://user-images.githubusercontent.com/61714427/181161754-5501d649-23f1-4aef-b094-f83a2d3d3417.gif">

Up to this point, this is all normal. The unusual part is that our tests don't actually check that the output is close to the solution. Instead, they check that the _norm_ of the output is close to the _norm_ of the accepted answer. That is, any point in the purple ring below will pass the `EXPECT_NEAR(norm(accepted_solution), norm(approximate_solution), tolerance)` check.
 
<img width="600" alt="expect_near_1" src="https://user-images.githubusercontent.com/61714427/181162162-c1166779-e05b-4e0f-995e-f61a51eabf22.png">

This is bad because practically all of those points in the purple ring are clearly not even close to the solution, but our test suite would classify them all as correct, which gives the developer the wrong impression that everything is working. For example, an incorrect implementation of those residual equations (represented by dashed lines) might still be inside that purple ring:

<img width="500" alt="expect_near_2" src="https://user-images.githubusercontent.com/61714427/181162562-9e3c88e3-faa6-40d5-b5e4-dfc47869f3b4.png">

Another bad practice is that we continue to define the "accepted solution" as the output of the code the first time the test is run. This is especially bad because the initial implementation of practically anything will inevitably have bugs, so defining a baseline from the output of an early implementation practically ensures that the baselines will be incorrect, which makes everything really confusing. For example, if the flawed implementation (dashed) was used to set a baseline:

[...]

PR #779, opened 2022-08-24
Test solid mechanics with patch tests

btalamini:
This PR was spurred by #754. The main change is to add more rigorous tests of the solid mechanics module. These are patch tests, which ensure that the numerical approximation of the PDEs are consistent. The structure should also allow generalization to manufactured solutions, which would allow us to increase our meaningful test coverage and to verify high-order convergence.

I can see that these tests could benefit from some refactoring to reduce boilerplate, to increase flexibility of options (such as geometric nonlinearity on/off, material type), and to make testing over the combinations of these options more automatic. I'm looking for help here from the code review.

Once we settle on the framework, it should be reused to check other features:

1. Dynamics
2. Potentially, other flavors of statics options: linear kinematics, parameterized models, and materials with state
3. The heat conduction module

We may want to leave issue #754 open until we're satisfied with the coverage, possibly amending it with a checklist.

These tests have already caught 3 significant bugs:

1. The residual calculation contracted the wrong index of the stress with the shape function derivatives. A stopgap solution is put in this PR that makes the residual correct, and issue #773 was opened to track a long-term fix.
2. Several of the material models were templated on spatial dimension, but the 2D versions gave nonsensical results. The stress returned was neither plane strain nor plane stress. I removed the template parameter on the material and moved it to the constitutive output call, with the 2D version giving the plane strain response. The comments have been updated to reflect this.
3. The geometric nonlinearity option was causing the reference configuration to sometimes be in an incorrect state. This led to issue #760 (that has since been closed, thanks @jamiebramwell).


src/serac/physics/state/finite_element_vector.hpp
```diff
@@ -370,4 +370,13 @@ double min(const FiniteElementVector& fe_vector);
  */
 double norm(const FiniteElementVector& state, const double p = 2);
 
+/**
+ * @brief Find the L2 norm of the error of a finite element vector with respect to an exact solution
+ *
+ * @param state The state variable to compute a norm of
```
samuelpmishLLNL: it looks like this description of `state` was copied from `norm()` and needs to be updated. I never know what to put down for doxygen descriptions of arguments like this, almost everything ends up sounding tautological.

src/serac/physics/tests/test_solid_functional_statics.cpp
```diff
@@ -0,0 +1,290 @@
...
+  AffineSolution():
+    A(dim), b(dim)
+  {
+    // clang-format off
+    A(0, 0) = 0.110791568544027; A(0, 1) = 0.230421268325901;
+    A(1, 0) = 0.198344644470483; A(1, 1) = 0.060514559793513;
+    if constexpr (dim == 3) {
+      A(0, 2) = 0.15167673653354;
+      A(1, 2) = 0.084137393813728;
+      A(2, 0) = 0.011544253485023; A(2, 1) = 0.060942846497753; A(2, 2) = 0.186383473579596;
```
samuelpmishLLNL: I think I see the point of this formatting, to make the code graphically represent the matrix terms that are affected. It could also be possible to do something like:
```suggestion
                                                                A(0, 2) = 0.15167673653354;
                                                                A(1, 2) = 0.084137393813728;
      A(2, 0) = 0.011544253485023; A(2, 1) = 0.060942846497753; A(2, 2) = 0.186383473579596;
```

Also, if the values of these entries is not important, then it's possible to make this initialization a little more succinct (using a different arbitrary choice of values)
```cpp
for (int i = 0; i < dim; i++) {
  for(int j = 0; j < dim; j++) {
    A(i,j) = 1.0 / (i + 2 * j);
  }
  b[i] = sin(i);
}
```
would have a similar effect

src/serac/physics/tests/test_solid_functional_statics.cpp
```diff
@@ -0,0 +1,290 @@
...
+  /**
+   * @brief MFEM-style coefficient function corresponding to this solution
+   *
+   * @param X Coordinates of point in reference configuration at which solution is sought
+   * @param u Exact solution evaluated at \p X
+   */
+  void operator()(const mfem::Vector& X, mfem::Vector& u) const
+  {
+    u = 0.0;
+    A.Mult(X, u);
```
samuelpmishLLNL: I think `Operator::Mult()` zeroes out the output-vector for us (?), which would make line 62 is unnecessary

src/serac/physics/tests/test_solid_functional_statics.cpp
```diff
@@ -0,0 +1,290 @@
...
+ *
+ * @tparam dim Spatial dimension
+ *
+ * @param b Kind of boundary conditions to apply in the problem
+ * @return std::set<int> Boundary attributes for the essential boundary condition
+ */
+template <int dim>
+std::set<int> essentialBoundaryAttributes(PatchBoundaryCondition bc)
+{
+  if constexpr (dim == 2) {
```
samuelpmishLLNL: It's okay to pass `dim` as a template parameter, but this function doesn't appear to really need compile-time information of the value of `dim`. So, it could be passed just as well as a conventional argument (this would require changing the `if constexpr`s into regular `if`s). Either implementation is acceptable as far as I'm concerned.

data/meshes/patch3D.mesh
```diff
@@ -0,0 +1,71 @@
...
+#1 1 1
+#0 1 1
+#0.25 0.333333333333333 0.25
+#0.62 0.25 0.25
+#0.75 0.75 0.25
+#0.25 0.75 0.25
+#0.25 0.25 0.75
+#0.75 0.25 0.75
+#0.75 0.75 0.75
+#0.25 0.75 0.8
```
samuelpmishLLNL: Can we delete these extra coordinates?

samuelpmishLLNL:
For reference, a visualization of the 3D patch test mesh is the red version below:
patch_test_mesh

In contrast to the blue version which has a lot of symmetries, the red version has the interior nodes perturbed to break the symmetries and ensure that the element geometries are not axis-aligned.

samuelpmishLLNL (approved):

data/meshes/patch3D.mesh
btalamini: Oops, that was careless. Deleted them.

src/serac/physics/tests/test_solid_functional_statics.cpp
btalamini: This is a good suggestion. In the long run, it would be even better to make the essential boundary condition calls use `tensor`, which would make this code more compact and readable. It would also increase the uniformity of the interface with the other loading functions.

src/serac/physics/tests/test_solid_functional_statics.cpp
btalamini: Thanks. You're right. Fixed.

src/serac/physics/tests/test_solid_functional_statics.cpp
btalamini: Good point. I'll leave it for now, just so the `static_assert` on the spatial dimension in the test body doesn't break. Not sure that is the right place to test this condition, but that's probably a subject for another PR.

src/serac/physics/tests/test_solid_functional_statics.cpp
```diff
@@ -0,0 +1,289 @@
...
+template <int p, int dim, typename ExactSolution>
+double solution_error(const ExactSolution& exact_displacement, PatchBoundaryCondition bc)
+{
+  MPI_Barrier(MPI_COMM_WORLD);
+
+  // Create DataStore
+  axom::sidre::DataStore datastore;
+  serac::StateManager::initialize(datastore, "solid_functional_static_solve");
+
+  // BT: shouldn't this assertion be in the physics module?
```
jamiebramwell, line 176: Agreed. I'll add an issue.

jamiebramwell (approved):
This is fantastic (and well past due)! Thank you for doing this!

src/serac/physics/state/finite_element_vector.hpp
btalamini: Thanks. I just updated it. I agree, but tautological is better than wrong. In a way, it's a good sign, because it means the function is probably self-documenting.

Issue #774, opened 2022-08-20
Quality-of-Life improvements and bugfixes for LiDO 


samuelpmishLLNL:
Kenny has been wrapping the solid mechanics module in LiDO and mentioned that he ran into a few issues, I'm summarizing a few of them here so I don't forget before the weekend:

1. Some missing implementation for 0th-order (constant) L2 spaces. Some of our polynomial bases routines start at order 1 instead of zero, we just need to add a few lines of code to support this.
2. Be more explicit about not supporting L2 spaces on boundary integrals. Right now, the failure mode is a long compiler error message, but this should be easy to address with a `static_assert` in `addBoundaryIntegral()`. There's some relevant discussions about why mfem doesn't want to support boundary elements for L2 here:
- https://github.com/mfem/mfem/issues/2320
- https://mfem.slack.com/archives/C0258BVPQP3/p1629304286057700

It may still be possible to implement a workaround to enable boundary integrals that involve L2 spaces, but unfortunately it would require a considerable amount of work to do, since it's not supported in mfem.

3. LiDO is using quite a few parameters, and it's tedious to make every material and traction take them as arguments. This would be addressed by some of the ideas described in the content/discussion of [this PR](https://github.com/LLNL/serac/pull/769#issuecomment-1219860934)

4. Confusion about [these types](https://github.com/LLNL/serac/blob/5aeeea93f029d3b3afcbe30e9ce39253c1ea77e8/src/serac/physics/materials/solid_functional_material.hpp#L220-L236) for specifying loads and tractions, e.g.

```cpp
template <int dim>
struct PressureFunction {
  std::function<double(const tensor<double, dim>&, const double)> pressure_func_;

  SERAC_HOST_DEVICE double operator()(const tensor<double, dim>& x, const double t) const
  {
    return pressure_func_(x, t);
  }
};
```
It is not a requirement to use functors like this as arguments to `SolidMechanics::addPiolaTraction()`. 

And to clarify: lambdas or `std::function`s for materials, tractions, loads, etc can be passed in directly, but will not work on a GPU, since `std::function` can only represent CPU-specific functions and "extended" lambdas can't be generic. For defining materials and loads that work on a GPU, you can put the implementation inside a functor object and mark its `operator()` with `SERAC_HOST_DEVICE`. However, `PressureFunction`, `TractionFunction` and so on, will not work on a GPU, because they are implemented in terms of the CPU-specific `std::function`.

[...]

PR #782, opened 2022-08-30
Improvements for LIDO

samuelpmishLLNL:
Re: https://github.com/LLNL/serac/issues/774

Here's a summary of some of the issues and which have been addressed so far in this PR (:heavy_check_mark: => addressed, :x: => still working):

:heavy_check_mark: 1. (Bug) Some missing implementation for 0th-order (constant) L2 spaces.
:heavy_check_mark: 2. (QoL) Be more explicit about not supporting L2 spaces on boundary integrals.
:heavy_check_mark: 3. (QoL) User-specified arguments per q-function.
<details>
  <summary>:heavy_check_mark: 3. (QoL) User-specified arguments per q-function.</summary>

---------

Before: q-functions had to accept all parameters present in the problem:

```cpp
SolidFunctional<p, dim, Parameters<This, That, TheOtherThing> > solid_solver(...);
solid_solver.setMaterial([](auto x, auto du_dx, auto this_q, auto that_q, auto the_other_thing_q){
   ...
});
```

Now, you can write the following instead:

```cpp
SolidFunctional<p, dim, Parameters<This, That, TheOtherThing> > solid_solver(...);
solid_solver.setMaterial(DependsOn<0, 2>{}, [](auto x, auto du_dx, auto this_q, auto the_other_thing_q){
   ...
});
```
  
</details>

:heavy_check_mark: 4. (QoL) Clarify [these types](https://github.com/LLNL/serac/blob/5aeeea93f029d3b3afcbe30e9ce39253c1ea77e8/src/serac/physics/materials/solid_functional_material.hpp#L220-L236) for specifying loads and tractions
:x: ~5. (QoL) Member functions that require explicit template parameters~ (I'm going to skip this for now, it's a breaking interface change that wouldn't actually provide much benefit)
:heavy_check_mark: 6. (Bug) some boundary integral arrays were being sized incorrectly
:heavy_check_mark: 7. (Bug) update signatures for quantity of interest member function aliases `AddVolumeIntegral()` and `AddAreaIntegral()` 
:heavy_check_mark: 8. (Bug) fix indexing bug when using vector-valued trial spaces in volume integrals for quantities of interest.
:heavy_check_mark: 9. (QoL) update readthedocs entries for Functional to include expectations of q-function interface for Quantities of Interest
<details>
  <summary>:heavy_check_mark: 10. (QoL) give users an option to bypass <code>setMaterial</code> calls and register a custom q-function to the residual.</summary>

---------

Before: material models make a lot of assumptions (materials don't depend on spatial coordinate or displacement vector explicitly, homogeneous mass density in reference configuration, ... )

```cpp
SolidFunctional<p, dim> solid_solver(...);
solid_solver.setMaterial(NeoHookean{rho, K, G});
```

Now, there's another option, where you can write something like:

```cpp
double lambda = ...;
double mu = ...;
solid_solver.addCustomDomainIntegral(DependsOn<>{}, [=](auto x, auto displacement, auto acceleration){
  auto du_dx = serac::get<1>(displacement);

  auto I       = Identity<dim>();
  auto epsilon = 0.5 * (transpose(du_dx) + du_dx);
  auto stress = lambda * tr(epsilon) * I + 2.0 * mu * epsilon;
 
  auto d2u_dt2 = serac::get<0>(acceleration);
  double rho = 1.0 + x[0]; // spatially-varying density
 
  return serac::tuple{rho * d2u_dt2, stress};
});
```
Note: this feature is experimental, and isn't properly tested yet.
  
</details>

------------

For reference: much of the credit for these bugfixes goes to @kswartz92, for uncovering a lot of gaps in our test suite coverage


samuelpmishLLNL:
new contributions to the developer guide (related to "quantities of interest") can be viewed here:
https://serac.readthedocs.io/en/mish2-improvements_for_lido/sphinx/dev_guide/functional.html

samuelpmishLLNL:
some timings comparing compilation of C++ files compiled with `CXX` language in CMake:

```
sam@provolone:~/code/serac/build-GCC-Debug$ time make -j
...
[100%] Built target functional_qoi
[100%] Built target functional_comparisons
[100%] Built target functional_nonlinear

real    0m20.359s
user    1m42.744s
```

as opposed to as `CUDA` language in CMake

```
sam@provolone:~/code/serac/build-GCC-Debug$ time make -j
...
[100%] Built target functional_qoi
[100%] Built target functional_comparisons
[100%] Built target functional_nonlinear

real    0m36.634s
user    3m20.530s
```

src/docs/sphinx/dev_guide/functional.rst
```diff
@@ -16,33 +16,33 @@ a solid mechanics simulation may look something like:
...
    constexpr int order = 2; // the polynomial order of the basis functions
-   constexpr int dim = 3; // the dimension of our problem
+   constexpr int dim = 3; // the number of components per node
    using test = H1<order, dim>;
    using trial = H1<order, dim>;
-   Functional< test(trial) > residual(test_fes, trial_fes);
+   Functional< test(trial) > residual(&test_fes, {&trial_fes});
 
-where ``test_fes``, ``trial_fes`` are the ``mfem`` finite element spaces for the problem. 
+where ``test_fes``, ``trial_fes`` are the ``mfem::FiniteElementSpace``s for the problem.
```
jamiebramwell: This rendered weird on readthedocs. Maybe it needs an extra space?

src/docs/sphinx/dev_guide/functional.rst
```diff
@@ -141,24 +189,113 @@ Putting these snippets together without the verbose comments, we have (note: the
...
-   residual.AddSurfaceIntegral([](auto x, auto disp /* unused */){ return traction(x); }, surface_mesh);
+    residual.AddSurfaceIntegral([](auto x, auto disp /* unused */){ return traction(x); }, surface_mesh);
 
-   auto r = residual(displacements);
+    auto r = residual(displacements);
 
 So, in only a few lines of code, we can create optimized, custom finite
 element kernels!
 
+Quantities of Interest
```
jamiebramwell, line 214: This is a really nice addition!

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -128,7 +128,7 @@ void functional_solid_test_static_J2()
 
   auto state = solid_solver.createQuadratureDataBuffer(initial_state);
 
-  solid_solver.setMaterial(mat, state);
+  solid_solver.setMaterial(DependsOn<>{}, mat, state);
```
jamiebramwell: Is the `DependsOn<>{}` required when there are no parameterizations?

jamiebramwell: Looks like the answer is yes. Should this be deleted?

src/serac/physics/solid_functional.hpp
```diff
@@ -272,6 +272,41 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
+   *    auto d2u_dt2 = serac::get<0>(acceleration);
+   *    double rho = 1.0 + x[0]; // spatially-varying density
+   *
+   *    return serac::tuple{rho * d2u_dt2, stress};
+   *  });
+   *
+   * ~~~
+   */
+  template <int... active_parameters, typename callable, typename StateType = Nothing>
+  void addCustomDomainIntegral(DependsOn<active_parameters...>, callable qfunction,
```
jamiebramwell, line 303: Is this tested?

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -177,28 +177,6 @@ struct ConstantTraction {
   }
 };
 
-/// Function-based traction boundary condition model
```
jamiebramwell, line 180: Was this deleted in favor of the custom integrals? I assume its due to GPU incompatibility.

src/serac/numerics/functional/boundary_integral.hpp
```diff
@@ -104,54 +111,89 @@ class BoundaryIntegral<test(trials...), exec> {
...
+   * @param[in] which_trial_space the index of the argument being differentiated
    */
-  void ComputeElementGradients(ExecArrayView<double, 3, ExecutionSpace::CPU> K_b, std::size_t which) const
+  void ComputeElementGradients(ExecArrayView<double, 3, ExecutionSpace::CPU> K_b, std::size_t which_trial_space) const
   {
-    element_gradient_[which](K_b);
+    int which_local_trial_space = -1;
+    for (size_t i = 0; i < argument_indices.size(); i++) {
+      if (which_trial_space == std::size_t(argument_indices[i])) {
+        which_local_trial_space = int(i);
```
jamiebramwell: Which is preferable: functional casts or static casts? I tend to use the latter, but I'm honestly not qualified to say which we should use.

jamiebramwell (commented):
These are some great quality of life improvements!

src/serac/numerics/functional/boundary_integral.hpp
samuelpmishLLNL: I don't have a preference. The Google C++ style guide says to prefer `static_cast`, so I'll change these accordingly.

src/serac/physics/materials/solid_functional_material.hpp
samuelpmishLLNL: These were a source of confusion. To clarify: a q-function can be one of the following
- a lambda function (cpu-only) 
- a `std::function`(cpu-only) 
- a functor object (cpu and gpu)

The main reason to choose the functor approach is when you need GPU support, but the implementation of these functors doesn't work on the GPU. So, these classes were incurring unnecessary overhead, and making the requirements of q-function less clear.

src/serac/physics/solid_functional.hpp
samuelpmishLLNL: I don't have a good test for these right now. I believe Kenny's actual use case involves manually calculating the inertial forces from acceleration along with a user-specified density field, but the dynamics capability is likely not working.

src/serac/numerics/functional/boundary_integral.hpp
```diff
@@ -26,41 +26,48 @@
...
-
   /**
    * @brief Constructs an @p BoundaryIntegral from a user-provided quadrature function
    * @tparam dim The dimension of the element (2 for quad, 3 for hex, etc)
+   * @tparam test the test function space
+   * @tparam trials the trial function space(s)
    * @tparam qpt_data_type The type of the data to store for each quadrature point
    * @param[in] num_elements The number of elements in the mesh
    * @param[in] J The Jacobians of the element transformations at all quadrature points
    * @param[in] X The actual (not reference) coordinates of all quadrature points
```
btalamini, line 46: I presume "actual" means "the coordinates of the quadrature points in physical space" as opposed to  "coordinates in the element parametric coordinates". In solid mechanics, I would think this would be the coordinates of the material point in the reference configuration. Is this correct? In that case, the "not refererence" comment is misleading. We may need to settle our terminology and use it consistently. Not a change for this PR, but something to keep in mind.

src/docs/sphinx/dev_guide/functional.rst
```diff
@@ -141,24 +189,113 @@ Putting these snippets together without the verbose comments, we have (note: the
...
+----------------------
+
+``Functional`` can also be used to represent scalar-valued integral expressions. These can be used
+to represent objective functions, constraints, or other "quantities of interest". To make a ``Functional``
+with a scalar-valued output, use ``double`` as the test space in its function signature:
+
+.. code-block:: cpp
+
+    using test = H1<order, dim>;
```
btalamini: Are the `trial` and `test` symbols backwards here? Since the signature is `test(trial)`, I would have expected `test` to be `double` in this case.

src/serac/numerics/functional/tests/functional_comparison_L2.cpp
```diff
@@ -119,6 +128,7 @@ void functional_test(mfem::ParMesh& mesh, L2<p> test, L2<p> trial, Dimension<dim
   EXPECT_NEAR(0., mfem::Vector(g1 - g2).Norml2() / g1.Norml2(), 1.e-14);
 }
 
+TEST(L2, 2DConstant) { functional_test(*mesh2D, L2<0>{}, L2<0>{}, Dimension<2>{}); }
```
btalamini, line 131: Cool, you're testing this new case. :)

src/serac/physics/solid_functional.hpp
samuelpmishLLNL: Would you prefer that I try to add a test for this capability now, or later on, after revisiting dynamics?

src/serac/numerics/functional/tests/functional_qoi.cpp
```diff
@@ -356,6 +356,101 @@ void qoi_test(mfem::ParMesh& mesh, H1<p1> trial1, H1<p2> trial2, Dimension<dim>)
...
+  mfem::HypreParVector  U   = *tmp;
+  U_gf.GetTrueDofs(U);
+
+  // Define the types for the test and trial spaces using the function arguments
+  using trial_space = H1<p>;
+
+  Functional<double(trial_space)> measure({&fespace});
+  measure.AddAreaIntegral(
+      DependsOn<>{}, [&](auto /*x*/) { return 1.0; }, mesh);
+  double relative_error = (measure(U) - measure_mfem(mesh)) / measure(U);
```
btalamini, line 430: I may not have understood the intent of this test, so I apologize if this is off base: It looks like this test verifies that the Functional implementation gets the same answer as a native MFEM implementation. I think for unit tests we should prefer exact solutions to code-to-code comparisons. The former gives us a clearer signal if it fails.

src/serac/physics/solid_functional.hpp
btalamini: I say let's get patch-type tests in for dynamics first.

btalamini (approved):

src/serac/numerics/functional/boundary_integral.hpp
samuelpmishLLNL: Yeah, this is a bad description. Unfortunately, I think mfem uses "reference" to mean what is usually called the "parent" / "master" element. I agree that we should adopt a more consistent terminology, I like:

- parent element / isoparametric coordinates to refer to $\xi$
- physical coordinates to refer to either $x$ (current configuration) or $X$ (reference, but not in the mfem sense, configuration)

And for reference (no pun intended), the line of code in question is $X$, the reference configuration of the physical coordinates.

kswartz92:
are there plans to add a `setPressure` type method to solid functional? If so, would this PR be the appropriate place to do so?

kswartz92:
what's the current status and plan moving forward for supporting parameters on L2 spaces? I'm mostly concerned with material parameters since they are commonly parameterized on L2 in topology optimization

src/docs/sphinx/dev_guide/functional.rst
sethwatts: Yes, it needs a space. This is one of my pet peeves with ReST. It makes the compiled text look like trash, too, IMO.

samuelpmishLLNL:
> are there plans to add a `setPressure` type method to solid functional? If so, would this PR be the appropriate place to do so?

Currently, (Piola) pressures can be applied as a special case of the `setPiolaTraction` member function. We could add a convenience method that wraps a user-specified pressure function and multiplies it by $\hat{n}$. One concern I have is that we end up with a lot of overloads, e.g.

- `setPiolaTraction(traction_function)`
- `setPiolaTraction(DependsOn<...>, traction_function)`
- `setCauchyTraction(traction_function)`
- `setCauchyTraction(DependsOn<...>, traction_function)`
- `setPiolaPressure(pressure_function)`
- `setPiolaPressure(DependsOn<...>, pressure_function)`
- `setCauchyPressure(pressure_function)`
- `setCauchyPressure(DependsOn<...>, pressure_function)`

which makes it more tedious to maintain. The other point is that deeply-nesting lambdas may impact performance, but we should collect data before making any decisions on that criterion.

> what's the current status and plan moving forward for supporting parameters on L2 spaces? I'm mostly concerned with material parameters since they are commonly parameterized on L2 in topology optimization

L2 spaces are supported for domain integrals. This PR adds support for 0th-order (one value per element) L2 spaces. Getting L2 to work on the boundary integrals is more challenging, as mfem seems to believe users shouldn't be able to do this at all. I don't agree with their philosophy, but regardless, it means that we would be on our own for coming up with a hack to get the appropriate information necessary to enable the L2 trial spaces on the boundary.

For reference, the tools we need to enable L2 spaces on the boundary are:
- L->E gather operation
- E->L scatter-add operation
- a way to get the dof ids associated with a given boundary element (must be consistent with the L->E and E->L mappings as well). 

see: https://github.com/mfem/mfem/issues/2495

kswartz92:
> > are there plans to add a `setPressure` type method to solid functional? If so, would this PR be the appropriate place to do so?
> 
> Currently, (Piola) pressures can be applied as a special case of the `setPiolaTraction` member function. We could add a convenience method that wraps a user-specified pressure function and multiplies it by n^. One concern I have is that we end up with a lot of overloads, e.g.
> 
> * `setPiolaTraction(traction_function)`
> * `setPiolaTraction(DependsOn<...>, traction_function)`
> * `setCauchyTraction(traction_function)`
> * `setCauchyTraction(DependsOn<...>, traction_function)`
> * `setPiolaPressure(pressure_function)`
> * `setPiolaPressure(DependsOn<...>, pressure_function)`
> * `setCauchyPressure(pressure_function)`
> * `setCauchyPressure(DependsOn<...>, pressure_function)`
> 
> which makes it more tedious to maintain. The other point is that deeply-nesting lambdas may impact performance, but we should collect data before making any decisions on that criterion.
> 
> > what's the current status and plan moving forward for supporting parameters on L2 spaces? I'm mostly concerned with material parameters since they are commonly parameterized on L2 in topology optimization
> 
> L2 spaces are supported for domain integrals. This PR adds support for 0th-order (one value per element) L2 spaces. Getting L2 to work on the boundary integrals is more challenging, as mfem seems to believe users shouldn't be able to do this at all. I don't agree with their philosophy, but regardless, it means that we would be on our own for coming up with a hack to get the appropriate information necessary to enable the L2 trial spaces on the boundary.
> 
> For reference, the tools we need to enable L2 spaces on the boundary are:
> 
> * L->E gather operation
> * E->L scatter-add operation
> * a way to get the dof ids associated with a given boundary element (must be consistent with the L->E and E->L mappings as well).
> 
> see: [mfem/mfem#2495](https://github.com/mfem/mfem/issues/2495)

Personally, I'm okay with using the traction interface to define pressure loads. This also makes our lido wrappers slightly easier to maintain. I'm guessing most people writing a custom load function would understand how to use the normal. @sethwatts do you have any objections?

btalamini:
> Currently, (Piola) pressures can be applied as a special case of the `setPiolaTraction` member function. 

This is semantics, but to me "pressure"  implies a follower load. That is, pressure is a traction that acts only in the *current* normal direction. I think this is worth supporting with a specialized function because of the rotation of the normal. I would vote that we don't need a `setPiolaPressure()` method, if by that you mean "constant times reference normal vector".

sethwatts:
> Personally, I'm okay with using the traction interface to define pressure loads. This also makes our lido wrappers slightly easier to maintain. I'm guessing most people writing a custom load function would understand how to use the normal. @sethwatts do you have any objections?

No objections. I agree that anyone writing a custom load should be on the hook for understanding how to use normals. @kswartz92 , are you intending to support a pressure load as a standard one in the lido wrapper?

src/docs/sphinx/dev_guide/functional.rst
samuelpmishLLNL: good catch!

src/serac/physics/solid_functional.hpp
jamiebramwell: Agreed, we can test it later. I'll add an issue.

jamiebramwell (approved):

PR #783, opened 2022-08-31
Update MFEM/Axom

white238:
* Updates MFEM and Axom to newer points in their develop branches
* Removes option to turn off Lumberjack when needing a parallel logger (noone used it and it's no longer needed)
* Removes our reliance on the soon to be deprecated (and eventually removed) implicit conversion in `mfem::Vector` to `double *`
* Adds temporary work around for MFEM deleting the operator + and - on `mfem::Vector`.  This is being reverted in an already created PR so I left in the, now commented out, original line.


samuelpmishLLNL (approved):

scripts/spack/packages/axom/package.py
```diff
@@ -43,7 +43,7 @@ class Axom(CachedCMakePackage, CudaPackage, ROCmPackage):
     # Note: Make sure this sha coincides with the git submodule
     # Note: We add a number to the end of the real version number to indicate that we have
     #  moved forward past the release. Increment the last number when updating the commit sha.
-    version('0.6.1.5', commit='d38862ac909fccbfb3d828a920063b4be4ec6a70', submodules="True")
```
jamiebramwell, line 46: Shouldn't the MFEM SHA also be changed?

white238: yes good catch.... too many things all at once

jamiebramwell (approved):

src/serac/numerics/odes.cpp
```diff
@@ -171,9 +171,12 @@ void SecondOrderODE::Solve(const double time, const double c0, const double c1,
   bool implicit = (c0 != 0.0 || c1 != 0.0);
   if (implicit) {
     if (enforcement_method_ == DirichletEnforcementMethod::DirectControl) {
-      d2U_dt2_ = (U_ - u) / c0;
-      dU_dt_   = du_dt;
-      U_       = u;
+      // TODO: MFEM PR#3064 explicitly deleted the const vector operator-.  This
+      // is in active discusion and may be un-deleted. Original line commented.
+      // d2U_dt2_ = (U_ - u) / c0;
```
white238, line 176: Confirmed these can revert back to the commented code after the mfem pr goes through.

cmake/SeracBasics.cmake
```diff
@@ -18,8 +18,6 @@ option(SERAC_ENABLE_CODEVELOP
        "Enable Serac's codevelop build (MFEM and Axom included as CMake subdirectories)"
        OFF)
 
-option(SERAC_ENABLE_LUMBERJACK "Enable Axom's Lumberjack component" ON)
```
white238, line 21: This was originally optional due to a bug (and really a lack of production level proving) in Lumberjack.  It is now more trusted and tested than the alternative in SLIC, so I am removing the option which noone ever exercised.

white238:
This is now ready to be reviewed.

src/serac/numerics/functional/functional_qoi.inl
```diff
@@ -453,7 +453,7 @@ private:
         for (axom::IndexType e = 0; e < K_elem.shape()[0]; e++) {
           for (axom::IndexType j = 0; j < K_elem.shape()[2]; j++) {
             auto [index, sign] = LUT(e, j);
-            gradient_L_[index] += sign * K_elem(e, 0, j);
+            gradient_L_.GetData()[index] += sign * K_elem(e, 0, j);
```
samuelpmishLLNL: wait, what-- `mfem::Vector` doesn't define `operator[]` (?!)

samuelpmishLLNL (approved):

src/serac/numerics/functional/functional_qoi.inl
white238: This one was less "easy" than the others and I might have gotten the wrong answer.  Let me reproduce the warning/error I was getting.

samuelpmishLLNL: I wasn't complaining, just surprised to realize that the `operator[]` must have been coming from the implicit conversion to `double *`.

samuelpmishLLNL: I guess they have `mfem::Vector::operator()(int i)`, does that work?

jamiebramwell: I believe the answer is yes, so we should be using that.

white238: Pushed a better fix.

jamiebramwell:
Looks good! Thanks!

PR #795, opened 2022-09-12
use the right index when selecting gradient kernel

samuelpmishLLNL:


src/serac/numerics/functional/domain_integral.hpp
```diff
@@ -205,7 +205,7 @@ class DomainIntegral {
       }
     }
     if (which_local_trial_space != -1) {
-      action_of_gradient_[which_trial_space](input_E, output_E);
+      action_of_gradient_[which_local_trial_space](input_E, output_E);
```
btalamini: I don't fully understand the code yet (not because it's unclear, only because I haven't examined it before), though I can see the gist of the error. Maybe a unit test is in order? It's a trivial fix, and I don't want to hold it up. On the other hand, I could see someone breaking this in the future who isn't as familiar with it (like me). What do you think? Overkill for this?

samuelpmishLLNL: By unit test do you mean a test fixture for just `DomainIntegral` (rather than `Functional` or `SolidMechanics` module) that confirms this part is working? I guess we don't have proper unit tests for these classes in that sense, they almost all start at `Functional` scope or higher.

This conditional logic that changed is being tested through the physics modules, but we didn't have a test case that caught this issue.

samuelpmishLLNL: I should probably just change this implementation to be a little bit clearer though. Something like a `std::map` from the "global" argument index to the "local" index used by the `DomainIntegral` would work just as well and make more sense.

btalamini: Using a map sounds like a good idea. 

For the test suggestion, I'm applying a rule of thumb: if a user catches a bug that we didn't see in the code despite looking at it several times, it's probably the kind of misunderstanding that we could make again,  and so deserves a test.  I'm fine with the test being at another scope. I want to guard against a regression if someone refactors this code in the future, especially someone less familiar with its design.

I could be convinced that it isn't necessary in this case - I trust your judgment.

kswartz92: Perhaps adding the situation that uncovered the bug (using a parameterized material with a parameterized body force) to one of the functional tests would be a good idea?

samuelpmishLLNL:
For reference, this bug showed up when adding the feature to let the integrals specify which arguments they required. Before, `serac::Functional` essentially represented calculations like:

Screen Shot 2022-09-13 at 8 07 49 AM

where $f$ is the `serac::Functional` and $g_i$ are the integral calculations registered to it. After letting each integral specify which arguments it wants, then it looks more like:

Screen Shot 2022-09-13 at 8 07 56 AM

So, now the $i^{th}$ argument of $f$ is not the same as the $i^{th}$ argument of $g$, so there needs to be a way to translate between them. I wrote that index translation thing once, and then copied it to the 6 different places it was required, but neglected to actually use the translated index in 2 places, and that produced the bug that Kenny saw.

src/serac/numerics/functional/domain_integral.hpp
btalamini: The explanation you just posted makes it clear. Would you mind putting a comment to this effect inside the implementation? Or, if you encode this into a map as you've suggested, the docstring for that will automatically cover it.

btalamini:
Side note: while we're in `src/serac/numerics/functional/domain_integral.hpp`, I noticed that the `argument_indices` member variable doesn't have the trailing underscore.

src/serac/numerics/functional/domain_integral.hpp
samuelpmishLLNL: I'm in the process of modifying one of the existing tests to cover this, and changing the implementation to (hopefully) make clearer what's going on with the index translation.

kswartz92:
I made these changes on my local branch and confirmed that it fixed the bug and that the gradient calculations are correct.

kswartz92 (approved):

src/serac/numerics/functional/boundary_integral.hpp
```diff
@@ -48,25 +48,28 @@ class BoundaryIntegral {
    * @param[in] X The actual (not reference) coordinates of all quadrature points
    * @param[in] N The unit normals of all quadrature points
    * @param[in] qf The user-provided quadrature function
-   * @param[in] arg_indices indices used to select which trail spaces to use in evaluation kernels
+   * @param[in] active_arguments indices used to select which trail spaces to use in evaluation kernels
```
btalamini: typo: `trail` -> `trial`

samuelpmishLLNL: oops, I got the one in the domain_integral, but missed this one

src/serac/physics/tests/serac_solid_functional.cpp
```diff
@@ -376,20 +376,28 @@ void functional_parameterized_solid_test(double expected_disp_norm)
...
   solid_solver.completeSetup();
 
   // Perform the quasi-static solve
   double dt = 1.0;
   solid_solver.advanceTimestep(dt);
 
+  // the calculations peformed in these lines of code
+  // are not used, but running them as part of this test
+  // checks the index-translation part of the derivative
+  // kernels is working
```
btalamini, line 398: nice.

btalamini (approved):

PR #798, opened 2022-09-13
Improve gitlab integration tests logging on failure

### LINKED ISSUE #793: Improve integration test logging in gitlab
In failed pipelines, it is not possible to see the output of the failing tests.  The output goes to an error log and we should create a script or pipe some commands together to output any failing test outputs to the screen.

chapman39:
Fixes #793


chapman39:
Ready for review. You can only see the results on `quartz-clang_10_0_0-src` if it fails. Once everything is ready to be merged I'll revert the `tests` submodule back to normal so everything will pass.

jamiebramwell (approved):
This is really nice! Thanks!

scripts/gitlab/build_and_test.sh
```diff
@@ -12,5 +12,22 @@ if [ $? -ne 0 ]; then { echo "ERROR: build_src.py failed." ; exit 1; } fi
...
-    if [ $? -ne 0 ]; then { echo "ERROR: ATS failed." ; exit 1; } fi
+    if [ $? -ne 0 ]; then 
+        # Print each failing test's logs
+        cd *.*.logs
+        FAILING_TEST_NUMBERS=$(awk '/#[0-9]* FAIL/ {print $1}' ats.log | cut -c 2-)
+        for num in $FAILING_TEST_NUMBERS ; do
+            LOG_FILENAMES=(*$num*.log *$num*.log.err)
+            for f in ${LOG_FILENAMES[@]} ; do
+                if [ -s $f ] ; then
+                    echo "\n======== START $f START ========"
```
white238:
```suggestion
                    echo ; echo "======== START $f START ========"
```
Apparently echo doesn't like '\n'.  This is what you get for trusting random things I send you in teams that I didn't bother to test. 

Apparently you can also use `printf "\n foo"` but `echo -e "\n foo"` isn't on all *nix platforms.

white238 (approved):
Approved pending my last code suggestion and reverting the broken test (obviously).

Thanks @chapman39 !

PR #801, opened 2022-09-23
Minimize Axom build size

white238:
* Removes submodules (previous fix had some very helpful quotes)
* Turns off `tools` variant in our use of Axom. This removes the remaining executables produce by the Axom package
* Remove the clearing of Spack mirrors. I am told that what I ran into previously was a fluke and has been solved.

This does not change the Axom library itself and so I will not be rebuilding.


white238:
This is already in a WIP PR in Axom and Spack that I was working on last week.  It just hadn't gotten here yet.

samuelpmishLLNL (approved):

jamiebramwell (approved):

PR #804, opened 2022-09-28
Add missing headers from install

white238:
This is definitely required and was causing a build failure in a downstream customer:

```
physics/common.hpp
```

I noticed these were not being installed. If they shouldn't be installed please tell me before I merge it:

```
physics/materials/green_saint_venant_thermoelastic.hpp
physics/materials/liquid_crystal_elastomer.hpp
physics/materials/material_verification_tools.hpp
```

### DIFF
--- a/src/serac/physics/CMakeLists.txt
+++ b/src/serac/physics/CMakeLists.txt
@@ -27,12 +27,13 @@ set(physics_sources
 
 set(physics_headers
     base_physics.hpp
+    common.hpp
     solid.hpp
     solid_functional.hpp
     thermal_conduction.hpp
     thermal_conduction_functional.hpp
-    thermal_solid.hpp
     thermal_mechanics_functional.hpp
+    thermal_solid.hpp
     )
 
 set(physics_dependencies
--- a/src/serac/physics/materials/CMakeLists.txt
+++ b/src/serac/physics/materials/CMakeLists.txt
@@ -15,14 +15,17 @@ set(materials_sources
     )
 
 set(materials_headers
-    hyperelastic_material.hpp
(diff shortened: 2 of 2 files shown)


jamiebramwell (approved):

samuelpmishLLNL:
Out of curiosity, is there a benefit to manually specifying every .hpp and .cpp file in our `CMakeLists.txt`s, as opposed to something like `file(GLOB ... *.hpp)`? It seems much easier to get out of sync (like this PR is demonstrating) by putting that burden of listing every file on the developer.

samuelpmishLLNL (approved):

white238:
> Out of curiosity, is there a benefit to manually specifying every .hpp and .cpp file in our `CMakeLists.txt`s, as opposed to something like `file(GLOB ... *.hpp)`? It seems much easier to get out of sync (like this PR is demonstrating) by putting that burden of listing every file on the developer.

File globbing is against the best practices of CMake for source lists.  This is from their documentation:

Note: We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild. 


It also makes me paranoid that we will add a file that should not be installed.  I am going come up with either a script that checks that all headers are installed and has a way of ignoring headers we know don't need to be installed or improve our install test to compile a more interesting problem that hits more headers.  The latter will probably not hit all headers though.

Issue #806, opened 2022-09-28
Shape sensitivity bugs in solid mechanics


btalamini:
Dan T found what he and I think are bugs in the shape sensitivity calculations. I'm mad at myself with not catching these when we reviewed it.

The first is here: 
https://github.com/LLNL/serac/blob/7cf216a1be68719dc997f393d741070f1dba4052/src/serac/physics/solid_functional.hpp#L365-L375

UPDATE: OK, now I see the first potential issue is not a bug. The Piola transformation is not modifying the stress, it's being attached to the stress to modify the shape function gradient (from the original reference configuration to the perturbed one) when it finally encounters it in the residual integral (in the machinery of `functional` at a higher level in the call stack). Sorry, I think this is the second time I've made this mistake. On the other hand, I think that lends support to the proposal to move the shape sensitivity modifications to functional.

Second issue is here:
https://github.com/LLNL/serac/blob/7cf216a1be68719dc997f393d741070f1dba4052/src/serac/physics/solid_functional.hpp#L383
 
There should be a `det(I + dp_dX)` factor on the first item of the tuple, just as there is on the second (which is applied on 375). I think this bug may have escaped notice because we don't have dynamic shape optimization yet.

Both of these lead me to a suggestion:
I think we should consider moving the shape sensitivity modifications into the integration and interpolation operations of `functional`. We could update the nodal coordinates instead of carrying the shape displacement `p` around, and make the shape function gradients dual numbers to catch the sensitivities to the coordinates.

samuelpmishLLNL:
I'm bumping into a related issue with the boundary integrals, since the derivatives terms for boundary integrals are w.r.t. isoparametric coordinates (so dp_dX, which should be named dp_dxi, is rectangular instead of square), so some of these calculations need to be reworked:

https://github.com/LLNL/serac/blob/7cf216a1be68719dc997f393d741070f1dba4052/src/serac/physics/solid_functional.hpp#L476-L503

The solution to this problem is less clear, since mfem doesn't provide $\displaystyle \frac{\partial X}{\partial \xi}$ in GetFaceGeometricFactors (despite having a flag that appears to let you ask for it), which I believe is necessary to compute the quantities we want.

btalamini:
[...]

PR #807, opened 2022-09-29
Shape parameterization fixes

jamiebramwell:
Partially addresses https://github.com/LLNL/serac/issues/806

### DIFF
--- a/src/serac/physics/solid_functional.hpp
+++ b/src/serac/physics/solid_functional.hpp
@@ -362,10 +362,10 @@ public:
           // reference configuration dx'/dX = I + du/dX + dp/dX. If we are not including geometric
           // nonlinearities, we ignore the du/dX factor.
 
-          auto deformation_grad = 0.0 * du_dX + 0.0 * dp_dX + I;
+          auto deformation_grad = 0.0 * du_dX + dp_dX + I;
 
           if (geom_nonlin_ == GeometricNonlinearities::On) {
-            deformation_grad = deformation_grad + du_dX + dp_dX;
+            deformation_grad = deformation_grad + du_dX;
           }
 
           // Note that the jacobian needs the fixup for the shape displacement.
--- a/src/serac/physics/solid_functional.hpp
+++ b/src/serac/physics/solid_functional.hpp
@@ -380,7 +380,7 @@ public:
           // stress with the shape function gradients.
           //
           // TODO: fix the residual implementation and remove this transpose.
-          return serac::tuple{material.density * d2u_dt2 * det(deformation_grad), transpose(flux)};
+          return serac::tuple{material.density * d2u_dt2, transpose(flux)};
         },
         mesh_, qdata);
   }
--- a/src/serac/physics/solid_functional.hpp
+++ b/src/serac/physics/solid_functional.hpp
@@ -362,10 +362,10 @@ public:
           // reference configuration dx'/dX = I + du/dX + dp/dX. If we are not including geometric
           // nonlinearities, we ignore the du/dX factor.
 
-          auto deformation_grad = 0.0 * du_dX + dp_dX + I;
+          auto deformation_grad = 0.0 * du_dX + 0.0 * dp_dX + I;
 
           if (geom_nonlin_ == GeometricNonlinearities::On) {
-            deformation_grad = deformation_grad + du_dX;
+            deformation_grad = deformation_grad + du_dX + dp_dX;
           }
 
           // Note that the jacobian needs the fixup for the shape displacement.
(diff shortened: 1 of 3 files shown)


src/serac/physics/tests/serac_solid_functional_shape.cpp
```diff
@@ -161,10 +160,15 @@ TEST(solidFunctionalShape, manualNodes)
 
   for (int i = 0; i < shape_displacement.Size(); ++i) {
     // Check the final displacement norm
-    EXPECT_NEAR(shape_displacement[i], pure_displacement[i], 1.0e-11);
+
+    // TODO why is this tolerance so high? It should be closer to machine precision.
+    EXPECT_NEAR(shape_displacement[i], pure_displacement[i], 2.0e-10);
```
btalamini: For the nonlinear case, the problem could be that the Newton tolerances are too loose, so the iterations quit before machine precision is reached.

Does the linear case also require the loose tolerance? That's more worrisome.

btalamini: Just checked - the default tolerance (which is used here) is only $10^{-4}$ on the relative residual  norm reduction, which seems very unlikely to give machine precision displacement errors. Try making this smaller.

src/serac/physics/solid_functional.hpp
```diff
@@ -380,7 +380,7 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
           // stress with the shape function gradients.
           //
           // TODO: fix the residual implementation and remove this transpose.
-          return serac::tuple{material.density * d2u_dt2, transpose(flux)};
+          return serac::tuple{material.density * d2u_dt2 * det(I + dp_dX), transpose(flux)};
```
btalamini, line 383: It's good that you fixed this, but I don't like that we don't have a test that can tell that it's right. It's  not reasonable to address that in this PR, but I don't want to lose track of it. Maybe I should split up the issues so that we can close the part that is done. Looks like I have to strike the part about `dp_dx` on the linear case anyway.

btalamini (changes requested):

src/serac/physics/tests/serac_solid_functional_shape.cpp
```diff
@@ -161,10 +160,15 @@ TEST(solidFunctionalShape, manualNodes)
 
   for (int i = 0; i < shape_displacement.Size(); ++i) {
     // Check the final displacement norm
-    EXPECT_NEAR(shape_displacement[i], pure_displacement[i], 1.0e-11);
+
+    // TODO why is this tolerance so high? It should be closer to machine precision.
```
samuelpmishLLNL: We should be testing relative errors, it's hard to understand if `2.0e-10` is a significant absolute error or not without knowing the characteristic dimensions of the problem

btalamini: Agree with relative. Though in this case, I doubt the displacements are large enough to explain the problem - looking at the loads and the elastic constants, I don't think the displacements are $~10^5$. I'm pretty sure the problem is the nonlinear solver tolerance.

btalamini: If the linearized kinematics case also needs the large(ish) tolerances, I don't know what the heck the problem is.

src/serac/physics/tests/serac_solid_functional_shape.cpp
jamiebramwell: Can do! This is why I shouldn't do technical work this late. I'm not a teenager anymore! 😆

src/serac/physics/tests/serac_solid_functional_shape.cpp
jamiebramwell: This error also exists in the linear case. My best guess is the projection I'm using to the mesh nodes is not the same as simple addition. I'll investigate this more tomorrow.

samuelpmishLLNL: > Though in this case, I doubt the displacements are large enough to explain the problem

My point was that the displacements in this problem are small, so the relative errors would be even larger than the absolute errors (but not clear how much bigger)

samuelpmishLLNL: Also: we don't need to be doing componentwise relative error calculations, we can just do 
$$\frac{||\boldsymbol{u}_s - \boldsymbol{u}||}{||\boldsymbol{u}||}$$

and not have to worry about division by zero per component.

samuelpmishLLNL: After tightening the nonlinear solver tolerances,
```cpp
  SolverOptions options = {DirectSolverOptions{.print_level = 0}, solid_mechanics::default_nonlinear_options};

  options.nonlinear.abs_tol = 1.0e-14;
  options.nonlinear.rel_tol = 1.0e-14;
```
the relative errors for the linear and nonlinear cases are

```cpp
  double error = pure_displacement.DistanceTo(shape_displacement.GetData());
  double relative_error = error / pure_displacement.Norml2();
  std::cout << relative_error << std::endl;
```

4.10296e-16 and 2.04338e-15, respectively.

edit: it's worth noting that these tighter tolerances require 2 newton iterations for the linear problem

btalamini (approved):

samuelpmishLLNL (approved):

PR #808, opened 2022-09-30
Revert HDF5 build type change

white238:
Reverting this change due to the following two reasons:

- hdf5+shared causes Axom to not find hdf5
- hdf5 build_type=Release causes netcdf-c to not find hdf5 on ubuntu 20

### DIFF
--- a/scripts/spack/packages/serac/package.py
+++ b/scripts/spack/packages/serac/package.py
@@ -155,10 +155,10 @@ class Serac(CachedCMakePackage, CudaPackage):
 
     # Optional (require when="+profile")
     for dep in ["adiak", "caliper"]:
-        depends_on("{0} build_type=Debug".format(dep), when="+profiling+debug".format(dep))
-        depends_on("{0} build_type=Release".format(dep), when="+profiling~debug".format(dep))
-        depends_on("{0}+shared".format(dep), when="+profiling+shared".format(dep))
-        depends_on("{0}~shared".format(dep), when="+profiling~shared".format(dep))
+        depends_on("{0} build_type=Debug".format(dep), when="+profiling+debug")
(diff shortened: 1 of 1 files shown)


jamiebramwell (approved):
Thanks for looking into this!

scripts/spack/packages/serac/package.py
```diff
@@ -161,12 +161,12 @@ class Serac(CachedCMakePackage, CudaPackage):
         depends_on("{0}~shared".format(dep), when="+profiling~shared".format(dep))
 
     # Required
-    for dep in ["axom", "conduit", "hdf5", "metis", "parmetis", "superlu-dist"]:
+    # TODO: don't put hdf5 in this list, for the following reasons:
```
chapman39: small thing but i feel like this is more of a note than a todo

samuelpmishLLNL:
>     * hdf5+shared causes Axom to not find hdf5

Is this a problem with hdf5 or a problem that should be resolved in Axom?

edit: whoops, just saw the comment 
`# TODO: hdf5+shared causes Axom to not find hdf5, should possibly fix at a later date`

------------

>     * hdf5 build_type=Release causes netcdf-c to not find hdf5 on ubuntu 20

I've been using the release build of hdf5 successfully for about 2 weeks on my local ubuntu 20 machine. What is the error you're observing?

white238:
hdf5 outputted config (in our old version) is very brittle.  I am sure this can be worked around but it is low priority due to a one line fix.

The error in netcdf was "configure: error: Can't find or link to the hdf5 library. Use --disable-hdf5, or see config.log for errors"

This can be investigated after we are done unblocking two people.

chapman39 (approved):

PR #810, opened 2022-09-30
Use Cauchy stress instead of Kirchhoff stress for solid mechanics material models

jamiebramwell:
This hopefully clarifies the residual calculation calls. It is also more common for lab material libraries to use Cauchy stress vs. reference configuration stress measures.

### DIFF
--- a/src/serac/physics/materials/parameterized_solid_functional_material.hpp
+++ b/src/serac/physics/materials/parameterized_solid_functional_material.hpp
@@ -89,7 +89,7 @@ struct ParameterizedNeoHookeanSolid {
     auto           G         = G0 + get<0>(DeltaG);
     auto           lambda    = K - (2.0 / dim) * G;
     auto           B_minus_I = du_dX * transpose(du_dX) + transpose(du_dX) + du_dX;
-    auto J = det(I + du_dX);
+    auto           J         = det(I + du_dX);
     return (lambda * log(J) * I + G * B_minus_I) / J;
   }
 
--- a/src/serac/physics/materials/parameterized_solid_functional_material.hpp
+++ b/src/serac/physics/materials/parameterized_solid_functional_material.hpp
@@ -89,7 +89,8 @@ struct ParameterizedNeoHookeanSolid {
     auto           G         = G0 + get<0>(DeltaG);
     auto           lambda    = K - (2.0 / dim) * G;
     auto           B_minus_I = du_dX * transpose(du_dX) + transpose(du_dX) + du_dX;
-    return (lambda * log(det(I + du_dX)) * I + G * B_minus_I) / det(I + du_dX);
+    auto J = det(I + du_dX);
+    return (lambda * log(J) * I + G * B_minus_I) / J;
   }
 
   /**
--- a/src/serac/physics/materials/parameterized_solid_functional_material.hpp
+++ b/src/serac/physics/materials/parameterized_solid_functional_material.hpp
@@ -34,7 +34,7 @@ struct ParameterizedLinearIsotropic {
    * @param du_dX Displacement gradient with respect to the reference configuration (displacement_grad)
    * @param DeltaK The bulk modulus offset
    * @param DeltaG The shear modulus offset
-   * @return The calculated material response (density, Kirchoff stress) for the material
+   * @return The calculated material response (Cauchy stress) for the material
    */
   template <typename DisplacementType, typename DispGradType, typename BulkType, typename ShearType>
   SERAC_HOST_DEVICE auto operator()(const DispGradType& du_dX, const BulkType& DeltaK, const ShearType& DeltaG) const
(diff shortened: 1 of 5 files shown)


src/serac/physics/solid_functional.hpp
```diff
@@ -368,17 +367,16 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
             deformation_grad = deformation_grad + du_dX;
           }
 
-          // Note that the jacobian needs the fixup for the shape displacement.
-          // The material returns Kirchoff stress, which contains det(du / dX'). We want
-          // The final Jacobian to be det(du / dX), so we compute
-          // det(du / dX) = det(du / dX') * det(dX' / dX) = det(du / dX') * det(I + dp / dX)
-          auto flux = dot(stress, transpose(inv(deformation_grad))) * det(I + dp_dX);
+          auto flux = dot(stress, transpose(inv(deformation_grad))) * det(deformation_grad);
```
btalamini, line 370: Don't we still need to multiply by `det(I + dp_dX)`? I agree that the `det(deformation_grad)` used to be absorbed in the stress, but now it isn't, so we have to multiply by it. But I don't see why we get to drop the  `det(I + dp_dX)` factor.

btalamini: Don't we still need to multiply by `det(I + dp_dX)`? I agree that the `det(deformation_grad)` used to be absorbed in the stress, but now it isn't, so we have to multiply by it. But I don't see why we get to drop the `det(I + dp_dX)` factor.

src/serac/physics/solid_functional.hpp
```diff
@@ -368,17 +367,16 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
```
btalamini, line 371: For future reference, if we're shifting to Cauchy stress, we should think about using an updated Lagrangian approach, computing the shape function derivatives w.r.t. current coordinates. Then we would cleanly return Cauchy stress from here. (We still have the shape displacement stuff to deal with until we move that into the integration as well).

src/serac/physics/solid_functional.hpp
```diff
@@ -368,17 +367,16 @@ class SolidFunctional<order, dim, Parameters<parameter_space...>, std::integer_s
...
           // This transpose on the stress in the following line is a
           // hack to fix a bug in the residual operator. The stress
           // should be transposed in the contraction of the Piola
           // stress with the shape function gradients.
           //
+          // Note that the mass part of the return is integrated in the perturbed reference
+          // configuration, hence the det(I + dp_dx) = det(dX'/dX)
+          //
           // TODO: fix the residual implementation and remove this transpose.
           return serac::tuple{material.density * d2u_dt2 * det(I + dp_dX), transpose(flux)};
```
btalamini, line 381: As I said in the earlier comment, I believe the PR is missing a factor of `det(I + dp_dX)` on the stress. If I'm right, maybe we should do that multiplication on this line so that there is symmetry between the D'Alembert force and the flux. Also, one comment then covers both.

src/serac/physics/solid_functional.hpp
jamiebramwell: We don't. This his now included in `det(deformation_grad) = det(dx_dX) = det(dx_dX') * det(dX'_dX) =  det(dx_dX') * det(I + dp_dX)`.

btalamini: Ok, from our verbal conversation, I got the piece I was missing: `deformation_grad` is not the "real" deformation gradient for the problem $\frac{\partial x}{\partial X'}$, but the gradient with respect to the original reference configuration coordinates $X$.

btalamini (approved):

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -74,7 +74,7 @@ struct NeoHookean {
     constexpr auto I         = Identity<dim>();
     auto           lambda    = K - (2.0 / 3.0) * G;
     auto           B_minus_I = du_dX * transpose(du_dX) + transpose(du_dX) + du_dX;
-    return lambda * log(det(I + du_dX)) * I + G * B_minus_I;
+    return (lambda * log(det(I + du_dX)) * I + G * B_minus_I) / det(I + du_dX);
```
samuelpmishLLNL:
```suggestion
    auto J = det(I + du_dX);
    return (lambda * log(J) * I + G * B_minus_I) / J;
```

src/serac/physics/materials/parameterized_solid_functional_material.hpp
```diff
@@ -89,7 +89,7 @@ struct ParameterizedNeoHookeanSolid {
     auto           G         = G0 + get<0>(DeltaG);
     auto           lambda    = K - (2.0 / dim) * G;
     auto           B_minus_I = du_dX * transpose(du_dX) + transpose(du_dX) + du_dX;
-    return lambda * log(det(I + du_dX)) * I + G * B_minus_I;
+    return (lambda * log(det(I + du_dX)) * I + G * B_minus_I) / det(I + du_dX);
```
samuelpmishLLNL:
```suggestion
    auto J = det(I + du_dX);
    return (lambda * log(J) * I + G * B_minus_I) / J;
```

samuelpmishLLNL (approved):
Sorry, I didn't realize I was holding this one up.

PR #812, opened 2022-10-04
Restore accidentally disabled traction patch tests

btalamini:
Sam noticed that I had accidentally turned off the patch tests that used traction loading. This turns them back on.

To be more precise, I had refactored the patch tests so that there is an `enum` determining whether the tests apply purely essential BCs, or a mixture of natural and essential BCs. Due to a copy and paste error, the natural-and-essential test cases were actually duplicates of the essential BC tests.

The traction tests should have failed when Jamie changed the stress measure from Kirchhoff to Cauchy. I verified that they did in fact fail when activated. This is good, as we want the tests to be sufficiently sensitive to catch common finite deformation mistakes like this. I then updated the stress transformation and now they now pass again, as expected.

### DIFF
--- a/src/serac/physics/materials/solid_functional_material.hpp
+++ b/src/serac/physics/materials/solid_functional_material.hpp
@@ -175,11 +175,9 @@ auto KirchhoffToPiola(const tensor<T1, dim, dim>& kirchhoff_stress, const tensor
 template <typename T1, typename T2, int dim>
 auto CauchyToPiola(const tensor<T1, dim, dim>& cauchy_stress, const tensor<T2, dim, dim>& displacement_gradient)
 {
-  auto kirchhoff_stress = det(displacement_gradient + Identity<dim>())*cauchy_stress;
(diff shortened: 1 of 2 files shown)


jamiebramwell (approved):
Thank you for fixing this!

src/serac/physics/materials/solid_functional_material.hpp
```diff
@@ -161,6 +161,24 @@ auto KirchhoffToPiola(const tensor<T1, dim, dim>& kirchhoff_stress, const tensor
   return transpose(linear_solve(displacement_gradient + Identity<dim>(), kirchhoff_stress));
 }
 
+/**
+ * @brief Transform the Cauchy stress to the Piola stress
+ *
+ * @tparam T1 number-like type of the displacement gradient components
+ * @tparam T1 number-like type of the Cauchy stress components
```
samuelpmishLLNL:
```suggestion
 * @tparam T1 number-like type of the Cauchy stress components
 * @tparam T2 number-like type of the displacement gradient components
```

samuelpmishLLNL (approved):

PR #820, opened 2022-10-07
Simplify Dependencies

### LINKED ISSUE #788: untangling target/test dependencies
Chris and I noticed that a lot of the CMake targets in serac have unnecessarily repeated dependencies:

serac_target_dependencies

In blue are the current explicitly defined intra-target dependencies, and in red are the essential dependencies after transitive reduction. Note: the targets ("physics_materials", "coefficients" and "integrators") are associated with the pre-`Functional` implementation and may be deprecated / removed.

Unfortunately, cleaning this up is trickier than expected, as the tests for each of these targets don't respect their associated locations in the project hierarchy. For example, the target `serac_infrastructure` is a leaf-node of the serac project (it has some external dependencies), but the tests in `serac/infrastructure/tests` depend on `serac_physics` (which itself depends on essentially everything in serac).

This is a bad habit, and I think it would be beneficial to untangle the tests such that their location in the project reflect the features under test. Tests should be placed in most nested directory as appropriate (i.e. the narrowest scope). So, a test that depends only on `serac_infrastructure` should be in `serac/infrastructure/tests` rather than `serac/tests`. Similarly, if a test depends on a lot more than `serac_infrastructure`, it should be in a parent directory, rather than in `serac/infrastructure/tests`.

chapman39:
Fixes #788

Here's a before-and-after of one of the graphs:
Screen Shot 2022-10-14 at 2 48 42 PM
Screen Shot 2022-10-14 at 2 48 54 PM

The only unneeded connections are with gtest. I tried making `test_utils` depend on `gtest_main` instead, but that gave me a bunch of gtest errors, so I decided not to mess with it.

`serac_functional` also has some private dependencies. If they were public these graphs could be simplified further (for example in the pic above `serac_physics_integrators->serac_numerics` could be removed), but I assumed they were private for a reason.

I've also created a python script (`generate_images.py`) that generates a bunch of images of the dependency graphs. It might be useful, but if not I can remove it from this PR. If you want me to keep it in I can add documentation for it.

Note this doesn't re-organize the tests it just removes the unnecessary dependencies.


samuelpmishLLNL:
> Note this doesn't re-organize the tests it just removes the unnecessary dependencies.

This PR is a step in the right direction, but until the tests are reorganized, we won't be able to remove most of their dependencies (but this can be done in a later PR).

For example, `serac/infrastructure/tests/serac_error_handling.cpp` has a test `TEST(SeracErrorHandling, BcRetrieveScalarCoef)` that depends on `BoundaryCondition`, which is not part of `serac/infrastructure`. So, in order for it to compile, we broadly made all the tests in `serac/infrastructure` depend on `serac_physics` to resolve `BoundaryCondition`. 

To me, it makes more sense to move that test to `serac/physics/tests` or `serac/physics/boundary_conditions/tests`, rather than `serac/infrastructure/tests`. Doing this would make it so that the dependencies of `serac/infrastructure/tests` will be smaller.

---------

Another way of putting it: I think our test directories should all have a CMakeLists.txt that has the following pattern (using `serac/infrastructure/tests/CMakeLists.txt` as an example):

```cmake
# note: the tests should only depend on the parent directory target and gtest
set(infrastructure_test_dependencies gtest serac_infrastructure)
```

and if a test like `serac_error_handling.cpp` doesn't compile because it needs to know about `serac_physics`, the solution should not be to expand `infrastructure_test_dependencies` to include all of serac, but to just move the test (either the whole file, or just the actual `gtest` function) to a different location in the project that has the appropriate dependencies.

scripts/graphviz/generate_images.py
```diff
@@ -0,0 +1,54 @@
...
+# other Serac Project Developers. See the top-level LICENSE file for
+# details.
+#
+# SPDX-License-Identifier: (BSD-3-Clause)
+
+"exec" "python3" "-u" "-B" "$0" "$@"
+
+#
+# usage:
+# python3 ./config-build.py -hc ./host-configs/rzgenie-toss_3_x86_64_ib-gcc@8.3.1.cmake --graphviz=graphviz/graph.dot
```
samuelpmishLLNL, line 13: interesting, I didn't know CMake a built-in option for dependency graph output!

chapman39:
@samuelpmishLLNL yeah that makes sense. we can try and get this pr in and in a separate pr i could start re-organizing the tests. i can remove the "fixes 788" thing for now since this doesn't resolve everything in the issue.

samuelpmishLLNL (approved):

src/serac/infrastructure/tests/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause) 
 
-set(test_dependencies gtest serac_physics serac_state)
+set(test_dependencies gtest serac_physics)
```
white238, line 7: This goes back to @samuelpmishLLNL 's comments, and outside the scope of this PR, about our unit tests being oddly over dependent.

This directory should not be dependent on `serac_physics` but is.

src/serac/numerics/functional/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause)
 
-set(functional_depends mfem mpi serac_infrastructure serac_mesh serac_numerics)
+set(functional_depends serac_mesh serac_numerics)
```
white238, line 7: I find the directory structure here a bit odd. I feel that subdirectories should children of their parents. If `serac_functional` depends on `serac_numerics` should it be pulled out into its own directory or maybe have this directory moved into a `numerics_core` or something?

white238: except tests.... they should be under the thing they are testing.

src/serac/physics/CMakeLists.txt
```diff
@@ -37,15 +37,9 @@ set(physics_headers
...
 
 set(physics_dependencies
-    serac_infrastructure
     serac_coefficients
-    serac_functional
     serac_physics_integrators
-    serac_physics_materials
-    serac_numerics
     serac_boundary_conditions
-    serac_state
```
white238, line 47: `serac_state` and `serac_materials` should be put back.

src/serac/physics/CMakeLists.txt
```diff
@@ -37,15 +37,9 @@ set(physics_headers
     )
 
 set(physics_dependencies
-    serac_infrastructure
     serac_coefficients
-    serac_functional
     serac_physics_integrators
```
white238, line 41: the fact that some have `serac_physics_whatever` and others don't is inconsistent here...

white238 (approved):

PR #821, opened 2022-10-12
Update spack import cmd

chapman39:
LiDO was having issues building due the new way Spack does the import command, so I figured it'd be good to update it here to prevent having to re-update it every time LiDO updates its Smith/ Serac version.


jamiebramwell (approved):
Thank you for fixing this!

scripts/spack/packages/adiak/package.py
```diff
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack import *
```
white238, line 6: This is a reversion of a recent change in Spack.  This should be fixed on the LiDO side not ours or better understood why this is a problem.

https://github.com/spack/spack/pull/30404

chapman39: It is being fixed on the LiDO side [here](https://lc.llnl.gov/gitlab/topopt/lido-2.0/-/merge_requests/420). And I thought it'd make sense to apply the fix here, so we don't have to every time LiDO needs a Smith/ Serac update.

I'm not sure what the issue was, but on building the tpls in Serac, it seemed to have made no difference.

samuelpmishLLNL (approved):

PR #827, opened 2022-10-19
patch tests for dynamics

btalamini:
Tests that solid mechanics physics module is able to exactly reproduce fields that are affine functions of space and time, which is essentially a test of consistency (in the sense of numerical approximation of differential equations).

Also implements a constant (in time) acceleration test, which tests second-order accuracy. This feature tests the body forcing term.

Bugs fixed in this PR:

- Sign convention in body forces
- Passes the correct time value from the ODE solver to the residual function
- Switches the solid mechanics default ODE integrator from `AverageAcceleration`, which turns out to be some [HHT-like method](https://github.com/mfem/mfem/blob/c644f3d1744af600f1a90c03e730c321543b58c5/linalg/ode.hpp#L785-L796), to the standard Newmark average acceleration method. This may not qualify as a "bug", but I was unable to get `AverageAcceleration` to pass the patch test. I do not yet know if this is an error in the MFEM implementation or if we are calling it incorrectly.

From here, we can replace the MFEM Newmark implementation with our own if we wish. I think it's good to have these tests in place first to reduce the chance of a regression.

This partially addresses #754. I'd leave that issue open until we make patch tests for `HeatTransfer`.


jandrej:
Randomly saw this. `mfem::SecondOrderODESolver` is mostly user contributed. No exhaustive tests are done.

jamiebramwell:
Good to know @jandrej ! In general, we think having hooks to the fancier integrators is probably necessary for stiff systems like electrochemistry, but might be overkill for things like solid mechanics.

src/serac/physics/tests/solid_dynamics_patch.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+using solid_mechanics::direct_dynamic_options;
+
+/**
+ * @brief Specify the kinds of boundary condition to apply
+ */
+enum class PatchBoundaryCondition
+{
+  Essential,
+  Mixed_essential_and_natural
```
samuelpmishLLNL: The second enum value name format looks strange to me, it looks like most of serac's `enum class` definitions are using `PascalCase` for their values. It looks like I'm guilty of writing the most enums that don't follow this pattern, so I should go and fix those too!

src/serac/physics/tests/solid_dynamics_patch.cpp
```diff
@@ -0,0 +1,438 @@
...
+      tensor<double, dim, dim> sigma = material(state, H);
+      auto                     F     = Identity<dim>() + H;
+      auto                     J     = det(F);
+      auto                     P     = J * dot(sigma, inv(transpose(F)));
+      // We don't have a good way to restrict the tractions to the
+      // complement of the essential boundary segments.
+      // The following matches the case when the top and left surfaces
+      // have essential boundary conditions:
+      //
+      // auto T = (n0[0] > 0.99 || n0[1] < -0.99)? dot(P, n0) : 0.0*n0;
```
samuelpmishLLNL, line 245: some of these predicates might be easier to specify in terms of the spatial coordinates, rather than normals
```cpp
auto T = (x[0] < x[1]) * dot(P, n0);
```

src/serac/numerics/odes.hpp
```diff
@@ -49,6 +49,8 @@ class SecondOrderODE : public mfem::SecondOrderTimeDependentOperator {
    * used by the residual operator
    */
   struct State {
+    double& time;
```
white238, line 55: doxygen comment required

src/serac/physics/tests/solid_dynamics_patch.cpp
btalamini: Thanks, just updated it. Somewhat relatedly, this type is duplicated in the statics patch tests. We should consider factoring out some of these utilities out into test fixtures.

src/serac/physics/solid_legacy.hpp
```diff
@@ -581,6 +581,8 @@ class SolidLegacy : public BasePhysics {
    */
   mfem::Vector previous_;
 
+  double ode_time_point_;
```
samuelpmishLLNL:
```suggestion
  /// @brief The value of time at which the ODE solver wants to evaluate the residual
  double ode_time_point_;
```

src/serac/physics/tests/solid_dynamics_patch.cpp
btalamini: That is true. But really, neither of those are the right way. Since this is just commented debug output, I prefer to leave it until we replace it with submesh boundary sets.

samuelpmishLLNL (approved):

src/serac/numerics/odes.hpp
btalamini: While I'm in there, I'll edit the doxygen comments on the other members that are misleading or incorrect.

src/serac/physics/solid_legacy.cpp
```diff
@@ -38,7 +38,8 @@ SolidLegacy::SolidLegacy(int order, const SolverOptions& options, GeometricNonli
                                                         .name       = detail::addPrefix(name, "adjoint_displacement")}),
       geom_nonlin_(geom_nonlin),
       keep_deformation_(keep_deformation),
-      ode2_(displacement_.space().TrueVSize(), {.c0 = c0_, .c1 = c1_, .u = u_, .du_dt = du_dt_, .d2u_dt2 = previous_},
+      ode2_(displacement_.space().TrueVSize(),
+            {.time = ode_time_point_, .c0 = c0_, .c1 = c1_, .u = u_, .du_dt = du_dt_, .d2u_dt2 = previous_},
```
white238, line 42: I can't find where `ode_time_point_` is getting initialized.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -132,10 +132,12 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
             FiniteElementState::Options{
                 .order = order, .vector_dim = mesh_.Dimension(), .name = detail::addPrefix(name, "shape_displacement")},
             sidre_datacoll_id_)),
-        nodal_forces_(mesh_, displacement_.space(), "nodal_forces"),
+        reactions_(mesh_, displacement_.space(), "reactions"),
         shape_sensitivity_(mesh_, displacement_.space(), "shape_sensitivity"),
-        ode2_(displacement_.space().TrueVSize(), {.c0 = c0_, .c1 = c1_, .u = u_, .du_dt = du_dt_, .d2u_dt2 = previous_},
+        ode2_(displacement_.space().TrueVSize(),
+              {.time = ode_time_point_, .c0 = c0_, .c1 = c1_, .u = u_, .du_dt = du_dt_, .d2u_dt2 = previous_},
```
white238, line 138: Wouldn't all these member variables be uninitialized at this point? I realize this was an existing problem...

src/serac/physics/solid_legacy.cpp
btalamini: As per Sam's suggestion I moved `ode_time_point_` to the `BasePhysics` module. It is now initialized there.

src/serac/physics/solid_legacy.hpp
btalamini: I moved this to the `BasePhysics` class and added the doxygen comment.

src/serac/physics/tests/solid_dynamics_patch.cpp
jamiebramwell: Speaking of, we should have all the pieces to do that now.

src/serac/numerics/odes.cpp
```diff
@@ -150,6 +150,7 @@ void SecondOrderODE::Solve(const double time, const double c0, const double c1,
 {
   // assign these values to variables with greater scope,
   // so that the residual operator can see them
+  state_.time  = time;
```
jamiebramwell, line 153: We should re-evaluate this interface strategy for ODE solvers. Tracking down the references and ownership of data can get confusing.

jamiebramwell (approved):
Thank you for tracking down the bugs!

PR #828, opened 2022-10-19
Cleanup legacy output logic

white238:
There was a lot of dead code that was left over from previous output formats.  If this was more heavy handed than it should have been tell me and I will revert it.


samuelpmishLLNL (approved):

src/serac/physics/tests/legacy_thermal_expansion.cpp
```diff
@@ -71,9 +71,6 @@ TEST(SolidLegacy, ThermalExpansion)
 
   solid_solver.setThermalExpansion(std::move(cte), std::move(ref_temp), temp);
```
white238: Should this file exist at this point? It is not being actually added as a test... or should I add it to the CMake file?

jamiebramwell: This was probably a miss at some point. As we're about to delete capability anyway, I say delete it.

jamiebramwell (approved):
Thanks for cleaning up my mess!

PR #832, opened 2022-10-24
Add more functions to `dual.hpp`

jamiebramwell:
This adds asin, acos, atan, min, and max to the dual class.

ping @kswartz92

### DIFF
--- a/src/serac/numerics/functional/dual.hpp
+++ b/src/serac/numerics/functional/dual.hpp
@@ -201,7 +201,7 @@ SERAC_HOST_DEVICE constexpr auto& operator-=(dual<gradient_type>& a, double b)
 
 /**
  * @brief Implementation of absolute value function for dual numbers
- * @note This is not differentiable at x = 0.0
+ * @note This is not differentiable at x = 0.0. At that point, the gradient is calculated as the gradient of x.
  */
 template <typename gradient_type>
 SERAC_HOST_DEVICE auto abs(dual<gradient_type> x)
@@ -211,7 +211,7 @@ SERAC_HOST_DEVICE auto abs(dual<gradient_type> x)
 
 /**
  * @brief Implementation of max for dual numbers
- * @note This is not differentiable at a == b
+ * @note This is not differentiable at a == b. At that point, the gradient is calculated as the gradient of b.
  */
 template <typename gradient_type>
 SERAC_HOST_DEVICE auto max(dual<gradient_type> a, double b)
@@ -220,10 +220,7 @@ SERAC_HOST_DEVICE auto max(dual<gradient_type> a, double b)
   return (a > b_dual) ? a : b_dual;
 }
 
-/**
- * @brief Implementation of max for dual numbers
- * @note This is not differentiable at a == b
- */
+/// @overload
 template <typename gradient_type>
 SERAC_HOST_DEVICE auto max(double a, dual<gradient_type> b)
 {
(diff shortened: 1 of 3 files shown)


src/serac/numerics/functional/tests/test_tensor_ad.cpp
```diff
@@ -20,14 +20,45 @@ TEST(Tensor, Norm)
...
+{
+  auto xd = max(make_dual(x), make_dual(y));
+  EXPECT_DOUBLE_EQ(std::max(x, y), xd.value);
+}
+
+TEST(DualNumberTensor, Min)
+{
+  auto xd = min(make_dual(x), make_dual(y));
+  EXPECT_DOUBLE_EQ(std::min(x, y), xd.value);
+}
```
samuelpmishLLNL: These tests should also check that the derivative has the correct value. i.e.

```cpp
const double x = 0.5;
const double y = 0.25;
...
auto xd = max(dual{x, 1.0}, dual{y, 2.0});
EXPECT_DOUBLE_EQ(x, xd.value);
EXPECT_DOUBLE_EQ(1.0, xd.gradient);
```

src/serac/numerics/functional/dual.hpp
```diff
@@ -206,6 +206,48 @@ SERAC_HOST_DEVICE auto abs(dual<gradient_type> x)
...
+{
+  return (a.value > b) ? a : b;
+}
+
+/** @brief implementation of max for dual numbers */
+template <typename gradient_type>
+SERAC_HOST_DEVICE auto max(double a, dual<gradient_type> b)
+{
+  return (a > b.value) ? a : b;
+}
```
samuelpmishLLNL: We should add tests for these mixed dual/double overloads. I don't think this implementation compiles-- the different branches of the ternary operator have different types.

Same for the dual/double `min` implementation.

jamiebramwell:
I think this is ready to go now. Any other comments @samuelpmishLLNL @white238 ?

white238 (approved):
Approved, pending random documentation thought

src/serac/numerics/functional/dual.hpp
```diff
@@ -199,13 +199,80 @@ SERAC_HOST_DEVICE constexpr auto& operator-=(dual<gradient_type>& a, double b)
...
+ */
 template <typename gradient_type>
 SERAC_HOST_DEVICE auto abs(dual<gradient_type> x)
 {
   return (x.value >= 0) ? x : -x;
 }
 
+/**
+ * @brief Implementation of max for dual numbers
+ * @note This is not differentiable at a == b
```
white238: Would these be better as `\overload brief description`?

src/serac/numerics/functional/dual.hpp
```diff
@@ -199,13 +199,80 @@ SERAC_HOST_DEVICE constexpr auto& operator-=(dual<gradient_type>& a, double b)
...
+template <typename gradient_type>
+SERAC_HOST_DEVICE auto max(dual<gradient_type> a, double b)
+{
+  auto b_dual = dual{b, 0.0};
+  return (a > b_dual) ? a : b_dual;
+}
+
+/**
+ * @brief Implementation of max for dual numbers
+ * @note This is not differentiable at a == b
```
samuelpmishLLNL: this note is good, maybe we could also say what it does in the case of equality (i.e. just returns the derivative of the second argument)

Another option might be to average the derivatives of the two operands when equal (this is probably overthinking things)

samuelpmishLLNL (approved):

PR #833, opened 2022-10-27
adding atan2 to dual.hpp

kswartz92:


samuelpmishLLNL:
we should add/test overloads for `atan2(dual<T> y, double x)` and `atan2(double y, dual<T> x)`, too

jamiebramwell (approved):
Thanks for the patch @kswartz92 !

white238 (approved):

PR #840, opened 2022-11-07
Add shape sensitivity to the thermal module

jamiebramwell:
- Adds shape displacement capabilities to the thermal module
- Refactors the thermal module to use a single `Functional` residual object
- Fixes a bug in the `FiniteElementVector` assignment operator to keep it from swapping out the associated `FiniteElementSpace`
- Removes the redundant mesh input in the `FiniteElementSpace`-based `FiniteElementVector` constructor
- Refactors the thermal materials to directly return heat capacity and thermal flux
- Fixes the first order time integrators to correctly update the sampling time in the physics driver
- Adds static and dynamic patch tests for `HeatTransfer`

NOTE: This branch contains https://github.com/LLNL/serac/pull/852 to ensure consistency. That PR should be merged first.


src/serac/physics/heat_transfer.hpp
```diff
@@ -237,34 +262,28 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+          // configuration and p is the shape displacement. We need the gradient with
+          // respect to the perturbed reference configuration x = X + p for the material model. Therefore, we calculate
+          // du/dx = du/dX * dX/dx = du/dX * (dx/dX)^-1 = du/dX * (I + dp/dX)^-1
 
-          auto response = material(x, temp, temp_grad, params...);
+          auto du_dx = dot(du_dX, inv(I + dp_dX));
 
-          auto source = response.specific_heat_capacity * response.density * u;
+          auto response = material(x + p, u, du_dx, params...);
+          auto source   = response.specific_heat_capacity * response.density * du_dt;
```
samuelpmishLLNL: `density` appears in the response struct and also the material type as a member variable.

Would it be appropriate to remove `density` from the response and just get it through the material model? (this is what we're doing in the solid mechanics module)

src/serac/physics/heat_transfer.hpp
```diff
@@ -237,34 +262,28 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
 
-          // Return the source and the flux as a tuple
-          return serac::tuple{source, flux};
+          // Note that the return is integrated in the perturbed reference
+          // configuration, hence the det(I + dp_dx) = det(dx/dX)
+          return serac::tuple{source * det(I + dp_dX),
+                              -1.0 * dot(response.heat_flux, inv(transpose(I + dp_dX))) * det(I + dp_dX)};
```
samuelpmishLLNL: I think we can simplify the long `dot(...)` expression slightly from

```
dot(response.heat_flux, inv(transpose(I + dp_dX)))
```
to
```
dot(inv(I + dp_dX), response.heat_flux)
```
-----------------

also, `det(I + dp_dX)` and `inv(I + dp_dX)` appear a couple of times in this kernel. It's probably worth only computing them once and storing it as temporary. I'm not sure what a good name for that matrix `I + dp_dX` would be.

src/serac/physics/heat_transfer.hpp
samuelpmishLLNL: another possible option is to make the thermal materials responsible for outputting the "source" term directly (which would require them to take in the time rate of the temperature, too).

src/serac/physics/heat_transfer.hpp
```diff
@@ -502,12 +556,35 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+   * problem with respect to the shape displacement field
+   *
+   * @return The sensitivity with respect to the shape displacement
+   *
+   * @pre `solveAdjoint` with an appropriate adjoint load must be called prior to this method.
+   */
+  FiniteElementDual& computeShapeSensitivity()
+  {
+    auto drdshape = serac::get<DERIVATIVE>((*residual_)(DifferentiateWRT<2>{}, temperature_, zero_, shape_displacement_,
+                                                        *parameter_states_[parameter_indices]...));
```
```suggestion
    auto drdshape = serac::get<DERIVATIVE>((*residual_)(DifferentiateWRT<SHAPE>{}, temperature_, zero_, shape_displacement_,
                                                        *parameter_states_[parameter_indices]...));
```

and define `SHAPE` up at the top with the other `static constexpr`s

src/serac/physics/heat_transfer.hpp
jamiebramwell: I just changed the thermal material interface to look more like the solid one (i.e. the `()` call returns flux and heat capacity and density are public struct members). I think further iterations on this interface should involve both modules to ensure they are consistent.

samuelpmishLLNL: I think density w.r.t. reference volume makes sense for a public member variable (i.e. it's homogeneous within a material block), but isn't heat capacity is a function of temperature and other stuff (@btalamini knows better than I do)?

btalamini: Yes, it's a response function, and can't be assumed constant in general.

btalamini: > another possible option is to make the thermal materials responsible for outputting the "source" term directly (which would require them to take in the time rate of the temperature, too).

I agree, there is an argument to be made for this. It would put the time and space derivatives on an equal footing in the q-function signature. Howver, it's hard for me to see if one way is better than the other yet. I'm also ok with making the heat capacity an output of the q-function.

samuelpmishLLNL: I think I'm leaning toward outputting the heat capacity

jamiebramwell: I'll take a stab at this change.

jamiebramwell: Now the thermal material returns a serac tuple of volumetric heat capacity and thermal flux directly. Let me know what you think!

src/serac/physics/materials/green_saint_venant_thermoelastic.hpp
```diff
@@ -74,7 +74,7 @@ struct GreenSaintVenantThermoelasticMaterial {
 
     state.strain_trace = get_value(trEg);
 
-    return serac::tuple{sigma, C_p * density, s0, q0};
+    return serac::tuple{sigma, C_v * density, s0, q0};
```
btalamini: This material got changed from when I wrote it. The `C_v` is meant to be the volumetric heat capacity (at constant volume), so it shouldn't be multiplied by the density.

The pattern of `rho*c` applies when the `c` is a heat capacity per unit mass (that is, specific heat). Multiplying by density turns it into volumetric heat capacity.

TL;DR: remove the `density` factor on the second output.

btalamini (changes requested):

src/serac/physics/materials/green_saint_venant_thermoelastic.hpp
jamiebramwell: Thanks! I was wondering about that and about to crack open my materials book to check.

src/serac/physics/heat_transfer.hpp
```diff
@@ -24,52 +24,52 @@
...
-          .print_level = 0,
-          .max_iter    = 200,
-          .lin_solver  = LinearSolver::CG,
-          .prec        = HypreSmootherPrec{mfem::HypreSmoother::Jacobi}};
-}
+const IterativeSolverOptions default_linear_options = {.rel_tol     = 1.0e-6,
+                                                       .abs_tol     = 1.0e-12,
+                                                       .print_level = 0,
+                                                       .max_iter    = 200,
+                                                       .lin_solver  = LinearSolver::CG,
```
btalamini: The heat transfer operator will often be self-adjoint, so CG will *usually* be a good choice. However, this isn't guaranteed in general, and the linear solver could have asymmetric matrices to work with. Maybe we should stay on the safe side for defaults and use GMRES. What do you think?

src/serac/physics/heat_transfer.hpp
```diff
@@ -224,47 +250,54 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
   }
 
   /**
-   * @brief Set the thermal flux and mass properties for the physics module
+   * @brief Set the thermal flux and heat capacity properties for the physics module
```
btalamini: The wording of this is a little misleading. The heat flux and the heat capacity are *outputs* of the material model, not properties to set. Maybe make this simpler, like "Set the material model for this region" or something like that.

src/serac/physics/heat_transfer.hpp
```diff
@@ -297,23 +330,42 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+          // Note that the current configuration x = X + p, where X is the original reference
+          // configuration and p is the shape displacement. We need the gradient with
+          // respect to the perturbed reference configuration x = X + p for the material model. Therefore, we calculate
+          // du/dx = du/dX * dX/dx = du/dX * (dx/dX)^-1 = du/dX * (I + dp/dX)^-1
 
-          auto flux = serac::zero{};
+          auto du_dx = dot(du_dX, inv(I_plus_dp_dX));
 
-          auto source = -1.0 * source_function(x, time_, u, du_dx, params...);
+          auto source = -1.0 * source_function(x + p, time_, u, du_dx, params...);
```
btalamini: This is reminding me to make patch tests which will check the sign convention of the loads (and more importantly, ensure it doesn't get broken in the future).

src/serac/physics/heat_transfer.hpp
```diff
@@ -328,17 +380,36 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+   *    2. `tensor<T,dim> n` the outward-facing unit normal for the quadrature point
+   *    3. `double t` the time (note: time will be handled differently in the future)
+   *    4. `T temperature` the current temperature at the quadrature point
+   *    4. `tuple{value, derivative}`, a variadic list of tuples (each with a values and derivative),
+   *            one tuple for each of the trial spaces specified in the `DependsOn<...>` argument.
+   *
+   * @note The actual types of these arguments passed will be `double`, `tensor<double, ... >` or tuples thereof
+   *    when doing direct evaluation. When differentiating with respect to one of the inputs, its stored
+   *    values will change to `dual` numbers rather than `double`. (e.g. `tensor<double,3>` becomes `tensor<dual<...>,
+   * 3>`)
```
btalamini, line 369: Just noticing that, because of the nice re-usable form of `Functional`, this comment is boilerplate that we copy and maintain separately for each physics module. Is there a way we could write this comment once (suitably genericized) and have it appear over these calls? I'm not a doxygen expert.

src/serac/physics/heat_transfer.hpp
```diff
@@ -328,17 +380,36 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
   void setFluxBCs(DependsOn<active_parameters...>, FluxType flux_function)
   {
-    K_functional_->AddBoundaryIntegral(
-        Dimension<dim - 1>{}, DependsOn<0, active_parameters + 1 ...>{},
-        [flux_function](auto x, auto n, auto u, auto... params) { return flux_function(x, n, u, params...); }, mesh_);
+    residual_->AddBoundaryIntegral(
+        Dimension<dim - 1>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
+        [this, flux_function](auto x, auto n, auto u, auto /* dtemp_dt */, auto shape, auto... params) {
+          auto p = get<VALUE>(shape);
+          return flux_function(x + p, n, ode_time_point_, u, params...);
```
btalamini: It's slightly suspicious that there is a minus sign on the heat source but not on the surface flux, but that will all come out in the wash once there is a patch test.

btalamini (approved):

src/serac/physics/heat_transfer.hpp
jamiebramwell: Sounds good to me. In fact, I think that's normally what I suggest to people. 😄

src/serac/physics/heat_transfer.hpp
jamiebramwell: Not sure. I'll look into it. However, these call lists will be subtlety different for each physics constitutive response/load, so it might be worth the repetition.

src/serac/physics/heat_transfer.hpp
jamiebramwell: I'll do a quick sanity check to verify.

src/serac/physics/heat_transfer.hpp
jamiebramwell: It looks like you can, but you have to use the entire docstrings. See [here](https://doxygen.nl/manual/commands.html#cmdcopydoc).

btalamini: That's a good find. We could replace "temperature" with "primal field", etc. in that block and only lose a tiny bit of specificity. Let's shelve that for now, but keep it in mind.

src/serac/physics/heat_transfer.hpp
jamiebramwell: Yep - this sign is wrong. Fix and updated test incoming...

src/serac/physics/tests/thermal.cpp
```diff
@@ -155,10 +154,10 @@ void functional_test_dynamic(double expected_temp_norm)
...
 }
 
-TEST(Thermal, 2DLinearStatic) { functional_test_static<1, 2>(2.2909240); }
-TEST(Thermal, 2DQuadStatic) { functional_test_static<2, 2>(2.29424403); }
-TEST(Thermal, 3DLinearStatic) { functional_test_static<1, 3>(46.6285642); }
-TEST(Thermal, 3DQuadStatic) { functional_test_static<2, 3>(46.6648538); }
+TEST(Thermal, 2DLinearStatic) { functional_test_static<1, 2>(3.0184961962491594); }
+TEST(Thermal, 2DQuadStatic) { functional_test_static<2, 2>(3.0186175426112518); }
+TEST(Thermal, 3DLinearStatic) { functional_test_static<1, 3>(4.0564914381870958); }
+TEST(Thermal, 3DQuadStatic) { functional_test_static<2, 3>(4.0573746914550073); }
```
samuelpmishLLNL: I don't want to be too disagreeable, but I did say that I was not going to approve PRs that introduce new tests with magic number answers based on the norm of the solution. Can we update this test to:

1. use one of the patch test meshes[^1].
2. use an analytic solution
3. demand that the relative error between the exact and approximate solutions is small.

When all is said and done, it should only amount to a difference of 5-10 lines of code, but it improves the value of the test significantly.

[^1]: In general, I don't see a compelling reason to use the beam/star meshes for any of the correctness tests, since they contain symmetries that make it possible for the finite element machinery to be wrong and still get the right answer by accident.

jamiebramwell: I was thinking that would be a separate PR, but I can take a stab at it here.

samuelpmishLLNL: oops, [9ad6528](https://github.com/LLNL/serac/pull/840/commits/9ad6528f1daf46b41fd5261ececf8dcd8f6d87f2) was a bigger change than I envisioned, but that is even more thorough!

jamiebramwell: Thanks go to @btalamini for making is patch test infrastructure so adaptable! I'm getting an error in the flux case that I need to investigate.

jamiebramwell:
Static patch test is done and passing.

samuelpmishLLNL (approved):

jamiebramwell:
There is something weird going on in the CI, so I will investigate this afternoon.

src/serac/physics/heat_transfer.hpp
```diff
@@ -337,7 +337,7 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
           auto source = source_function(x + p, ode_time_point_, u, du_dx, params...);
 
           // Return the source and the flux as a tuple
-          return serac::tuple{source * det(I_plus_dp_dX), serac::zero{}};
+          return serac::tuple{-1.0 * source * det(I_plus_dp_dX), serac::zero{}};
```
btalamini, line 340: I thought that the previous sign convention was opposite the one in solid mechanics, but I hadn't gotten the chance to double check it yet. Good catch.

jamiebramwell: This is what we have in the theory docs, so I put it in here. I can go check the solid one.

jamiebramwell: Confusingly, the solid docs have it the other way. So at least we're consistent with the documentation?

src/serac/physics/heat_transfer.hpp
```diff
@@ -126,24 +134,27 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
     parameters_.resize(sizeof...(parameter_space));
 
     // Create a pack of the primal field and parameter finite element spaces
-    std::array<const mfem::ParFiniteElementSpace*, sizeof...(parameter_space) + 1> trial_spaces;
+    mfem::ParFiniteElementSpace* test_space = &temperature_.space();
+
+    std::array<const mfem::ParFiniteElementSpace*, sizeof...(parameter_space) + NUM_STATE_VARS> trial_spaces;
     trial_spaces[0] = &temperature_.space();
+    trial_spaces[1] = &temperature_.space();
```
white238, line 141: Is temperature going to both trial_space 0 and 1 correct here?

src/serac/physics/tests/thermal_finite_diff.cpp
```diff
@@ -143,6 +140,124 @@ TEST(Thermal, FiniteDifference)
...
+TEST(HeatTransfer, FiniteDifferenceShape)
+{
+  MPI_Barrier(MPI_COMM_WORLD);
+
+  int serial_refinement   = 0;
+  int parallel_refinement = 0;
+
+  // Create DataStore
+  axom::sidre::DataStore datastore;
+  serac::StateManager::initialize(datastore, "thermal_parameterized_shape_sensitivities");
```
white238, line 152: If we are going from `Thermal` to `HeatTransfer` should these be renamed?

white238 (approved):

PR #844, opened 2022-11-09
Add the parameters to the output state arrays

### LINKED ISSUE #842: Parameters not being output
The output file does not contain the parameters defined in the drivers.

Sam's comment after looking at the code: the base physics class iterates over its solution fields and duals, but the parameters live in the derived physics modules.

jamiebramwell:
Resolves https://github.com/LLNL/serac/issues/842

### DIFF
--- a/src/serac/physics/heat_transfer.hpp
+++ b/src/serac/physics/heat_transfer.hpp
@@ -179,6 +179,9 @@ public:
     parameter_states_[i] = &parameter_state;
     parameter_sensitivities_[i] =
         StateManager::newDual(parameter_state.space(), parameter_state.name() + "_sensitivity");
+
+    // Add the parameter to the BasePhysics containers for output
+    states_.push_back(parameter_state);
     duals_.push_back(*parameter_sensitivities_[i]);
   }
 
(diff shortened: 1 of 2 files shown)


barreracruz:
Thank you for addressing this so quickly!

white238 (approved):

barreracruz (approved):

samuelpmishLLNL:
this was merged too quickly, in my opinion. The code

```cpp
  void setParameter(FiniteElementState& parameter_state, size_t i)
  {
    parameter_states_[i] = &parameter_state;
    parameter_sensitivities_[i] =
        StateManager::newDual(parameter_state.space(), parameter_state.name() + "_sensitivity");

    // Add the parameter to the BasePhysics containers for output
    states_.push_back(parameter_state);
    duals_.push_back(*parameter_sensitivities_[i]);
  }
```

is counterintuitive, it calls `vector::push_back(...)` in a setter. I think any C++ developer that sees a function named `setSomethingOrRather` should assume that it's admissible to call that as many times as necessary to update the value of `SomethingOrRather`, but that won't work here.

The number of parameters is known a priori, so we should just size the container correctly at construction and make this a true setter (with no other side-effects).

PR #851, opened 2022-11-10
clarify q-function requirements for solid mechanics module

samuelpmishLLNL:
Also update the links on the doxygen main page to point to the actual physics modules, rather than empty forward declarations.

### DIFF
--- a/src/docs/doxygen/mainpage.md
+++ b/src/docs/doxygen/mainpage.md
@@ -3,8 +3,8 @@
 #### Physics Simulation Modules ####
 
 * [Base physics](@ref serac::BasePhysics)
-* [Heat Transfer](@ref serac::HeatTransfer)
-* [Solid mechanics](@ref serac::SolidMechanics)
+* [Heat Transfer](classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html)
+* [Solid mechanics](classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html)
 * [Thermomechanics](@ref serac::Thermomechanics)
 
 #### Simulation Utilities ####
@@ -23,7 +23,7 @@
 
 #### Functional Interface ####
 
-* <a style="font-weight:bold" href="classserac_1_1Functional_3_01test_07trial_08_4.html">Functional</a>
+* [Functional](classserac_1_1Functional_3_01test_07trials_8_8_8_08_00_01exec_01_4.html)
 * [Domain integral](@ref serac::DomainIntegral)
 * [Boundary integral](@ref serac::BoundaryIntegral)
 
--- a/src/docs/doxygen/mainpage.md
+++ b/src/docs/doxygen/mainpage.md
@@ -3,8 +3,8 @@
 #### Physics Simulation Modules ####
 
 * [Base physics](@ref serac::BasePhysics)
-* <a style="font-weight:bold" href="classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html">Heat Transfer</a>
-* <a style="font-weight:bold" href="classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html">Solid Mechanics</a>
+* [**Heat Transfer**][1]
+* [**Solid Mechanics**][2]
 * [Thermomechanics](@ref serac::Thermomechanics)
 
 #### Simulation Utilities ####
@@ -23,7 +23,7 @@
 
 #### Functional Interface ####
 
-* <a style="font-weight:bold" href="classserac_1_1Functional_3_01test_07trials_8_8_8_08_00_01exec_01_4.html">Functional</a>
+* [**Functional**][3]
 * [Domain integral](@ref serac::DomainIntegral)
 * [Boundary integral](@ref serac::BoundaryIntegral)
 
@@ -36,3 +36,8 @@
 * <a style="font-weight:bold" href="logger_8hpp.html">Logging</a>
 * <a style="font-weight:bold" href="profiling_8hpp.html">Profiling</a>
 * <a style="font-weight:bold" href="terminator_8hpp.html">Terminator</a>
+
+<!-- Reference Links -->
+[1]: classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html
+[2]: classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html
+[3]: classserac_1_1Functional_3_01test_07trials_8_8_8_08_00_01exec_01_4.html
--- a/src/docs/doxygen/mainpage.md
+++ b/src/docs/doxygen/mainpage.md
@@ -3,8 +3,8 @@
 #### Physics Simulation Modules ####
 
 * [Base physics](@ref serac::BasePhysics)
-* [Heat Transfer](classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html)
-* [Solid mechanics](classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html)
+* <a style="font-weight:bold" href="classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html">Heat Transfer</a>
+* <a style="font-weight:bold" href="classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html">Solid Mechanics</a>
 * [Thermomechanics](@ref serac::Thermomechanics)
 
 #### Simulation Utilities ####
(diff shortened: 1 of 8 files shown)


samuelpmishLLNL:
Note: the `HeatTransfer` module should have something similar done, but #840 is changing some of those interfaces so this PR only does the solid mechanics module.

src/docs/doxygen/mainpage.md
```diff
@@ -3,8 +3,8 @@
 #### Physics Simulation Modules ####
 
 * [Base physics](@ref serac::BasePhysics)
-* [Heat Transfer](@ref serac::HeatTransfer)
-* [Solid mechanics](@ref serac::SolidMechanics)
+* [Heat Transfer](classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html)
+* [Solid mechanics](classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html)
```
btalamini: Is there a reason to remove the doxygen `@ref` commands? Is it not functioning, or was it the wrong thing to do in the first place? There's an incongruity between the unmodified links and the new explicit references to (very ugly) html files. Possibly related: when I build the documentation on my system, the links that are modified look different, being rendered in regular typeface instead of bold.

src/docs/doxygen/mainpage.md
```diff
@@ -23,7 +23,7 @@
 
 #### Functional Interface ####
 
-* <a style="font-weight:bold" href="classserac_1_1Functional_3_01test_07trial_08_4.html">Functional</a>
+* [Functional](classserac_1_1Functional_3_01test_07trials_8_8_8_08_00_01exec_01_4.html)
```
btalamini: Same comment as above.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -349,6 +349,16 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
    * @tparam MaterialType The solid material type
    * @tparam StateType the type that contains the internal variables for MaterialType
    * @param material A material that provides a function to evaluate stress
+   * @pre body_force must be a object that can be called with the following arguments:
```
jamiebramwell:
```suggestion
   * @pre material must be a object that can be called with the following arguments:
```

src/serac/physics/solid_mechanics.hpp
```diff
@@ -349,6 +349,16 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   *    1. `tensor<T,dim> x` the spatial coordinates for the quadrature point
```
jamiebramwell: I believe this should be state, displacement gradient, parameters....

src/serac/physics/solid_mechanics.hpp
```diff
@@ -349,6 +349,16 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   *    2. `double t` the time (note: time will be handled differently in the future)
+   *    3+. `tuple{value, derivative}`, a tuple of values and derivatives for each of the trial spaces
```
btalamini: Is the `+` sign after the `3` a typo?

jamiebramwell:
> Note: the `HeatTransfer` module should have something similar done, but #840 is changing some of those interfaces so this PR only does the solid mechanics module.

I'll add the similar changes directly in #840

src/serac/physics/solid_mechanics.hpp
```diff
@@ -349,6 +349,16 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   *            specified in the `DependsOn<...>` argument. 
+   *
+   * @note The actual types of these arguments passed will be `double`, `tensor<double, ... >` or tuples thereof
```
btalamini, line 358: This note is helpful. However, the arrow between the template classes is a little confusing, because it looks like broken C++. maybe replace it with "to" if we can't get an actual arrow glyph?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -349,6 +349,16 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   *    when doing direct evaluation. When differentiating with respect to one of the inputs, its stored
+   *    values will change to `dual` numbers rather than `double`. (e.g. `tensor<double,3>` -> `tensor<dual<...>, 3>`) 
+   *
    * @param qdata the buffer of material internal variables at each quadrature point
```
btalamini, line 363: Should the prereq for `operator()` now demand that it returns Cauchy stress instead of Kirchhoff? (Note to ourselves, we need to run a spellcheck over the whole docs to look for missing h's in "Kirchhoff" 😁)

src/serac/physics/solid_mechanics.hpp
```diff
@@ -459,9 +469,15 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
    * @tparam BodyForceType The type of the body force load
-   * @param body_force A source function for a prescribed body load
+   * @pre body_force must be a object that can be called with the following arguments:
+   *    1. `tensor<T,dim> x` the spatial coordinates for the quadrature point
+   *    2. `double t` the time (note: time will be handled differently in the future)
+   *    3. `tuple{value, derivative}`, a variadic list of tuples (each with a values and derivative), 
+   *            one tuple for each of the trial spaces specified in the `DependsOn<...>` argument. 
+   * @note The actual types of these arguments passed will be `double`, `tensor<double, ... >` or tuples thereof
+   *    when doing direct evaluation. When differentiating with respect to one of the inputs, its stored
+   *    values will change to `dual` numbers rather than `double`. (e.g. `tensor<double,3>` -> `tensor<dual<...>, 3>`)
```
btalamini: Same comment about arrow symbol.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -491,7 +507,16 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   * @pre TractionType must be a object that can be called with the following arguments:
+   *    1. `tensor<T,dim> x` the spatial coordinates for the quadrature point
+   *    2. `tensor<T,dim> n` the outward-facing unit normal for the quadrature point
+   *    3. `double t` the time (note: time will be handled differently in the future)
+   *    4. `tuple{value, derivative}`, a variadic list of tuples (each with a values and derivative), 
+   *            one tuple for each of the trial spaces specified in the `DependsOn<...>` argument. 
+   *
+   * @note The actual types of these arguments passed will be `double`, `tensor<double, ... >` or tuples thereof
+   *    when doing direct evaluation. When differentiating with respect to one of the inputs, its stored
+   *    values will change to `dual` numbers rather than `double`. (e.g. `tensor<double,3>` -> `tensor<dual<...>, 3>`)
```
btalamini: ditto.

btalamini (commented):
The documentation on `Functional` is nice - we should all shoot for this level of completeness throughout the code.

src/docs/doxygen/mainpage.md
samuelpmishLLNL: `@ref` didn't seem to work when I tried initially, but you're right that the formatting is different-- presumably why the links in lower in the file resort to HTML rather than markdown link syntax to ensure things are rendered bold. I'll look again and see if I can make things stylistically consistent again.

src/serac/physics/solid_mechanics.hpp
samuelpmishLLNL: I was looking for a way to communicate that arguments 3 and above were the variadic list, and apparently not doing a good job!

Furthermore, I think the "+" screws up the list formatting, so I should just remove it.

src/serac/physics/solid_mechanics.hpp
samuelpmishLLNL: This is font-dependent, a typeface with ligatures (e.g. Fira Code) does render an arrow glyph when the characters `->` appear adjacent.

samuelpmishLLNL: But I can still change it to something else if you prefer

src/serac/physics/solid_mechanics.hpp
samuelpmishLLNL: Whoops, good point! This is the main source of the confusion: although `Functional` has a clear set of rules for arguments are handled, our physics modules have no consistency on what they expect from the q-functions.

src/serac/physics/solid_mechanics.hpp
jamiebramwell: See https://github.com/LLNL/serac/pull/853

jamiebramwell (approved):
Really great! Thanks!

src/docs/doxygen/mainpage.md
```diff
@@ -3,8 +3,8 @@
 #### Physics Simulation Modules ####
 
 * [Base physics](@ref serac::BasePhysics)
-* [Heat Transfer](classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html)
-* [Solid mechanics](classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html)
+* <a style="font-weight:bold" href="classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html">Heat Transfer</a>
+* <a style="font-weight:bold" href="classserac_1_1SolidMechanics_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_0a4b58643425b41433838b2828455c49.html">Solid Mechanics</a>
```
btalamini: If I change the relevant link to

```markdown
* [Solid Mechanics](@ref serac::SolidMechanics)
```

everything works as expected; the type is rendered in boldface on the web page and the link jumps to the right spot. I'm building the `serac_doxygen` target. Your system is behaving differently, I suppose?

src/docs/sphinx/dev_guide/tensor_dual.rst
```diff
@@ -343,16 +343,16 @@ pattern as before:
    tensor<double,3,3> L = ...;
 
    // promote the arguments to dual numbers with make_dual()
-   std::tuple dual_args = make_dual(p, v, L);
+   tuple dual_args = make_dual(serac::tuple{p, v, L});
```
btalamini, line 346: Thanks for fixing this section.

src/serac/numerics/functional/tests/tuple_arithmetic_unit_tests.cpp
```diff
@@ -95,6 +95,37 @@ TEST(TupleArithmeticUnitTests, TensorOutputWithTupleInput)
   EXPECT_NEAR(norm(df1 - df0) / norm(df0), 0.0, 2.0e-8);
 }
 
+TEST(TupleArithmeticUnitTests, ReadTheDocsExample)
```
btalamini, line 98: Great idea. Not quite as nifty as Rust parsing the dosctrings, but a good compromise for c++ to help us keep the docs and the code in synch.

src/docs/doxygen/mainpage.md
btalamini: Ok, I understand the problem now. The auto-generated doxygen link is going to the wrong place.

btalamini (approved):

PR #852, opened 2022-11-11
Unify parameter and sensitivity methods

### LINKED ISSUE #850: setParameter fix added bugs and can cause duplicate values
```cpp
  void setParameter(FiniteElementState& parameter_state, size_t i)
  {
    parameter_states_[i] = &parameter_state;
    parameter_sensitivities_[i] =
        StateManager::newDual(parameter_state.space(), parameter_state.name() + "_sensitivity");

    // Add the parameter to the BasePhysics containers for output
    states_.push_back(parameter_state);
    duals_.push_back(*parameter_sensitivities_[i]);
  }
```

is counterintuitive, it calls `vector::push_back(...)` in a setter. I think any C++ developer that sees a function named `setSomethingOrRather` should assume that it's admissible to call that as many times as necessary to update the value of `SomethingOrRather`, but that won't work here.

The number of parameters is known a priori, so we should just size the container correctly at construction and make this a true setter (with no other side-effects).

_Originally posted by @samuelpmishLLNL in https://github.com/LLNL/serac/issues/844#issuecomment-1310790491_

jamiebramwell:
This creates a common interface for parameters and sensitivities between physics modules for easier use by optimization libraries. Some key points:

- This pushes as many of the methods as possible up to the `BasePhysics` class. However, the compute sensitivity method must be a template and therefore cannot be in the base class. This could be relaxed if we compute all of the available sensitivities at once. Any thoughts? UPDATE: I was able to move the sensitivity methods up to the base class. This means that an optimizer code should be able to work on the base class methods.
- It adds `FiniteElementState` accessors by name and a routine to return the valid names.
- Adds a `generateParameter` factory method where the physics module can build an appropriate `FiniteElementState` based on the finite element basis given as a template parameter. The physics user is still expected to manage the lifetime of this object.
- Consolidates the parameter information into a single struct.
- Moves the shape displacement information from the physics module to the `StateManager`. This is because the shape parameter is tied to the mesh instead of the specific physics.

Resolves https://github.com/LLNL/serac/issues/850


src/serac/physics/base_physics.hpp
```diff
@@ -99,13 +92,91 @@ class BasePhysics {
...
+  /**
+   * @brief Generate a finite element state object for the given parameter index
+   *
+   * @param parameter_index The index of the parameter to generate
+   * @param parameter_name The name of the parameter to generate
+   *
+   * @note The user is responsible for managing the lifetime of this object. It is required
+   * to exist whenever this physics module is called.
+   */
+  virtual std::unique_ptr<FiniteElementState> generateParameter(size_t             parameter_index,
```
kswartz92: I think it would be more consistent with `setParameter()` to make the index the second argument

src/serac/physics/base_physics.hpp
```diff
@@ -99,13 +92,91 @@ class BasePhysics {
...
+   */
+  void setParameter(FiniteElementState& parameter_state, size_t parameter_index);
+
+  /**
+   * @brief Get the parameter field of the physics module
+   *
+   * @param parameter_index The parameter index to retrieve
+   * @return The FiniteElementState representing the user-defined parameter
+   */
+  virtual FiniteElementState& getParameter(size_t parameter_index)
```
kswartz92: Is there a reason to make this method virtual? I'm wondering if this method should not be virtual so that derived classes cannot change its behavior. Same comment for the const version of this method

jamiebramwell: Good point. I think with the new dynamic vectors of parameters I can do this.

kswartz92:
LGTM

kswartz92 (approved):

samuelpmishLLNL:
regarding 

>However, the compute sensitivity method must be a template and therefore cannot be in the base class. This could be relaxed if we compute all of the available sensitivities at once.

I think there's another approach, since the actual signature of the `computeSensitivity()` method is independent of the parameter:

https://godbolt.org/z/jo5Ga1roK

i.e. keep an array of `std::function`s of the partially-evaluated residual calls, and use the parameter index to select the right one.

it's not elegant, but it might be worth considering, if it helps LIDO.

jamiebramwell:
> regarding
> 
> > However, the compute sensitivity method must be a template and therefore cannot be in the base class. This could be relaxed if we compute all of the available sensitivities at once.
> 
> I think there's another approach, since the actual signature of the `computeSensitivity()` method is independent of the parameter:
> 
> https://godbolt.org/z/jo5Ga1roK
> 
> i.e. keep an array of `std::function`s of the partially-evaluated residual calls, and use the parameter index to select the right one.
> 
> it's not elegant, but it might be worth considering, if it helps LIDO.

@kswartz92 @sethwatts @tepperly , would this help your interface?

sethwatts:
> @kswartz92 @sethwatts @tepperly , would this help your interface?

In general, computing all the available sensitivities _may_ be wasteful if we won't use all of them, but with the new wrapper @kswartz92 is developing, I think this is less of a problem than it used to be, since it's now much easier to only define parameters relevant to a given problem. (There is still the possibility of computing unused sensitivities, but it would be fewer than before.) I don't know whether leaving the sensitivity call in the derived classes hurts that interface, so I will defer to @kswartz92. If it does, @samuelpmishLLNL 's solution would be more efficient at runtime than computing all available sensitivities, but I don't know how that balances against maintenance cost in Serac.

kswartz92:
> > @kswartz92 @sethwatts @tepperly , would this help your interface?
> 
> In general, computing all the available sensitivities _may_ be wasteful if we won't use all of them, but with the new wrapper @kswartz92 is developing, I think this is less of a problem than it used to be, since it's now much easier to only define parameters relevant to a given problem. (There is still the possibility of computing unused sensitivities, but it would be fewer than before.) I don't know whether leaving the sensitivity call in the derived classes hurts that interface, so I will defer to @kswartz92. If it does, @samuelpmishLLNL 's solution would be more efficient at runtime than computing all available sensitivities, but I don't know how that balances against maintenance cost in Serac.

I don't think it's a big concern either way. I think It would be nice to move as many methods as possible to the base class to ensure compatibility of all derived physics modules, but as long as we can call the required methods (`getParameter()`, `getState()`, and `computeSensitivity()`) then our interface will work as we're hoping.

Regarding the efficiency aspect, my understanding is that only required sensitivities would be calculated, whether or not this method is moved into the base physics. Am I understanding this correct?

jamiebramwell:
> Regarding the efficiency aspect, my understanding is that only required sensitivities would be calculated, whether or not this method is moved into the base physics. Am I understanding this correct?

Yes, if we use @samuelpmishLLNL 's method.

sethwatts:
If we do not use @samuelpmishLLNL 's method and instead compute all available gradients rather than just those LiDO or another optimization library wants (which per @kswartz92 's comment, may not even be necessary), the worst is that we do some unnecessary quadrature/assembly operations. The expensive part almost always going to be the adjoint solve. As @kswartz92 says, it's not a huge concern one way or the other. I think whatever is cleanest for Serac makes sense here.

jamiebramwell:
Using @samuelpmishLLNL 's solution, I was able to make the `computeSensitivity` a base class method. Let me know what you think!

jamiebramwell:
This branch is now ready for review. It should go in before https://github.com/LLNL/serac/pull/840 .

src/serac/physics/base_physics.hpp
```diff
@@ -99,13 +93,146 @@ class BasePhysics {
...
+   * @note The user is responsible for managing the lifetime of this object. It is required
+   * to exist whenever advanceTimestep, solveAdjoint, or computeSensitivity is called.
+   *
+   * @note The finite element space for this object is generated from the parameter
+   * discretization space (e.g. L2, H1) and the computational mesh given in the physics module constructor.
+   */
+  std::unique_ptr<FiniteElementState> generateParameter(const std::string& parameter_name, size_t parameter_index);
+
+  /**
+   * @brief register the provided FiniteElementState object as the source of values for parameter `i`
```
white238:
```suggestion
   * @brief Register the provided FiniteElementState object as the source of values for parameter `i`
```

src/serac/physics/base_physics.hpp
```diff
@@ -99,13 +93,146 @@ class BasePhysics {
...
+
+  /**
+   * @brief Get a vector of the finite element state solution variable names
+   *
+   * @return The solution variable names
+   */
+  virtual std::vector<std::string> stateNames() = 0;
+
+  /**
+   * @brief Generate a finite element state object for the given parameter index
```
white238, line 112: There is some inconsistency between "finite element state" vs "FiniteElementState" vs "Finite Element State".  I am sure this existed before as well.  We may want to do a bulk doxygen cleanup PR.

src/serac/physics/base_physics.hpp
```diff
@@ -161,17 +288,39 @@ class BasePhysics {
...
+  /// @brief The information needed for the physics parameters stored as Finite Element State fields
+  struct ParameterInfo {
+    /// The trial spaces used for the Functional object
+    std::unique_ptr<mfem::ParFiniteElementSpace> trial_space;
+
+    /// The finite element states representing user-defined and owned parameter fields
+    serac::FiniteElementState* state;
+
+    /**
+     * @brief The sensitivities (dual vectors) with repect to each of the input parameter fields
```
white238:
```suggestion
     * @brief The sensitivities (dual vectors) with respect to each of the input parameter fields
```

src/serac/physics/base_physics.hpp
```diff
@@ -161,17 +288,39 @@ class BasePhysics {
...
+     * @note this optional as FiniteElementDuals are not default constructable and
```
white238:
```suggestion
     * @note This is optional as FiniteElementDuals are not default constructable and
```

src/serac/physics/heat_transfer.hpp
```diff
@@ -368,6 +351,35 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+   */
+  const FiniteElementState& state(const std::string& state_name) override
+  {
+    if (state_name == "temperature") {
+      return temperature_;
+    } else if (state_name == "adjoint_temperature") {
+      return adjoint_temperature_;
+    }
+
+    SLIC_ERROR_ROOT(axom::fmt::format("State {} requestion from solid mechanics module {}, but it doesn't exist",
```
white238:
```suggestion
    SLIC_ERROR_ROOT(axom::fmt::format("State '{}' requested from solid mechanics module '{}', but it doesn't exist",
```

src/serac/physics/solid_mechanics.hpp
```diff
@@ -759,43 +746,41 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
    */
-  template <int parameter_field>
-  FiniteElementDual& computeSensitivity()
+  FiniteElementDual& computeSensitivity(size_t parameter_field) override
   {
-    auto drdparam =
-        serac::get<DERIVATIVE>((*residual_)(DifferentiateWRT<parameter_field + NUM_STATE_VARS>{}, displacement_, zero_,
-                                            shape_displacement_, *parameter_states_[parameter_indices]...));
+    SLIC_ASSERT_MSG(parameter_field < sizeof...(parameter_indices),
+                    axom::fmt::format("Invalid parameter index {} reqested for sensitivity."));
```
white238:
```suggestion
                    axom::fmt::format("Invalid parameter index {} requested for sensitivity."));
```

src/serac/physics/state/state_manager.cpp
```diff
@@ -76,30 +81,50 @@ void StateManager::initialize(axom::sidre::DataStore& ds, const std::string& out
...
     state.setFromGridFunction(*grid_function);
   }
   named_states_[name] = grid_function;
+}
+
+FiniteElementState StateManager::newState(FiniteElementVector::Options&& options, const std::string& mesh_tag)
+{
+  SLIC_ERROR_ROOT_IF(!ds_, "Serac's data store was not initialized - call StateManager::initialize first");
+  SLIC_ERROR_ROOT_IF(datacolls_.find(mesh_tag) == datacolls_.end(),
+                     axom::fmt::format("Mesh tag \"{}\" not found in the data store", mesh_tag));
```
white238: We, not specifically this PR, really are all over the place quoting injected values... sigh.

src/serac/physics/state/state_manager.cpp
```diff
@@ -140,23 +177,8 @@ FiniteElementDual StateManager::newDual(const mfem::ParFiniteElementSpace& space
...
-                       axom::fmt::format("Serac's datacollection was already given a field named '{0}'", dual_name));
-    // Create a new grid function with unallocated data. This will be managed by sidre.
-    grid_function = new mfem::ParGridFunction(&dual.space(), static_cast<double*>(nullptr));
-    datacoll.RegisterField(dual_name, grid_function);
-    std::unique_ptr<mfem::HypreParVector> true_dofs(grid_function->GetTrueDofs());
-    dual = *true_dofs;
-  }
-  named_duals_[dual_name] = grid_function;
+  auto dual = FiniteElementDual(mesh(mesh_tag), space, dual_name);
+  storeDual(dual);
```
white238, line 181: 👍

src/serac/physics/base_physics.hpp
samuelpmishLLNL: Would you prefer "finite element state" in these kind of comments, or `FiniteElementState`?

src/serac/physics/state/state_manager.cpp
samuelpmishLLNL: Is this saying you would prefer to not put quotes around them? It seems appropriate here, since it makes clear the exact value of the `mesh_tag` string.

white238: I prefer to quote the values that are given in error messages.

Example:
```
Error: The value 'foo' given by user is super duper bad.
```

samuelpmishLLNL: I've made changes to most of the formatted strings in serac (I skipped a few in serac/infrastructure). I use single quotes around the substitutions to avoid having to escape the double quote character in the string. For some substitutions with numeric output (rather than indices, strings)

```cpp
axom::fmt::format("Bulk sensitivity vector norm: {}", bulk_norm)
```

I just left the single quotes off, as it seems to produce cleaner output.

src/serac/physics/base_physics.hpp
```diff
@@ -99,13 +93,148 @@ class BasePhysics {
...
+   * @return The computed adjoint finite element state
+   */
+  virtual const serac::FiniteElementState& solveAdjoint(FiniteElementDual& /*adjoint_load */,
+                                                        FiniteElementDual* /* dual_with_essential_boundary */ = nullptr)
+  {
+    SLIC_ERROR_ROOT(axom::fmt::format("Adjoint analysis not defined for physics module {}", name_));
+
+    // Return a dummy state value to quiet the compiler. This will never get used.
+    return *states_[0];
+  }
```
samuelpmishLLNL, line 236: I think some of these functions should be pure virtual in the future (i.e. once we remove the "Legacy" physics modules), so that the compiler will emit an error message if a new physics module forgets to implement one, instead of defaulting to these incorrect placeholder implementations.

samuelpmishLLNL (approved):

src/serac/physics/base_physics.hpp
white238: I generally prefer the class name so its easy to look up but consistency is most important. We can fix it after in a general doxygen pr.

white238 (approved):

PR #853, opened 2022-11-14
Fix the Kirchhoff typos

jamiebramwell:
Also fix the docs that we now return Cauchy stress.

### DIFF
--- a/src/serac/physics/materials/solid_material.hpp
+++ b/src/serac/physics/materials/solid_material.hpp
@@ -152,7 +152,7 @@ struct J2 {
  * @tparam dim number of spatial dimensions
  *
  * @param displacement_gradient Displacement gradient
- * @param kirchhoff_stress Kirchoff stress
+ * @param kirchhoff_stress Kirchhoff stress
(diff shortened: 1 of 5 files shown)


samuelpmishLLNL (approved):

src/serac/physics/materials/solid_material.hpp
```diff
@@ -152,7 +152,7 @@ struct J2 {
  * @tparam dim number of spatial dimensions
  *
  * @param displacement_gradient Displacement gradient
- * @param kirchhoff_stress Kirchoff stress
```
btalamini, line 155: doh! I wrote this one. Got the name right in the variable but wrong in the comment. Probably copied and pasted it like a drone.

btalamini (approved):

src/serac/physics/materials/solid_material.hpp
jamiebramwell: Well, considering my terrible spelling is probably to blame I can't get too mad. 😆

PR #855, opened 2022-11-16
Feature/talamini/differentiable newton

btalamini:
Implement an adjoint-capable iterative solver for scalar-valued nonlinear equations.

This PR implements a similar algorithm as the well-known "RTSAFE" routine from Numerical Recipes, which globalizes Newton's method with fallback bisection steps. However, this version also implements the adjoint solve, so that derivatives of the unknown with respect to additional parameters are computed (and computed efficiently). The nonlinear equation to solve should have the form $f(x, p_1, p_2, \ldots) = 0$, where $x$ is the unknown and $p_i$ are parameters, which may be of arbitrary tensor character. If any of the $p_i$ are dual number-valued, the output of the solver is also a dual number with the appropriate directional derivative.

Here is a concrete example, taken from the unit tests (it's admittedly contrived, but illustrative):

```c++
// compute the Frobenius norm using Newton's method
auto my_norm = [](auto A) {
    double tolerance = 1e-10;
    double lower = 1e-3;
    double upper = 20.0;
    double x0 = get_value(inner(A, A));
    auto [x, status] = solve_scalar_equation(
        [](auto x, auto P){ return x*x - inner(P, P); }, x0, tolerance, lower, upper, A);
    return x;
};

tensor< double, 3, 3 > A = {{{1.0, 2.0, 3.0}, 
                             {4.0, 5.0, 6.0}, 
                             {7.0, 8.0, 9.0}}};

// compute ||A|| and d(||A||)/dA
auto [normA, dnormA_dA] = my_norm(make_dual(A));
```

Be aware that the code in `tuple.hpp`, `tensor.hpp`, and `dual.hpp` have been refactored in order to break a circular dependency (thanks, @samuelpmishLLNL).


src/serac/numerics/functional/tensor.hpp
```diff
@@ -244,6 +240,22 @@ SERAC_HOST_DEVICE zero get(const zero&)
...
+}
+
+/** @brief let a scalar be accessed like a degenerate tuple */
+template <int i>
+SERAC_HOST_DEVICE double& get(double& x)
+{
+  static_assert(i == 0);
+  return x;
+}
+
```
samuelpmishLLNL: Sorry, this one is on me. Do we still need these? I believe we deleted the code that called these.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -756,6 +692,18 @@ SERAC_HOST_DEVICE constexpr auto inner(const tensor<S, m, n>& A, const tensor<T,
...
+SERAC_HOST_DEVICE constexpr auto inner(const tensor<S, m>& A, const tensor<T, m>& B)
+{
+  decltype(S{} * T{}) sum{};
+  for (int i = 0; i < m; i++) {
+    sum += A[i] * B[i];
+  }
+  return sum;
+}
+
+SERAC_HOST_DEVICE constexpr auto inner(double A, double B) { return A * B; }
```
samuelpmishLLNL: same here, I'm not sure these are used any more. Maybe they're still worth keeping

samuelpmishLLNL (approved):

src/serac/numerics/functional/tensor.hpp
jamiebramwell: I think is still worth keeping as a general purpose capability.

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -0,0 +1,673 @@
...
+        converged   = (x == xl);
+      } else {  // use Newton step
+        delta_x_old = delta_x;
+        delta_x     = fval / df_dx;
+        auto temp   = x;
+        x -= delta_x;
+        converged = (x == temp);
+      }
+
+      // std::cout << "iter " << iterations << " x = " << x << std::endl;
```
jamiebramwell: dead code

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -0,0 +1,673 @@
...
+
+  // Accumulate derivatives so that the user can get derivatives
+  // with respect to parameters, subject to constraing that f(x, p) = 0 for all p
+  // Conceptually, we're doing the following:
+  // [fval, df_dp] = f(get_value(x), p)
+  // df = 0
+  // for p in params:
+  //   df += inner(df_dp, dp)
+  // dx = -df / df_dx
+  constexpr bool contains_duals =
```
jamiebramwell, line 653: Is this useful enough to be it's own function? Essentially is anything in the parameter pack a dual?

jamiebramwell (approved):
Thanks @btalamini !

src/serac/numerics/functional/tensor.hpp
btalamini: This is not being used. I'll delete it, unless there any objections.

src/serac/numerics/functional/tensor.hpp
btalamini: Because of my poor understanding of metaprogramming, I'm unsure of all of the uses this function may have. Is this only for the case of a scalar product of two (arbitrary dimension) vectors? If so, is the motivation for removing it that it does the same the same thing as `dot(...)`? If the answer to this is also "yes", I'm with Jamie, I think we should keep it.

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
samuelpmishLLNL: > Is this useful enough to be it's own function?

I'd say not yet. Brandon and I noticed that these type traits like `is_dual` are almost never used in serac, so there's not a huge demand for them right now.

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
btalamini: How fine-grained are the logging tools? Output of this kind can be useful for debugging. You wouldn't want to turn it on in a finite element simulation (even in debug) where this is operative at quadrature points, because you'll be flooded with data. On the other hand, if you're debugging a material model, you might want this information.

I'll remove it for now, we can discuss this when we we need it.

PR #860, opened 2022-11-30
WIP: Reorganizing tests

### LINKED ISSUE #788: untangling target/test dependencies

chapman39:
Starting out reorganizing tests in a hierarchical way @samuelpmishLLNL described in issue https://github.com/LLNL/serac/issues/788 and PR https://github.com/LLNL/serac/pull/820. Pretty much tests will be moved to its minimum highest dependency to eliminate unnecessary dependencies.

Fixes #788

### DIFF
--- a/src/serac/infrastructure/tests/CMakeLists.txt
+++ b/src/serac/infrastructure/tests/CMakeLists.txt
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause) 
 
-set(test_dependencies gtest serac_physics)
+set(infrastructure_test_dependencies gtest serac_infrastructure)
 blt_list_append( TO test_dependencies ELEMENTS caliper adiak IF ${SERAC_ENABLE_PROFILING} )
 
 set(infrastructure_tests
@@ -13,4 +13,4 @@ set(infrastructure_tests
     profiling.cpp)
 
 serac_add_tests( SOURCES ${infrastructure_tests}
-                 DEPENDS_ON ${test_dependencies})
+                 DEPENDS_ON ${infrastructure_test_dependencies})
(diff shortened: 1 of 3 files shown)


src/serac/infrastructure/tests/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause) 
 
-set(test_dependencies gtest serac_physics)
+set(infrastructure_test_dependencies gtest serac_infrastructure)
 blt_list_append( TO test_dependencies ELEMENTS caliper adiak IF ${SERAC_ENABLE_PROFILING} )
```
white238, line 8:
```suggestion
blt_list_append( TO infrastructure_test_dependencies ELEMENTS caliper adiak IF ${SERAC_ENABLE_PROFILING} )
```

white238:
@chapman39 

Github marked this PR merged and closed due to the b647863af commit being in the history (then reverted) of the SERAC_USE_CAMP branch.

You will need to recreate this work in a new branch/PR.

PR #863, opened 2022-12-07
Feature/talamini/nonlinear hardening

btalamini:
* Adds a J2 plasticity model with nonlinear hardening models.
* Gives an example of a parameterized J2 model that efficiently computes sensitivities to design parameters.

Uses the differentiable Newton solver to make the return map operation (efficiently) differentiable, using an adjoint solve with the converged solution (and derivative). The kinematics are small strain. Two widely used hardening laws are implemented, and the material model is template-parameterized on the hardening model so that users can add new ones.

As a bonus, this PR revives the uniaxial stress material point tester. Sam and I had written this in a branch a while back, but I inadvertently failed to merge parts of it into the `develop` branch at the time.

Next steps:
* *Rate effects*: Add optional rate sensitivity so that model can be viscoplastic. Will require slight interface change to get time information into constitutive models so that ODEs can be integrated.
* *Finite deformation*: Using Multiplicative kinematics $(\mathbf{F}^e \mathbf{F}^p)$. Will need to finish the skeleton implementation of 3x3 spectral decomp and matrix log/exp operators.


src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -665,4 +676,39 @@ auto solve_scalar_equation(function&& f, double x0, double tolerance, double low
...
+ *
+ * @tparam function Type for the functor object
+ * @tparam n Vector dimension of the equation
+ * @param f A callable representing the function of which a root is sought. Must take an n-vector
+ * argument and return an n-vector
+ * @param x0 Initial guess for root. Must be an n-vector.
+ * @return A root of @p f.
+ */
+template <typename function, int n>
+auto find_root(function&& f, tensor<double, n> x0)
```
samuelpmishLLNL, line 692: we have these two methods `find_root` and `solve_scalar_equation` that are equivalent mathematically, I wonder if they might be combined in some way. 

One possible way is to overload on the type of the options:
```cpp
struct NewtonOptions {
  double tolerance;
  int max_iterations;
};

template <typename function, int n>
auto find_root(function&& f, tensor<double, n> x0, NewtonOptions opt) {
  ...
}

struct NewtSafeOptions {
  double tolerance;
  int max_iterations;
  double bounds[2];
};

template <typename function, int n>
auto find_root(function&& f, double x0, NewtSafeOptions opt) {
  ...
}
```
I don't think this needs to happen in this PR, but it might be worthwhile in the longer run

src/serac/physics/materials/parameterized_solid_material.hpp
```diff
@@ -105,4 +105,95 @@ struct ParameterizedNeoHookeanSolid {
...
+                  const SaturationStrength sigma_sat, const StrainConstant strain_constant) const
+  {
+    // The output stress tensor should use dual numbers if any of the parameters are dual.
+    // This slightly ugly trick to accomplishes that by picking up any dual number types
+    // from the parameters into the dummy variable "one".
+    // Another possiblity would be to cast the results to the correct type at the end, which
+    // would avoid doing any unneccessary arithmetic with dual numbers.
+    using T = typename underlying_scalar<YieldStrength, SaturationStrength, StrainConstant>::type;
+    T one;
+    one = 1.0;
```
samuelpmishLLNL, line 145: first, this ugly workaround was my suggestion, so I apologize.

>    // Another possiblity would be to cast the results to the correct type at the end, which
>    // would avoid doing any unneccessary arithmetic with dual numbers.

Two ways I can think of to do this are mocked up here: https://godbolt.org/z/v4hK9EoTP

-Option 0 (implicit conversions) make the written code a little bit cleaner to look at, but have the usual drawbacks of implicit conversions (potentially unintended type changes). That being said, this kind of implicit conversion is the least problematic-- analogous to how `float f = 3.0;` is technically an implicit conversion, but is harmless in practice. This approach also has the downside of requiring us to reimplement the compiler-generated constructors for the tensor class.

-Option 1 (explicit conversions) make the written code a little bit more verbose, but also makes it clearer about where potential type changes are occurring.

src/serac/physics/materials/tests/parameterized_nonlinear_J2_material.cpp
```diff
@@ -0,0 +1,87 @@
...
+  auto internal_state_new = internal_state;
+
+  auto stress = material(internal_state_new, H, sigma_y, make_dual(sigma_sat), strain_constant);
+
+  double h                   = 1e-4 * sigma_sat;
+  auto   internal_state_copy = internal_state;
+  auto   stress_p            = material(internal_state_copy, H, sigma_y, sigma_sat + h, strain_constant);
+  internal_state_copy        = internal_state;
+  auto stress_m              = material(internal_state_copy, H, sigma_y, sigma_sat - h, strain_constant);
+  auto finite_difference     = 1. / (2 * h) * (stress_p - stress_m);
```
samuelpmishLLNL: I trust that this line is correct as written, but I always worry about the order-dependence of these operators
```suggestion
  auto finite_difference     = (1.0 / (2 * h)) * (stress_p - stress_m);
```
or 
```suggestion
  auto finite_difference     = (stress_p - stress_m) / (2.0 * h);
```

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
btalamini: I thought about this, too, and agree we should spend some time thinking about how much should be unified. One could argue that the top-level solvers for the finite element systems are also equivalent. To me, the scalar function solver is the most different. The bounds are not really options, for example (and would be meaningless for n > 1). But your point is well taken.

src/serac/physics/materials/tests/parameterized_nonlinear_J2_material.cpp
btalamini: It does do what we want, since division and multiplication have the same precedence and have left-to-right associativity. I'm used to reading these things, but I have no objections to the extra parentheses from your first example to make the intent clearer. I'll push that change.
The second suggestion is different, in that it will cause 9 floating point divisions (one for each stress component) instead of 9 multiplications. I reflexively avoid FP divides as an old ingrained habit, but you would know better than me if that is still warranted on modern hardware. (This particular use is obviously not performance-critical, but I'd like to know if I  should change my patterns).

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
samuelpmishLLNL: just ran a couple tests on 2 machines:
- `matrix_mult_scalar` implements operator* by doing `B(i,j) = A(i,j) * scale);`
- `matrix_div1_scalar` implements operator/ by doing `B(i,j) = A(i,j) / scale);`
- `matrix_div2_scalar` implements operator/ by doing `inv_scale = 1.0 / scale;` first, then `B(i,j) = A(i,j) * inv_scale);`

Desktop x86 processor (i9-13900KF, -O3)
```
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
matrix_mult_scalar      0.527 ns        0.527 ns   1000000000
matrix_div1_scalar       3.53 ns         3.53 ns    198139978
matrix_div2_scalar      0.529 ns        0.529 ns   1000000000
```

Desktop x86 processor (i9-13900KF, -O3 -ffast-math -march=native)
```
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
matrix_mult_scalar      0.356 ns        0.356 ns   1000000000
matrix_div1_scalar      0.365 ns        0.365 ns   1000000000
matrix_div2_scalar      0.363 ns        0.363 ns   1000000000
```

Apple arm processor (M1, -O3)
```
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
matrix_mult_scalar      0.786 ns        0.786 ns    701262272
matrix_div1_scalar      0.788 ns        0.788 ns    890200168
matrix_div2_scalar      0.786 ns        0.786 ns    890381338
```

Apple arm processor (M1, -O3  -ffast-math)
```
-------------------------------------------------------------
Benchmark                   Time             CPU   Iterations
-------------------------------------------------------------
matrix_mult_scalar      0.320 ns        0.320 ns   1000000000
matrix_div1_scalar      0.313 ns        0.313 ns   1000000000
matrix_div2_scalar      0.314 ns        0.314 ns   1000000000
```

----------

so, empirically: 
- no difference in either case when `-ffast-math` enabled
- the x86 version without `-ffast-math` is considerably slower
- all implementations on the arm processor are identical 

Either way, we should probably just implement the `operator/` overloads to compute `1.0 / scale` first, since we can't assume `-ffast-math` is always enabled.

btalamini: Awesome! Thanks for this valuable data point. Agree with your conclusion.

In the LLVM [fast-math details](https://llvm.org/docs/LangRef.html#fast-math-flags), note the `arcp` sub-option, which explicitly allows converting division to multiplication by the reciprocal. I think this sub-option, as well as all the others, are activated when clang is executed with `-ffast-math`, which would explain the identical results on x86 in that mode.

Some of the sub-options tradeoffs are generally acceptable for us, others in restricted cases, but I think the `reassoc` one will ruin the numerical stability of many algorithms (in the floating-point arithmetic sense). If it's true that all these options are activated in one fell swoop, we would need to use `-ffast-math` *very* narrowly.

samuelpmishLLNL (approved):

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -665,4 +676,39 @@ auto solve_scalar_equation(function&& f, double x0, double tolerance, double low
...
+  static_assert(std::is_same_v<decltype(f(x0)), tensor<double, n>>,
+                "error: f(x) must have the same number of equations as unknowns");
+
+  double epsilon        = 1.0e-8;
+  int    max_iterations = 10;
+
+  auto x = x0;
+
+  int k = 0;
+  while (k++ < max_iterations) {
```
jamiebramwell: Is there any reason this can't be a for loop? I think it would be more readable.

src/serac/physics/materials/parameterized_solid_material.hpp
```diff
@@ -105,4 +105,95 @@ struct ParameterizedNeoHookeanSolid {
   double G0;       ///< base shear modulus
 };
 
+/**
+ * @brief Infers type resulting from algebraic expressions of a group of variables
+ *
+ * Useful if one needs to create a variable that is dual-valued if any operands are dual.
+ */
+template <typename... T>
+struct underlying_scalar {
```
jamiebramwell, line 114: `scalar_sum` maybe? This name does not state that it is the type of the sum of the inputs, so it could be misleading.

src/serac/physics/materials/parameterized_solid_material.hpp
jamiebramwell: I think I'm leaning toward option 0, but I'm fine either way. I think this should happen outside of this PR though.

jamiebramwell (approved):

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
btalamini: Just changed it. Thanks, that was a bit odd.

src/serac/physics/materials/parameterized_solid_material.hpp
btalamini: I agree this is not well named. `scalar_sum` more narrowly describes *how* it works,  but it doesn't express *what* you would use it for. We could have used multiplication, for instance, instead of addition operators. The point is that if any of them were dual number-valued, the resulting type would be dual number-valued. I'm definitely open to suggestions on this. Maybe it will go away with one of Sam's suggested refactoring options.

jamiebramwell: I'd say `scalar_type`, but then it would be `scalar_type::type`, so who knows. This is a temporary work around, so I'm ok with whatever as long as it's documented.

btalamini: Ok, I'll leave it for now.

PR #868, opened 2022-12-17
Make all `FiniteElementState`s use `mfem::Ordering::byNODES`

jamiebramwell:
For clarity, this ensures all finite element states and spaces use nodal instead of vdim ordering (e.g. xxxyyyzzz instead of xyzxyzxyz).

### DIFF
--- a/src/serac/physics/state/state_manager.cpp
+++ b/src/serac/physics/state/state_manager.cpp
@@ -52,7 +52,17 @@ double StateManager::newDataCollection(const std::string& name, const std::optio
     datacoll.UpdateMeshAndFieldsFromDS();
 
     // Functional needs the nodal grid function and neighbor data in the mesh
+
+    // This mfem call ensures the mesh contains an H1 grid function describing nodal
+    // cordinates. The parameters do the following:
+    // 1. Sets the order of the mesh to  p = 1
+    // 2. Uses a continuous (i.e. H1) finite element space
+    // 3. Uses the spatial dimension as the mesh dimension (i.e. it is not a lower dimension manifold)
+    // 4. Uses nodal instead of VDIM ordering (i.e. xxxyyyzzz instead of xyzxyzxyz)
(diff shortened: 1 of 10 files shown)


sethwatts:
> For clarity, this ensures all finite element states and spaces use nodal instead of vdim ordering (e.g. xxxyyyzzz instead of xyzxyzxyz).

Honestly, if MFEM had just used the naming "xxxyyyzzz" and "xyzxyzxyz," it would have reduced confusion so much...

I think always using a single Ordering is a good idea. At the risk of heated discussion, why xxxyyyzzz instead of xyzxyzxyz?

jamiebramwell:
> I think always using a single Ordering is a good idea. At the risk of heated discussion, why xxxyyyzzz instead of xyzxyzxyz?

This is consistent with all of the other finite element codes I've worked with in the past. It makes it much easier to grab the component vectors separately.

samuelpmishLLNL (approved):

src/serac/physics/state/state_manager.cpp
```diff
@@ -218,7 +218,7 @@ mfem::ParMesh* StateManager::setMesh(std::unique_ptr<mfem::ParMesh> pmesh, const
 
   // Functional needs the nodal grid function and neighbor data in the mesh
   auto& new_pmesh = mesh(mesh_tag);
-  new_pmesh.EnsureNodes();
+  new_pmesh.SetCurvature(1, false, -1, mfem::Ordering::byNODES);
```
samuelpmishLLNL: What is this `SetCurvature()` call doing, and what are the magic integers and boolean conveying?

jamiebramwell: I'll add a comment about this method. Essentially, it does the same thing as `EnsureNodes`, but it is not hardcoded to use `mfem::Ordering::byVDIM`.

white238 (approved):

PR #869, opened 2022-12-19
Force at least one linear solve per nonlinear solve

jamiebramwell:
This creates a modified version of the native `mfem::NewtonSolver` that optionally forces at least one linear solve per nonlinear solve. This is needed for downstream optimization customers to avoid the optimizer thinking it has hit a flat response of the parameter fields.

### DIFF
--- a/src/serac/numerics/CMakeLists.txt
+++ b/src/serac/numerics/CMakeLists.txt
@@ -14,6 +14,7 @@ set(numerics_headers
     equation_solver.hpp
     expr_template_impl.hpp
     expr_template_ops.hpp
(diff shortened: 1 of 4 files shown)


sethwatts:
LGTM

jamiebramwell:
Talking to @samuelpmishLLNL and @kswartz92 , a better option might be to either perturb or reset the displacement field before calling to the solver. As we move to more advanced nonlinear solvers, it is likely this issue won't go away.

jamiebramwell:
This PR is closed in favor of manually resetting the current displacement guess.

PR #873, opened 2023-01-11
support more element geometries and integration types

samuelpmishLLNL:
Serac currently uses `mfem::ElementRestriction` and `mfem::GeometricFactors` under the hood to enable users to specify different kinds of integral calculations, but those underlying mfem classes do not support many of the possible choices of element geometry and function space (:heavy_check_mark: : supported,   :x: : unsupported, :question: : I'm unsure):

| Domain Integral Support | H1                 | Hcurl              | L2                 |
| ----------------------- | ------------------ | ------------------ | ------------------ |
| Triangle                | :question:                | :x:                | :question:                |
| Quadrilateral           | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
| Tetrahedron             | :question:               | :x:                | :question:                |
| Hexahedron              | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |

| Boundary Integral Support | H1                 | Hcurl | L2   |
| ------------------------- | ------------------ | ----- | ---- |
| Triangle                  | :question:                 | :x:   | :x:  |
| Quadrilateral             | :heavy_check_mark: | :x:   | :x:  |
| Tetrahedron               | :question:               | :x:   | :x:  |
| Hexahedron                | :heavy_check_mark: | :x:   | :x:  |

The point of this PR is to implement the functionality that is missing in mfem to cover all 24 of the combinations above.

**(May 31) Update:**  most of the work has been done for Hcurl support, but it's not quite finished and will be added in a future PR.

-------------

Much of the complexity in this PR is related to deciphering the largely undocumented practice of how mfem enumerates the dofs on the different kinds of elements. For clarity, here are some images showing the node numbering convention that serac will follow:

H1 / L2 (linear, quadratic, cubic): 
----

H1_tri
H1_quad
H1_tet
H1_hex

Note: these images assume a right-handed coordinate system, where "x" is always the direction from node 0 (the origin) to node 1


samuelpmishLLNL:
for future reference:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108975

src/serac/numerics/functional/tests/simplex_basis_function_unit_tests.cpp
```diff
@@ -155,11 +148,111 @@ TEST(CubicTetrahedron, Gradients) {
...
+
+TEST(CubicTetrahedron, mass_integration_inexact) {
+  using element_type = finite_element< mfem::Geometry::TETRAHEDRON, H1<3> >;
+  verify_mass_matrix_integration< element_type, 4 >(0.02);
+}
+
+TEST(CubicTetrahedron, mass_integration_exact) {
+  using element_type = finite_element< mfem::Geometry::TETRAHEDRON, H1<3> >;
+  verify_mass_matrix_integration< element_type, 5 >(2.0e-15);
+}
```
samuelpmishLLNL: @jamiebramwell @btalamini
These tests show the difference in exactness of integration for our current default and the "next-higher" quadrature rules

src/serac/numerics/functional/debug_print.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include "axom/core.hpp"
+
+#include <string>
+#include <fstream>
+#include <iomanip>
+#include <vector>
+
+#include "serac/numerics/functional/dof_numbering.hpp"
+
+template <typename T>
```
jamiebramwell: This file needs doxygen.

src/serac/numerics/functional/debug_print.hpp
```diff
@@ -0,0 +1,90 @@
...
+  for (int i = 0; i < v.size(); i++) {
+    outfile << v[i] << std::endl;
+  }
+  outfile.close();
+}
+
+void write_to_file(mfem::Vector v, std::string filename)
+{
+  std::ofstream outfile(filename);
+  for (int i = 0; i < v.Size(); i++) {
```
jamiebramwell: Why not `mfem::Vector::Print(...)`?

src/serac/numerics/functional/debug_print.hpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+std::ostream& operator<<(std::ostream& out, serac::SignedIndex i)
+{
+  out << "{" << i.index_ << ", " << i.sign_ << "}";
+  return out;
+}
+
+template <typename T>
+void write_to_file(axom::Array<T, 2, axom::MemorySpace::Host> arr, std::string filename)
```
jamiebramwell: Does axom have native print functions to call?

src/serac/numerics/functional/tests/capacitor.cpp
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2019-2023, Lawrence Livermore National Security, LLC and
```
jamiebramwell: While this is a nice example, is it testing anything that we don't test elsewhere?

src/serac/numerics/functional/tests/check_gradient.hpp
```diff
@@ -13,100 +13,172 @@
...
-  U_minus.Add(-epsilon, dU);
-
-  mfem::Vector df1 = f(U_plus);
-  df1 -= f(U_minus);
-  df1 /= (2 * epsilon);
-
-  auto [value, dfdU] = f(serac::differentiate_wrt(U));
-  mfem::Vector df2   = dfdU(dU);
+  auto [value, dfdU]   = f(serac::differentiate_wrt(U));
+  mfem::Vector df_jvp1 = dfdU(dU);
```
jamiebramwell: I assume this stands for `jacobian_vector_product`. A more descriptive name would be nice.

jamiebramwell: I would also note that this is the matrix-free product in the variable name (or a comment).

src/serac/numerics/functional/tests/check_gradient.hpp
```diff
@@ -13,100 +13,172 @@
...
 
   std::unique_ptr<mfem::HypreParMatrix> dfdU_matrix = assemble(dfdU);
+  mfem::Vector                          df_jvp2     = (*dfdU_matrix) * dU;
 
-  mfem::Vector df3 = (*dfdU_matrix) * dU;
-
-  double relative_error1 = df1.DistanceTo(df2.GetData()) / df1.Norml2();
-  double relative_error2 = df1.DistanceTo(df3.GetData()) / df1.Norml2();
+  if (df_jvp1.Norml2() != 0) {
```
jamiebramwell, line 32: When would this equal zero?

src/serac/numerics/functional/tests/check_gradient.hpp
```diff
@@ -13,100 +13,172 @@
...
+    double relative_error = df_jvp1.DistanceTo(df_jvp2.GetData()) / df_jvp1.Norml2();
+    EXPECT_NEAR(0., relative_error, 5.e-6);
+  }
 
-  EXPECT_NEAR(0., relative_error1, 5.e-6);
-  EXPECT_NEAR(0., relative_error2, 5.e-6);
+  mfem::Vector f_values[5];
```
jamiebramwell, line 38: I would add a comment about what the entries of `f_values` are. E.g. small perturbations of output from F, like f(U - 2 * eps * dU) ... f(U + 2 * eps * dU).

src/serac/numerics/functional/tests/check_gradient.hpp
```diff
@@ -13,100 +13,172 @@
...
     std::unique_ptr<mfem::HypreParMatrix> dfdU_matrix = assemble(dfdU);
 
-    mfem::Vector df3 = (*dfdU_matrix) * dU;
-
-    double relative_error1 = df1.DistanceTo(df2.GetData()) / df1.Norml2();
-    double relative_error2 = df1.DistanceTo(df3.GetData()) / df1.Norml2();
-
-    EXPECT_NEAR(0., relative_error1, 5.e-6);
-    EXPECT_NEAR(0., relative_error2, 5.e-6);
+    mfem::Vector df_jvp2        = (*dfdU_matrix) * dU;
```
jamiebramwell: Same documentation comments as the other `check_gradient` function.

src/serac/numerics/functional/tests/functional_basic_h1_scalar.cpp
```diff
@@ -25,32 +25,32 @@
...
 
-  auto mesh = mesh::refineAndDistribute(buildMeshFromFile(meshfile), 1);
+  auto                        trial_fec = mfem::H1_FECollection(ptrial, dim);
+  mfem::ParFiniteElementSpace trial_fespace(mesh.get(), &trial_fec);
 
-  // Create standard MFEM bilinear and linear forms on H1
-  auto                        fec = mfem::H1_FECollection(p, dim);
-  mfem::ParFiniteElementSpace fespace(mesh.get(), &fec);
+  mfem::Vector U(trial_fespace.TrueVSize());
+  // U.Randomize();
```
jamiebramwell: dead code

src/serac/numerics/functional/tests/functional_basic_h1_scalar.cpp
```diff
@@ -25,32 +25,32 @@
...
 
-  mfem::Vector U(fespace.TrueVSize());
-  U.Randomize();
+  mfem::ParGridFunction     U_gf(&trial_fespace);
+  mfem::FunctionCoefficient x_squared([](mfem::Vector x) { return x[0] * x[0]; });
+  U_gf.ProjectCoefficient(x_squared);
+  // mfem::FunctionCoefficient x_coord([](mfem::Vector x) { return x[0]; });
```
jamiebramwell: dead code

src/serac/numerics/functional/tests/functional_basic_h1_scalar.cpp
```diff
@@ -78,15 +78,41 @@ void thermal_test()
...
+
+TEST(basic, thermal_tris) { thermal_test<1, 1>("/data/meshes/patch2D_tris.mesh"); }
+TEST(basic, thermal_quads) { thermal_test<1, 1>("/data/meshes/patch2D_quads.mesh"); }
+TEST(basic, thermal_tris_and_quads) { thermal_test<1, 1>("/data/meshes/patch2D_tris_and_quads.mesh"); }
+
+TEST(basic, thermal_tets) { thermal_test<1, 1>("/data/meshes/patch3D_tets.mesh"); }
+TEST(basic, thermal_hexes) { thermal_test<1, 1>("/data/meshes/patch3D_hexes.mesh"); }
+TEST(basic, thermal_tets_and_hexes) { thermal_test<1, 1>("/data/meshes/patch3D_tets_and_hexes.mesh"); }
+
+// TEST(basic, thermal_test_3D) { thermal_test<1, 1, 3>(); }
```
jamiebramwell: dead code.

src/serac/numerics/functional/tests/functional_boundary_test.cpp
```diff
@@ -148,27 +148,37 @@ void boundary_test(mfem::ParMesh& mesh, L2<p> test, L2<p> trial, Dimension<dim>)
...
         // tensor<double,dim> b{sin(x[0]), x[0] * x[1]};
-        return x[0] * x[1] + /*dot(b, n) +*/ rho * u;
+        return x[0] * x[1] + /* dot(b, n) +*/ rho * u;
       },
       mesh);
 
   mfem::Vector r1 = (*J) * U + (*F);
   mfem::Vector r2 = residual(U);
 
+  // std::cout << std::setprecision(16);
```
jamiebramwell: dead code

src/serac/numerics/functional/tests/functional_boundary_test.cpp
```diff
@@ -186,13 +196,12 @@ TEST(FunctionalBoundary, 2DQuadratic) { boundary_test(*mesh2D, H1<2>{}, H1<2>{},
...
 TEST(FunctionalBoundary, 3DQuadratic) { boundary_test(*mesh3D, H1<2>{}, H1<2>{}, Dimension<3>{}); }
 
-// TODO: mfem treats L2 differently w.r.t. boundary elements, need to figure out how to get
-// the appropriate information (dofs, dof_ids for each boundary element) before these can be reenabled
-// TEST(boundaryL2, 2DLinear) { boundary_test(*mesh2D, L2<1>{}, L2<1>{}, Dimension<2>{}); }
-// TEST(boundaryL2, 2DQuadratic) { boundary_test(*mesh2D, L2<2>{}, L2<2>{}, Dimension<2>{}); }
-//
-// TEST(boundaryL2, 3DLinear) { boundary_test(*mesh3D, L2<1>{}, L2<1>{}, Dimension<3>{}); }
-// TEST(boundaryL2, 3DQuadratic) { boundary_test(*mesh3D, L2<2>{}, L2<2>{}, Dimension<3>{}); }
+TEST(boundaryL2, 2DLinear) { boundary_test(*mesh2D, L2<1>{}, L2<1>{}, Dimension<2>{}); }
```
jamiebramwell, line 203: 🎉

src/serac/numerics/functional/tests/functional_boundary_test.cpp
```diff
@@ -203,14 +212,20 @@ int main(int argc, char* argv[])
...
-  std::string meshfile3D = SERAC_REPO_DIR "/data/meshes/beam-hex.mesh";
+  std::string meshfile2D = SERAC_REPO_DIR "/data/meshes/patch2D_tris_and_quads.mesh";
+  std::string meshfile3D = SERAC_REPO_DIR "/data/meshes/patch3D_tets_and_hexes.mesh";
+  // std::string meshfile3D = SERAC_REPO_DIR "/data/meshes/onetet.mesh";
   mesh2D = mesh::refineAndDistribute(buildMeshFromFile(meshfile2D), serial_refinement, parallel_refinement);
   mesh3D = mesh::refineAndDistribute(buildMeshFromFile(meshfile3D), serial_refinement, parallel_refinement);
 
+  // by default, mfem::Meshes aren't completely initialized on construction,
+  // so we have to manually initialize some of it
+  mesh2D->EnsureNodes();
```
jamiebramwell, line 229: Should we just put this in the `buildMeshFromFile` methods?

src/serac/numerics/functional/tests/functional_boundary_test.cpp
```diff
@@ -203,14 +212,20 @@ int main(int argc, char* argv[])
...
 
-  int serial_refinement   = 0;
+  int serial_refinement   = 1;
   int parallel_refinement = 0;
 
-  std::string meshfile2D = SERAC_REPO_DIR "/data/meshes/star.mesh";
-  std::string meshfile3D = SERAC_REPO_DIR "/data/meshes/beam-hex.mesh";
+  std::string meshfile2D = SERAC_REPO_DIR "/data/meshes/patch2D_tris_and_quads.mesh";
+  std::string meshfile3D = SERAC_REPO_DIR "/data/meshes/patch3D_tets_and_hexes.mesh";
+  // std::string meshfile3D = SERAC_REPO_DIR "/data/meshes/onetet.mesh";
```
jamiebramwell: dead code

src/serac/numerics/functional/tests/functional_comparisons.cpp
```diff
@@ -357,13 +353,13 @@ TEST(Thermal, 3DLinear) { functional_test(*mesh3D, H1<1>{}, H1<1>{}, Dimension<3
...
 TEST(Thermal, 3DCubic) { functional_test(*mesh3D, H1<3>{}, H1<3>{}, Dimension<3>{}); }
 
-TEST(Hcurl, 2DLinear) { functional_test(*mesh2D, Hcurl<1>{}, Hcurl<1>{}, Dimension<2>{}); }
-TEST(Hcurl, 2DQuadratic) { functional_test(*mesh2D, Hcurl<2>{}, Hcurl<2>{}, Dimension<2>{}); }
-TEST(Hcurl, 2DCubic) { functional_test(*mesh2D, Hcurl<3>{}, Hcurl<3>{}, Dimension<2>{}); }
-
-TEST(Hcurl, 3DLinear) { functional_test(*mesh3D, Hcurl<1>{}, Hcurl<1>{}, Dimension<3>{}); }
-TEST(Hcurl, 3DQuadratic) { functional_test(*mesh3D, Hcurl<2>{}, Hcurl<2>{}, Dimension<3>{}); }
-TEST(Hcurl, 3DCubic) { functional_test(*mesh3D, Hcurl<3>{}, Hcurl<3>{}, Dimension<3>{}); }
+// TEST(Hcurl, 2DLinear) { functional_test(*mesh2D, Hcurl<1>{}, Hcurl<1>{}, Dimension<2>{}); }
```
jamiebramwell, line 358: I can help with the Hcurl basis functions after this goes in.

src/serac/physics/tests/solid_dynamics_patch.cpp
```diff
@@ -401,21 +243,343 @@ class ConstantAccelerationSolution {
   mfem::Vector acceleration;  /// Constant acceleration vector of solution
 };
 
-TEST(SolidMechanicsDynamic, ConstantAcceleration2dQ1TractionBcs)
+/**
```
jamiebramwell, line 246: Doxygen needs updating.

src/serac/physics/tests/solid_statics_patch.cpp
```diff
@@ -187,7 +328,8 @@ double solution_error(const ExactSolution& exact_displacement, PatchBoundaryCond
 
   SolidMechanics<p, dim> solid(std::move(equation_solver), solid_mechanics::default_quasistatic_options, GeometricNonlinearities::On, "solid");
 
-  solid_mechanics::NeoHookean mat{.density=1.0, .K=1.0, .G=1.0};
+  //solid_mechanics::NeoHookean mat{.density=1.0, .K=1.0, .G=1.0};
```
jamiebramwell: Why is this using a linear material instead of a neo-Hookean one?

src/serac/physics/tests/thermal_shape.cpp
```diff
@@ -45,11 +45,18 @@ TEST(HeatTransfer, MoveShape)
...
-  auto options = heat_transfer::default_nonlinear_options;
+  auto nonlinear_options         = heat_transfer::default_nonlinear_options;
+  nonlinear_options.absolute_tol = 1.0e-14;
+  nonlinear_options.relative_tol = 1.0e-14;
 
   // Use tight tolerances as this is a machine precision test
-  options.absolute_tol = 1.0e-14;
-  options.relative_tol = 1.0e-14;
+  //
+  // Sam: we're setting a really small abs tolerance here to
```
jamiebramwell, line 54: I've also used direct solvers to get around this in the past. We could also move to using the Hypre-native CG and GMRES.

src/serac/numerics/functional/integral.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include "serac/infrastructure/accelerator.hpp"
+#include "serac/numerics/functional/geometric_factors.hpp"
+#include "serac/numerics/functional/function_signature.hpp"
+#include "serac/numerics/functional/domain_integral_kernels.hpp"
+#include "serac/numerics/functional/boundary_integral_kernels.hpp"
+
+namespace serac {
+
+struct Integral {
```
jamiebramwell, line 24: This file needs doxygen.

src/serac/numerics/functional/integral.hpp
```diff
@@ -0,0 +1,217 @@
...
+    }
+  }
+
+  Type type;
+
+  using eval_func = std::function<void(const std::vector<const double*>&, double*, bool)>;
+  std::map<mfem::Geometry::Type, eval_func>               evaluation_;
+  std::vector<std::map<mfem::Geometry::Type, eval_func> > evaluation_with_AD_;
+
+  using jvp_func = std::function<void(const double*, double*)>;
```
jamiebramwell: I would add a more descriptive name (e.g. `jacobian_vector_product`)

src/serac/numerics/functional/integral.hpp
```diff
@@ -0,0 +1,217 @@
...
+  std::vector<std::map<mfem::Geometry::Type, jvp_func> > jvp_;
+
+  using grad_func = std::function<void(ExecArrayView<double, 3, ExecutionSpace::CPU>)>;
+  std::vector<std::map<mfem::Geometry::Type, grad_func> > element_gradient_;
+
+  std::vector<uint32_t> active_trial_spaces;
```
jamiebramwell: The member variable needs a trailing underscore.

src/serac/numerics/functional/integral.hpp
```diff
@@ -0,0 +1,217 @@
...
+  static constexpr std::size_t num_types = Type::_size;
+
+  Integral(Type t, std::vector<uint32_t> trial_space_indices) : type(t), active_trial_spaces(trial_space_indices)
+  {
+    std::size_t num_trial_spaces = trial_space_indices.size();
+    evaluation_with_AD_.resize(num_trial_spaces);
+    jvp_.resize(num_trial_spaces);
+    element_gradient_.resize(num_trial_spaces);
+
+    for (size_t i = 0; i < num_trial_spaces; i++) {
```
jamiebramwell: Some documentation on the `functional_to_integral_` container would greatly improve readability.

jamiebramwell: maybe `functional_to_indegral_index_`?

src/serac/numerics/functional/integral.hpp
```diff
@@ -0,0 +1,217 @@
...
+      functional_to_integral_[active_trial_spaces[i]] = static_cast<int>(i);
+    }
+  }
+
+  void Mult(const std::vector<mfem::BlockVector>& input_E, mfem::BlockVector& output_E, int differentiation_index,
+            bool update_state) const
+  {
+    output_E = 0.0;
+
+    bool  with_AD = (functional_to_integral_.count(differentiation_index) > 0 && differentiation_index != -1);
```
jamiebramwell: When would `count` be greater than 1?

src/serac/numerics/functional/geometry.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+  if (g == mfem::Geometry::TRIANGLE || g == mfem::Geometry::SQUARE) {
+    return 2;
+  }
+
+  if (g == mfem::Geometry::TETRAHEDRON || g == mfem::Geometry::CUBE) {
+    return 3;
+  }
+
+  return -1;
```
jamiebramwell, line 63: Should this be an error?

src/serac/numerics/functional/geometry.hpp
```diff
@@ -0,0 +1,87 @@
...
+  if (g == mfem::Geometry::SQUARE) {
+    return q * q;
+  }
+  if (g == mfem::Geometry::TETRAHEDRON) {
+    return (q * (q + 1) * (q + 2)) / 6;
+  }
+  if (g == mfem::Geometry::CUBE) {
+    return q * q * q;
+  }
+  return -1;
```
jamiebramwell, line 42: Error?

src/serac/numerics/functional/debug_print.hpp
samuelpmishLLNL: All of the mfem containers have different methods and interfaces for writing things out to file. I don't know if axom has a preferred method, and stl containers don't have any way of doing it.

So, the point of this was just to have a common way to dump any of our supported containers to file for debugging.

src/serac/numerics/functional/debug_print.hpp
samuelpmishLLNL: I don't know, but I'll probably still end up use a wrapper like this, to avoid having to figure out each container's preferred output function.

src/serac/numerics/functional/tests/capacitor.cpp
samuelpmishLLNL: Oops, I never meant to add this file to the PR.

src/serac/numerics/functional/tests/check_gradient.hpp
samuelpmishLLNL: `jvp` and `vjp` are how `jax` describes these kind of operations. I was considering changing our terminology to be more consistent with `jax`, to avoid having a bunch of incompatible terminology for the same ideas.

I'd prefer to not have to annotate `Functional::Gradient::operator()` output at its call site, but I can document that method more thoroughly to clarify that it isn't forming a sparse matrix.

src/serac/numerics/functional/tests/check_gradient.hpp
samuelpmishLLNL: we have a few unit tests that have constant qfunctions (e.g. surface_area = $\iint 1 dA$, volume = $\iiint 1 dV$)

src/serac/numerics/functional/tests/functional_boundary_test.cpp
samuelpmishLLNL: I would prefer that. 

I don't know the rules of when this should be called (i.e. should it be in `buildMeshFromFile`, or `refineAndDistribute`), so if you can help me understand that part, then I'll make the appropriate changes. Also, if we put it in `buildMeshFromFile`, then the other mesh builders (Rectangle, Cuboid, Disk, etc) should get it as well.

src/serac/physics/tests/solid_statics_patch.cpp
btalamini: Is this where we were trying to get higher order patch tests working? If so, then this is an artifact of our experiments and should be reverted to the nonlinear model. Sam and I decided to postpone those tests for now, because we'll need to select higher precision quadrature rules to make it work.

src/serac/physics/tests/thermal_shape.cpp
samuelpmishLLNL: it sounds like mfem is open to the idea of using consistent norms too. I'll leave it as-is currently, with a TODO.

src/serac/numerics/functional/integral.hpp
samuelpmishLLNL: `count(...)` will never return something greater than 1, making `> 0` and `== 1` are practically equivalent here (so I went with the shorter one).

src/serac/numerics/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ set(numerics_sources
     odes.cpp
     )
 
-set(numerics_depends serac_infrastructure)
+set(numerics_depends serac_infrastructure serac_functional)
```
white238, line 28: This is creating a circular dependency between numerics and functional

white238: Looking below it seems like that was already there implicitly through headers. That is unfortunate.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -109,237 +62,86 @@ template <int i>
...
-
-      // write out the q-function derivatives after multiplying by J_e so that
-      // won't need to be applied in the action_of_gradient and element_gradient kernels
+    // write out the q-function derivatives after applying the
+    // physical_to_parent transformation, so that those transformations
+    // won't need to be applied in the action_of_gradient and element_gradient kernels
+    if constexpr (differentiation_index != serac::NO_DIFFERENTIATION) {
       for (int q = 0; q < leading_dimension(qf_outputs); q++) {
-        qf_derivatives_(e, q) = get_gradient(qf_outputs[q]);
+        qf_derivatives[e * qpts_per_elem + uint32_t(q)] = get_gradient(qf_outputs[q]);
```
jamiebramwell, line 129: Any ideas how to handle the casting back and forth @white238 ? The issue is that MFEM containers use `int` and std containers use `uint32_t`. The conversions clutter the code, but I'm not sure if anything can be done about it.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -109,237 +62,86 @@ template <int i>
 struct DerivativeWRT {
 };
 
-template <int Q, Geometry g, typename test, typename... trials>
+template <int Q, mfem::Geometry::Type g, typename test, typename... trials>
```
jamiebramwell: Is `Q` documented anywhere? This is a spot where I can imagine a lot of confusion between 1D quadrature points vs total quadrature points.

src/serac/numerics/CMakeLists.txt
samuelpmishLLNL: Where is the circular dependency? 

I thought our directory structure was using the pattern where parent directories depended on the targets defined in their subdirectories (e.g. `serac` depends on `serac/numerics`, which depends on `serac/numerics/functional`)

white238: https://github.com/LLNL/serac/blob/9df86949f20f34fad863d7f1df054c5551c65d06/src/serac/numerics/functional/CMakeLists.txt#L7

white238: I agree based on directory structure what you added was "correct" but the existing one I just linked is the incorrect one.

samuelpmishLLNL: it looks like the quadrature data is the only piece being used from `numerics`, so it just should be moved

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -386,17 +188,17 @@ auto batch_apply_chain_rule(derivative_type* qf_derivatives, const tensor<T, n>&
  * @see mfem::GeometricFactors
  * @param[in] num_elements The number of elements in the mesh
  */
-template <Geometry g, typename test, typename trial, int Q, typename derivatives_type>
-void action_of_gradient_kernel(const mfem::Vector& dU, mfem::Vector& dR,
-                               CPUArrayView<derivatives_type, 2> qf_derivatives, std::size_t num_elements)
+template <int Q, mfem::Geometry::Type geom, typename test, typename trial, typename derivatives_type>
```
jamiebramwell, line 183: How is `Q` interpreted for non-tensor-product elements? Does the comment need to be updated?

src/serac/numerics/CMakeLists.txt
white238: Thank you for taking care of this. It is something that is easy to sneak in.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -460,6 +262,40 @@ void element_gradient_kernel(ExecArrayView<double, 3, ExecutionSpace::CPU> dK,
...
+    std::shared_ptr<derivative_type> qf_derivatives, uint32_t num_elements)
+{
+  return [=](const std::vector<const double*>& inputs, double* outputs, bool /* update state */) {
+    evaluation_kernel_impl<wrt, Q, geom>(s, inputs, outputs, positions, jacobians, qf, qf_derivatives.get(),
+                                         num_elements, s.index_seq);
+  };
+}
+
+template <int wrt, int Q, mfem::Geometry::Type geom, typename signature, typename derivative_type>
+std::function<void(const double*, double*)> jvp_kernel(signature, std::shared_ptr<derivative_type> qf_derivatives,
```
jamiebramwell: Maybe `jacobian_vec_product`?

src/serac/numerics/functional/debug_print.hpp
jamiebramwell: Any thoughts @white238 ?

src/serac/numerics/functional/detail/segment_L2.inl
```diff
@@ -10,22 +10,30 @@
...
  */
 
-// specialization of finite_element for L2 on segment geometry
-//
 // this specialization defines shape functions (and their gradients) that
 // interpolate at Gauss-Lobatto nodes for the appropriate polynomial order
 //
 // note: mfem assumes the parent element domain is [0,1]
 // for additional information on the finite_element concept requirements, see finite_element.hpp
 /// @cond
```
white238, line 18:
```suggestion
```

white238: I think .. this could be wrong.

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -96,8 +96,8 @@ struct finite_element<Geometry::Hexahedron, H1<p, c> > {
   template <bool apply_weights, int q>
   static constexpr auto calculate_B()
   {
-    constexpr auto                  points1D  = GaussLegendreNodes<q>();
-    [[maybe_unused]] constexpr auto weights1D = GaussLegendreWeights<q>();
+    constexpr auto                  points1D  = GaussLegendreNodes<q, mfem::Geometry::SEGMENT>();
```
jamiebramwell, line 99: Is it possible to rename `q` to `q1D`? As you can tell by my comments, this has bitten me in the past.

src/serac/numerics/functional/detail/triangle_H1.inl
```diff
@@ -0,0 +1,332 @@
...
+ *
+ * @brief Specialization of finite_element for H1 on triangle geometry
+ */
+
+// this specialization defines shape functions (and their gradients) that
+// interpolate at Gauss-Lobatto nodes for the appropriate polynomial order
+//
+// note: mfem assumes the parent element domain is the convex hull of {{0,0}, {1,0}, {0,1}}
+// for additional information on the finite_element concept requirements, see finite_element.hpp
+/// @cond
```
white238, line 21: Are these large exclusions necessary/wanted?

src/serac/numerics/functional/detail/tetrahedron_H1.inl
```diff
@@ -0,0 +1,425 @@
...
+      typename std::conditional<components == 1, tensor<double, ndof>, tensor<double, ndof, components> >::type;
+
+  using dof_type = tensor<double, c, ndof>;
+
+  using value_type = typename std::conditional<components == 1, double, tensor<double, components> >::type;
+  using derivative_type =
+      typename std::conditional<components == 1, tensor<double, dim>, tensor<double, components, dim> >::type;
+  using qf_input_type = tuple<value_type, derivative_type>;
+
+  SERAC_HOST_DEVICE static constexpr double shape_function([[maybe_unused]] tensor<double, dim> xi, int i)
```
jamiebramwell, line 46: Maybe add a reference for how these are calculated?

src/serac/numerics/functional/geometry.hpp
samuelpmishLLNL: these functions are used in `constexpr` and (eventually) `__device__` contexts, so our error handling options are limited

src/serac/physics/tests/solid_statics_patch.cpp
```diff
@@ -162,21 +283,41 @@ std::set<int> essentialBoundaryAttributes(PatchBoundaryCondition bc)
  * solid functional that should lead to the exact solution
  * See AffineSolution for an example
  */
-template <int p, int dim, typename ExactSolution>
-double solution_error(const ExactSolution& exact_displacement, PatchBoundaryCondition bc)
+template < typename element_type>
```
white238, line 277: The tparams above are now incorrect.

src/serac/numerics/functional/dof_numbering.hpp
```diff
@@ -215,8 +225,11 @@ struct DofNumbering {
           uint32_t dof_id         = static_cast<uint32_t>(fabs(dof_ids_h[index]));  // note: 1-based index
           int      dof_sign       = dof_ids[index] > 0 ? +1 : -1;
           bdr_element_dofs_(e, i) = {dof_id - 1, dof_sign};  // subtract 1 to get back to 0-based index
+
+          // std::cout << dof_id - 1 << " " << boundary_dofs.dof_info(e, i).index() << std::endl;
```
jamiebramwell: dead code

src/serac/numerics/functional/tests/simplex_basis_function_unit_tests.cpp
```diff
@@ -0,0 +1,709 @@
+#include <gtest/gtest.h>
```
white238:
```suggestion
// Copyright (c) 2019-2023, Lawrence Livermore National Security, LLC and
// other Serac Project Developers. See the top-level LICENSE file for
// details.
//
// SPDX-License-Identifier: (BSD-3-Clause)

#include <gtest/gtest.h>
```

src/serac/numerics/functional/tests/functional_comparisons.cpp
```diff
@@ -357,13 +354,13 @@ TEST(Thermal, 3DLinear) { functional_test(*mesh3D, H1<1>{}, H1<1>{}, Dimension<3
...
```
white238, line 358: Can we add a TODO/comment why this is disabled?

src/serac/numerics/functional/detail/triangle_H1.inl
samuelpmishLLNL: the `detail/*.inl` files all implement different template specializations of the same class, so `@cond` is used here to avoid writing the same doxygen descriptions on each specialization.

There is a more detailed docstring on the "base class":
https://github.com/LLNL/serac/blob/71e4dadb5365ccfef7a30f5f342098948c0a30da/src/serac/numerics/functional/finite_element.hpp#L324-L348

src/serac/numerics/functional/element_restriction.hpp
```diff
@@ -0,0 +1,255 @@
...
+};
+
+/// a struct of metadata (index, sign, orientation) associated with a degree of freedom
+struct DoF {
+  // sam: I wanted to use a bitfield for this type, but a 10+ year-old GCC bug
+  // makes it practically impossible to assign to bitfields without warnings/errors
+  // with -Wconversion enabled, see: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39170
+  // So, we resort to masks and bitshifting instead.
+
+  static constexpr uint64_t sign_mask        = 0x8000'0000'0000'0000;       ///< bits for sign field
```
jamiebramwell, line 37: Can we use `std::vector<bool>` or `std::bitset`?  Is this required because it is used on the device?

src/serac/numerics/functional/functional.hpp
```diff
@@ -629,6 +598,8 @@ class Functional<test(trials...), exec> {
                              form_.input_L_[which_argument].Size(), sparse_matrix_frees_graph_ptrs,
                              sparse_matrix_frees_values_ptr, col_ind_is_sorted);
 
+      // write_to_file(J_local, "J.mtx");
```
jamiebramwell: dead code

src/serac/numerics/functional/debug_print.hpp
white238: There is but it doesn't pretty print it out based on dimension.

https://github.com/LLNL/axom/blob/8eca639582081a5e5177fc13ab4a44fb2378473d/src/axom/core/ArrayBase.hpp#L611-L644

src/serac/numerics/functional/element_restriction.hpp
samuelpmishLLNL: `std::vector<bool>` is notoriously weird and I don't think there's much of an advantage to `std::bitset` (in addition to the `__device__` compatibility issues). Were there specific concerns about this implementation?

jamiebramwell: not really - just bitset syntax always takes me a bit longer to parse.

src/serac/numerics/functional/boundary_integral_kernels.hpp
samuelpmishLLNL: If you look at the positions of the quadrature points in an element, "Q" is the number of points along one edge, e.g.

```
6 quadrature points for a triangle with Q = 3
.
.  .
.  .  .

16 quadrature points for a quadrilateral with Q = 4
.  .  .  .
.  .  .  .
.  .  .  .
.  .  .  .
```

I don't know a good term for this pattern (the Qth triangle number for tris, the Qth tetrahedron number for tets, Qth perfect square for quads, Qth perfect cube for hexes, etc).

I'm not sure where to put this description, I don't want to duplicate this explanation in too many places.

src/serac/numerics/functional/detail/hexahedron_H1.inl
samuelpmishLLNL: Q1D makes sense for tensor product elements, but there is no underlying 1D quadrature rule for simplices

src/serac/numerics/functional/debug_print.hpp
white238: My only thoughts are it is weird to have these sitting inside of functional. Can we move them into a utilities directory? I assume they are aren't in infrastructure due to needing types and dependencies?

src/serac/numerics/functional/boundary_integral_kernels.hpp
white238: I can't think of a way of handling this other than implicit conversions (obvious no) or wrapping the types to convert internally.  Does this bleed into the user code or is it just us?

src/serac/numerics/functional/boundary_integral_kernels.hpp
samuelpmishLLNL: Oops, I forgot we had this function: 
```cpp
constexpr int num_quadrature_points(mfem::Geometry::Type g, int Q)
{
  if (g == mfem::Geometry::SEGMENT) {
    return Q;
  }
  if (g == mfem::Geometry::TRIANGLE) {
    return (Q * (Q + 1)) / 2;
  }
  if (g == mfem::Geometry::SQUARE) {
    return Q * Q;
  }
  if (g == mfem::Geometry::TETRAHEDRON) {
    return (Q * (Q + 1) * (Q + 2)) / 6;
  }
  if (g == mfem::Geometry::CUBE) {
    return Q * Q * Q;
  }
  return -1;
}
```
I'll try to refer to this implementation to clarify what "Q" means instead (although it's used in quite a few places).

samuelpmishLLNL: Also, I don't know why Doxygen isn't warning about a missing entry for that kernel, the functions immediately below it created warnings when they didn't have comments.

src/serac/numerics/functional/boundary_integral_kernels.hpp
samuelpmishLLNL: This affects any code that mixes some combination of containers from mfem/axom/stl. 

I don't believe there's any way to prevent implicit conversion between integer types, because that's part of the C++ language specification. Serac uses `-Wsign-conversion` which warns on such conversions, and we also use `-Werror` which makes those warnings into errors, so we're forced to cast everything back and forth. 

The ideal solution is to use containers that follow a consistent convention, so the casting is unnecessary and there's no implicit conversion, but that requires cooperation from mfem and axom. I think axom's index types are configurable, so we could ask it to use stl-compatible integer types, but mfem hardcodes `int` for everything.

related mfem issue: https://github.com/mfem/mfem/issues/3229

src/serac/numerics/functional/detail/hexahedron_H1.inl
samuelpmishLLNL: oops, I think I misunderstood the comment: maybe you meant for just the tensor product elements?

src/serac/numerics/functional/debug_print.hpp
samuelpmishLLNL: I moved this header and the `mpi_fstream` files into infrastructure

src/serac/numerics/functional/boundary_integral_kernels.hpp
samuelpmishLLNL: I'm not going to worry about this too much, since the quadrature rules are being revamped imminently (and many of these template parameters will be gone).

src/serac/numerics/functional/detail/hexahedron_H1.inl
jamiebramwell: yes - I believe my thought was this is only applicable to the tensor product elements. Really, anything to help clarify what `q` means in it's local context is appreciated.

src/serac/numerics/functional/boundary_integral_kernels.hpp
white238: I vote we make an issue and discuss at a later date. It is more important to get this PR in.

src/serac/numerics/functional/tests/functional_boundary_test.cpp
jamiebramwell: This should probably be in `buildMeshFromFile`. I believe the parallel mesh will build a nodal grid function if the serial mesh had one. I agree that all the mesh builders should have this call. We assume we have a nodal grid function in a lot of places.

src/serac/numerics/functional/integral.hpp
jamiebramwell: I would use the clearer `== 1` or add a comment. I think it's worth it to avoid the extra cognitive load.

src/serac/numerics/functional/boundary_integral_kernels.hpp
jamiebramwell: As long as it's consistent and documented in a logical place, that's good enough for me!

jamiebramwell (approved):

jamiebramwell:
Really awesome work @samuelpmishLLNL !

white238 (approved):

PR #874, opened 2023-01-13
Use the correct integration rule for runtime spaces

jamiebramwell:
This fixes a bug where the runtime (MFEM) integration rule was calculated incorrectly for multiple trial spaces of different order. It also adds a test based on @kswartz92 's reproducer.

### DIFF
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -281,7 +281,7 @@ public:
     }
 
     const mfem::FiniteElement&   el = *test_space_->GetFE(0);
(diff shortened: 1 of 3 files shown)


jamiebramwell:
ping @sethwatts . This should fix the integration order issues.

white238 (approved):

kswartz92:
I built LiDO in codevelop using this branch and confirmed that it fixes both issues I was previously observing. Specifically
1. I can now use quadratic shape parameter fields with linear material parameter fields in functional
2. I am now getting convergence with linear, H1 parameter fields in a topology optimization

kswartz92 (approved):

PR #875, opened 2023-01-14
Simplify finite element state creation

### LINKED ISSUE #872: simplify `FiniteElementState` creation
This code (thanks @barreracruz for the bug report) currently has a problem with it:

```cpp
// Parameter 1
FiniteElementState orderParam(StateManager::newState(FiniteElementState::Options{.order = p, .name = "orderParam"}));
orderParam = max_order_param;

// Parameter 2
auto fec = std::unique_ptr<mfem::FiniteElementCollection>(new mfem::L2_FECollection(p, dim));
FiniteElementState gammaParam(StateManager::newState(FiniteElementState::Options{.order = p, .coll = std::move(fec), .name = "gammaParam"}));
auto gammaFunc = [](const mfem::Vector& x, double) -> double { return (x[1] > 0.5) ? M_PI_2 : 0.0; };
mfem::FunctionCoefficient gammaCoef(gammaFunc);
gammaParam.project(gammaCoef);

// Parameter 3
FiniteElementState etaParam(StateManager::newState(FiniteElementState::Options{.order = p, .coll = std::move(fec), .name = "etaParam"}));
auto etaFunc = [](const mfem::Vector& /*x*/, double) -> double { return 0.0; };
mfem::FunctionCoefficient etaCoef(etaFunc);
etaParam.project(etaCoef);
```

I briefly looked at this code when reviewing the bug report and thought it looked reasonable (note: there were no compilation warnings/errors).

Oddly enough, it doesn't even produce any noticeable problems until running a simulation in parallel with a large enough mesh, at which point NaNs appear in the residual or a segfault may occur (but not consistently). 

The problem is that the 3rd parameter `std::move`s the finite element collection a second time, so it gets a bad value.

------------

We should make sure that other users don't hit this problem in the future by either:

- improving error checking on this (and other similar) interfaces. If we take pointers that are expected to be non-null, we should verify that as soon as possible, and alert the user if something is wrong.

- simplifying the creation of new `FiniteElementState`s. Currently, the user has to repeat themself by specifying the polynomial order in two different places (required to be specified in the `mfem::FiniteElementCollection` as well as the `FiniteElementState::Options`-- which one is actually used?), and creating separate `std::unique_ptr<mfem::FiniteElementCollection>`s for each state. This feels like too much complexity for what could be just an enum (e.g. `enum class Family{ H1, Hcurl, Hdiv, DG };` ). The enum approach doesn't require move semantics, so it's much harder to shoot yourself in the foot.

jamiebramwell:
This replaces the unique pointer to a finite element collection with an enum denoting the basis function selection in the construction of a `FiniteElementVector`.

Note that a new basis function enum class was needed due to resolution conflicts between `dual.hpp` and `expr_template_ops.hpp`, which both define templated `operator*` methods. 

Resolves https://github.com/LLNL/serac/issues/872

### DIFF
--- a/src/serac/physics/state/finite_element_vector.hpp
+++ b/src/serac/physics/state/finite_element_vector.hpp
@@ -18,7 +18,6 @@
 #include "mfem.hpp"
 
 #include "serac/infrastructure/variant.hpp"
-#include "serac/numerics/functional/functional.hpp"
 
 namespace serac {
 
@@ -26,6 +25,17 @@ namespace serac {
  * @brief A sum type for encapsulating either a scalar or vector coeffient
  */
 using GeneralCoefficient = variant<std::shared_ptr<mfem::Coefficient>, std::shared_ptr<mfem::VectorCoefficient>>;
+
+/// @brief The type of a finite element basis function
+/// @note This class is used instead of the Family class from functional do to incompatibilities with Vector expression
+/// templates and the dual number class.
+enum class ElementType
+{
+  H1,     ///< Nodal scalar valued basis functions
+  HCURL,  ///< Nedelec (continuous tangent) vector-valued basis functions
+  HDIV,   ///< Raviart-Thomas (continuous normal) vector-valued basis functions
+  L2      ///< Discontinuous scalar valued basis functions
+};
 
 /**
  * @brief Class for encapsulating the data associated with a vector derived
@@ -58,7 +68,7 @@ public:
      *
      * Options are H1, HCURL, HDIV, or L2.
      */
-    Family element_type = Family::H1;
+    ElementType element_type = ElementType::H1;
 
     /**
      * @brief The name of the field encapsulated by the state object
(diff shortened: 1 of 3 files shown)


src/serac/physics/state/finite_element_vector.hpp
```diff
@@ -51,10 +62,14 @@ class FiniteElementVector : public mfem::HypreParVector {
      * Defaults to scalar valued spaces.
      */
     int vector_dim = 1;
+
     /**
-     * @brief The FECollection to use - defaults to an H1_FECollection
+     * @brief Enum deoniting type of basis functions to use
```
white238: denoting maybe?
```suggestion
     * @brief Enum denoting type of basis functions to use
```

maybe?

white238 (approved):

samuelpmishLLNL (approved):

PR #877, opened 2023-01-20
Update MFEM and Axom

white238:
This includes a bunch of CUDA build fixes in Axom.

Updates to the MFEM 4.5.2 release candidate.

### DIFF
--- a/scripts/spack/configs/config.yaml
+++ b/scripts/spack/configs/config.yaml
@@ -33,12 +33,8 @@ config:
   template_dirs:
     - $spack/share/spack/templates
 
-
-  # Locations where different types of modules should be installed.
-  module_roots:
-    tcl:    $spack/share/spack/modules
-    lmod:   $spack/share/spack/lmod
-
+  # Directory where licenses should be located
+  license_dir: $spack/etc/spack/licenses
 
   # Temporary locations Spack can try to use for builds.
   #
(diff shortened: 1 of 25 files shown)


scripts/spack/packages/mfem/package.py
```diff
@@ -7,19 +7,18 @@
 import shutil
 import sys
 
-from spack import *
+from spack.package import *
```
jamiebramwell, line 10: any word on MFEM adding a cmake-based variant?

jamiebramwell (approved):

scripts/spack/packages/mfem/package.py
white238: none that i have heard.

scripts/spack/configs/config.yaml
```diff
@@ -130,7 +126,7 @@ config:
   # If set to 4, for example, `spack install` will run `make -j4`.
   # If not set, all available cores are used by default.
   # for uberenv, limit build_jobs to 8
```
chapman39, line 128: comment says 8 is limit but switched to 16

chapman39 (approved):

scripts/spack/packages/axom/package.py
```diff
@@ -7,13 +7,12 @@
 import socket
 from os.path import join as pjoin
 
-from spack import *
+from spack.package import *
```
chapman39, line 10: adding `.package` is needed now?

white238: I'm pretty sure that is the new spack standard. I just copy what is in the newer packages.

PR #880, opened 2023-01-26
Bugfix: some tensors in shape function routines were incorrectly shaped

samuelpmishLLNL:
discovered by Kenny

### DIFF
--- a/src/serac/numerics/functional/detail/hexahedron_Hcurl.inl
+++ b/src/serac/numerics/functional/detail/hexahedron_Hcurl.inl
@@ -459,7 +459,7 @@ struct finite_element<Geometry::Hexahedron, Hcurl<p>> {
 
   //////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#if defined(__CUDACC__)
+#if 0
 
   template <int q>
   static SERAC_DEVICE auto interpolate(const dof_type& element_values, const tensor<double, dim, dim>& J,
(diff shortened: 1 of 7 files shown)


kswartz92:
Building lido with this serac branch fixed my issues, thanks @samuelpmishLLNL

src/serac/numerics/functional/detail/hexahedron_Hcurl.inl
```diff
@@ -442,41 +421,45 @@ struct finite_element<Geometry::Hexahedron, Hcurl<p>> {
...
+      auto A1 = contract< x, 0 >(A20, B2) - contract< x, 0 >(A21, G2);
+      element_residual[0].z += contract< z, 0 >(A1, B1);
+    }
     // clang-format on
   }
 
   //////////////////////////////////////////////////////////////////////////////////////////////////////
 
-#if defined(__CUDACC__)
+#if 0
```
jamiebramwell, line 462: Is this disabled for now because we can't dynamically allocate the temporary tensors on device? What is the plan for this moving forward?

samuelpmishLLNL: it's not really disabling it now, the code wasn't being used before. 

> What is the plan for this moving forward?

When I refactor these kernels to use runtime-configurable polynomial orders, I'd like to get everything up and running on CUDA again (but there are still some additional changes that need to be made to the physics modules to work with GPUs).

jamiebramwell: I would add a comment to reflect this, but that makes sense to me.

kswartz92 (approved):

src/serac/numerics/functional/detail/hexahedron_Hcurl.inl
samuelpmishLLNL: noted here: https://github.com/LLNL/serac/issues/881

jamiebramwell (approved):

PR #884, opened 2023-02-07
Periodic mesh bugfix and test

jamiebramwell:
This ensures that the mesh nodal grid function uses discontinuous basis functions when the supplied mesh is periodic. It also fixes a bug that the nodal grid function could be double deleted between the `mfem::Mesh` and the `StateManager`.

### DIFF
--- a/src/serac/physics/state/state_manager.cpp
+++ b/src/serac/physics/state/state_manager.cpp
@@ -60,6 +60,9 @@ double StateManager::newDataCollection(const std::string& name, const std::optio
     auto nodes            = mesh(name).GetNodes();
     if (nodes) {
       is_discontinuous = nodes->FESpace()->FEColl()->GetContType() == mfem::FiniteElementCollection::DISCONTINUOUS;
+      SLIC_WARNING_ROOT(
+          "Periodic mesh detected! This will only work on translational periodic surfaces for vector H1 fields and "
+          "has not been thoroughly tested. Proceed at your own risk.");
     }
 
     // This mfem call ensures the mesh contains an H1 grid function describing nodal
@@ -240,6 +243,9 @@ mfem::ParMesh* StateManager::setMesh(std::unique_ptr<mfem::ParMesh> pmesh, const
   auto nodes            = pmesh->GetNodes();
   if (nodes) {
     is_discontinuous = nodes->FESpace()->FEColl()->GetContType() == mfem::FiniteElementCollection::DISCONTINUOUS;
+    SLIC_WARNING_ROOT(
+        "Periodic mesh detected! This will only work on translational periodic surfaces for vector H1 fields and "
+        "has not been thoroughly tested. Proceed at your own risk.");
   }
 
   // This mfem call ensures the mesh contains an H1 grid function describing nodal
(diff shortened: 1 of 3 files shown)


src/serac/physics/state/state_manager.cpp
```diff
@@ -221,6 +233,26 @@ void StateManager::save(const double t, const int cycle, const std::string& mesh
 
 mfem::ParMesh* StateManager::setMesh(std::unique_ptr<mfem::ParMesh> pmesh, const std::string& mesh_tag)
 {
+  // Determine if the existing nodal grid function is discontinuous. This
+  // indicates that the mesh is periodic and the new nodal grid function must also
+  // be discontinuous.
+  bool is_discontinuous = false;
+  auto nodes            = pmesh->GetNodes();
```
samuelpmishLLNL, line 243: Can't `GetNodes()` return `NULL` for other reasons (unrelated to periodicity of the mesh)?

I've asked mfem to clarify the preconditions of this function before, but it's still essentially undocumented.

jamiebramwell: Yes. However, if it is a periodic mesh this will be non-null as it needs the discontinuous basis function. That is why I have the default of false on line 239.

samuelpmishLLNL:
Some other thoughts: 

1. I'm not convinced that periodic meshes in mfem work reliably, as discussed here: https://github.com/mfem/web/issues/211

To summarize, even the example meshes featured on mfem website will not enforce periodicity correctly in many use cases. I think it's worth managing expectations by warning the user (if we detect a periodic mesh) that things may not work, for reasons out of our control.

2. It's not obvious to me what mfem is doing regarding periodicity, so if we want to eventually support that as an official feature, we should do the due diligence of verifying that it works for different combinations of test/trial spaces.

src/serac/physics/state/state_manager.cpp
samuelpmishLLNL: Ah, I misunderstood what this was doing

samuelpmishLLNL (approved):

jamiebramwell:
> Some other thoughts:
> 
>     1. I'm not convinced that periodic meshes in mfem work reliably, as discussed here: [Clarify limitations of periodic meshes mfem/web#211](https://github.com/mfem/web/issues/211)
> 
> 
> To summarize, even the example meshes featured on mfem website will not enforce periodicity correctly in many use cases. I think it's worth managing expectations by warning the user (if we detect a periodic mesh) that things may not work, for reasons out of our control.
> 
>     2. It's not obvious to me what mfem is doing regarding periodicity, so if we want to eventually support that as an official feature, we should do the due diligence of verifying that it works for different combinations of test/trial spaces.

Good point. I'll add a "buyer beware" warning if a user supplies a periodic mesh.

src/serac/physics/state/state_manager.cpp
```diff
@@ -53,13 +53,28 @@ double StateManager::newDataCollection(const std::string& name, const std::optio
 
     // Functional needs the nodal grid function and neighbor data in the mesh
 
+    // Determine if the existing nodal grid function is discontinuous. This
+    // indicates that the mesh is periodic and the new nodal grid function must also
+    // be discontinuous.
+    bool is_discontinuous = false;
```
white238, line 59: Can these two duplicate sections be unified into a function w/o too much effort?

white238 (approved):

barreracruz:
I double checked in my branch with a representative test and these changes seem to fix the issues I had. Thank you!

barreracruz (approved):

PR #892, opened 2023-02-23
WIP: Fix the shape displacement for periodic meshes

jamiebramwell:
Previously, the shape displacement grid function was H1 even if the underlying mesh was periodic, and therefore had an L2 nodal grid function. This ensures consistency between the two fields.


jamiebramwell:
After thinking about this, the shape fields should be in H1 even for periodic meshes. I'm closing this for now.

PR #893, opened 2023-02-24
Expose the solid mechanics stiffness matrix

jamiebramwell:


src/serac/physics/solid_mechanics.hpp
```diff
@@ -557,6 +557,24 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+  /**
+   * @brief Return the assembled stiffness matrix
+   *
+   * This method returns a pair {K, K_e} representing the last computed linearized stiffness matrix.
+   * The K matrix has the essential degree of freedom rows and columns zeroed with a
+   * 1 on the diagonal and K_e contains the zeroed rows and columns, e.g. K_total = K + K_e.
+   *
+   * @warning This interface is not stable and may change in the future.
+   *
+   * @return A pair of the eliminted stiffness matrix and a matrix containing the eliminated rows and cols
```
white238:
```suggestion
   * @return A pair of the eliminated stiffness matrix and a matrix containing the eliminated rows and cols
```

white238 (approved):

barreracruz:
FYI, I pull these changes into my branch and extracted the matrices in Matlab for a simple problem. 
K:
image
and Ke:
image

For some reason, I had to extract one column at a time since Matlab gave me this error when trying to load the files directly:
image
and
image
Not sure why this is happening, but I guess this is more of an MFEM thing?

samuelpmishLLNL (approved):

samuelpmishLLNL:
> For some reason, I had to extract one column at a time since Matlab gave me this error when trying to load the files directly ...

In case anyone wants this in the future, here's a little MATLAB function to read whatever kind of format HypreParMatrix is outputting:

```matlab
# load_hypre.m
function K = load_hypre(filename)
  
  [f, err] = fopen(filename);
  
  % read header, not used here but necessary for combining separate files
  dimensions = fscanf(f, '%u', [1, 4]);
 
  % read data (0-based COO format)
  IJV = reshape(fscanf(f, '%d %d %f'), 3, [])';
  
  K = sparse(IJV(:,1)+1, IJV(:,2)+1, IJV(:,3));

```

I only tested it in octave, so I'm not actually sure it works in MATLAB. Also, I imagine the full HypreParMatrix is split into separate files for distributed problems, so you'll likely need to load the blocks of rows for each processor and combine them.

jamiebramwell:
> I only tested it in octave, so I'm not actually sure it works in MATLAB. Also, I imagine the full HypreParMatrix is split into separate files for distributed problems, so you'll likely need to load the blocks of rows for each processor and combine them.

I think in that case, the global numberings aren't guaranteed anyway. See https://github.com/mfem/mfem/issues/2992

samuelpmishLLNL:
> > I only tested it in octave, so I'm not actually sure it works in MATLAB. Also, I imagine the full HypreParMatrix is split into separate files for distributed problems, so you'll likely need to load the blocks of rows for each processor and combine them.
> 
> I think in that case, the global numberings aren't guaranteed anyway. See [mfem/mfem#2992](https://github.com/mfem/mfem/issues/2992)

I don't really understand the discussion in that issue. Is it implying that calling `HypreParMatrix::Print()` from each processor does not output enough information to reconstruct the matrix?

PR #894, opened 2023-02-24
Better nonlinear solvers with KINSOL

jamiebramwell:
This PR is a step toward providing more robust nonlinear solver options. Specifically it:

- Refactors the `EquationSolver` interface to physics modules. Instead of passing option structs to physics modules, they are passed via dependency injection. See [here](https://serac.readthedocs.io/en/feature-bramwell-nonlinear_solvers/sphinx/dev_guide/equation_solver.html) for built docs.
- Removes the confusing `std::optional` and `std::variant` idioms in the solver configuration structs
- Updates Sundials to 6.4.1
- Adds the ability to have a custom nonlinear solver similar to the linear solver option
- Adds missing mfem initialization calls 
- Adds KINSOL-based tests to ensure correctness
- Adds a work-around to avoid move assignment operators for `mfem::Vector` within the nonlinear function evaluation (residual) calls (see https://github.com/mfem/mfem/issues/3531)
- Removes an unused test driver routine


jamiebramwell:
I am in the process of updating all of our TPL installs to use the Sundials 6.4.1.

scripts/spack/packages/serac/package.py
```diff
@@ -87,7 +87,8 @@ class Serac(CachedCMakePackage, CudaPackage):
     depends_on('py-sphinx', when="+devtools")
     depends_on('py-ats', when="+devtools")
 
-    depends_on("sundials@5.7.0+hypre+monitoring~examples~examples-install",
+    # MFEM is deprecating the monitoring support with sundials v6.0 and later
+    depends_on("sundials@6.4.1+hypre~monitoring~examples~examples-install",
```
white238, line 91: #906

src/serac/infrastructure/initialize.cpp
```diff
@@ -54,6 +56,12 @@ std::pair<int, int> initialize(int argc, char* argv[], MPI_Comm comm)
   // Start the profiler (no-op if not enabled)
   profiling::initialize(comm);
 
+  mfem::Hypre::Init();
+
+#ifdef MFEM_USE_SUNDIALS
```
white238, line 61: We should have our own config variable for this.

https://github.com/LLNL/serac/blob/develop/src/serac/serac_config.hpp.in
https://github.com/LLNL/serac/blob/develop/cmake/serac-config.cmake.in#L66
https://github.com/LLNL/serac/blob/develop/cmake/SeracConfigHeader.cmake#L39

We should probably clean this up and not have it in so many places... sigh.

src/serac/numerics/equation_solver.cpp
```diff
@@ -19,7 +19,7 @@ EquationSolver::EquationSolver(MPI_Comm comm, const LinearSolverOptions& lin_opt
     lin_solver_ = BuildIterativeLinearSolver(comm, *iter_options);
   }
   // If it's a custom solver, check that the mfem::Solver* is not null
-  else if (auto custom = std::get_if<CustomSolverOptions>(&lin_options)) {
+  else if (auto custom = std::get_if<CustomLinearSolverOptions>(&lin_options)) {
```
white238: 👍

src/serac/infrastructure/initialize.cpp
white238: Are we ever going to use sundials directly or only through MFEM?

src/serac/numerics/solver_config.hpp
```diff
@@ -282,7 +300,15 @@ struct NonlinearSolverOptions {
...
+  NonlinearSolver nonlin_solver = NonlinearSolver::Newton;
 };
 
+/**
+ * @brief Parameters for a nonlinear solver
+ *
+ * This can either be a user-constructed and owned solver or an iterative solver that serac constructs and builds
+ * itself.
+ */
+using NonlinearSolverOptions = std::variant<CustomNonlinearSolverOptions, IterativeNonlinearSolverOptions>;
```
white238: Can you explain why one is "Custom" and the other is not? Does "Custom" imply non-iterative? Should it be `SeracNonlinearSolverOptions` or just `NonlinearSolverOptions`?

src/serac/physics/heat_transfer.hpp
```diff
@@ -475,7 +475,12 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
           temperature_.space().TrueVSize(),
 
           [this](const mfem::Vector& u, mfem::Vector& r) {
-            r = (*residual_)(u, zero_, shape_displacement_, *parameters_[parameter_indices].state...);
+            const mfem::Vector res =
+                (*residual_)(u, zero_, shape_displacement_, *parameters_[parameter_indices].state...);
+
+            // TODO this copy is required as the sundials solvers do not allow move assignments because of their memory
```
white238, line 535: Link to MFEM issue solving this?

src/serac/physics/solid_legacy.hpp
```diff
@@ -100,7 +100,7 @@ class SolidLegacy : public BasePhysics {
      * @brief The options for the inlucded nonlinear solve
```
white238:
```suggestion
     * @brief The options for the included nonlinear solve
```

src/serac/infrastructure/initialize.cpp
jamiebramwell: We're only going to use it through MFEM.

src/serac/numerics/solver_config.hpp
jamiebramwell: Custom means a user builds their own MFEM solver object and passes it in as a raw pointer. I can add some better verbiage on this.

white238 (approved):

jamiebramwell:
@btalamini @samuelpmishLLNL Can you take a look at this one? Thanks!

src/serac/numerics/solver_config.hpp
```diff
@@ -103,9 +103,12 @@ enum class LinearSolver
...
 {
-  MFEMNewton,               /**< Newton-Raphson */
-  KINFullStep,              /**< KINFullStep */
-  KINBacktrackingLineSearch /**< KINBacktrackingLineSearch */
+  Newton,                    /**< MFEM-native Newton-Raphson */
+  LBFGS,                     /**< MFEM-native Limited memory BFGS */
+  KINFullStep,               /**< KINSOL Full Newton (Sundials must be enabled) */
+  KINBacktrackingLineSearch, /**< KINSOL Newton with Backtracking Line Search (Sundials must be enabled) */
+  KINPicard,                 /**< KINSOL Picard (Sundials must be enabled) */
+  KINFP                      /**< KINSOL Fixed Point (Sundials must be enabled) */
```
samuelpmishLLNL: I think these methods only work for specific kinds of nonlinear equations, so they are likely not suitable for solving residual equations in general (unless KINSOL is doing something I don't understand)

samuelpmishLLNL: Also, I think Jorge-Luis mentioned earlier that Newton and KINFullStep were producing significantly different results, which seems unexpected

jamiebramwell: That definitely seems weird. I'll investigate this.

jamiebramwell: The differences were due to the linear solver not converging and the two nonlinear solvers having different behavior under that case. When the linear solver converges, the two Newton methods produce the same results.

Picard works for a general nonlinear solution scheme, but fixed point does not. I've removed the later and added Picard as a solver for one of our tests to verify this.

src/serac/numerics/equation_solver.hpp
```diff
@@ -61,12 +61,18 @@ class EquationSolver : public mfem::Solver {
...
+  }
 
   /**
    * @overload
    */
-  const mfem::IterativeSolver& NonlinearSolver() const { return *nonlin_solver_; }
+  const mfem::NewtonSolver& NonlinearSolver() const
+  {
+    return std::visit([](auto&& solver) -> const mfem::NewtonSolver& { return *solver; }, nonlin_solver_);
+  }
```
samuelpmishLLNL: What is this doing? Is `nonlin_solver_` always an `mfem::NewtonSolver`?

src/serac/numerics/equation_solver.hpp
```diff
@@ -169,7 +175,7 @@ class EquationSolver : public mfem::Solver {
   /**
    * @brief The optional nonlinear Newton-Raphson solver object
    */
-  std::unique_ptr<mfem::NewtonSolver> nonlin_solver_;
+  std::variant<std::unique_ptr<mfem::NewtonSolver>, mfem::NewtonSolver*> nonlin_solver_;
```
samuelpmishLLNL: this also looks unusual, are the KINSOL solver options all derived from `mfem::NewtonSolver`?

jamiebramwell:
I think the initial pass of refactoring the physics module/`EquationSolver` interface is done. Hopefully this is a bit clearer now!

src/serac/numerics/solver_config.hpp
```diff
@@ -169,120 +158,86 @@ struct AMGXPrec {
...
 
-/**
- * @brief Parameters for a custom solver (currently just a non-owning pointer to the solver)
- * @note This is preferable to unique_ptr or even references because non-trivial copy constructors
- * and destructors are a nightmare in this context
- */
-struct CustomSolverOptions {
   /**
-   * @brief A non-owning pointer to the custom mfem solver to use
+   * @brief Deubbing print level for the preconditioner
```
samuelpmishLLNL:
```suggestion
   * @brief Debugging print level for the preconditioner
```

src/serac/physics/heat_transfer.hpp
```diff
@@ -29,45 +29,31 @@ namespace heat_transfer {
...
-/**
- * @brief Reasonable defaults for quasi-static thermal conduction simulations
- */
-const SolverOptions default_static_options = {default_linear_options, default_nonlinear_options};
-
-/**
- * @brief Reasonable defaults for quasi-static thermal conduction simulations with a direct solver
- */
-const SolverOptions direct_static_options = {direct_linear_options, default_nonlinear_options};
+    .relative_tol = 1.0e-4, .absolute_tol = 1.0e-8, .max_iterations = 500, .print_level = 1};
```
samuelpmishLLNL: Can we explicitly specify which method is being used here?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -31,39 +31,33 @@ namespace solid_mechanics {
...
+const LinearSolverOptions direct_linear_options = {.linear_solver = LinearSolver::SuperLU, .print_level = 0};
 
 /**
  * @brief default iteration limits, tolerances and verbosity for solving the
  * systems of nonlinear equations that show up in implicit
  * solid mechanics simulations
  */
 const NonlinearSolverOptions default_nonlinear_options = {
-    .rel_tol = 1.0e-4, .abs_tol = 1.0e-8, .max_iter = 10, .print_level = 1};
+    .relative_tol = 1.0e-4, .absolute_tol = 1.0e-8, .max_iterations = 10, .print_level = 1};
```
samuelpmishLLNL: here too

src/serac/physics/solid_mechanics.hpp
```diff
@@ -31,39 +31,33 @@ namespace solid_mechanics {
...
-/// solver options that use a direct linear solver for (quasi-)static analyses
-const SolverOptions direct_static_options = {direct_linear_options, default_nonlinear_options};
-
-/// the default solver and time integration options for dynamic analyses
-const SolverOptions default_dynamic_options = {
-    default_linear_options, default_nonlinear_options,
-    TimesteppingOptions{TimestepMethod::Newmark, DirichletEnforcementMethod::RateControl}};
+/**
+ * @brief default quasistatic timestepping options for solid mechanics
+ */
```
samuelpmishLLNL: is this equivalent to 

```cpp
/// default quasistatic timestepping options for solid mechanics
```

if so, the single-line version is preferable to me (but I don't know if this is covered by our style guide)

src/serac/physics/solid_mechanics.hpp
```diff
@@ -647,7 +655,8 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
 
     // the ~20 lines of code below are essentially equivalent to the 1-liner
     // u += dot(inv(J), dot(J_elim[:, dofs], (U(t + dt) - u)[dofs]));
-    {
+    auto* lin_solver = nonlin_solver_->LinearSolver();
+    if (lin_solver) {
```
samuelpmishLLNL: I can't remember if this block of code was only for improving the first newton iterate, or if it also served to set the essential degree of freedom values. If it does the latter, then putting it behind a conditional would be a mistake.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -744,7 +751,11 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     // Add the sign correction to move the term to the RHS
     adjoint_load_vector *= -1.0;
 
-    auto& lin_solver = nonlin_solver_.LinearSolver();
+    auto* lin_solver = nonlin_solver_->LinearSolver();
+
+    SLIC_ERROR_ROOT_IF(
+        !lin_solver,
+        "Adjoint solves currently need a linear solver associated with the supplied forward nonlinear solver");
```
samuelpmishLLNL: is it worthwhile for serac to support solver options that prevent using its differentiation features?

src/serac/physics/tests/parameterized_thermomechanics_example.cpp
```diff
@@ -88,8 +86,15 @@ TEST(Thermomechanics, ParameterizedMaterial)
     serac::StateManager::setMesh(std::move(mesh));
   }
 
-  SolidMechanics<p, dim, Parameters<H1<p>, H1<p> > > simulation(default_static_options, GeometricNonlinearities::On,
-                                                                "thermomechanics_simulation");
+  serac::NonlinearSolverOptions nonlin_options{.nonlin_solver  = NonlinearSolver::KINPicard,
+                                               .relative_tol   = 1.0e-12,
+                                               .absolute_tol   = 1.0e-12,
+                                               .max_iterations = 30,
+                                               .print_level    = 1};
```
samuelpmishLLNL: why Picard here?

jamiebramwell: Just to make sure the Picard option is working. I'm happy to move it to another test if you have ideas.

src/serac/physics/solid_mechanics.hpp
jamiebramwell: I think I used the /** */ method mostly because my doxygen plug in defaults to that. I can make a pass at changing this, but it might be best to be in a separate PR for clarity. I'll change this one though.

src/serac/physics/solid_mechanics.hpp
jamiebramwell: Any thoughts @btalamini? I know you were interested in implementing custom nonlinear solvers that may or may not have internal linear solvers. My thinking is that you can still define a custom `EquationSolver` with appropriate linear (adjoint) solvers in that case. We just need to make sure it has been defined. I guess the big question is do we want to allow equation solvers that do not have linear solvers? If not, I can ensure that the given pointer in the `EquationSolver` constructor is non-null and just return a reference here.

src/serac/physics/solid_mechanics.hpp
jamiebramwell: Yes, you are right! Good catch!

jamiebramwell: Yes! This is a bug. Nice catch!

src/serac/physics/tests/parameterized_thermomechanics_example.cpp
samuelpmishLLNL: Do we already have a dedicated set of tests for just the `EquationSolver` class? That would be a more natural place to figure out if its different options are working properly. 

I think files labeled with "example" should focus more on demonstrating how to use serac effectively, instead of testing solver options.

jamiebramwell: That's an excellent point. I will add some unit tests for the `EquationSolver` class.

src/serac/physics/solid_mechanics.hpp
btalamini: I don't want to over-design it. I think it's ok to require that a nonlinear solver has to have a linear solver. We can reorganize later if necessary. There is a very strong incentive to re-use the linear solver (since the martix is essentially the same) that it makes sense to push the user in this direction.

jamiebramwell:
I swapped to `shared_ptr` within the `EquationSolver` class to enable user-facing value semantics. Let me know what you think!

jamiebramwell:
I believe I've addressed all the comments from this PR and our Friday discussion. Let me know if you have any other questions or concerns.

jamiebramwell:
I think I'm finally happy with where this ended up. Thank you @samuelpmishLLNL for your ideas on how to simplify the interface!

src/docs/sphinx/dev_guide/equation_solver.rst
```diff
@@ -0,0 +1,177 @@
...
+.. literalinclude:: ../../../../src/serac/numerics/solver_config.hpp
+   :start-after: _preconditioners_start
+   :end-before: _preconditioners_end
+   :language: C++
+
+As an example, the default solid mechanics ``EquationSolver`` can be built using the following syntax:
+
+.. code-block:: cpp
+
+  const serac::NonlinearSolverOptions nonlinear_options = {.nonlin_solver  = NonlinearSolver::Newton,
```
white238: Is this section anywhere in the code so we can ensure that it does not get out of date?

src/docs/sphinx/dev_guide/equation_solver.rst
```diff
@@ -0,0 +1,177 @@
...
+.. code-block:: cpp
+
+  auto nonlinear_solver = std::make_unique<mfem::NewtonSolver>(mpi_comm);
+  auto linear_solver    = std::make_unique<mfem::GMRESSolver>(mpi_comm);
+  auto preconditioner   = std::make_unique<mfem::OperatorJacobiSmoother>();
+
+  serac::SolidMechanics<p, dim> solid_solver(EquationSolver(std::move(nonlinear_solver), std::move(linear_solver), std::move(preconditioner)),
+                                             solid_mechanics::default_quasistatic_options);
+
+Note that each physics module must have its own ``EquationSolver``. They cannot be reused between modules.
```
white238:
```suggestion
.. warning:: Each physics module must have its own ``EquationSolver``. They cannot be reused between modules. 
```

src/docs/sphinx/dev_guide/equation_solver.rst
jamiebramwell: Let me add this to a test...

src/serac/physics/thermomechanics.hpp
```diff
@@ -156,15 +186,11 @@ class Thermomechanics : public BasePhysics {
...
-      SLIC_ERROR_ROOT_IF(std::abs(dt - initial_dt) > 1.0e-6,
-                         "Operator split coupled solvers cannot adaptively change the timestep");
-    } else {
-      SLIC_ERROR_ROOT("Only operator split coupling is currently implemented");
-    }
+    double initial_dt = dt;
+    thermal_.advanceTimestep(dt);
+    solid_.advanceTimestep(dt);
+    SLIC_ERROR_ROOT_IF(std::abs(dt - initial_dt) > 1.0e-6,
+                       "Operator split coupled solvers cannot adaptively change the timestep");
```
white238, line 192: Are we always operator split coupled now? (whatever that means)

jamiebramwell: Yes - the complexities of handling operator split and monolithic solves in one module would be a pretty big challenge. If we decide to go down that route, it will be a separate (big) PR.

src/serac/numerics/tests/equationsolver.cpp
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    AllEquationSolverTests, EquationSolverSuite,
+    testing::Combine(testing::Values(NonlinearSolver::Newton, NonlinearSolver::LBFGS, NonlinearSolver::KINFullStep,
+                                     NonlinearSolver::KINBacktrackingLineSearch, NonlinearSolver::KINPicard),
+                     testing::Values(LinearSolver::CG, LinearSolver::GMRES, LinearSolver::SuperLU),
+                     testing::Values(Preconditioner::HypreJacobi, Preconditioner::HypreL1Jacobi,
+                                     Preconditioner::HypreGaussSeidel, Preconditioner::HypreAMG)));
```
samuelpmishLLNL, line 114: I'm not a gtest superuser-- does this generate a test for each possible combination of {nonlinear, linear, preconditioner} values?

jamiebramwell: Yes - it tests all of the possible combinations.

samuelpmishLLNL: Cool, I look forward to using this sort of pattern in some of our other test suites too!

jamiebramwell: Yep! It's super handy. 
image

src/serac/physics/thermomechanics.hpp
```diff
@@ -32,20 +32,52 @@ template <int order, int dim, typename... parameter_space>
...
+   * @brief Construct a new coupled Thermal-SolidMechanics object
    *
-   * @param thermal_options The options for the linear, nonlinear, and ODE solves of the thermal operator
-   * @param solid_options The options for the linear, nonlinear, and ODE solves of the thermal operator
+   * @param thermal_nonlin_opts The nonlinear solver options for solving the nonlinear heat conduction residual
+   * equations
+   * @param thermal_lin_opts The linear solver options for solving the linearized Jacobian heat transfer equations
+   * @param thermal_timestepping The timestepping options for the heat transfer operator
+   * @param solid_nonlin_opts The nonlinear solver options for solving the nonlinear solid mechanics residual equations
+   * @param solid_lin_opts The linear solver options for solving the linearized Jacobian solid mechanics equations
```
white238:
```suggestion
   * @param thermal_nonlin_opts The options for solving the nonlinear heat conduction residual equations
   * @param thermal_lin_opts The options for solving the linearized Jacobian heat transfer equations
   * @param thermal_timestepping The timestepping options for the heat transfer operator
   * @param solid_nonlin_opts The options for solving the nonlinear solid mechanics residual equations
   * @param solid_lin_opts The options for solving the linearized Jacobian solid mechanics equations
```

Little less wordy

src/serac/physics/tests/thermal_mechanics.cpp
```diff
@@ -226,20 +226,20 @@ void parameterized()
 
   // define the solid solver configurations
   // no default solver options for solid yet, so make some here
-  const IterativeSolverOptions default_linear_options = {.rel_tol     = 1.0e-6,
-                                                         .abs_tol     = 1.0e-10,
-                                                         .print_level = 0,
-                                                         .max_iter    = 500,
-                                                         .lin_solver  = LinearSolver::GMRES,
-                                                         .prec        = HypreBoomerAMGPrec{}};
+  const LinearSolverOptions default_linear_options = {.linear_solver  = LinearSolver::GMRES,
```
white238: If these are "default" should they be in a header somewhere and not just in a test?

src/serac/physics/tests/thermal_dynamics_patch.cpp
```diff
@@ -106,12 +104,11 @@ double dynamic_solution_error(const ExactSolution& exact_solution, PatchBoundary
...
   // Construct a heat transfer solver
-  auto solver_options                        = direct_dynamic_options;
-  solver_options.nonlinear.abs_tol           = 5e-13;
-  solver_options.nonlinear.rel_tol           = 5e-13;
-  solver_options.dynamic->timestepper        = TimestepMethod::BackwardEuler;
-  solver_options.dynamic->enforcement_method = DirichletEnforcementMethod::DirectControl;
-  HeatTransfer<p, dim> thermal(solver_options, "thermal");
+  NonlinearSolverOptions nonlinear_opts{.relative_tol = 5.0e-13, .absolute_tol = 5.0e-13};
+  TimesteppingOptions    dyn_opts{.timestepper        = TimestepMethod::BackwardEuler,
+                               .enforcement_method = DirichletEnforcementMethod::DirectControl};
```
white238, line 110: kind of surprised by this formatting...

src/serac/physics/tests/thermal_mechanics.cpp
jamiebramwell: This is a bad stale variable name. I will fix it.

jamiebramwell: actually, we should be able to just use the default, so I'll do that.

src/serac/physics/solid_mechanics_input.cpp
```diff
@@ -82,7 +82,7 @@ serac::SolidMechanicsInputOptions FromInlet<serac::SolidMechanicsInputOptions>::
                        "Unrecognized enforcement method: " << enforcement_method);
     dyn_options.enforcement_method = enforcement_methods.at(enforcement_method);
 
-    result.solver_options.dynamic = std::move(dyn_options);
+    result.timestepping_options = std::move(dyn_options);
```
white238: `dyn_options` seems like a wrongly named variable at this point.

samuelpmishLLNL (approved):

src/serac/physics/solid_mechanics.hpp
```diff
@@ -32,39 +32,32 @@ namespace solid_mechanics {
...
-const SolverOptions default_dynamic_options = {
-    default_linear_options, default_nonlinear_options,
-    TimesteppingOptions{TimestepMethod::Newmark, DirichletEnforcementMethod::RateControl}};
-
-/// the direct solver and time integration options for dynamic analyses
-const SolverOptions direct_dynamic_options = {
-    direct_linear_options, default_nonlinear_options,
-    TimesteppingOptions{TimestepMethod::Newmark, DirichletEnforcementMethod::RateControl}};
+/// default implicit dynamic timestepping options for solid mechanics
+const TimesteppingOptions default_dynamic_options = {TimestepMethod::Newmark, DirichletEnforcementMethod::RateControl};
```
white238: Another swapping of `dynamic` and `timestepping`

src/serac/physics/heat_transfer_input.cpp
```diff
@@ -72,7 +72,7 @@ serac::HeatTransferInputOptions FromInlet<serac::HeatTransferInputOptions>::oper
                        "Unrecognized enforcement method: " << enforcement_method);
     dyn_options.enforcement_method = enforcement_methods.at(enforcement_method);
 
-    result.solver_options.dynamic = std::move(dyn_options);
+    result.timestepping_options = dyn_options;
```
white238: dynamic vs timestepping

src/serac/physics/heat_transfer.hpp
```diff
@@ -104,14 +93,35 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+  /**
+   * @brief Construct a new heat transfer object
+   *
+   * @param[in] solver The nonlinear equation solver for the heat transfer equations
+   * @param[in] dynamic_opts The timestepping options for the heat transfer ordinary differential equations
+   * @param[in] name An optional name for the physics module instance
+   * used by an underlying material model or load
+   * @param[in] pmesh The mesh to conduct the simulation on, if different than the default mesh
+   */
+  HeatTransfer(std::unique_ptr<serac::mfem_ext::EquationSolver> solver, const serac::TimesteppingOptions dynamic_opts,
```
white238: dynamic vs timestepping

src/serac/physics/heat_transfer.hpp
```diff
@@ -30,45 +30,34 @@ namespace heat_transfer {
...
+                                                          .max_iterations = 500,
+                                                          .print_level    = 1};
 
 /**
  * @brief Reasonable defaults for dynamic thermal conduction simulations
  */
-const SolverOptions default_dynamic_options = {
-    default_linear_options, default_nonlinear_options,
-    TimesteppingOptions{TimestepMethod::BackwardEuler, DirichletEnforcementMethod::RateControl}};
+const TimesteppingOptions default_dynamic_options = {TimestepMethod::BackwardEuler,
```
white238: dynamic vs timestepping

src/serac/physics/heat_transfer.hpp
```diff
@@ -104,14 +93,35 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+   * @param[in] nonlinear_opts The nonlinear solver options for solving the nonlinear residual equations
+   * @param[in] lin_opts The linear solver options for solving the linearized Jacobian equations
+   * @param[in] dynamic_opts The timestepping options for the heat transfer ordinary differential equations
    * @param[in] name An optional name for the physics module instance
    * used by an underlying material model or load
    * @param[in] pmesh The mesh to conduct the simulation on, if different than the default mesh
    */
-  HeatTransfer(const SolverOptions& options, const std::string& name = {}, mfem::ParMesh* pmesh = nullptr)
+  HeatTransfer(const NonlinearSolverOptions nonlinear_opts, const LinearSolverOptions lin_opts,
+               const serac::TimesteppingOptions dynamic_opts, const std::string& name = "",
```
white238: dynamic vs timestepping

src/serac/numerics/equation_solver.hpp
```diff
@@ -235,4 +266,4 @@ template <>
 struct FromInlet<serac::mfem_ext::EquationSolver> {
   /// @brief Returns created object from Inlet container
   serac::mfem_ext::EquationSolver operator()(const axom::inlet::Container& base);
-};
+};
```
white238:
```suggestion
};

```

src/serac/numerics/equation_solver.hpp
```diff
@@ -92,116 +121,118 @@ class EquationSolver : public mfem::Solver {
...
+   * as a member variable for lifetime purposes
    */
-  std::unique_ptr<mfem::NewtonSolver> nonlin_solver_;
+  mutable std::unique_ptr<mfem::SuperLURowLocMatrix> superlu_mat_;
 
   /**
-   * @brief Whether the solver (linear solver) has been configured with the nonlinear solver
-   * @note This is a workaround as some nonlinear solvers require SetOperator to be called
-   * before SetSolver
+   * @brief The underlying MFEM-based superLU solver. It requires a special
```
white238:
```suggestion
   * @brief The underlying MFEM-based SuperLU solver. It requires a special
```

src/serac/numerics/equation_solver.hpp
```diff
@@ -92,116 +121,118 @@ class EquationSolver : public mfem::Solver {
...
 
-    /**
-     * @brief The underlying MFEM-based superLU solver. It requires a special
-     * superLU matrix type which we store in this object. This enables compatibility
-     * with HypreParMatrix when used as an input.
-     */
-    mfem::SuperLUSolver superlu_solver_;
-  };
+/**
+ * @brief A wrapper class for using the MFEM super LU solver with a HypreParMatrix
```
white238:
```suggestion
 * @brief A wrapper class for using the MFEM SuperLU solver with a HypreParMatrix
```

src/serac/numerics/equation_solver.cpp
```diff
@@ -333,10 +364,13 @@ serac::NonlinearSolverOptions FromInlet<serac::NonlinearSolverOptions>::operator
...
+  auto lin    = base["linear"].get<LinearSolverOptions>();
+  auto nonlin = base["nonlinear"].get<NonlinearSolverOptions>();
+
+  auto [linear_solver, preconditioner] = serac::mfem_ext::buildLinearSolverAndPreconditioner(lin, MPI_COMM_WORLD);
+
+  serac::mfem_ext::EquationSolver eq_solver(serac::mfem_ext::buildNonlinearSolver(nonlin, MPI_COMM_WORLD),
+                                            std::move(linear_solver), std::move(preconditioner));
+
+  return eq_solver;
+}
```
white238:
```suggestion
}

```

src/serac/numerics/equation_solver.cpp
```diff
@@ -88,133 +212,41 @@ std::unique_ptr<mfem::AmgXSolver> configureAMGX(const MPI_Comm comm, const AMGXP
...
-    default:
-      SLIC_ERROR_ROOT("Linear solver type not recognized.");
-      exitGracefully(true);
-  }
-
-  iter_lin_solver->SetRelTol(lin_options.rel_tol);
-  iter_lin_solver->SetAbsTol(lin_options.abs_tol);
-  iter_lin_solver->SetMaxIter(lin_options.max_iter);
-  iter_lin_solver->SetPrintLevel(lin_options.print_level);
+  std::unique_ptr<mfem::Solver> preconditioner_ptr;
```
white238: There are several variables here that end in`_ptr`. We don't do that anywhere else in the code. We should be consistent unless there is a reason.

src/serac/numerics/equation_solver.hpp
```diff
@@ -92,116 +121,118 @@ class EquationSolver : public mfem::Solver {
...
    */
-  std::unique_ptr<mfem::NewtonSolver> nonlin_solver_;
+  mutable std::unique_ptr<mfem::SuperLURowLocMatrix> superlu_mat_;
 
   /**
-   * @brief Whether the solver (linear solver) has been configured with the nonlinear solver
-   * @note This is a workaround as some nonlinear solvers require SetOperator to be called
-   * before SetSolver
+   * @brief The underlying MFEM-based SuperLU solver. It requires a special
+   * superLU matrix type which we store in this object. This enables compatibility
```
white238:
```suggestion
   * SuperLU matrix type which we store in this object. This enables compatibility
```

white238 (approved):

PR #895, opened 2023-02-27
Refactor `solveAdjoint` interface to be compatible with systems of equations

jamiebramwell:
The previous interface assumed only one primal field in the adjoint solve. This new interface allows for block systems of equations.

Note @kswartz92 , this will change your wrapper interface slightly.

### DIFF
--- a/src/serac/physics/heat_transfer.hpp
+++ b/src/serac/physics/heat_transfer.hpp
@@ -569,7 +569,8 @@ public:
       // If the essential adjoint load container does not have a temperature dual but it has a non-zero size, the
       // user has supplied an incorrectly-named dual vector.
       SLIC_ERROR_IF(duals_with_essential_boundary.size() != 0,
-                    "Essential adjoint boundary condition given for an unexpected primal field. Expected adjoint boundary condition named \"temperature\".");
+                    "Essential adjoint boundary condition given for an unexpected primal field. Expected adjoint "
+                    "boundary condition named \"temperature\".");
     }
 
     for (const auto& bc : bcs_.essentials()) {
--- a/src/serac/physics/heat_transfer.hpp
+++ b/src/serac/physics/heat_transfer.hpp
@@ -528,7 +528,7 @@ public:
    *
    * @param adjoint_loads An unordered map containing finite element duals representing the RHS of the adjoint equations
    * indexed by their name
-   * @param duals_with_essential_boundary An unordered map containing finite element duals representing the
+   * @param adjoint_with_essential_boundary An unordered map containing finite element states representing the
    * non-homogeneous essential boundary condition data for the adjoint problem indexed by their name
    * @return An unordered map of the adjoint solutions indexed by their name. It has a single entry named
    * "adjoint_temperature"
--- a/src/serac/physics/heat_transfer.hpp
+++ b/src/serac/physics/heat_transfer.hpp
@@ -530,6 +530,21 @@ public:
    * boundary condition data for the adjoint problem
    * @return The computed adjoint finite element state
    */
+
+  /**
+   * @brief Solve the adjoint problem
+   * @pre It is expected that the forward analysis is complete and the current temperature state is valid
+   * @pre The adjoint load maps are expected to contain a single entry named "temperature"
+   * @note If the essential boundary dual is not specified, homogeneous essential boundary conditions are applied to
+   * the adjoint system
+   *
+   * @param adjoint_loads An unordered map containing finite element duals representing the RHS of the adjoint equations
+   * indexed by their name
+   * @param duals_with_essential_boundary A unordered map containing finite element duals representing the
+   * non-homogenous essential boundary condition data for the adjoint problem indexed their name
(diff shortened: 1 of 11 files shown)


src/serac/physics/heat_transfer.hpp
```diff
@@ -550,8 +561,15 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
-      adjoint_essential = *dual_with_essential_boundary;
+    auto essential_adjoint_temp = duals_with_essential_boundary.find("temperature");
+
+    if (essential_adjoint_temp != duals_with_essential_boundary.end()) {
+      adjoint_essential = essential_adjoint_temp->second;
+    } else {
+      // If the essential adjoint load container does not have a temperature dual but it has a non-zero size, the
+      // user has supplied an incorrectly-named dual vector.
+      SLIC_ERROR_IF(duals_with_essential_boundary.size() != 0,
+                    "Essential adjoint boundary condition given for an unexpected primal field.");
```
white238:
```suggestion
                    "Essential adjoint boundary condition given for an unexpected primal field. Expected adjoint boundary condition named \"temperature\".");
```

Or something to that extent. I prefer when error messages nudge you in the direction to fix your problem. We could even output the given names.

white238 (approved):

jamiebramwell:
@samuelpmishLLNL @btalamini @kswartz92 Can you take a look at this? Thanks!

kswartz92 (approved):
this looks to be an easy to change to accommodate on the LiDO side

src/serac/physics/heat_transfer.hpp
```diff
@@ -522,26 +522,37 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
    *
-   * @param[in] adjoint_load The dual state that contains the right hand side of the adjoint system (d quantity of
-   * interest/d temperature)
-   * @param[in] dual_with_essential_boundary A optional finite element dual containing the non-homogenous essential
-   * boundary condition data for the adjoint problem
-   * @return The computed adjoint finite element state
+   * @param adjoint_loads An unordered map containing finite element duals representing the RHS of the adjoint equations
+   * indexed by their name
+   * @param duals_with_essential_boundary A unordered map containing finite element duals representing the
+   * non-homogenous essential boundary condition data for the adjoint problem indexed their name
```
```suggestion
   * @param duals_with_essential_boundary An unordered map containing finite element duals representing the
   * non-homogenous essential boundary condition data for the adjoint problem indexed by their name
```
I can't remember if it's "homogeneous" or "homogenous".

On a side note: what is this argument (`duals_with_essential_boundary`) for? If it relates to the essential boundary condition values for a given time, should we be asking the user for that information or just keep track of it ourselves? If it is just the values of the essential bcs, should it be a dual or a primal field?

samuelpmishLLNL:
It seems like we're using a lot of competing conventions for specifying field information in these interfaces:

- lookup mutable ref by name
   `FiniteElementState& state(const std::string& state_name)`
- lookup mutable ref by index
   `FiniteElementState& parameter(const size_t parameter_index)`
- lookup mutable ref by named functions
   `FiniteElementState& shapeDisplacement() { return shape_displacement_; }`
- separate getter/setter lookup by index:
   `void setParameter(const size_t parameter_index, FiniteElementState& parameter_state);`
   `void setShapeDisplacement(FiniteElementState& shape_displacement);`
- using `std::unordered_map` of names
   `const std::unordered_map<...> solveAdjoint(std::unordered_map<...> , std::unordered_map<...>);`

I think we should consider refactoring this eventually to make things consistent.

src/serac/physics/heat_transfer.hpp
jamiebramwell: The adjoint system can have non-zero essential boundary conditions if the boundary values are part of the quantity of interest. This is common for parameter estimation in inverse problems where the quantity of interest is a particular experimental data point. You are correct that I believe this should be a `FiniteElementState` instead of a `FiniteElementDual`.

jamiebramwell:
> It seems like we're using a lot of competing conventions for specifying field information in these interfaces:
> 
>     * lookup mutable ref by name
>       `FiniteElementState& state(const std::string& state_name)`
> 
>     * lookup mutable ref by index
>       `FiniteElementState& parameter(const size_t parameter_index)`
> 
>     * lookup mutable ref by named functions
>       `FiniteElementState& shapeDisplacement() { return shape_displacement_; }`
> 
>     * separate getter/setter lookup by index:
>       `void setParameter(const size_t parameter_index, FiniteElementState& parameter_state);`
>       `void setShapeDisplacement(FiniteElementState& shape_displacement);`
> 
>     * using `std::unordered_map` of names
>       `const std::unordered_map<...> solveAdjoint(std::unordered_map<...> , std::unordered_map<...>);`
> 
> 
> I think we should consider refactoring this eventually to make things consistent.

I opened a new issue to track this. I agree a clean look at the interfaces is warranted.

samuelpmishLLNL (approved):

PR #903, opened 2023-03-15
Bugfix/white238/fix doxygen yet again

white238:
It's frustrating that readthedocs keeps breaking this. This should be more robust and is now using their environment variable.

https://serac.readthedocs.io/en/bugfix-white238-fix_doxygen_yet_again/


samuelpmishLLNL (approved):

jamiebramwell (approved):
Thank you for figuring this out!

PR #911, opened 2023-03-27
Fix parallel issues in the parameterized thermomechanics example

jamiebramwell:
As I was updating this example to use the Picard solver, I noted a few issues that are corrected by this PR.

- Uses a parallel norm calculation for the adjoint QOI value. If this is not used, it reports the local norm and is therefore incorrect for parallel runs.
- Uses parallel-aware SLIC logging calls for readability of output.
- Adds a correctness check as a test.


samuelpmishLLNL:
For my benefit, what was the cause of the dot product being carried out locally? Does

```cpp
mfem::InnerProduct(mfem::Vector, mfem::HypreParVector) 
```

return the local dot product rather than the global one?

jamiebramwell:
> For my benefit, what was the cause of the dot product being carried out locally? Does
> 
> ```c++
> mfem::InnerProduct(mfem::Vector, mfem::HypreParVector) 
> ```
> 
> return the local dot product rather than the global one?

If the vector is an `mfem::Vector` instead of a `mfem::HypreParVector`, it uses the local version of the inner product. In other words, it uses https://github.com/mfem/mfem/blob/92bf531fab08deff5d6ba687e26bc276a33e1650/linalg/vector.hpp#L679-L682 instead of https://github.com/mfem/mfem/blob/92bf531fab08deff5d6ba687e26bc276a33e1650/linalg/hypre.cpp#L414-L417 . The pure `mfem::Vector` version does not perform an MPI reduce.

jamiebramwell:
Another fix would be passing the MPI communicator to an `InnerProduct` method with a different signature: https://github.com/mfem/mfem/blob/92bf531fab08deff5d6ba687e26bc276a33e1650/linalg/vector.hpp#L689

samuelpmishLLNL:
Ah, I see. There is no such overload `mfem::InnerProduct(mfem::Vector, mfem::HypreParVector)`, so the second  argument is upcast to `mfem::Vector` to hit the `mfem::InnerProduct(mfem::Vector, mfem::Vector)` overload. That upcast seems dangerous, since it's an implicit conversion between types with different semantics (local vs. global), but it's good to know about!

samuelpmishLLNL (approved):

src/serac/physics/tests/parameterized_thermomechanics_example.cpp
```diff
@@ -188,20 +189,35 @@ int main(int argc, char* argv[])
...
 }
 
 // output:
 // vertical displacement integrated over the top surface: 0.000883477
-// total area of the top surface: 0.441959
+// total area of the top surface: 0.441077
 // exact area of the top surface: 0.441786
 // average vertical displacement: 0.001999
 // expected average vertical displacement: 0.002
-// directional derivative of QoI by adjoint-state method: 0.441931
```
btalamini, line 206: Do these changes to the directional derivatives make sense? It's potentially concerning that the gold result was wrong and we didn't notice.

jamiebramwell: Yes - it was taking the directional derivative of a random direction. Now that I'm using a seed for the randomize call, it is reproducible. 

The parallel error showed up in the output when running over more than one rank. We just weren't checking it, so I added a gtest.

btalamini: Ok, I saw the pseudorandom seed setting. Got it.

btalamini (approved):

PR #912, opened 2023-03-28
Update Hypre and add an ILU preconditioner

jamiebramwell:
This updates Hypre to v2.26.0 and the latest version of MFEM master. It also adds an Hypre-based ILU preconditioner as a solver option and enables a separate print level flag for HypreILU and HypreBoomerAMG.


jamiebramwell:
I still need to update our pre-installed IPLs, but the functionality is good to go.

src/serac/physics/tests/solid_legacy_sensitivity.cpp
```diff
@@ -28,8 +28,8 @@ TEST(SolidLegacy, FiniteDiff)
   mfem::ParMesh* mesh = serac::StateManager::setMesh(std::make_unique<mfem::ParMesh>(MPI_COMM_WORLD, cuboid));
 
   // Setup some solver options
-  serac::IterativeSolverOptions const default_linear_options = {.rel_tol     = 1.0e-8,
-                                                                .abs_tol     = 1.0e-12,
+  serac::IterativeSolverOptions const default_linear_options = {.rel_tol     = 1.0e-10,
```
btalamini: Did these tolerances need to be tightened to avoid some specific bad outcome? ( Like a linear solver that doesn't converge and breaks the nonlinear solver, etc.). If so, a comment could help us later. What do you think - is this overkill?

btalamini (approved):

chapman39:
How come sundials isn't installed for certain systems now (rzgenie, quartz, ubuntu docker)? I'm still not fully sure what sundials is, if Serac uses it, or if this will affect LiDO?

jamiebramwell:
Weird! I thought it was being installed by default. I will investigate...

jamiebramwell:
I believe this is now ready to go.

chapman39 (approved):

host-configs/docker/clang@10.0.0.cmake
```diff
@@ -57,29 +57,29 @@ set(ENABLE_OPENMP ON CACHE BOOL "")
 # TPLs
 #------------------------------------------------------------------------------
 
-set(TPL_ROOT "/home/serac/serac_tpls/clang-10.0.0" CACHE PATH "")
+set(TPL_ROOT "/home/serac/serac_tpls/spack/opt/spack/linux-ubuntu20.04-x86_64/clang-10.0.0" CACHE PATH "")
```
white238, line 60: This alludes to a missing `config` section in `spack.yaml` for the docker images.

white238: @chapman39 can you fix this in a follow up PR?

white238 (approved):

host-configs/docker/clang@10.0.0.cmake
chapman39: i was wondering about that. yeah i can do that

PR #918, opened 2023-04-10
adding operator name prefix to dual names

kswartz92:
I was trying to use parameter fields with multiple physics operators, which resulted in repeated dual names in the data manager. Adding the physics operator name as a prefix to the dual name fixes this issue.


samuelpmishLLNL:
Can you describe the use case and problem in more detail?

jamiebramwell:
I think I understand the problem. If you have two physics modules active with the same parameter names, they will clobber each other in the `StateManager`. Thank you for the fix @kswartz92 !

jamiebramwell (approved):

samuelpmishLLNL:
I'm trying to understand if this change will be duplicating the parameter output for each physics module it is used by

jamiebramwell:
> I'm trying to understand if this change will be duplicating the parameter output for each physics module it is used by

That's a good question. @kswartz92 , are you using the same `FiniteElementState` as a parameter in multiple modules, or are you generating new parameters for each module that just happen to have the same name?

kswartz92:
> > I'm trying to understand if this change will be duplicating the parameter output for each physics module it is used by
> 
> That's a good question. @kswartz92 , are you using the same `FiniteElementState` as a parameter in multiple modules, or are you generating new parameters for each module that just happen to have the same name?

I am using the same `FiniteElementState` as a parameter in multiple modules. As @samuelpmishLLNL points out, I think this could result in undesirable names (e.g. "heat_solid_parameter0_sensitivity") if a state was generated by a solid mechanics module and then set by a heat transfer module. I'm wondering now if the  `addPrefix` should be removed from `generateParameter()`.

Am I correct in thinking that parameter (state) names shouldn't have anything to do with the module, but sensitivity (dual) names should have the module name?

jamiebramwell:
> 
> Am I correct in thinking that parameter (state) names shouldn't have anything to do with the module, but sensitivity (dual) names should have the module name?

That makes sense to me.

btalamini:
> > Am I correct in thinking that parameter (state) names shouldn't have anything to do with the module, but sensitivity (dual) names should have the module name?
> 
> That makes sense to me.

Can I pull on this thread a bit? Why do we want the sensitivity fields prefixed with the module name?

jamiebramwell:
> 
> Can I pull on this thread a bit? Why do we want the sensitivity fields prefixed with the module name?

Because the sensitivity is directly connected to a specific adjoint solution owned by a specific physics module.

btalamini:
> > Can I pull on this thread a bit? Why do we want the sensitivity fields prefixed with the module name?
> 
> Because the sensitivity is directly connected to a specific adjoint solution owned by a specific physics module.

There is another option, which is that Serac returns the total sensitivity of a QOI wrt a parameter field. Someone has to sum these sensitivity contributions, either LiDO or Serac - is it obvious that it should be LiDO? I don't want to drag us through design design decisions that have already been settled. Maybe we can have a quick online meeting?

samuelpmishLLNL (approved):
Sorry, I forgot about this one. I hope I wasn't holding anything up.

PR #920, opened 2023-04-11
Add true dof list essential boundary condition option

jamiebramwell:
This adds the capability for a spatially-defined essential boundary condition for the `SolidMechanics` module. Instead of specifying a boundary attribute set, you specify a callback function which returns whether or not a node is constrained given its spatial position.

This also fixes some bugs in how we handle true degree of freedom lists in the boundary condition manager.

Results from test showing a pinned corner and fixed z displacement at the bottom with negative z displacement at the top.
MicrosoftTeams-image


src/serac/physics/solid_mechanics.hpp
```diff
@@ -283,6 +283,26 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   * The @a true_dofs list can be determined using functions from the @a mfem::ParFiniteElementSpace class.
+   *
+   * @note The coefficient is required to be vector-valued. However, only the dofs specified in the @a true_dofs
+   * array will be set. This means that if the @a true_dofs array only contains dofs for a specific vector component in
+   * a vector-valued finite element space, only that component will be set.
+   */
+  void setDisplacementBCs(const mfem::Array<int>                                          true_dofs,
+                          std::function<void(const mfem::Vector&, double, mfem::Vector&)> disp)
+  {
+    disp_bdr_coef_ = std::make_shared<mfem::VectorFunctionCoefficient>(disp);
```
kswartz92: I think you're missing a `dim` argument here

jamiebramwell: Yep, you are correct. Fix (and test) incoming...

jamiebramwell:
Question - should we merge this as-is for flexibility or add a `pinNode` condition that does a spatial query and determines the dof list itself?

jamiebramwell:
- [x] Change to `pinNode` boundary condition

jamiebramwell:
I believe this is good to go now.

jamiebramwell:
ping @kswartz92 @barreracruz

src/serac/physics/solid_mechanics.hpp
```diff
@@ -363,6 +372,83 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   * @param disp The vector function containing the prescribed displacement values
+   *
+   * The @a true_dofs list can be determined using functions from the @a mfem::ParFiniteElementSpace class.
+   *
+   * @note The coefficient is required to be vector-valued. However, only the dofs specified in the @a true_dofs
+   * array will be set. This means that if the @a true_dofs array only contains dofs for a specific vector component in
+   * a vector-valued finite element space, only that component will be set.
+   */
+  void setDisplacementBCs(const mfem::Array<int>                                          true_dofs,
+                          std::function<void(const mfem::Vector&, double, mfem::Vector&)> disp)
```
samuelpmishLLNL: if I write

```cpp
foo.setDisplacementBCs({1, 2, 3}, bar);
```

it's not immediately clear which overload is being called. 

Does the initializer list in the first argument create a `std::set` (i.e. dirichlet bc by attribute) or an `mfem::Array<int>` (dirichlet b.c. by dof index)?

We could use a more descriptive name (e.g. `setDisplacementBCsByDofList` or something) to disambiguate, although that does make the function name pretty long.

jamiebramwell: I was wondering if this method should even be publicly available at all, but I can imagine times where someone might want to grab a random assortment of dofs they calculated themselves. I agree that at the very least it should have a different name as it is fundamentally different.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -1073,6 +1211,65 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
                             shape_displacement_, *parameters_[parameter_indices].state...);
       }...};
+
+  /**
+   * @brief Calculate a list of constrained dofs in the true displacement vector from a function that
+   * returns true if a physical coordinate is in the constrained set
+   *
+   * @param is_node_constrained A function that takes a point in physical space and returns true if the contained
+   * degrees of freedom should be constrained
+   * @param component A marker for whetehr
```
samuelpmishLLNL:
```suggestion
   * @param component which component is constrained (uninitialized implies all components are constrained)
```

samuelpmishLLNL: I'm not sure if my suggestion is actually correct

src/serac/physics/solid_mechanics.hpp
```diff
@@ -1073,6 +1211,65 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+            }
+          }
+
+          if (is_active_component) {
+            // Get the local dof number for the prescribed component
+            int local_vector_dof = mfem::Ordering::Map<mfem::Ordering::byNODES>(
+                nodal_positions.FESpace()->GetNDofs(), nodal_positions.FESpace()->GetVDim(), i, d);
+
+            // Save the spatial position for this coordinate dof
+            node_coords(d) = nodal_positions(local_vector_dof);
```
samuelpmishLLNL: I think we want to initialize each component of the nodal position vector unconditionally

src/serac/physics/tests/solid.cpp
samuelpmishLLNL, line 1: This test could be a patch test / analytic solution. As written, it's not clear that the expected answer is correct, but I don't think this needs to be addressed to merge the PR in.

src/serac/physics/tests/solid.cpp
```diff
@@ -320,14 +499,16 @@ void functional_parameterized_solid_test(double expected_disp_norm)
...
-TEST(SolidMechanics, 2DLinearPressure)
-{
-  functional_solid_test_boundary<1, 2>(0.028525698834671667, TestType::Pressure);
-}
+TEST(SolidMechanics, SpatialBoundaryCondition) { functional_solid_spatial_essential_bc(); }
+
+// TEST(SolidMechanics, 2DLinearPressure)
+//{
+//  functional_solid_test_boundary<1, 2>(0.028525698834671667, TestType::Pressure);
+//}
```
samuelpmishLLNL: can these tests be re-enabled?

samuelpmishLLNL (approved):

src/serac/physics/solid_mechanics.hpp
```diff
@@ -363,6 +372,137 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   * @brief Set the displacement essential boundary conditions on a set of true degrees of freedom
+   *
+   * @param true_dofs A set of true degrees of freedom to set the displacement on
+   * @param disp The vector function containing the prescribed displacement values
+   *
+   * The @a true_dofs list can be determined using functions from the @a mfem::ParFiniteElementSpace class.
+   *
+   * @note The coefficient is required to be vector-valued. However, only the dofs specified in the @a true_dofs
+   * array will be set. This means that if the @a true_dofs array only contains dofs for a specific vector component in
+   * a vector-valued finite element space, only that component will be set.
```
barreracruz: I do not have any case in which I want to constrain one or two components only, but good to know that I have control over that with this method for less common configurations. I guess that defining the coefficient as be vector-valued may confuse a new user (that doesn't read the documentation)... Just trying to add a meaningful comment here, but not a big deal for me. Happy to be able to use it now :). Thanks!

src/serac/physics/tests/solid.cpp
```diff
@@ -100,6 +100,177 @@ void functional_solid_test_static_J2()
...
+
+  // Perform the quasi-static solve
+  double dt = 1.0;
+  solid_solver.advanceTimestep(dt);
+  solid_solver.outputState();
+
+  auto [size, rank] = serac::getMPIInfo();
+
+  mfem::Vector rank_0_displacement(54);
+  rank_0_displacement(0)  = -1.0617852403956275e-15;
```
white238: :dizzy_face:

src/serac/physics/tests/solid.cpp
jamiebramwell: Oh yes - this was a shortcut for testing.

src/serac/physics/tests/solid.cpp
jamiebramwell: I honestly just copy/pasted from the computed solution. I can go back through and clean this up.

jamiebramwell:
I think this PR is ready to go now.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -363,6 +372,141 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   *
+   * For the displacement function, the argument is the input position and the output is the value of the component of
+   * the displacement.
+   *
+   * @note The displacement function is required to be vector-valued. However, only the dofs specified in the @a
+   * true_dofs array will be set. This means that if the @a true_dofs array only contains dofs for a specific vector
+   * component in a vector-valued finite element space, only that component will be set.
+   */
+  void setDisplacementBCsByDofList(const mfem::Array<int>                                          true_dofs,
+                                   std::function<void(const mfem::Vector&, double, mfem::Vector&)> disp)
```
btalamini, line 392: This is a nitpick: the signature of the `disp` functor isn't described. In particular, the time argument isn't mentioned, and I think this is the only way this overload differs from the other one.

btalamini (approved):

PR #922, opened 2023-04-19
Small cleanup

white238:
I noticed some small things when I pulled our recent uberenv update changes over to Axom.

These are all noop changes.


jamiebramwell (approved):

chapman39:
should we add `mirrors:: {}` to all config files?

chapman39 (approved):

white238:
> should we add `mirrors:: {}` to all config files?

I don't think thats necessary. This was a temporary work around because LC was blocking the Spack mirror. It was fixed a while ago.

PR #933, opened 2023-05-10
Allow block systems to be solved with `SuperLU`

jamiebramwell:
This adds the ability for form a monolithic system from a block operator when `SuperLU` is requested for block systems. In other words, direct solvers can now be used for block systems.


lihanyu97 (approved):

src/serac/numerics/equation_solver.cpp
```diff
@@ -62,10 +62,41 @@ void SuperLUSolver::Mult(const mfem::Vector& x, mfem::Vector& y) const
...
+                           "Trying to use SuperLU on a block operator that does not contain HypreParMatrix blocks.");
+
+        hypre_blocks(i, j) = hypre_block;
+      }
+    }
+
+    // Note that MFEM passes ownership of this matrix to the caller
+    monolithic_mat_ = std::unique_ptr<mfem::HypreParMatrix>(mfem::HypreParMatrixFromBlocks(hypre_blocks));
+
+    superlu_mat_ = std::make_unique<mfem::SuperLURowLocMatrix>(*monolithic_mat_);
```
samuelpmishLLNL: I'm not familiar with this part of mfem, but 

https://docs.mfem.org/html/classmfem_1_1SuperLURowLocMatrix.html#a42d2e1b4f6d76f9c39f2985acc4c4928

says that a copy is made when calling this `SuperLURowLocMatrix` ctor, so the new data member `monolithic_mat_` may not be needed.

samuelpmishLLNL (approved):

PR #935, opened 2023-05-10
Disable Sundials tests when not built with Sundials

jamiebramwell:


white238 (approved):

jamiebramwell:
Looks like clang doesn't like this solution. I'll fix it when I get back to California.

jamiebramwell:
I believe this should work with clang compilers now.

chapman39 (approved):

PR #939, opened 2023-05-15
Update Caliper and Adiak spack packages

tepperly:
* Add versions 2.8.0, 2.9.0 and 2.9.1 of Caliper (update to a release in Jan 24, 2023).
* Add version 0.2.2 of Adiak that is newer than the 0.3.0-alpha tag.

### DIFF
--- a/scripts/spack/packages/caliper/package.py
+++ b/scripts/spack/packages/caliper/package.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
@@ -6,12 +6,10 @@
 import os
 import sys
 
-from llnl.util import tty
-
-from spack import *
+from spack.package import *
 
 
-class Caliper(CMakePackage, CudaPackage):
+class Caliper(CMakePackage, CudaPackage, ROCmPackage):
     """Caliper is a program instrumentation and performance measurement
     framework. It is designed as a performance analysis toolbox in a
     library, allowing one to bake performance analysis capabilities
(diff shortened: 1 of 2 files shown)


chapman39:
@tepperly do you know if this builds on clang compilers? im trying newer version on my Toss4 pr, but i seem to be getting libunwind errors

```
     589    /usr/WS2/meemee/serac/serac_libs/toss_4_x86_64_ib/2023_05_12_17_52_
            06/spack/lib/spack/env/clang/clang++ -O3 -DNDEBUG CMakeFiles/cali-s
            tat.dir/cali-stat.cpp.o -o cali-stat  ../util/libcaliper-tools-util
            .a ../../libcaliper-serial.a /usr/lib64/libdw.so /usr/tce/packages/
            clang/clang-14.0.6/lib/libunwind.so /usr/WS2/meemee/serac/serac_lib
            s/toss_4_x86_64_ib/2023_05_12_17_52_06/clang-14.0.6/adiak-0.2.2-rys
            bzdxdzdche5tjsqz7zy46kquzknbb/lib/libadiak.a /usr/tce/packages/mvap
            ich2/mvapich2-2.3.6-clang-14.0.6/lib/libmpi.so /usr/tce/packages/mv
            apich2/mvapich2-2.3.6-clang-14.0.6/lib/libmpicxx.so -ldl /usr/lib64
            /librt.so -pthread
     590    /opt/rh/gcc-toolset-10/root/usr/lib/gcc/x86_64-redhat-linux/10/../.
            ./../../bin/ld: ../../libcaliper-serial.a(Callpath.cpp.o): in funct
            ion `(anonymous namespace)::Callpath::callpath_service_register(cal
            i::Caliper*, cali::Channel*)':
  >> 591    Callpath.cpp:(.text+0x491): undefined reference to `_Ux86_64_getcon
            text'
```

scripts/spack/packages/caliper/package.py
```diff
@@ -1,193 +1,178 @@
...
-    conflicts('+libdw', '@:2.4')
+    conflicts("+sosflow", "@2.0.0:2.9")
+    conflicts("+adiak", "@:2.1")
+    conflicts("+libdw", "@:2.4")
+    conflicts("+rocm", "@:2.7")
+    conflicts("+rocm+cuda")
 
-    patch('for_aarch64.patch', when='target=aarch64:')
-    # SERAC EDIT START
-    patch('no_static_cupti.patch', when='+cuda~shared')
```
chapman39, line 94: @white238 do we still need this patch?

white238: It should be in the newer versions of caliper.

chapman39:
i think we can just use the builtin spack recipes, since the only difference was the static patch.

scripts/spack/packages/caliper/package.py
tepperly: It looks like our spack.yml says that libunwind is unbuildable. It may be that we need a more recent version.

white238: libunwind is a bit of a mystery.. older external versions work on toss3, the toss4 external failed but has the symbol they are looking for (i verified via nm).

@chapman39 was able to move forward just removing the external libunwind on toss4

tepperly: https://github.com/libunwind/libunwind/blob/master/src/x86_64/getcontext.S#L38

white238: Yea its there:
```
[quartz:~]$ nm -D /usr/lib64/libunwind.so | grep -i getcontext
0000000000005534 T _Ux86_64_getcontext
```

jamiebramwell (approved):

white238:
Closing due to being taken care of in the TOSS4 PR by alex. Thanks for pointing this out @tepperly !

Issue #940, opened 2023-05-26
KINSOL exits prematurely on simple nonlinear problem


samuelpmishLLNL:
here's a sourcefile with a reproducer:
[beam_bending.zip](https://github.com/LLNL/serac/files/11578278/beam_bending.zip)

When running this problem with the Kinsol Newton option, it performs a few iterations and exits with the error message:

```
 [KINSOL] KINSol
   Return value: -7 (KIN_MXNEWT_5X_EXCEEDED)
[KINSOL ERROR]  KINSol
  Five consecutive steps have been taken that satisfy a scaled step length test.
```

Using the naive Newton method converges monotonically in ~10 steps. It's unclear to me if this is indicative of a bug in our wrapper of Kinsol or if we just need to adjust some knobs.

jandrej:
You have to set `KINSetNumMaxIters` to something larger as well (the default is 200, is that value kept?). You probably also want to allow KINSol to rebuild the Jacobian for every nonlinear iteration for a test. E.g. set https://sundials.readthedocs.io/en/latest/kinsol/Usage/index.html#c.KINSetMaxSetupCalls to 1

PR #941, opened 2023-05-26
fix failing thermomechanics test

samuelpmishLLNL:
the relevant fix is the small change to solid_mechanics.hpp, where our extrapolated initial guess for the Newton solver wasn't respecting the essential boundary conditions. Was there a related issue for this failing test?

I'm also adding the reproducer file from #940 to facilitate future debugging of that issue.

### DIFF
--- a/src/serac/physics/tests/beam_bending.cpp
+++ b/src/serac/physics/tests/beam_bending.cpp
@@ -1,0 +1,109 @@
+// Copyright (c) 2019-2023, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "serac/physics/solid_mechanics.hpp"
+
+#include <functional>
+#include <fstream>
+#include <set>
+#include <string>
+
+#include "axom/slic/core/SimpleLogger.hpp"
+#include <gtest/gtest.h>
+#include "mfem.hpp"
+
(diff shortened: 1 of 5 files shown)


btalamini:
This is a great catch. It's troubling that none of our other tests were sensitive to this. I guess that our patch tests were passing because they were taking more than 1 Newton iteration, and we recovered the essential boundary conditions on the second and subsequent Newton iterations.
I think we would catch this if we disallow Newton iterations on the patch tests. The patch solutions are linear, so the warm start should hit the exact solution without any subsequent Newton iterations.

btalamini (approved):

src/serac/physics/tests/beam_bending.cpp
```diff
@@ -0,0 +1,108 @@
...
+  axom::sidre::DataStore datastore;
+  serac::StateManager::initialize(datastore, "beam_bending_data");
+
+  // Construct the appropriate dimension mesh and give it to the data store
+  std::string filename = SERAC_REPO_DIR "/data/meshes/beam-quad.mesh";
+
+  auto mesh = mesh::refineAndDistribute(buildMeshFromFile(filename), 0, 0);
+  serac::StateManager::setMesh(std::move(mesh));
+
+  // _solver_params_start
```
jamiebramwell: I believe these are copied from the other example that I pull into the docs page. They should be deleted here.

jamiebramwell (approved):
Thank you for figuring this out!

PR #945, opened 2023-06-08
Optimize code check globbing

white238:
Previous code was globbing every file in our repository, including build directories and submodules, then including/excluding based on good lists. This reworks that logic to only glob known source directories.

### DIFF
--- a/cmake/SeracMacros.cmake
+++ b/cmake/SeracMacros.cmake
@@ -24,44 +24,28 @@
 macro(serac_add_code_checks)
 
     set(options)
-    set(singleValueArgs PREFIX )
-    set(multiValueArgs  INCLUDES EXCLUDES)
+    set(singleValueArgs PREFIX)
+    set(multiValueArgs)
 
     # Parse the arguments to the macro
     cmake_parse_arguments(arg
          "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN})
 
-    set(_all_sources)
+    # Create file globbing expressions that only include directories that contain source
+    set(_base_dirs "tests" "src" "examples")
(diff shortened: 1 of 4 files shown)


scripts/azure-pipelines/linux-check.sh
```diff
@@ -73,7 +73,7 @@ if [[ "$DO_DOCS_CHECK" == "yes" ]] ; then
 fi
 
 if [[ "$DO_STYLE_CHECK" == "yes" ]] ; then
-    or_die make VERBOSE=1 check
+    or_die make VERBOSE=1 clangformat_check
```
white238, line 76: This was technically only doing clang format before because it was the only check executable in the image but this makes me feel better.

jamiebramwell (approved):

jamiebramwell:
@barreracruz This should fix your compile time problem.

cmake/SeracMacros.cmake
```diff
@@ -24,44 +24,28 @@
...
+    set(singleValueArgs PREFIX)
+    set(multiValueArgs)
 
     # Parse the arguments to the macro
     cmake_parse_arguments(arg
          "${options}" "${singleValueArgs}" "${multiValueArgs}" ${ARGN})
 
-    set(_all_sources)
+    # Create file globbing expressions that only include directories that contain source
+    set(_base_dirs "tests" "src" "examples")
```
chapman39, line 35: does this add `examples` to be code-checked? was it not previously? (just testing my understanding)

chapman39 (approved):

cmake/SeracMacros.cmake
jamiebramwell: I believe yes, hence the small style updates.

PR #952, opened 2023-06-27
add differentiable eigenvalue calculation for small matrices

samuelpmishLLNL:


kswartz92 (approved):
I've been successfully using this functionality in LiDO. I believe @samuelpmishLLNL mentioned this wouldn't work on GPUs, but in the meantime it's very useful for computing maximum principal stress.

samuelpmishLLNL:
lgtm

jamiebramwell (approved):
This will be super useful! Thanks @samuelpmishLLNL !

samuelpmishLLNL:
@jamiebramwell 
thanks for doing the `make style` for me, I don't have a machine with the right version of clang-format any more, and apparently clang-format-11 gives slightly different results

jamiebramwell:
> @jamiebramwell thanks for doing the `make style` for me, I don't have a machine with the right version of clang-format any more, and apparently clang-format-11 gives slightly different results

Yeah, the style thing is annoying. We should probably just do a batch clang update on all of our platforms.

PR #954, opened 2023-06-29
Pass dX/dxi to boundary q-functions

samuelpmishLLNL:
Following a discussion on 7/22, this PR aims to add support for passing additional information to boundary integrals, to enable differentiating them with respect to mesh nodes. 

---------

Previously, the reference configuration unit normal (let $\text{normalize}(\square) \coloneqq \frac{\square}{||\square||})$

$$
\boldsymbol{\hat{n}}(\boldsymbol{X}) = \text{normalize}(\frac{d\boldsymbol{X}}{d\xi} \times \frac{d\boldsymbol{X}}{d\eta})
$$

was precalculated at each quadrature point on the surface and passed in to boundary q-functions. However, if the user wanted to apply a displacement, $`\boldsymbol{p}(\boldsymbol{X})`$, to the mesh coordinates, it wasn't possible to calculate the updated surface normal from inside the q-function with the information available ($`\displaystyle\boldsymbol{\hat{n}}(\boldsymbol{X}), \frac{d\boldsymbol{p}}{d\xi}, \frac{d\boldsymbol{p}}{d\eta}`$) since the expression for the updated surface normal

$$
\begin{align}
\boldsymbol{\hat{n}}(\boldsymbol{X} + \boldsymbol{p}) 
&= 
\text{normalize}(\frac{d(\boldsymbol{X} + \boldsymbol{p})}{d\xi} \times \frac{d(\boldsymbol{X} + \boldsymbol{p})}{d\eta})  \\ \\
&= 
\text{normalize}\big(
(\frac{d\boldsymbol{X}}{d\xi} \times \frac{d\boldsymbol{X}}{d\eta}) + 
(\frac{d\boldsymbol{X}}{d\xi} \times \frac{d\boldsymbol{p}}{d\eta}) + 
(\frac{d\boldsymbol{p}}{d\xi} \times \frac{d\boldsymbol{X}}{d\eta}) + 
(\frac{d\boldsymbol{p}}{d\xi} \times \frac{d\boldsymbol{p}}{d\eta})
\big)
\end{align}
$$

includes terms that require information not present in the q-function.

This PR makes it so that boundary q-functions no longer receive the reference unit surface normal-- instead, they are passed in the entire surface jacobian ($`\frac{d\boldsymbol{X}}{d\xi}, \frac{d\boldsymbol{X}}{d\eta}`$) of the mesh coordinates, so that users have full control of the surface normal calculation.

As a result, this is an interface breaking change, and boundary q-functions passed directly to `serac::Functional` will need to be changed slightly (q-functions passed to physics modules don't need to change):

```cpp
// before:
[](auto X, auto n, ...) {
  return f(X, n);
}

// after:
[](auto X, /* auto n, */ ...) {
  auto n = normalize(cross(get<DERIVATIVE>(X)));
  return f(get<VALUE>(X), n);
}
```

That is: the mesh coordinates and their surface jacobian are now passed in as a tuple, like the other user-requested arguments.


samuelpmishLLNL:
Note: there are also some changes to the `QuadratureData` class here. because I noticed a bug in the implementation where meshes with both simplex and tensor product elements would not handle the quadrature data buffers correctly. This is conceptually orthogonal to the name of this PR, but is included here anyway.

src/serac/physics/tests/beam_bending.cpp
```diff
@@ -77,7 +77,7 @@ TEST(BeamBending, TwoDimensional)
   solid_solver.setDisplacement(bc);
 
   solid_solver.setPiolaTraction(
-      [](const auto& x, const tensor<double, dim>& n, const double) { return -0.01 * n * (x[1] > 0.99); });
```
jamiebramwell, line 80: Is there a reason why `auto` doesn't work anymore? Or is this just for clarity?

src/serac/physics/tests/lce_Brighenti_tensile.cpp
```diff
@@ -150,15 +150,20 @@ TEST(LiquidCrystalElastomer, Brighenti)
   Functional<double(H1<p, dim>)> avgYDispQoI({&solid_solver.displacement().space()});
   avgYDispQoI.AddSurfaceIntegral(
       DependsOn<0>{},
-      [=](auto x, auto n, auto displacement) {
+      [=](auto position, auto displacement) {
+        auto [X, dX_dxi] = position;
         auto [u, du_dxi] = displacement;
-        return dot(u, n) * ((x[1] > 0.99 * ly) ? 1.0 : 0.0);
+        auto n = normalize(cross(dX_dxi));
```
jamiebramwell: Should this be a (simple) function called `calculate_normal` or something? I can imagine it will be used often.

src/serac/numerics/functional/quadrature_data.hpp
```diff
@@ -44,6 +44,59 @@ struct Nothing {
 struct Empty {
 };
 
+template <typename T>
```
jamiebramwell, line 47: Am I interpreting this correctly that this is a specialization of axom array for the `serac::Nothing` type? If so, a few comments stating why this is needed would be appreciated.

src/serac/numerics/functional/quadrature_data.hpp
```diff
@@ -54,54 +107,64 @@ struct Empty {
...
-  /// dtor, deallocates memory
-  ~QuadratureData() { delete data; }
 
-  /// access a mutable reference to the quadrature data at element `i`, quadrature point `j`
-  SERAC_HOST_DEVICE T& operator()(size_t i, size_t j) { return data[i * stride + j]; }
+  using geom_array_t = std::array< uint32_t, mfem::Geometry::NUM_GEOMETRIES >;
 
-  /// access a const reference to the quadrature data at element `i`, quadrature point `j`
-  SERAC_HOST_DEVICE const T& operator()(size_t i, size_t j) const { return data[i * stride + j]; }
+  using tmp_t = axom::Array<T, 2>;
```
jamiebramwell: What does this temp type contain? Why is it needed?

src/serac/numerics/functional/quadrature_data.hpp
```diff
@@ -44,6 +44,59 @@ struct Nothing {
 struct Empty {
 };
 
+template <typename T>
+struct QuadratureData;
```
jamiebramwell, line 48: This file needs doxygen.

src/serac/numerics/functional/quadrature_data.hpp
jamiebramwell: Maybe `quadrature_storage_data_t`? Any other descriptive name would help.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -599,9 +598,11 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
   {
     residual_->AddBoundaryIntegral(
         Dimension<dim - 1>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
-        [this, traction_function](auto x, auto n, auto, auto, auto shape, auto... params) {
-          auto p = get<VALUE>(shape);
-          return -1.0 * traction_function(x + p, n, ode_time_point_, params...);
+        [this, traction_function](auto X, auto /* displacement */, auto /* acceleration */, auto shape, auto... params) {
+          auto x = X + shape;
+          auto n = cross(get<DERIVATIVE>(x));
+          auto area_correction = norm(n) / norm(cross(get<DERIVATIVE>(X)));
```
jamiebramwell, line 764: Some comments on this calculation would be good (e.g. "This scales the dA term in the boundary integral to use the shape displaced mesh coordinates. The `norm(cross(get<DERIVATIVE>(X)))` term is already applied in the `Functional` core, so it must be adjusted using the new `x` coordinate and Jacobian).

src/serac/physics/heat_transfer.hpp
```diff
@@ -433,10 +433,12 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
     residual_->AddBoundaryIntegral(
         Dimension<dim - 1>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
-        [this, flux_function](auto x, auto n, auto u, auto /* dtemp_dt */, auto shape, auto... params) {
-          auto p    = get<VALUE>(shape);
+        [this, flux_function](auto X, auto u, auto /* dtemp_dt */, auto shape, auto... params) {
           auto temp = get<VALUE>(u);
-          return flux_function(x + p, n, ode_time_point_, temp, params...);
+          auto x = X + shape;
+          auto n = cross(get<DERIVATIVE>(x));
+          auto area_correction = norm(n) / norm(cross(get<DERIVATIVE>(X)));
```
jamiebramwell, line 449: Same comment as the solid module about this calculation.

src/serac/numerics/functional/quadrature_data.hpp
```diff
@@ -54,54 +107,64 @@ struct Empty {
...
+
+    constexpr std::array geometries = {
+      mfem::Geometry::SEGMENT,
+      mfem::Geometry::TRIANGLE,
+      mfem::Geometry::SQUARE,
+      mfem::Geometry::TETRAHEDRON,
+      mfem::Geometry::CUBE
+    };
+
+    for (auto geom : geometries) {
```
jamiebramwell, line 130: This is a bit hard to read. Some comments as to what these structures are would help.

src/serac/numerics/functional/quadrature_data.hpp
```diff
@@ -54,54 +107,64 @@ struct Empty {
...
+      }
+    }
+
+  }
+
+  axom::ArrayView<T, 2> operator[](mfem::Geometry::Type geom) {
+    return axom::ArrayView<T, 2>(data.at(geom));
+  }
+
+  std::map< mfem::Geometry::Type, tmp_t > data;
```
jamiebramwell: What are the indicies of this `data` object? I assume [geometry_type][element_number][quadrature_index].

src/serac/physics/tests/beam_bending.cpp
samuelpmishLLNL: it was changed to be `auto` instead of the specific tensor type. Before this, we didn't support differentiation w.r.t. normal direction so that argument didn't need to be `auto`.

src/serac/physics/tests/lce_Brighenti_tensile.cpp
samuelpmishLLNL: We could give it a name if you like, although I prefer the clarity of seeing the two operations taking place.

src/serac/numerics/functional/quadrature_data.hpp
samuelpmishLLNL: oops, I forgot to pick names for some of the temporary types

src/serac/numerics/functional/quadrature_data.hpp
samuelpmishLLNL: I'll go back and rename the things in this class that I forgot to do originally. 

---------

To answer your question so you don't have to go reread the comments after I write them:

Before, meshes were only allowed to have one element geometry, so when we allocated memory for quadrature point data it was just a 2D array, with indices describing:
(which element, which qpt within that element).

Now, meshes are allowed to have different kinds of element geometries. So, when we allocate space for that quadrature point data, it's a 3D array indexed by:
(which geometry, which element with that geometry, which qpt within that element)

This 3D array is implemented as a `std::map` of 2D axom arrays, since the leftmost index is "sparse" (only a couple of the geometry types will be used for a given mesh). `tmp_t` was a (bad, temporary) name given to the 2D array before I changed it to use `axom::Array`.

src/serac/numerics/functional/quadrature_data.hpp
samuelpmishLLNL: it looks like I clobbered the doxygen comments when updating the special storage classes for `serac::Nothing` and `serac::Empty`, so I'll go add them back.

The specializations are there so that all quadrature data types (including `Nothing` and `Empty`) can reuse the same kernel template, but there is a fast path for those special cases, since they don't actually need to store / access memory.

src/serac/numerics/functional/detail/tetrahedron_H1.inl
```diff
@@ -418,7 +418,7 @@ struct finite_element<mfem::Geometry::TETRAHEDRON, H1<p, c> > {
 
           for (int k = 0; k < ndof; k++) {
             element_residual[j * step](i, k) +=
-                source * shape_function(xi, k) + dot(flux, shape_function_gradient(xi, k)) * wt;
+                (source * shape_function(xi, k) + dot(flux, shape_function_gradient(xi, k))) * wt;
```
jamiebramwell, line 421: Ahhhh now I see the problem. Nice find!

src/serac/numerics/functional/tests/functional_shape_derivatives.cpp
```diff
@@ -0,0 +1,363 @@
+// Copyright (c) 2019-2023, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file functional_shape_derivatives.cpp
```
jamiebramwell, line 8: This is a really nice test! Thank you for implementing it!

jamiebramwell (approved):

jamiebramwell:
Nice! While we wait for this to merge in, I might sketch up a pressure load PR based on this one for @kswartz92 to try.

src/serac/physics/tests/lce_Brighenti_tensile.cpp
btalamini: +1 for providing the function. But it doesn't need to be in this PR.

btalamini (approved):

PR #957, opened 2023-07-05
Remove the `SetElasticityOptions` call

jamiebramwell:
The `SetElasticityOptions` for the BoomerAMG preconditioner in the solid mechanics module was seg faulting on `HYPRE_BoomerAMGSetInterpRefine(amg_precond, interp_refine)` in a user code. This turns off that feature in favor of a simpler system size adjustment.

It also fixes a bug in the `ParameterizedLinearIsotropicSolid` material object and adds a test to ensure it compiles.


src/serac/physics/solid_mechanics.hpp
```diff
@@ -196,7 +196,11 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     // to be the displacement
     auto* amg_prec = dynamic_cast<mfem::HypreBoomerAMG*>(nonlin_solver_->preconditioner());
     if (amg_prec) {
-      amg_prec->SetElasticityOptions(&displacement_.space());
+      // amg_prec->SetElasticityOptions(&displacement_.space());
+
+      // The above call was seg faulting in Hypre in the HYPRE_BoomerAMGSetInterpRefine(amg_precond, interp_refine)
+      // method as of version v2.26.0. Instead, we just set the system size for Hypre.
+      amg_prec->SetSystemsOptions(dim, true);
```
samuelpmishLLNL: Is this change equivalent to the old line, or is it a temporary workaround while the segfault is figured out?

src/serac/physics/tests/solid.cpp
```diff
@@ -320,7 +320,7 @@ void functional_parameterized_solid_test(double expected_disp_norm)
   EXPECT_NEAR(expected_disp_norm, norm(solid_solver.displacement()), 1.0e-6);
 }
 
-TEST(SolidMechanics, 2DQuadParameterizedStatic) { functional_parameterized_solid_test<2, 2>(2.1906312704664623); }
+TEST(SolidMechanics, 2DQuadParameterizedStatic) { functional_parameterized_solid_test<2, 2>(2.1773851975471392); }
```
samuelpmishLLNL, line 474: any idea what's causing the ~1% change in solution norm?

samuelpmishLLNL (approved):

src/serac/physics/tests/solid.cpp
jamiebramwell: I believe it's because the AMG preconditioner isn't as accurate anymore and it's hard to force the linear solve to continue iterating past this point.

src/serac/physics/solid_mechanics.hpp
jamiebramwell: Temporary. I will add a TODO.

jamiebramwell:
I believe this is ready to go now.

white238 (approved):

PR #964, opened 2023-07-28
WIP: Heat transfer transient adjoint

jamiebramwell:
Proposed workflow:

1. Implement a parameterized dynamic heat conduction physics module.
2. Complete a full forward simulation.
3. Obtain previous states using the new `previousTemperature(...)` method.
4. Calculate a QoI based on the temperature history.
5. Compute a possibly time-varying adjoint load (d QOI/d state)
6. Call `reverseAdjointTimestep` to perform one backward Euler step for the adjoint solve using the timestep from the forward pass.
7. Call `computeTimestepSensitivity` to calculate the sensitivity (d QOI/d parameter) for the current timestep.

Some notes:

- Currently hardcoded to only use backward Euler time integrators.
- It reads the previous temperature values from disk at every timestep (no checkpointing).
- It changed the quasi-static function signature from `solveAdjoint` to `reverseAdjointTimestep` to be consistent with the forward pass.
- It updates Axom to include the `axom::MFEMSidreDataCollection` bugfix needed to correctly read an `mfem::ParGridFunction` from disk.


src/serac/physics/state/state_manager.cpp
```diff
@@ -92,6 +92,28 @@ double StateManager::newDataCollection(const std::string& name, const std::optio
...
+  std::string coll_name = mesh_name + "_datacoll";
+
+  axom::sidre::MFEMSidreDataCollection previous_datacoll(coll_name);
+
+  previous_datacoll.SetComm(states_to_load.begin()->get().mesh().GetComm());
+  previous_datacoll.SetPrefixPath(output_dir_);
+  previous_datacoll.Load(cycle_to_load);
+
+  for (auto state : states_to_load) {
+    SLIC_ERROR_ROOT_IF(collectionID(&states_to_load.begin()->get().mesh()) != mesh_name,
```
tupek2, line 109: Should this be state->get().mesh()?  Currently its just accesses .begin() for every iteration, so its not really checking anything.

src/serac/numerics/odes.hpp
```diff
@@ -163,6 +163,8 @@ class SecondOrderODE : public mfem::SecondOrderTimeDependentOperator {
    */
   const State& GetState() { return state_; }
 
+  TimestepMethod GetTimestepper() { return timestepper_; }
```
tepperly, line 171: Doxygen?

src/serac/physics/base_physics.cpp
```diff
@@ -45,10 +47,14 @@ void BasePhysics::setTime(const double time) { time_ = time; }
 
 double BasePhysics::time() const { return time_; }
 
+double BasePhysics::adjointTime() const { return adjoint_time_; }
```
tepperly, line 50: This is the time during the adjoint stepping?

src/serac/physics/base_physics.cpp
```diff
@@ -45,10 +47,14 @@ void BasePhysics::setTime(const double time) { time_ = time; }
...
+
 void BasePhysics::setCycle(const int cycle) { cycle_ = cycle; }
 
 int BasePhysics::cycle() const { return cycle_; }
 
+int BasePhysics::adjointCycle() const { return adjoint_cycle_; }
```
tepperly, line 56: What's the adjoint cycle concept? Is it an iteration count or what?

src/serac/physics/heat_transfer.hpp
```diff
@@ -650,39 +646,179 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+
+    SLIC_ERROR_ROOT_IF(ode_.GetTimestepper() != TimestepMethod::BackwardEuler,
+                       "Only backward Euler implemented for transient adjoint heat conduction.");
+
+    SLIC_ERROR_ROOT_IF(adjoint_cycle_ == 0,
+                       "Maximum number of adjoint timesteps exceeded! The number of adjoint timesteps must equal the "
+                       "number of forward timesteps");
+
+    bool is_first_adjoint_timestep = (adjoint_cycle_ == -1);
+    bool is_last_adjoint_timestep  = (adjoint_cycle_ == 1);
```
tepperly: Presumably these are const

tepperly (commented):
Is the end always defined by the final time? LiDO's API can potentially support other ways of ending. This could be managed in the LiDO wrapper.

src/serac/physics/base_physics.cpp
jamiebramwell: It's the iteration count on the backward pass. I added some extra docs, but let me know if this isn't clear.

src/serac/physics/base_physics.cpp
jamiebramwell: yes - I thought it was better to have this be a separate field vs. the forward pass, but I am open to suggestions.

jamiebramwell:
I changed the interface to reflect our discussion yesterday. The sensitivities are no longer accumulated internally. This means only the sensitivity increment is returned to a user for each timestep. I also added a field for the timestep size to be consistent with the forward pass. Let me know what you think!

src/serac/physics/base_physics.cpp
tupek2: Having them be different implies they can sometimes go out of sync?  What is the situation where the adjointTime and time would return different values?

jamiebramwell:
Closed in favor of https://github.com/LLNL/serac/pull/998

PR #972, opened 2023-08-03
Adding more functional and functional detail headers

chapman39:
an extension of this pr: https://github.com/LLNL/serac/pull/971


white238 (approved):

samuelpmishLLNL:
To catch this sort of issue in the future, I think we should have tests that mimic how downstream users bring Serac into their CMake projects. There are two idiomatic ways to bring in dependencies with CMake: `add_subdirectory` and `find_package`. Ideally, we should be support both.

samuelpmishLLNL (approved):

PR #977, opened 2023-08-14
Clarify the inputs/outputs of a forward simulation

### LINKED ISSUE #898: Clean up field information interfaces
It seems like we're using a lot of competing conventions for specifying field information in these interfaces:

- lookup mutable ref by name
   `FiniteElementState& state(const std::string& state_name)`
- lookup mutable ref by index
   `FiniteElementState& parameter(const size_t parameter_index)`
- lookup mutable ref by named functions
   `FiniteElementState& shapeDisplacement() { return shape_displacement_; }`
- separate getter/setter lookup by index:
   `void setParameter(const size_t parameter_index, FiniteElementState& parameter_state);`
   `void setShapeDisplacement(FiniteElementState& shape_displacement);`
- using `std::unordered_map` of names
   `const std::unordered_map<...> solveAdjoint(std::unordered_map<...> , std::unordered_map<...>);`

I think we should consider refactoring this eventually to make things consistent.

_Originally posted by @samuelpmishLLNL in https://github.com/LLNL/serac/issues/895#issuecomment-1466627334_

jamiebramwell:
This PR is the first set of changes needed to clarify and unify the input and output of each physics module. Specifically, it:

1. Removes the ability for the underlying time integrator to adaptively change the timestep without user intervention.
2. Removes non-const accessors for the underlying state (e.g. temperature and displacement) and parameter fields.
3. Changes the data ownership model for parameters. Parameters are now only owned by the physics modules and must be updated using the `setParameter` method.
4. Adds finite element family constructors for `FiniteElementVector`s, similar to how it was handled in `Functional`.
5. Removes the ability to manually set a current simulation cycle and time.
6. Removes the concept of a "default" mesh. Instead, mesh tags are now required for physics module construction.
6. Uses the more descriptive name `outputStateToDisk` for the writing to file operation.
7. General improvements to the doxygen documentation.

All of these changes are open to discussion. Additionally as this will be an interface breaking change, should we deprecate but leave the old function signatures or break the interface in a single PR? Comments and thoughts are welcome!

Resolves https://github.com/LLNL/serac/issues/898


src/serac/physics/base_physics.cpp
```diff
@@ -69,7 +67,7 @@ std::unique_ptr<FiniteElementState> BasePhysics::generateParameter(const std::st
   return new_state;
 }
 
-void BasePhysics::setParameter(size_t parameter_index, FiniteElementState& parameter_state)
+void BasePhysics::registerParameter(size_t parameter_index, const FiniteElementState& parameter_state)
```
samuelpmishLLNL: why do we register parameters and set other kinds of input?

src/serac/physics/base_physics.hpp
```diff
@@ -58,38 +58,31 @@ class BasePhysics {
...
+   * @return The current forward-solution time
    */
   virtual double time() const;
 
   /**
-   * @brief Set the current cycle
+   * @brief Get the current forward-solution cycle iteration number
    *
-   * @param[in] cycle The cycle
+   * @return The current forward-solution cycle iteration number
```
samuelpmishLLNL, line 78: does this imply that cycle has no meaning in the adjoint pass?

src/serac/physics/base_physics.hpp
```diff
@@ -175,22 +167,25 @@ class BasePhysics {
...
   /**
-   * @brief Get the shape displacement of the associated mesh for this physics object
+   * @brief Get a mutable reference the shape displacement of the associated mesh for this physics object
    *
    * @return The associated shape displacement
+   *
+   * @note If the FiniteElementState returned by this function is modified, the updated shape displacement will be used
+   * in the physics module.
    */
   FiniteElementState& shapeDisplacement() { return shape_displacement_; }
```
samuelpmishLLNL: maybe we don't want this to be a mutable reference for encapsulation reasons

samuelpmishLLNL (commented):
I'd like if the solveAdjoint interface were more consistent with the rest of serac's interfaces.

Right now for the forward pass, serac is asking users to:
1. specify inputs with named setters
2. invoke advanceTimestep()
3. call named getters to see updated solution

but on the backward pass, the input/output specification is different:
1. aggregate all the inputs to `solveAdjoint` into a std::map (strings -> references to data)
2. call solveAdjoint
3. some of the outputs are returned directly from solveAdjoint in a std::map, some require getters (e.g. computeShapeSensitivity)

I think it would be clearer if we did one (setters/getters) or the other (arguments) for inputs/outputs, but not both. For example:

```cpp
// setup
...
simulation.completeSetup();

// forward
simulation.setParameter(myParameter(t));
simulation.setTimestep(...);
simulation.advanceTimestep();
qoi += func(simulation.getDisplacements());

...

// backward
simulation.setAdjointLoads(...);
simulation.solveAdjoint();
simulation.getShapeSensitivity();
```

xor

```cpp
// setup
...
simulation.completeSetup();

// forward
auto & u_next = simulation.advanceTimestep(dt, {"param1", myParameter(t)});
qoi += func(u_next);

...

// backward
auto & adj_u = simulation.solveAdjoint({{"displacement", adjoint_load}});
simulation.getShapeSensitivity(adj_u);
```

I lean toward setters/getters, since the user may not always want all of the outputs of advanceTimestep/solveAdjoint

src/serac/physics/base_physics.hpp
jamiebramwell: Yes - although the question remains of should the adjoint cycle and forward sim cycle be the same. I think I can argue that both ways.

src/serac/physics/base_physics.cpp
jamiebramwell: The problem is that I wrote this so a physics module doesn't own the parameters - a user does. This avoids a copy of a piece of field data, but it does make memory ownership patterns more irregular. I will sketch what it would look like if we just owned the parameters and set them by value. The cleaner interface might be worth the memory hit.

src/serac/physics/base_physics.hpp
sethwatts: To clarify, does "cycle" mean iteration count in the time-stepping, or does it have a more general sense?

tupek2: I'd like to unify the adjoint cycle/cycle, adjoint timestep/timestep, etc.  We can then have a method to ask for the final solution, or last cycle, or whatever if we need to re-fetch end of forward time integration data.

src/serac/physics/base_physics.cpp
tupek2: Hey, I'd like to push on this so we get the discussion moving again.  Seems like there are a variety of design decisions to be make regarding a more general interface (also, support for parameters which are quadrature fields, adjoint loads which are d_qoi_d_some_rate, etc.).  Maybe we can have a meeting with a small-ish group to make some initial decisions?

tupek2: After some thought, I might actually lean toward the second solution outlined.  The reason: after advanceTimestep is called, how does the user know which states are valid to now 'get'.  Can the user get the adjoint_temperature?  Because if they do, it will be invalid.  Alternatively, if the return of advanceTimestep is a class/struct will a smaller set of getters (or member variables), then it is clear what fields are computed during advanceTimestep.  Similarly, solveAdjoint could return a struct with handles to adjoint relative fields, but probably also checkpointed primal fields as well.  This again makes it clear what are the computed outputs from this operation.

samuelpmishLLNL: I spoke with @tupek2  yesterday afternoon about some of the ideas he suggested above, and one of the takeaways for me was that the current way we're managing the material-specific internal variables at each quadrature point is clumsy and complicating some of the interface choices. So, I'll look at coming up with a different container that can hopefully address those issues, specifically:
- no more shared ownership of quadrature data
- relaxing type safety enough to make specifying quadrature data as inputs/outputs simpler

tupek2: Also, I was going to try prototyping a slightly more 'functional' interface (in the sense of LISP, ocaml, jax, haskell, etc.) to the advanceTimestep, adjoint calls.  Ideally this will allow us to pass in the same arguments 2 times and get the same result both times!  This is also a useful first step if we plan to do dynamic checkpointing down the line, and I think will make it easier to reason about the advance logic and adjoint math.

tupek2 (approved):
All looks good to me.  Maybe with the exception of adding yet another setter (for the timestep).  I'd prefer to pass that into the advance, and keep reducing the number of internal states in the physics classes.  But, I think that can be done in a subsequent commit when we handle varying time-steps for the reverse mode transient problems.

jamiebramwell:
I believe this is ready for another look @samuelpmishLLNL @btalamini @tupek2 @kswartz92 @sethwatts . Let me know what you think! The easiest way to ingest this giant PR is probably to start with the test examples. I'd be happy to set up a meeting where we can walk through that if it would help.

src/serac/numerics/functional/functional.hpp
```diff
@@ -224,7 +224,8 @@ class Functional<test(trials...), exec> {
     }
 
     for (uint32_t i = 0; i < num_trial_spaces; i++) {
-      P_trial_[i] = trial_space_[i]->GetProlongationMatrix();
+      auto ts     = trial_space_[i];
+      P_trial_[i] = ts->GetProlongationMatrix();
```
samuelpmishLLNL: why break this into two lines?

src/serac/physics/base_physics.cpp
```diff
@@ -108,7 +91,7 @@ void BasePhysics::setShapeDisplacement(FiniteElementState& shape_displacement)
   shape_displacement_ = shape_displacement;
 }
 
-void BasePhysics::outputState(std::optional<std::string> paraview_output_dir) const
+void BasePhysics::outputStateToDisk(std::optional<std::string> paraview_output_dir) const
```
samuelpmishLLNL, line 91: I like the clearer function name here-- in a future PR could we consider renaming `FiniteElementState` and `material_model::State` to avoid having so many unrelated abstractions go by the name `State`? 

e.g. `FiniteElementState`->`Field`, `material_model::State`->`material_model::internal_variables`

src/serac/physics/base_physics.hpp
```diff
@@ -58,158 +67,136 @@ class BasePhysics {
...
-  virtual void setTime(const double time);
-
-  /**
-   * @brief Get the current time
+   * @brief Get the current forward-solution time
    *
-   * @return The current time
+   * @return The current forward-solution time
    */
   virtual double time() const;
```
samuelpmishLLNL, line 73: Is there a way to get the time in the adjoint pass? Would a user ever need to get the time in the adjoint pass?

tupek2 (approved):

samuelpmishLLNL (approved):

src/serac/physics/base_physics.cpp
tupek2: I agree!  And the name here outputStateToDisk seems to be the 'best' usage of the term 'state'.

src/serac/physics/base_physics.hpp
tupek2: Well, the user may need the time-step (if they forgot to store it off somewhere).  My slow moving PR for solid transient adjoints includes that ability, for what its worth.

src/serac/physics/base_physics.cpp
tupek2: maybe InternalVariables (I'm trying to see if we have a capitalization, _ naming style set).  Also, InternalVariableArray may be better.  Class names should basically never be plural.  Because then people make std::vector (InternalVariables) and have to say 'internal variables'-es.  I got in trouble with this naming some years ago writing a simd library, and everyone had a good laugh at me for a few years cause it was impossible to talk about in conversations.

tupek2: Sorry, last suggestion, I promise:

InternalVariableField

src/serac/numerics/functional/functional.hpp
jamiebramwell: I believe this is leftover from debugging. I will revert it.

src/serac/physics/base_physics.cpp
jamiebramwell: I'll put up an issue on this because I think these objects now have enough usage to come up with better names.

jamiebramwell:
My current plan is to clean up the `adjoint` interfaces in a follow-on PR, so I'll leave the cycle/time stuff for now in the interest of not making this PR bigger than it already is. Thank you for the reviews!

PR #978, opened 2023-08-16
Remove expression templates

jamiebramwell:
The expression template implementation was causing issues when interacting with the dual number class due to templated operator overloads. As a simpler implementation is possible and we would like consistent finite element space definitions, this removes the current expression template capability. It was mostly used for testing purposes.


src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -270,7 +270,7 @@ axom::Array<DoF, 2, axom::MemorySpace::Host> GetElementRestriction(const mfem::F
     return axom::Array<DoF, 2, axom::MemorySpace::Host>{};
   } else {
     uint64_t                                     dofs_per_elem = elem_dofs.size() / n;
-    axom::Array<DoF, 2, axom::MemorySpace::Host> output({n, dofs_per_elem});
+    axom::Array<DoF, 2, axom::MemorySpace::Host> output(n, dofs_per_elem);
```
jamiebramwell:
```suggestion
    axom::Array<DoF, 2, axom::MemorySpace::Host> output({n, dofs_per_elem});
```

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -381,7 +381,7 @@ axom::Array<DoF, 2, axom::MemorySpace::Host> GetFaceDofs(const mfem::FiniteEleme
     return axom::Array<DoF, 2, axom::MemorySpace::Host>{};
   } else {
     uint64_t                                     dofs_per_face = face_dofs.size() / n;
-    axom::Array<DoF, 2, axom::MemorySpace::Host> output({n, dofs_per_face});
+    axom::Array<DoF, 2, axom::MemorySpace::Host> output(n, dofs_per_face);
```
jamiebramwell:
```suggestion
    axom::Array<DoF, 2, axom::MemorySpace::Host> output({n, dofs_per_face});
```

samuelpmishLLNL (approved):

src/serac/numerics/functional/element_restriction.cpp
samuelpmishLLNL: out of curiosity, why does this change keep showing up in every pr and getting reverted?

jamiebramwell: I updated axom in my local environment at some point (but haven't pushed that changed through a PR). The initializer list constructor for `axom::Array` no longer exists in develop.

white238 (approved):

PR #982, opened 2023-08-18
Contact enforcement with Tribol

ebchin:
Serac implementation of enforcing contact constraints using Tribol's `SINGLE_MORTAR` method.  Based on code by @jamiebramwell.  `SINGLE_MORTAR` is designed for Lagrange multiplier enforcement; however, a penalty method is available using Tribol's constraint matrix and by updating pressures to be proportional to gaps.  Note, `SINGLE_MORTAR` does not contain a consistent tangent, so quadratic convergence in Newton iterations is not expected.

Includes:
- Multirank contact using the redecomp library
- Support for active sets
- Support for multiple contact constraints, each one capable of Lagrange multiplier or penalty enforcement (experimental)

Limitations:
- Only supports linear elements; higher-order meshes using Tribol's linear methods may be supported in the near future using MFEM's low-order refined transfer mapping
- KINSOL's line search can lead to NaNs when active set is enabled


data/meshes/beam-hex-with-contact-block.mesh
```diff
@@ -0,0 +1,119 @@
+MFEM mesh v1.0
```
jamiebramwell, line 1: A few comments about what these meshes represent would be nice.

src/serac/physics/contact/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    )
+
+set(contact_sources
+    contact_data.cpp
+    contact_pair.cpp
+    )
+
+set(contact_depends serac_infrastructure)
+blt_list_append(TO contact_depends ELEMENTS tribol IF TRIBOL_FOUND)
+blt_list_append(TO contact_depends ELEMENTS redecomp IF TRIBOL_FOUND AND ENABLE_MPI)
```
jamiebramwell, line 24: I don't think `ENABLE_MPI` is ever off. Any thoughts @white238 ?

src/serac/numerics/equation_solver.cpp
```diff
@@ -76,12 +76,16 @@ void SuperLUSolver::SetOperator(const mfem::Operator& op)
 
     for (int i = 0; i < row_blocks; ++i) {
       for (int j = 0; j < col_blocks; ++j) {
-        auto* hypre_block = const_cast<mfem::HypreParMatrix*>(
-            dynamic_cast<const mfem::HypreParMatrix*>(&block_operator->GetBlock(i, j)));
-        SLIC_ERROR_ROOT_IF(!hypre_block,
-                           "Trying to use SuperLU on a block operator that does not contain HypreParMatrix blocks.");
-
-        hypre_blocks(i, j) = hypre_block;
+        if (!block_operator->IsZeroBlock(i, j)) {
```
jamiebramwell: While it's fairly self-documenting, I would add a comment that this handles the case of zero blocks (like what happens in contact) appropriately.

src/serac/physics/tests/contact_beam.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  // Output the sidre-based plot files
+  solid_solver.outputState(paraview_name);
+
+  // Check the l2 norm of the displacement dofs
+  auto u_l2 = mfem::ParNormlp(solid_solver.displacement(), 2, MPI_COMM_WORLD);
+  EXPECT_NEAR(3.3257055635785537, u_l2, 1.0e-3);
+}
+
+// NOTE: if Penalty is first and Lagrange Multiplier is second, super LU gives a
```
jamiebramwell: This seems weird. Does that mean that superlu is saving some internal state even though we destruct the solver?

src/serac/physics/tests/contact_patch.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+  // Perform the quasi-static solve
+  double dt = 1.0;
+  solid_solver.advanceTimestep(dt);
+
+  // Output the sidre-based plot files
+  solid_solver.outputState(paraview_name);
+
+  // Check the l2 norm of the displacement dofs
+  auto u_l2 = mfem::ParNormlp(solid_solver.displacement(), 2, MPI_COMM_WORLD);
```
jamiebramwell: I realize this is my old stuff, but it would be nice to add a more analytic test of the individual dof values. That should be pretty easy to do for this patch test.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -836,29 +838,102 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+          // gradient of residual function
+          [this](const mfem::Vector& u) -> mfem::Operator& {
+            auto [r, drdu] = (*residual_)(differentiate_wrt(u), zero_, shape_displacement_,
+                                          *parameters_[parameter_indices].state...);
+            J_             = assemble(drdu);
+            J_e_           = bcs_.eliminateAllEssentialDofsFromMatrix(*J_);
+            return *J_;
+          });
+    } else {
+      return std::make_unique<mfem_ext::StdFunctionOperator>(
```
jamiebramwell: I would add a comment about the extra degrees of freedom introduced by the contact operator. This explains why the `StdFunctionOperator` needs to be constructed in a different way. We should think about if it is possible that these can be combined into a single call. Maybe some default values `num_pressure_dofs = 0` for the no contact case could help?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -836,29 +838,102 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+
+  /**
+   * @brief Add a mortar contact boundary condition
+   *
+   * @param pair_id Unique identifier for the ContactPair
+   * @param bdry_attr_surf1 MFEM boundary attributes for the first surface
+   * @param bdry_attr_surf2 MFEM boundary attributes for the second surface
+   * @param contact_opts Defines contact method, enforcement, type, and penalty
+   */
+  void addContactPair(int pair_id, const std::set<int>& bdry_attr_surf1, const std::set<int>& bdry_attr_surf2,
```
jamiebramwell: I would add a comment that this is only implemented for quasi-static simulations. A SLIC_ERROR checking for this would also be a good idea.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -896,13 +971,20 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+    }
+
     if (is_quasistatic_) {
       residual_with_bcs_ = buildQuasistaticOperator();
 
       // the residual calculation uses the old stiffness matrix
       // to help apply essential boundary conditions, so we
       // compute J here to prime the pump for the first solve
-      residual_with_bcs_->GetGradient(displacement_);
+      // TODO (EBC): looks like this is done when we call quasiStaticSolve()?
```
jamiebramwell: Yes - I think we can delete this since you moved it to that call. I'll check the logic to be sure...

src/serac/physics/solid_mechanics.hpp
```diff
@@ -896,13 +971,20 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     // Build the dof array lookup tables
     displacement_.space().BuildDofToArrays();
 
+    // create contact mesh and compute forces, pressures, and Jacobians
+    if (contact_.haveContactPairs()) {
+      double dt = 0.0;
+      contact_.update(0, 0.0, dt);
```
jamiebramwell: A comment on what the zero arguments are would be nice.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -958,10 +1040,18 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
-    auto [r, drdu] = (*residual_)(differentiate_wrt(displacement_), zero_, shape_displacement_,
-                                  *parameters_[parameter_indices].state...);
-    J_             = assemble(drdu);
-    J_e_           = bcs_.eliminateAllEssentialDofsFromMatrix(*J_);
+    if (!contact_.haveLagrangeMultipliers()) {
+      residual_with_bcs_->GetGradient(displacement_);
+    } else {
+      int               disp_size = displacement_.Size();
+      mfem::Array<int>  block_offsets({0, disp_size, disp_size + contact_.numPressureTrueDofs()});
+      mfem::BlockVector augmented_solution(block_offsets);
```
jamiebramwell: Whenever you create a block vector, I would add some comments about the memory semantics used. E.g. This block vector contains references to the independently owned and managed displacement and pressure fields.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -1004,12 +1094,37 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+
+      lin_solver.Mult(dr_, du_);
+      displacement_ += du_;
+
+      nonlin_solver_->solve(displacement_);
+    } else {
+      lin_solver.SetOperator(*J_contact_);
+
+      int               disp_size = displacement_.Size();
+      mfem::Array<int>  block_offsets({0, disp_size, disp_size + contact_.numPressureTrueDofs()});
```
jamiebramwell: A comment block showing the linearized system of equations that you are solving would be good.

src/serac/physics/contact/contact_data.hpp
```diff
@@ -0,0 +1,199 @@
...
+  /**
+   * @brief Have there been contact pairs added?
+   *
+   * @return true if contact pairs have been added
+   * @return false if there are no contact pairs
+   */
+  bool haveContactPairs() const;
+
+  /**
+   * @brief Get the contact constraint residual (i.e. nodal forces)
```
jamiebramwell: Are these always in true-vector format? If so, i would just call them `contactForces` and return a `mfem::HypreParVector`, or maybe a `serac::FiniteElementDual`.

src/serac/physics/contact/contact_data.hpp
```diff
@@ -0,0 +1,199 @@
...
+   * @return Nodal contact forces as a Vector
+   */
+  mfem::Vector trueContactForces() const;
+
+  /**
+   * @brief Returns pressures on Lagrange multiplier true degrees of freedom
+   *
+   * @return Pressure true degrees of freedom as a Vector
+   */
+  mfem::Vector truePressures() const;
```
jamiebramwell: Same comment, but `mfem::HypreParVector` or `serac::FiniteElementState`

src/serac/physics/contact/contact_data.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+  /**
+   * @brief Returns nodal gaps on true degrees of freedom
+   *
+   * @return Nodal gaps as a Vector
+   */
+  mfem::Vector trueGaps() const;
```
jamiebramwell: `FiniteElementDual`?

src/serac/physics/contact/contact_data.hpp
```diff
@@ -0,0 +1,199 @@
...
+
+  /**
+   * @brief Returns a 2x2 block Jacobian on displacement/pressure degrees of
+   * freedom from contact constraints
+   *
+   * @return Pointer to block Jacobian (2x2 BlockOperator of HypreParMatrix)
+   */
+  std::unique_ptr<mfem::BlockOperator> contactJacobian() const;
```
jamiebramwell: Is the Jacobian actually computed here or is it just handed off? I would add a comment (and maybe rename it `computeContactJacobian` if that's what's happening.

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,294 @@
...
+    have_lagrange_multipliers_ = true;
+    num_pressure_true_dofs_ += pairs_.back().numPressureTrueDofs();
+  }
+}
+
+bool ContactData::haveContactPairs() const { return !pairs_.empty(); }
+
+void ContactData::update(int cycle, double time, double& dt)
+{
+  tribol::updateMfemParallelDecomposition();
```
jamiebramwell, line 46: Is this the active set search?

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,294 @@
...
+mfem::Vector ContactData::trueGaps() const
+{
+  mfem::Vector g_true(numPressureTrueDofs());
+  auto         dof_offsets = pressureTrueDofOffsets();
+  for (size_t i{0}; i < pairs_.size(); ++i) {
+    if (pairs_[i].getContactOptions().enforcement == ContactEnforcement::LagrangeMultiplier) {
+      mfem::Vector g_pair_true;
+      g_pair_true.MakeRef(g_true, dof_offsets[static_cast<int>(i)],
+                          dof_offsets[static_cast<int>(i) + 1] - dof_offsets[static_cast<int>(i)]);
+      // NOTE: gaps are considered a dual field in MFEM and the correct operator
```
jamiebramwell: @srwopschall , is this the correct operator to use? I vaguely remember you having a similar issue before.

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,294 @@
...
+  }
+  return g_true;
+}
+
+std::unique_ptr<mfem::BlockOperator> ContactData::contactJacobian() const
+{
+  jacobian_offsets_    = mfem::Array<int>({0, reference_nodes_->ParFESpace()->GetTrueVSize(),
+                                        numPressureTrueDofs() + reference_nodes_->ParFESpace()->GetTrueVSize()});
+  auto block_J         = std::make_unique<mfem::BlockOperator>(jacobian_offsets_);
+  block_J->owns_blocks = true;
```
jamiebramwell, line 109: Why is this set to true and then immediately set to false below?

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,294 @@
...
+          block_J->SetBlock(0, 0, &pair_J->GetBlock(0, 0));
+        } else {
+          block_J->SetBlock(0, 0,
+                            mfem::Add(1.0, static_cast<mfem::HypreParMatrix&>(block_J->GetBlock(0, 0)), 1.0,
+                                      static_cast<mfem::HypreParMatrix&>(pair_J->GetBlock(0, 0))));
+        }
+        delete &pair_J->GetBlock(0, 0);
+      }
+    }
+    if (!pair_J->IsZeroBlock(1, 0)) {
```
jamiebramwell: I have to admit this is fairly hard to read. Can you add some comments to help walk through what you're doing?

src/serac/physics/contact/contact_pair.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "tribol/interface/mfem_tribol.hpp"
+
+namespace serac {
+
+ContactPair::ContactPair(int pair_id, const mfem::ParMesh& mesh, const std::set<int>& bdry_attr_surf1,
+                         const std::set<int>& bdry_attr_surf2, const mfem::ParGridFunction& current_coords,
+                         ContactOptions contact_opts)
+    : pair_id_{pair_id}, contact_opts_{contact_opts}, current_coords_{current_coords}
+{
+  tribol::registerMfemCouplingScheme(pair_id, 2 * pair_id, 2 * pair_id + 1, mesh, current_coords, bdry_attr_surf1,
```
jamiebramwell: What is the purpose of these `pair_id` modifiers?

src/serac/physics/contact/contact_pair.cpp
```diff
@@ -0,0 +1,119 @@
...
+    for (auto attr : bdry_attr_surf2) {
+      surf2_markers[attr - 1] = 1;
+    }
+    for (int e{0}; e < pressure_space.GetNE(); ++e) {
+      if (surf2_markers[pressure_space.GetAttribute(e) - 1]) {
+        mfem::Array<int> vdofs;
+        pressure_space.GetElementVDofs(e, vdofs);
+        for (int d{0}; d < vdofs.Size(); ++d) {
+          int k = vdofs[d];
+          if (k < 0) {
```
jamiebramwell: What is this index logic doing?

jamiebramwell (commented):
I think my biggest comment is that the function signatures returning `mfem::Vector` should be `serac::FIniteElementState/Dual` and we should remove `true` from the names as we always work on the MFEM "true dofs".

src/serac/physics/tests/contact_beam.cpp
ebchin: Possibly.  It looks like Serac is destructing everything properly; the `SolidMechanics` object is in the `TEST_P` scope, and it holds the `unique_ptr` to the equation solver which holds the `unique_ptr` to superlu.  I'll see if I can find something weird going on in superlu itself.

data/meshes/beam-hex-with-contact-block.mesh
```diff
@@ -0,0 +1,122 @@
...
+
+# a unit cube occupying [6.25,7.25]x[0,1]x[1,2] and an 8 element beam
+# occupying [0,8]x[0,1]x[0,1] where each element is a unit cube
```
samuelpmishLLNL, line 13:
```suggestion
#
#  z                        o---o
#  ^                        |   |
#  |                        o---o
#
#  o---o---o---o---o---o---o---o---o
#  |   |   |   |   |   |   |   |   |
#  o---o---o---o---o---o---o---o---o    ---> x
#
# a unit cube occupying [6.25,7.25]x[0,1]x[1,2] and an 8 element beam
# occupying [0,8]x[0,1]x[0,1] where each element is a unit cube
```

src/serac/physics/contact/contact_data.hpp
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * @brief Current coordinates of the mesh
+   */
+  mfem::ParGridFunction current_coords_;
+
+  /**
+   * @brief The contact boundary condition information
+   */
+  std::vector<ContactPair> pairs_;
+#endif
```
samuelpmishLLNL: Do these members need to be `#ifdef`ed out?

src/serac/physics/contact/contact_data.hpp
```diff
@@ -0,0 +1,202 @@
...
+                                                .type        = ContactType::Frictionless,
+                                                .penalty     = 1.0e3};
+
+}  // namespace contact
+
+/**
+ * @brief This class stores contact pair data and interacts with Tribol
+ */
+class ContactData {
+public:
```
samuelpmishLLNL: if the purpose of this class is to interact with Tribol, should we just name it "TribolInterface" or similar? `ContactData`, `ContactPairs` and `ContactConfig` all sound like the same thing.

src/serac/physics/contact/contact_data.hpp
```diff
@@ -0,0 +1,202 @@
...
+   * @return Pressure finite element state
+   */
+  mfem::Vector mergedPressures() const;
+
+  /**
+   * @brief Returns nodal gaps on true degrees of freedom
+   *
+   * @return Nodal gaps as a Vector
+   */
+  mfem::Vector mergedGaps() const;
```
samuelpmishLLNL: what kind of mfem::Vector? T/L/E, ordering, polynomial order, etc

are the nodal gaps vector-valued or just the normal component?

src/serac/physics/contact/contact_pair.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * @brief Container for all interactions with the Tribol contact library
+ *
+ * This class provides a wrapper for all of the information needed by Tribol to
+ * generate the mortar weights for the current penalty contact formulation. All
+ * of the secondary nodes are currently considered active and therefore we are
+ * enforcing an equality constraint.
+ **/
+class ContactPair {
```
samuelpmishLLNL: conceptually, what's the difference between `ContactPair` and `ContactData`?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -836,29 +838,102 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
-          auto [r, drdu] =
-              (*residual_)(differentiate_wrt(u), zero_, shape_displacement_, *parameters_[parameter_indices].state...);
-          J_   = assemble(drdu);
-          J_e_ = bcs_.eliminateAllEssentialDofsFromMatrix(*J_);
-          return *J_;
-        });
+            return *J_contact_;
+          });
+    }
+  }
```
samuelpmishLLNL: is it feasible to encapsulate these contact-specific solver modifications into one of the contact classes, rather than inlining it in the solid_mechanics module?

ebchin:
Thanks for the reviews @samuelpmishLLNL and @jamiebramwell! With the exception of some likely unrelated tolerance issues in `functional_shape_derivatives`, I think all of your comments have been addressed with the latest commits.

src/serac/physics/contact/contact_data.hpp
ebchin: The only member that needs to be `#ifdef`'d out is `pairs_`, since the entire `ContactPair` class is `#ifdef`'d out if built without Tribol.  I `#ifdef`'d the other members to deal with compiler warnings that the members aren't used when built without Tribol.

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,367 @@
...
+  // This function computes forces, gaps, and Jacobian contributions based on the current field quantities. Note the
+  // fields (with the exception of pressure) are stored on the redecomposed surface mesh until transferred by calling
+  // forces(), mergedGaps(), etc.
+  tribol::update(cycle, time, dt);
+}
+
+FiniteElementDual ContactData::forces() const
+{
+  FiniteElementDual f(*reference_nodes_->ParFESpace(), "contact force");
+  for (const auto& interaction : interactions_) {
```
tupek2, line 56: Is this this pulling forces from the parallel contact decomp?  And if so, 1 MPI round per interaction?

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,367 @@
...
+    p_blk.MakeRef(const_cast<mfem::Vector&>(u), disp_size, numPressureDofs());
+
+    mfem::Vector r_blk, g_blk;
+    r_blk.MakeRef(r, 0, disp_size);
+    g_blk.MakeRef(r, disp_size, numPressureDofs());
+
+    double dt = 1.0;
+    setDisplacements(u_blk);
+    // we need to call update first to update gaps
+    update(1, 1.0, dt);
```
tupek2: this function seems to do a lot? updates gaps, then updates other stuff after pressure is computed?  Can it be broken up into its different responsibilities and then named more concretely?

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,367 @@
...
+    // with updated gaps, we can update pressure for contact interactions with penalty enforcement
+    setPressures(p_blk);
+    // call update again with the right pressures
+    update(1, 1.0, dt);
+
+    orig_r(u_blk, r_blk);
```
tupek2: what is this doing?

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,367 @@
...
+  };
+}
+
+void ContactData::setPressures(const mfem::Vector& merged_pressures) const
+{
+  auto dof_offsets = pressureDofOffsets();
+  for (size_t i{0}; i < interactions_.size(); ++i) {
+    FiniteElementState p_interaction(interactions_[i].pressureSpace());
+    if (interactions_[i].getContactOptions().enforcement == ContactEnforcement::LagrangeMultiplier) {
+      mfem::Vector p_interaction_ref(const_cast<mfem::Vector&>(merged_pressures), dof_offsets[static_cast<int>(i)],
```
tupek2: why is there a const cast?  are merged_pressures changing in here?  If so, pass non-const?

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,367 @@
...
+    g_blk.Set(1.0, mergedGaps());
+  };
+}
+
+std::function<std::unique_ptr<mfem::BlockOperator>(const mfem::Vector&)> ContactData::jacobianFunction(
+    std::function<std::unique_ptr<mfem::HypreParMatrix>(const mfem::Vector&)> orig_J) const
+{
+  return [this, orig_J](const mfem::Vector& u) -> std::unique_ptr<mfem::BlockOperator> {
+    mfem::Vector u_blk;
+    u_blk.MakeRef(const_cast<mfem::Vector&>(u), 0, reference_nodes_->ParFESpace()->GetTrueVSize());
```
tupek2: please consider alternative without a const_cast? const_casts are slippery slopes

src/serac/physics/contact/contact_data.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+    auto J_contact         = std::make_unique<mfem::BlockOperator>(jacobian_offsets_);
+    J_contact->owns_blocks = true;
+    J_contact->SetBlock(0, 0, J.release());
+
+    return J_contact;
+  };
+}
+
+void ContactData::setPressures([[maybe_unused]] const mfem::Vector& true_pressures) const {}
```
tupek2, line 359: truePressure for naming consistency?

ebchin:
The contact-specific stuff has been moved to `solid_mechanics_contact.hpp`.  Let me know if this is preferable or if I should revert it back.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -903,12 +903,6 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
 
     if (is_quasistatic_) {
       residual_with_bcs_ = buildQuasistaticOperator();
-
-      // the residual calculation uses the old stiffness matrix
-      // to help apply essential boundary conditions, so we
-      // compute J here to prime the pump for the first solve
-      residual_with_bcs_->GetGradient(displacement_);
-
```
tupek2, line 1020: Do we want to remove this?  Seems unrelated to contact, maybe?

src/serac/physics/solid_mechanics_contact.hpp
```diff
@@ -0,0 +1,353 @@
...
+            // create block operator for constraints
+            J_constraint_e_ = std::make_unique<mfem::BlockOperator>(J_offsets_);
+            J_constraint_e_->SetBlock(0, 0, J_e_.get());
+            J_constraint_e_->SetBlock(1, 0, J_e_21_.get());
+
+            J_operator_ = J_constraint_.get();
+            return *J_constraint_;
+          });
+    } else {
+      return std::make_unique<mfem_ext::StdFunctionOperator>(
```
tupek2, line 165: Just reading quickly, i dont understand this case.  is it penalty, or some case where there are no multipliers on processor?  Maybe create two functions for the 2 cases with names making it clear what the two cases are doing?

src/serac/physics/solid_mechanics_contact.hpp
```diff
@@ -0,0 +1,353 @@
...
+    for (int i = 0; i < constrained_dofs.Size(); i++) {
+      int j = constrained_dofs[i];
+      du_[j] -= displacement_(j);
+    }
+
+    dr_ = 0.0;
+    mfem::EliminateBC(*J_, *J_e_, constrained_dofs, du_, dr_);
+
+    // Update the initial guess for changes in the parameters if this is not the first solve
+    for (std::size_t parameter_index = 0; parameter_index < parameters_.size(); ++parameter_index) {
```
tupek2: Is this essentially the same code in the base class for doing warm start?  If so, can we create a function that does it once and reuse it?

ebchin: Yes, it _should_ be the same.  The problem is changing the size of the linear solver's operator to do the warm start on just the displacement dofs then changing it back to include the Lagrange multiplier's blocks was giving me an error with SuperLU.  To get around this, I do the warm start with contact including the Lagrange multipliers.

I don't know enough about SuperLU to know if changing the operator size triggering an error is the appropriate behavior and it doesn't look like we can copy the linear solver easily.  Any thoughts on how this might be handled better @jamiebramwell?

src/serac/physics/solid_mechanics_contact.hpp
ebchin: Added some comments to clarify what each case does

src/serac/physics/solid_mechanics.hpp
ebchin: It is unrelated -- just addressing the comment above: https://github.com/LLNL/serac/pull/982#discussion_r1307804104 .  It probably should be a different PR, but since I'm making other changes around this code path, I think it's OK to leave it in.

src/serac/physics/contact/contact_data.cpp
ebchin: Active set search is here: 
https://github.com/LLNL/serac/blob/516c43ca5b29ce8030e04aaaff1dc749527c24a2/src/serac/physics/contact/contact_interaction.cpp#L117-L133
It's called when determining values for gap/pressure/Jacobian

src/serac/physics/contact/contact_data.cpp
ebchin: Yes and yes

src/serac/physics/contact/contact_data.cpp
ebchin: Yes, but these changes will need to be made upstream in Tribol (see https://github.com/LLNL/Tribol/issues/19 there).  Once they have been made, the corresponding code in Serac will be updated in a future PR.

src/serac/physics/contact/contact_data.cpp
ebchin: It's so we can create a `mfem::Vector`view of `u`.  Unfortunately, there isn't a way to create a const `mfem::Vector` view.  I added a comment and first did the const cast to a `mfem::Vector` called `u_const` so it's a bit clearer.

ebchin:
@tupek2, @jamiebramwell, and @samuelpmishLLNL, I think all comments have been addressed.  Contact-specific theory documentation can be viewed here: https://serac.readthedocs.io/en/feature-chin23-contact_mechanics/sphinx/theory_reference/solid.html#contact-mechanics.  Let me know if anything else needs to be taken care of before merging.  Thanks!

tupek2 (approved):

samuelpmish (approved):

samuelpmishLLNL (approved):

src/docs/sphinx/theory_reference/solid.rst
```diff
@@ -389,3 +389,131 @@ isotropic and kinematic hardening
 .. image:: ../figures/J2_both.gif
    :width: 80%
    :align: center
+
```
jamiebramwell, line 392: This is great! Thank you for writing this up!

src/serac/physics/solid_mechanics_contact.hpp
jamiebramwell: We talked about this offline, but I've noticed odd memory issues with SuperLU. I recommend trying Strumpack to see if that resolves the issue.

src/serac/physics/contact/CMakeLists.txt
jamiebramwell: ping @white238

jamiebramwell (approved):
Excellent work @ebchin !

ebchin:
Took care of a few outstanding issues in the latest push:
- Switch to strumpack from SuperLU in the examples
- Enable KINSOL in examples with equality constraints
- Moved warm start to a method that is used in `SolidMechanics` and `SolidMechanicsContact`

src/serac/physics/solid_mechanics_contact.hpp
ebchin: Fixed now that we're using Strumpack!

PR #764, opened 2022-08-11
Remove bloat from our binaries

samuelpmishLLNL:
re: https://github.com/LLNL/serac/issues/761

It looks like by default, `uberenv` builds everything as `RelWithDebInfo`, so the TPLs produce huge binaries:

```sh
sam@provolone:~/code/serac_libs_before/gcc-9.3.0/conduit-0.8.3-wqhg2i6ady424csssfpjlggcjflt42jx/lib$ ls -la
...
-rw------- 1 sam sam  64230114 Aug  4 13:00 libconduit.a
-rw------- 1 sam sam 185510866 Aug  4 13:00 libconduit_blueprint.a
-rw------- 1 sam sam  41864046 Aug  4 13:01 libconduit_blueprint_mpi.a
-rw------- 1 sam sam  47961410 Aug  4 13:01 libconduit_relay.a
-rw------- 1 sam sam   4944286 Aug  4 13:00 libconduit_relay_mpi.a
-rw------- 1 sam sam  27381396 Aug  4 13:01 libconduit_relay_mpi_io.a
```

vs.

```
sam@provolone:~/code/serac_libs_after/gcc-9.3.0/conduit-0.8.3-6ilpkq7rsb23lavuks2wuv45wppjkbv4/lib$ ls -la
...
-rw------- 1 sam sam 3797592 Aug 11 12:44 libconduit.a
-rw------- 1 sam sam 5196542 Aug 11 12:45 libconduit_blueprint.a
-rw------- 1 sam sam 2186540 Aug 11 12:45 libconduit_blueprint_mpi.a
-rw------- 1 sam sam 2122044 Aug 11 12:45 libconduit_relay.a
-rw------- 1 sam sam  318550 Aug 11 12:44 libconduit_relay_mpi.a
-rw------- 1 sam sam 1543462 Aug 11 12:45 libconduit_relay_mpi_io.a
```

so about a ~20x reduction in TPL binary size. Linking against these also was increasing the size of our unit tests by more than 100MB+, each. 

Before: (Release)

```
-rwxrwxr-x  1 sam sam 140083584 Aug 10 17:08 serac_solid_functional_boundary
-rwxrwxr-x  1 sam sam 140348240 Aug 10 17:08 serac_solid_functional_finite_diff
-rwxrwxr-x  1 sam sam 141273528 Aug 10 17:08 serac_solid_reuse_mesh
-rwxrwxr-x  1 sam sam 141249672 Aug 10 17:08 serac_solid_sensitivity
[...]

PR #994, opened 2023-08-25
update MFEM, superlu-dist, and Tribol

ebchin:
Contact support needs this recent fix: https://github.com/mfem/mfem/commit/71fab657ec8c4a25c761338d9df3e0cdf5faf24e and will benefit from this recently merged PR: https://github.com/mfem/mfem/pull/3557 in MFEM.

This more recent version of MFEM does not need an old superlu-dist, so the version specified in the serac `package.py` has been updated.

Also, Tribol has been updated to the most recent develop.


ebchin:
I think this one is ready to go now

chapman39 (approved):

scripts/spack/packages/serac/package.py
```diff
@@ -137,8 +137,6 @@ class Serac(CachedCMakePackage, CudaPackage):
     depends_on("adiak+mpi", when="+profiling")
     depends_on("caliper+mpi+adiak~papi", when="+profiling")
 
-    depends_on("superlu-dist@6.1.1")
```
jamiebramwell, line 140: @white238 @chapman39 , do you remember why this was here? I can't figure out why we need it, but I'm sure there was a reason.

jamiebramwell (approved):

scripts/spack/packages/serac/package.py
chapman39: looking at the blame, it was added in this pr https://github.com/LLNL/serac/pull/764, its not clear why though

edit: this was added years ago actually

ebchin: Looks like the version number was added in [this commit](https://github.com/LLNL/serac/commit/14de70691423bd6d5c2fbbe7617f1e591bd3f24a) to lock down package version numbers.  With that in mind, I'll add it back and update the version to the one spack decided to build (8.1.2).

host-configs/docker/clang@10.0.0.cmake
```diff
@@ -4,9 +4,9 @@
...
 
-set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "")
+set(CMAKE_BUILD_TYPE "Release" CACHE STRING "")
```
ebchin, line 9: @white238 @chapman39 Should the docker builds be changed back to debug?

chapman39: i think they should be release but lets see what chris says

PR #995, opened 2023-08-30
Feature/tupek/transient adjoint

tupek2:


tupek2:
I seem to be having trouble setting the submodules back to their 'develop' versions.  Or, perhaps I need to pull specific a specific sha-1 for each submodule?

PR #998, opened 2023-09-01
Feature/tupek/transient adj

tupek2:
Implementation of reverse time integration for heat equation with fix timestep backward Euler. Only considering initial temperature, and shape sensitivities in this PR.  Other extensions will follow.


src/serac/physics/heat_transfer.hpp
```diff
@@ -649,34 +655,103 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
     for (const auto& bc : bcs_.essentials()) {
-      bc.apply(*J_T, adjoint_load_vector, adjoint_essential);
+      bc.apply(*J_T, modified_RHS, adjoint_essential);
     }
 
     lin_solver.SetOperator(*J_T);
-    lin_solver.Mult(adjoint_load_vector, adjoint_temperature_);
+    lin_solver.Mult(modified_RHS, adjoint_temperature_);
+
+    m_mat->Mult(adjoint_temperature_, _d_temperature_start_of_step_);
```
jamiebramwell: I assume this was the term I was missing? I haven't dug out my notes yet...

src/serac/physics/heat_transfer.hpp
```diff
@@ -686,24 +761,35 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
   }
 
+  /**
+   * @brief Compute the implicit sensitivity of the quantity of interest with respect to the initial temperature
+   *
+   * @return The sensitivity with respect to the initial temperature
+   *
+   * @pre `reverseAdjointTimestep` must be called as many times as the forward solver was advanced before this is called
+   */
+  const std::unordered_map<std::string, const serac::FiniteElementDual&> computeInitialConditionSensitivity() override
```
jamiebramwell, line 822: As this is technically available at any point during the reverse timestep process, is it worth documenting what it is when accessed in that state? @sethwatts , do you think you would ever use it?

src/serac/physics/heat_transfer.hpp
```diff
@@ -720,9 +806,17 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+  /**
+   * @brief Rate of change in temperature at the current adjoint timestep
+   */
+  FiniteElementState temperature_rate_;
+
+  /// The adjoint temperature finite element states, the multiplier on the residual for a given timestep
   serac::FiniteElementState adjoint_temperature_;
 
+  // The total sensitivity of the qoi with respect to the start of the previos timestep's temperature
+  serac::FiniteElementDual _d_temperature_start_of_step_;
```
jamiebramwell: Any reason for the leading underscore?

src/serac/physics/heat_transfer.hpp
```diff
@@ -771,16 +865,16 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
   mfem::Vector u_predicted_;
 
   /// Previous value of du_dt used to prime the pump for the nonlinear solver
-  mfem::Vector previous_;
+  mfem::Vector previous_u_dot_;
```
jamiebramwell: 👍

src/serac/physics/state/finite_element_vector.cpp
```diff
@@ -150,4 +150,19 @@ double min(const FiniteElementVector& fe_vector)
...
+{
+  SLIC_ERROR_IF(
+      v1.Size() != v2.Size(),
+      axom::fmt::format("Finite element vector of size '{}' can not inner product with another vector of size '{}'",
+                        v1.Size(), v2.Size()));
+  SLIC_ERROR_IF(v1.comm() != v2.comm(),
+                "Can not compute inner products between vectors with different mpi communicators");
+
+  double global_ip;
+  double local_ip = mfem::InnerProduct(v1, v2);
```
jamiebramwell, line 181: Is there a reason this doesn't work? https://github.com/mfem/mfem/blob/1ccb31fde641a96c9fd2920bc123c35eee31d476/linalg/hypre.hpp#L333 Is it resolving to the non-reduced `mfem::Vector` implementation if we don't force it?

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+  d_qoi_d_temperature = temperature;
+  d_qoi_d_temperature *= dt;
+}
+
+std::unique_ptr<HeatTransfer<p, dim>> create_heat_transfer(
+    const NonlinearSolverOptions& nonlinear_opts, const TimesteppingOptions& dyn_opts,
+    const heat_transfer::IsotropicConductorWithLinearConductivityVsTemperature& mat)
+{
+  // eventually figure out how to clear out cider state
```
jamiebramwell: Thank you for leaving this here - I will dig into why this isn't working.

src/serac/physics/materials/thermal_material.hpp
```diff
@@ -65,6 +65,62 @@ struct LinearIsotropicConductor {
...
 };
 
+/// Nonlinear isotropic thermal conduction material model
+struct IsotropicConductorWithLinearConductivityVsTemperature {
+  /**
+   * @brief Construct a Isotropic Conductor with Conductivity linear with Temparture object
+   *
+   * @param density Density of the material (mass/volume)
+   * @param specific_heat_capacity Specific heat capacity of the material (energy / (mass * temp))
+   * @param conductivity Thermal conductivity of the material at temp = 0 (power / (length * temp))
```
jamiebramwell: "Reference thermal conductivity..." maybe?

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  // Solver options
+  NonlinearSolverOptions nonlinear_opts{.relative_tol = 5.0e-13, .absolute_tol = 5.0e-13};
+  TimesteppingOptions    dyn_opts{.timestepper        = TimestepMethod::BackwardEuler,
+                               .enforcement_method = DirichletEnforcementMethod::DirectControl};
+  heat_transfer::IsotropicConductorWithLinearConductivityVsTemperature mat{1.0, 1.0, 1.0, 2.0};
+  TimeSteppingInfo                                                     tsInfo{.totalTime = 0.5, .num_timesteps = 4};
+};
+
+TEST_F(HeatTransferSensitivityFixture, InitialTemperatureSensitivities)
```
jamiebramwell, line 255: Is a static parameterization (e.g. sensitivity with respect to a parameterized conductivity) test just not implemented or is there something fundamentally broken with that workflow?

jamiebramwell (commented):
Nice work @tupek2 !

src/serac/physics/heat_transfer.hpp
sethwatts: If the question is, would we use this method, then yes, I could definitely see doing inverse problems where we are trying to deduce initial conditions, of find optimal ones t evolve to a desired end state. If the question is, would you ever try using it before working to the end adjoint time / zero primal time, I think the answer is no.

src/serac/physics/heat_transfer.hpp
tupek2: Lack of creativity.  I was trying to invent some convention that leading _ means implicit derivative of the qoi.  But, that was before I noticed you called such things implicit sensitivities, or implicit derivatives (which is a term I like, though I hadn't heard before).  So I'll improve soon!

src/serac/physics/heat_transfer.hpp
tupek2: Before getting back to time zero, that quantity would be the implicit sensitivity of the qoi with respect to the temperature at the start of the most recent cycle/timestep.  Not something one would commonly be interested in, but it does have an interpretation.

src/serac/physics/heat_transfer.hpp
tupek2: yeah, I kind of re-derived things, so I'm not exactly sure how things match up with what was before.  This extra implicit sensitivity field certainly helps me reason through the math easier.  I think I'll chuck in a ~4 line comment for the derivation to set notation, and so maybe future people can figure what in the world I'm up to here.

src/serac/physics/state/finite_element_vector.cpp
tupek2: It didn't call that one by default, it was calling the non-reduced one for sure.  Honestly, I didn't know that one existed!  Not always clear which vector in mfem has what implemented for it.  I think there is likely a way to force it to call that one (now that I know it exists).  But, this exercise has given me a new mission for serac!  I would love to convince everyone that our FiniteElementVector should have a HypreParVector, not be one.  There are many reasons for this, e.g., good dependency management through interfaces, ..., but mostly it allows us to control exactly how common math operations work in a consistent way, without having to always think about which vector versions is being invoked in the background depending on the input types.  If we do not agree with certain mfem conventions (like not overloading +/-), we can simply do that for our interface layer independently of any current or changing mfem decisions on such things.

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
tupek2: I will try to add an explicit test for this, for completeness.

src/serac/physics/heat_transfer.hpp
```diff
@@ -602,40 +605,43 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
    * @note If the essential boundary dual is not specified, homogeneous essential boundary conditions are applied to
    * the adjoint system
    *
+   * @param dt The size of the adjoint timestep to take in the backward time integration calculation
    * @param adjoint_loads An unordered map containing finite element duals representing the RHS of the adjoint equations
    * indexed by their name
```
samuelpmishLLNL: Not part of this PR, but what are the requirements for this unordered_map @jamiebramwell? Does it treat unspecified RHS as zero, an error, or something else?

src/serac/physics/heat_transfer.hpp
```diff
@@ -602,40 +605,43 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
     // Add the sign correction to move the term to the RHS
-    adjoint_load_vector *= -1.0;
+    temperature_adjoint_load_vector *= -1.0;
+
+    mfem::HypreParVector temperature_rate_adjoint_load_vector(temperature_adjoint_load_vector);
+    temperature_rate_adjoint_load_vector = 0.0;
+    if (temp_rate_adjoint_load != adjoint_loads.end()) {
+      temperature_rate_adjoint_load_vector = temp_rate_adjoint_load->second;
+      temperature_rate_adjoint_load_vector *= -1.0;
+    }
```
samuelpmishLLNL, line 661: Do we have any theory reference for the adjoint stuff? It's hard for me to see the underlying math here, and know which terms should be getting minus signs.

edit: I don't mean a textbook or something-- a document specifically for serac that our implementation refers back to, to make everything consistent in notation, sign conventions, etc

src/serac/physics/heat_transfer.hpp
```diff
@@ -649,34 +655,103 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+      // sensitivities.
+
+      auto [r, drdu] = (*residual_)(differentiate_wrt(temperature_), zero_, shape_displacement_,
+                                    *parameters_[parameter_indices].state...);
+      auto jacobian  = assemble(drdu);
+      auto J_T       = std::unique_ptr<mfem::HypreParMatrix>(jacobian->Transpose());
+
+      for (const auto& bc : bcs_.essentials()) {
+        bc.apply(*J_T, temperature_adjoint_load_vector, adjoint_essential);
+      }
```
samuelpmishLLNL, line 693: is `adjoint_essential` actually some sort of adjoint quantity, or is it the temperature values that were used in the essential b.c.s on the forward pass?

src/serac/physics/heat_transfer.hpp
```diff
@@ -649,34 +655,103 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+
     return {{"adjoint_temperature", adjoint_temperature_}};
   }
 
+  FiniteElementState previousTemperature(int cycle) const
+  {
+    FiniteElementState previous_temperature(temperature_);
+    StateManager::loadPreviousStates(cycle, {previous_temperature});
+    return previous_temperature;
+  }
```
samuelpmishLLNL: This name is potentially confusing, it's not returning the temperature before `cycle`, it's returning any temperature (past, present or future).

It also overlaps conceptually with the existing temperature getter, maybe we can consolidate these in the future.

src/serac/physics/heat_transfer.hpp
samuelpmishLLNL: Does it need its own member function then, or is it equivalent to just computing sensitivities after returning to cycle zero in the adjoint pass?

src/serac/physics/heat_transfer.hpp
samuelpmishLLNL: there are several different naming conventions for similar ideas 
`u` vs `temperature`, 
`_dot` suffix vs `_rate` suffix,
`previous_` prefix vs `_start_of_step` suffix

src/serac/physics/heat_transfer.hpp
```diff
@@ -720,9 +806,17 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
-  /// The adjoint temperature finite element state
+  /**
+   * @brief Rate of change in temperature at the current adjoint timestep
+   */
+  FiniteElementState temperature_rate_;
+
+  /// The adjoint temperature finite element states, the multiplier on the residual for a given timestep
   serac::FiniteElementState adjoint_temperature_;
 
+  // The total sensitivity of the qoi with respect to the start of the previos timestep's temperature
```
```suggestion
  // The total sensitivity of the qoi with respect to the start of the previous timestep's temperature
```

src/serac/physics/heat_transfer.hpp
btalamini: @samuelpmishLLNL, I think [these notes](https://cs.stanford.edu/~ambrad/adjoint_tutorial.pdf) are brief and clear. We do need to clearly document our convention in our API eventually - whether our adjoint load is the partial of the QOI wrt the primal, or its negation. We don't necessarily need to do that in this PR.

src/serac/physics/heat_transfer.hpp
sethwatts: > Not something one would commonly be interested in, but it does have an interpretation.

Agreed on both points.

src/serac/physics/heat_transfer.hpp
btalamini: This looks right to me. I think we could re-factor later to avoid some of the matvecs, but that's an optimization for the future.

src/serac/physics/materials/thermal_material.hpp
```diff
@@ -65,6 +65,62 @@ struct LinearIsotropicConductor {
...
+   * @tparam T2 Temperature type
+   * @tparam T3 Temperature gradient type
+   * @param[in] temperature_gradient Temperature gradient
+   * @return The calculated material response (tuple of volumetric heat capacity and thermal flux) for a linear
+   * isotropic material
+   */
+  template <typename T1, typename T2, typename T3>
+  SERAC_HOST_DEVICE auto operator()(const T1& /* x */, const T2& temperature, const T3& temperature_gradient) const
+  {
+    auto currentConductivity = conductivity_ + d_conductivity_d_temperature_ * temperature;
```
btalamini: I realize the constitutive model is not the point of this PR, but should we have a check here to make sure the conductivity is positive? What is our convention for runtime consistency checks?

src/serac/physics/heat_transfer.hpp
tupek2: In principal, it could be re-computed with the 'adjoint temperature', and multiplying it by just the mass matrix transposed at reverse time=0.  However, this would force the user to essentially rederive (and recompute) a quantity that we already store.  And this multiply would presumably require an new API function anyway.  In my view, I think that breaks encapsulation too much to justify (potentially) reducing the number of public API functions.

tupek2: Also, I'm voting for not documenting what it is if accessed early.  This way we do not have to commit to it being available for all physics before the end of the adjoint pass? There could be future optimizations that take advantage of this flexibility.

src/serac/physics/heat_transfer.hpp
tupek2: I added the derivation to the doxygen to set internal notation, conventions, etc.

src/serac/physics/heat_transfer.hpp
tupek2: I had a different question there @jamiebramwell .  What situation would have the essential BCs not be 0 here?

src/serac/physics/heat_transfer.hpp
tupek2: I added more info here.

src/serac/physics/heat_transfer.hpp
tupek2: I consolidated the naming a bit as suggested, but kept the different u and temperature names for now as they are used a bit differently.

src/serac/physics/materials/thermal_material.hpp
tupek2: I changed this to a runtime check, for better or worse.

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
tupek2: parametric sensitivies can be added for the next PR.  There is some more work to do to make those work correctly

src/serac/physics/heat_transfer.hpp
tupek2: I changed this to loadCheckpointedTemperature, it cannot really load a future temperature, though it could try and fail and crash!

jamiebramwell:
After chatting in the meeting today, we decided to skip the parameter sensitivities for now as https://github.com/LLNL/serac/pull/977 will have a huge impact anyway. @btalamini @samuelpmishLLNL @sethwatts @samuelpmishLLNL , any comments are appreciated!

src/serac/physics/state/finite_element_vector.cpp
jamiebramwell: I think the biggest blocker for this is we would have to call the accessor for the underlying `mfem::HypreParVector` to interact with `Functional` or any of the nonlinear, linear, or ODE solvers. That said, it would make reasoning about these math operations much simpler for the reasons you described. I'll tag this as a discussion topic for Friday.

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
```diff
@@ -0,0 +1,329 @@
...
+  d_qoi_d_temperature = temperature;
+  d_qoi_d_temperature *= dt;
+}
+
+std::unique_ptr<HeatTransfer<p, dim>> create_nonlinear_heat_transfer(
+    axom::sidre::DataStore& /*data_store*/, const NonlinearSolverOptions& nonlinear_opts,
+    const TimesteppingOptions&                                                  dyn_opts,
+    const heat_transfer::IsotropicConductorWithLinearConductivityVsTemperature& mat)
+{
+  // eventually figure out how to clear out cider state
```
jamiebramwell:
```suggestion
  // eventually figure out how to clear out sidre state
```

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+  return std::make_tuple(qoi, initial_temperature_sensitivity, shape_sensitivity);
+}
+
+std::tuple<double, FiniteElementDual, FiniteElementDual> computeThermalConductivitySensitivity(
+    axom::sidre::DataStore& data_store, const NonlinearSolverOptions& nonlinear_opts,
+    const TimesteppingOptions& dyn_opts, const heat_transfer::ParameterizedLinearIsotropicConductor& mat,
+    const TimeSteppingInfo& ts_info)
+{
+  printf("a\n");
```
jamiebramwell: I assume these print statements are for debugging. They should be removed before this is merged.

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
```diff
@@ -0,0 +1,329 @@
...
+  // StateManager::reset();
+  // static int iter = 0;
+  // StateManager::initialize(data_store, "thermal_dynamic_solve"+std::to_string(iter++));
+  // std::string filename = std::string(SERAC_REPO_DIR) + "/data/meshes/star.mesh";
+  // mfem::ParMesh* mesh = StateManager::setMesh(std::move(saveMesh));
+  static int iter = 0;
+  auto       thermal =
+      std::make_unique<ParametrizedHeatTransferT>(nonlinear_opts, heat_transfer::direct_linear_options, dyn_opts,
+                                                  parametrized_thermal_prefix + std::to_string(iter++));
+  printf("one\n");
```
jamiebramwell: More debug prints to remove.

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
```diff
@@ -0,0 +1,329 @@
...
+  return thermal;
+}
+
+using ParametrizedHeatTransferT = HeatTransfer<p, dim, Parameters<H1<p>>, std::integer_sequence<int, 0>>;
+
+std::unique_ptr<ParametrizedHeatTransferT> create_parameterized_heat_transfer(
+    axom::sidre::DataStore& /*data_store*/, const NonlinearSolverOptions& nonlinear_opts,
+    const TimesteppingOptions& dyn_opts, const heat_transfer::ParameterizedLinearIsotropicConductor& mat)
+{
+  // eventually figure out how to clear out cider state
```
jamiebramwell:
```suggestion
  // eventually figure out how to clear out sidre state
```

src/serac/physics/heat_transfer.hpp
```diff
@@ -601,41 +604,68 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
    * @pre The adjoint load maps are expected to contain a single entry named "temperature"
    * @note If the essential boundary dual is not specified, homogeneous essential boundary conditions are applied to
    * the adjoint system
+   * @note We provide a quick derivation for the discrete adjoint equations and notation used here for backward
```
jamiebramwell, line 607: It looks like this isn't quite rendering correctly in doxygen: https://serac.readthedocs.io/en/feature-tupek-transient_adj/doxygen/html/classserac_1_1HeatTransfer_3_01order_00_01dim_00_01Parameters_3_01parameter__space_8_8_8_01_4_00b10147e2187ea0fff3ed4d400f86c2eb.html

jamiebramwell: I've enabled this branch in readhthedocs, so it should autobuild your new docs for each commit pushed.

src/serac/physics/heat_transfer.hpp
jamiebramwell: This would be non-zero if the quantity of interest was a function of the applied boundary values in the forward problem. This most commonly occurs in inverse scenarios.

src/serac/physics/heat_transfer.hpp
jamiebramwell: I would hope that sidre handles this at least somewhat gracefully. I will investigate.

jamiebramwell:
Looks good! Once the documentation rendering is fixed I think this is ready to go!

src/serac/physics/base_physics.hpp
```diff
@@ -115,7 +115,7 @@ class BasePhysics {
    * @param parameter_name The name of the parameter to generate
    *
    * @note The user is responsible for managing the lifetime of this object. It is required
-   * to exist whenever advanceTimestep, solveAdjoint, or computeSensitivity is called.
+   * to exist whenever advanceTimestep, reverseAdjointTimestep, or computeTimestepSensitivity is called.
```
sethwatts, line 118: Minor comment, but does "reverse adjoint timestep" have sort of a double negative meaning, in the sense that the adjoint approach is already moving backwards through time? What do you think about a more "symmetric" construction with `advanceTimestep` such as `reverseTimestep` or `retreatTimestep`? Or even `advanceAdjointTimestep`, although that might cause confusion for someone using Serac as a (primal) solver library.

src/serac/physics/heat_transfer.hpp
```diff
@@ -601,41 +604,68 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+   * u^{0}_{,d}, \f] where the multiplier/adjoint equations are solved backward in time starting at \f$n=N\f$,
+   * \f$\mu^{N+1} = 0\f$, and \f$u^{0}_{,d}\f$ is the sensitivity of the initial primal variable with respect to the
+   * parameters.\n We call the quantities \f$\pi^n_{,u}\f$ and \f$\pi^n_{,v}\f$ the temperature and temperature-rate
+   * adjoint loads, respectively.
    *
    * @param adjoint_loads An unordered map containing finite element duals representing the RHS of the adjoint equations
-   * indexed by their name
+   * indexed by their name.  It must have a load named "temperature", which is the sensitivity of the qoi over that
+   * timestep to the end-of-step temperature.  It may optionally have a load named "temperature_rate", which is the
+   * sensitivity of the qoi over that timestep with respect to the end-of_step temperature rate of change.
```
sethwatts, line 634: 👍

src/serac/physics/heat_transfer.hpp
sethwatts: 👍

src/serac/physics/state/finite_element_vector.cpp
```diff
@@ -150,4 +150,19 @@ double min(const FiniteElementVector& fe_vector)
   return global_min;
 }
 
+double innerProduct(const FiniteElementVector& v1, const FiniteElementVector& v2)
+{
+  SLIC_ERROR_IF(
+      v1.Size() != v2.Size(),
+      axom::fmt::format("Finite element vector of size '{}' can not inner product with another vector of size '{}'",
+                        v1.Size(), v2.Size()));
+  SLIC_ERROR_IF(v1.comm() != v2.comm(),
```
sethwatts, line 175: To be really pedantic, you could also check whether the two vectors are written against the same `mfem::(Par)FinieteElementSpace` - you could imagine scenarios where two different spaces on the same communicator coincidentally have the same sizes, but you're doing a non-nonsensical inner product.

src/serac/physics/heat_transfer.hpp
tupek2: This seems to be a fundamental issue with readthedocs (or at least how we are using it) and doxygen latex.  You can see the same rendering issue with existing in-code latex, e.g.: https://serac.readthedocs.io/en/feature-tupek-transient_adj/doxygen/html/classserac_1_1BoundaryCondition.html#ae8052ce133b3128b6879782d8d824812

After googling, I have no idea where to even begin fixing this.  It renders perfectly well with doxygen, and I have 0 understanding about the differences between that and readthedocs.

tupek2: image

tupek2: here it is in the build path with doxygen

src/serac/physics/base_physics.hpp
tupek2: Keeping name for now. Can discuss in a bug fix day.

src/serac/physics/state/finite_element_vector.cpp
tupek2: Done, I ended up essentially coping the LiDO implementation.  Maybe mfem should support == on their spaces.

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
tupek2: fixed

src/serac/physics/tests/dynamic_thermal_adjoint.cpp
tupek2: fixed

src/serac/physics/heat_transfer.hpp
jamiebramwell: Weird. I say we merge this and I'll investigate in another PR.

jamiebramwell (approved):

btalamini (approved):

PR #1001, opened 2023-09-19
Input Deck Material Rework

chapman39:
Reworks the input deck functionality back to a working state.

### Material types that can be used in an input deck:
  * NeoHookean
  * J2
  * J2Nonlinear (PowerLawHardening, VoceHardening)
  * LinearElastic
  * LinearIsotropicConductor
  * LinearConductor (dim=2 or 3)

### Features
* Specify multiple materials in a single physics
* List materials inlined or from a variable
* Adjust lua files to work with new material schema
* Generate input schema documentation
* Inlet verifies material section to ensure correctness (e.g. type, spelling of material model, unused parameters)

### Example of setting materials in lua
```lua
-- materials can be stored above here
material_j2 = { model = "J2", E = 100.0, nu = 0.25, hardening = 100.0 }
solid = {
    ...
    -- material parameters
    materials = {
        material_j2,        
        {
            model = "NeoHookean", -- or materials can be inlined here
            mu = 0.26,
            K = 5.5,
            density = 2
        },
        {
            model = "J2Nonlinear",
            E = 1,
            nu = 2,
            hardening = 
            {
                law = "PowerLawHardening",
                sigma_y = 1.1,
                n = 1.2,
                eps0 = 1.3
            },
            density = 2.4,
        },
    },
    ...
}
```


samuelpmishLLNL:
@chapman39 @white238 
here's a prototype of what I was trying to describe on the whiteboard on Friday: https://godbolt.org/z/veEh9xMqz

chapman39:
Ready for review. Could use feedback on variable names and inlet variable descriptions to more accurately represent the physics.

src/serac/physics/materials/thermal_material_input.cpp
```diff
@@ -0,0 +1,106 @@
...
+      serac::tensor<double, 2, 2>      cond = {{{v[0][0], v[0][1]}, {v[1][0], v[1][1]}}};
+      result = serac::heat_transfer::LinearConductor<2>(base["density"], base["cp"], cond);
+    } else if (dim == 3) {
+      std::vector<std::vector<double>> v    = {base["kappa_tensor"]["row1"], base["kappa_tensor"]["row2"],
+                                            base["kappa_tensor"]["row3"]};
+      serac::tensor<double, 3, 3>      cond = {
+          {{v[0][0], v[0][1], v[0][2]}, {v[1][0], v[1][1], v[1][2]}, {v[2][0], v[2][1], v[2][2]}}};
+      result = serac::heat_transfer::LinearConductor<3>(base["density"], base["cp"], cond);
+    }
+  }
```
samuelpmishLLNL, line 103: is it possible to query the type associated with a lua variable? If so, we could implement something like

```cpp
if(is_a_double(base["kappa"])) {
  result = serac::heat_transfer::LinearIsotropicConductor(base["density"], base["cp"], base["kappa"]);
}
if(is_an_array_of_arrays(base["kappa"])) {
  int num_rows = base["kappa"].size();
  if (dim == 2) {
      std::vector<std::vector<double>> v    = {base["kappa"]["row1"], base["kappa"]["row2"]};
      serac::tensor<double, 2, 2> conductivity = {{{v[0][0], v[0][1]}, {v[1][0], v[1][1]}}};
      result = serac::heat_transfer::LinearConductor<2>(base["density"], base["cp"], conductivity);
  }
  if (dim == 3) {
      std::vector<std::vector<double>> v    = {base["kappa"]["row1"], base["kappa"]["row2"], base["kappa"]["row3"]};
      serac::tensor<double, 3, 3> conductivity = {{
        {v[0][0], v[0][1], v[0][2]}, 
        {v[1][0], v[1][1], v[1][2]}, 
        {v[2][0], v[2][1], v[2][2]}
      }};
      result = serac::heat_transfer::LinearConductor<3>(base["density"], base["cp"], conductivity);
  }
}
```

Then the user wouldn't need to explicitly name which kind of LinearConductor it is in the input file, e.g.

```lua
materials = {
  -- we can tell that it's an isotropic model (LinearIsotropicConductor) since kappa is a floating point number
  { model = "LinearConductor", kappa = 0.5, cp = 1.0, density = 1.0 }, 

  -- we use the more general material model (LinearConductor) when kappa is a matrix
  -- also, we can figure out dim = 2 from the matrix dimensions, rather than asking the user to put that in the input file
  { model = "LinearConductor", kappa = [[1.0, 0.0], [0.0, 2.0]], cp = 1.0, density = 1.0 } 
}
```

src/serac/physics/materials/thermal_material_input.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <string>
+#include "serac/infrastructure/input.hpp"
+#include "serac/physics/materials/thermal_material.hpp"
+
+namespace serac {
+
+/// @brief Holds all possible heat transfer materials that can be utilized in our Input Deck
+using var_thermal_material_t = std::variant<heat_transfer::LinearIsotropicConductor, heat_transfer::LinearConductor<2>,
+                                            heat_transfer::LinearConductor<3>>;
```
samuelpmishLLNL, line 23: This is a reasonable implementation for what we currently offer, but this design isn't extensible for users that want to add their own material models.

I think we may want to reconsider this approach so that users can more easily write new materials and access them from input files without modifying serac. For example (although not necessarily a good implementation), instead of a `std::variant`, we could have a user-extendable `std::map<std::string, std::function< ... > >` of callbacks. After reading the name of the material model, it would invoke the callback in that std::map (where the user could write code to set up their custom material model, and then register it to the physics module with `setMaterial`).

src/serac/physics/materials/thermal_material_input.cpp
samuelpmishLLNL: We could also do something similar with the J2 models, so that the users just write "J2" in the model name, and then look at the "hardening" variable to figure out which one to use (or just remove our old "J2" model and replace it by a linear hardening rule in the more general model).

data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
```diff
@@ -51,7 +51,7 @@ thermal_conduction = {
     order = 2,
 
     -- material parameters
-    kappa = 0.5,
+    materials = { { model = "LinearIsotropicConductor", kappa = 0.5, cp = 0.5, density = 0.5 }, },
```
samuelpmishLLNL, line 54: I didn't see any tests with the `LinearConductor` model (with matrix-valued `kappa`), so it might be good to add another test for those materials

src/serac/physics/materials/thermal_material_input.cpp
chapman39: The problem is that we need to define `kappa`'s type in `ThermalMaterialInputOptions::defineInputFileSchema()` before we read from a lua file, which is why I unfortunately had to use different names.

data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
chapman39: I agree. All I really did was modify the previous lua files to work with the new material schema. We also need examples for all solid mechanics materials besides Neo-hookean.

src/serac/physics/materials/thermal_material_input.cpp
samuelpmishLLNL: > The problem is that we need to define kappa's type in ThermalMaterialInputOptions::defineInputFileSchema() before we read from a lua file

Where does this requirement come from? If I comment out 
```
container.addDouble("kappa", "The conductivity parameter");
```
am I not allowed to have a field `kappa` for that material model in the input file?

chapman39: Yes because then you wouldn't be able to access it in the `FromInlet operator()` like `base["kappa"]`.

chapman39:
I ran the integration tests, and found many were failing. However, they fail on develop as well:

> FAILED:  34 tolerance_test(dyn_direct_solve_serial), tolerance_test(dyn_direct_solve_parallel), tolerance_test(dyn_newmark_solve_serial), tolerance_test(dyn_newmark_solve_parallel), tolerance_test(dyn_newmark_solve_bending_serial), tolerance_test(dyn_newmark_solve_bending_parallel), tolerance_test(dyn_newmark_solve_bending_first_serial), tolerance_test(dyn_newmark_solve_bending_first_parallel), tolerance_test(dyn_solve_serial), tolerance_test(dyn_solve_parallel), serac(qs_direct_solve_serial), serac(qs_direct_solve_parallel), serac(qs_linear_serial), serac(qs_linear_parallel), serac(qs_solve_serial), serac(qs_solve_parallel), tolerance_test(dyn_exp_solve_serial), tolerance_test(dyn_exp_solve_parallel), tolerance_test(dyn_imp_solve_serial), tolerance_test(dyn_imp_solve_parallel), tolerance_test(dyn_imp_solve_restart_serial), tolerance_test(dyn_imp_solve_restart_parallel), tolerance_test(dyn_imp_solve_time_varying_serial), tolerance_test(dyn_imp_solve_time_varying_parallel), serac(static_reaction_exact_serial), serac(static_reaction_exact_parallel), tolerance_test(static_solve_serial), tolerance_test(static_solve_parallel), tolerance_test(static_solve_multiple_bcs_serial), tolerance_test(static_solve_multiple_bcs_parallel), tolerance_test(static_solve_repeated_bcs_serial), tolerance_test(static_solve_repeated_bcs_parallel), serac(thermal_expansion_serial), serac(thermal_expansion_parallel)
PASSED:   24
SKIPPED:  10

We may need to rebaseline or maybe something about the input deck is still messed up.

chapman39:
Docs https://serac.readthedocs.io/en/feature-chapman39-input_deck_update/sphinx/user_guide/input_schema.html

src/serac/physics/heat_transfer.hpp
```diff
@@ -216,31 +218,41 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
-                     physics_name, mesh_tag, {}, cycle, time)
+  HeatTransfer(const HeatTransferInputOptions& input_options, const std::string& physics_name,
+               const std::string& mesh_tag, int cycle = 0, double time = 0.0)
+      : HeatTransfer(input_options.nonlin_solver_options, input_options.lin_solver_options,
+                     input_options.timestepping_options, physics_name, mesh_tag, {}, cycle, time)
   {
-    if (options.initial_temperature) {
-      auto temp = options.initial_temperature->constructScalar();
+    // This is the only other options stored in the input file that we can use
+    // in the initialization stage
```
white238: I am not sure this comment is helpful at this point. I would say either remove it or add a more helpful doxygen comment saying which input options are used.

data/input_files/tests/solid/dyn_solve.lua
```diff
@@ -48,8 +48,7 @@ solid = {
     order = 1,
 
     -- neo-Hookean material parameters
-    mu = 0.25,
-    K  = 5.0,
+    materials = { { model = "NeoHookean", mu = 0.25, K = 5.0, density = 1.0 }, }
```
white238, line 51: Is `1.0` a sensible default? Can this be added to the Inlet definition call?

chapman39: do you mean setting a defaultValue when I define the schema? i think it would be fine to set it for density because its used for all of the material types, but for the others we wouldn't be able to tell whether or not the user entered the right parameters - so it would be difficult to verify

jamiebramwell:
> I ran the integration tests, and found many were failing. However, they fail on develop as well:
> 
> > FAILED:  34 tolerance_test(dyn_direct_solve_serial), tolerance_test(dyn_direct_solve_parallel), tolerance_test(dyn_newmark_solve_serial), tolerance_test(dyn_newmark_solve_parallel), tolerance_test(dyn_newmark_solve_bending_serial), tolerance_test(dyn_newmark_solve_bending_parallel), tolerance_test(dyn_newmark_solve_bending_first_serial), tolerance_test(dyn_newmark_solve_bending_first_parallel), tolerance_test(dyn_solve_serial), tolerance_test(dyn_solve_parallel), serac(qs_direct_solve_serial), serac(qs_direct_solve_parallel), serac(qs_linear_serial), serac(qs_linear_parallel), serac(qs_solve_serial), serac(qs_solve_parallel), tolerance_test(dyn_exp_solve_serial), tolerance_test(dyn_exp_solve_parallel), tolerance_test(dyn_imp_solve_serial), tolerance_test(dyn_imp_solve_parallel), tolerance_test(dyn_imp_solve_restart_serial), tolerance_test(dyn_imp_solve_restart_parallel), tolerance_test(dyn_imp_solve_time_varying_serial), tolerance_test(dyn_imp_solve_time_varying_parallel), serac(static_reaction_exact_serial), serac(static_reaction_exact_parallel), tolerance_test(static_solve_serial), tolerance_test(static_solve_parallel), tolerance_test(static_solve_multiple_bcs_serial), tolerance_test(static_solve_multiple_bcs_parallel), tolerance_test(static_solve_repeated_bcs_serial), tolerance_test(static_solve_repeated_bcs_parallel), serac(thermal_expansion_serial), serac(thermal_expansion_parallel)
> > PASSED:   24
> > SKIPPED:  10
> 
> We may need to rebaseline or maybe something about the input deck is still messed up.

I think re-enabling the integration tests and rebaselining should be a follow-on PR as this is already pretty big.

data/input_files/tests/solid/dyn_solve.lua
jamiebramwell: I don't think we should have density be a default value. That is clearly something we need as input to make reasonable simulations. It is mostly set to 1.0 since all of our problems are currently notional examples.

data/input_files/tests/thermal_conduction/dyn_exp_solve.lua
jamiebramwell: The additional tests should probably go in the integration test follow on PR.

src/serac/physics/materials/thermal_material_input.cpp
jamiebramwell: This is a good idea and I believe it will require modifications in Inlet.

src/serac/physics/materials/thermal_material_input.hpp
jamiebramwell: @white238 , we probably want this in the future. Is it best to leave it at the current implementation until we add this capability or would the `std::map` approach be better?

data/input_files/tests/solid/dyn_solve.lua
chapman39: What do you mean by user-set value? An option in Inlet?

samuelpmishLLNL: imo, defaults should only be used if there is a natural default value, but in this case
1. a lot of different materials may be approximated by the same model, so there is no obvious "default" material choice
2. the values of model parameters for a given material depend on the user's (arbitrary) choice of units

So, I'd vote for requiring users to specify all the model parameters to avoid confusion.

src/serac/physics/materials/thermal_material_input.hpp
samuelpmishLLNL: sorry for not clarifying in the initial comment: I'm not expecting any changes in this PR-- just that we might want to discuss and revisit this design choice at some point in the future

data/input_files/tests/solid/dyn_solve.lua
white238: Sounds great. I was just checking due to how much of a change from the previous input files. All of these had default values and it seemed odd in retrospect.

jamiebramwell: Sorry - I typo'd my original comment. I don't think density should have a default value.

src/serac/physics/materials/thermal_material_input.cpp
chapman39: issue in axom tracking this https://github.com/LLNL/axom/issues/1234

src/serac/physics/materials/hardening_input.cpp
```diff
@@ -0,0 +1,59 @@
...
+  // PowerLawHardening
+  container.addDouble("n", "Hardening index in reciprocal form");
+  container.addDouble("eps0", "Reference value of accumulated plastic strain");
+
+  // VoceHardening
+  container.addDouble("sigma_sat", "Saturation value of flow strength");
+  container.addDouble("strain_constant", "Constant dictating how fast the exponential decays");
+
+  // Verify
+  container.registerVerifier([](const axom::inlet::Container& c) -> bool {
```
white238, line 26: 👍 +1

src/serac/physics/materials/hardening_input.hpp
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) 2019-2023, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file solid_material_input.hpp
```
white238:
```suggestion
 * @file hardening_input.hpp
```

src/serac/physics/materials/hardening_input.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+#pragma once
+
+#include <string>
+#include "serac/infrastructure/input.hpp"
+#include "serac/physics/materials/solid_material.hpp"
+
+namespace serac {
+
+/// @brief Holds all possible isotropic hardening laws that can be utilized in our Input Deck
```
white238:
```suggestion
/// @brief Holds all possible isotropic hardening laws that can be utilized in our input file
```

src/serac/physics/materials/solid_material_input.hpp
```diff
@@ -0,0 +1,49 @@
...
+#pragma once
+
+#include <string>
+#include "serac/infrastructure/input.hpp"
+#include "serac/physics/materials/solid_material.hpp"
+#include "serac/physics/materials/hardening_input.hpp"
+
+namespace serac {
+
+/// @brief All possible solid mechanics materials that can be utilized in our Input Deck
```
white238:
```suggestion
/// @brief All possible solid mechanics materials that can be utilized in our input file
```

src/serac/physics/materials/thermal_material_input.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  // For LinearIsotropicConductor
+  container.addDouble("kappa", "The conductivity parameter");
+
+  // For LinearConductor
+  container.addInt("dim", "Dimension of conductivity tensor parameter");
+  auto& kappa_tensor_container = container.addStruct("kappa_tensor", "The conductivity tensor parameter");
+  kappa_tensor_container.addDoubleArray("row1", "First row of the conductivity tensor parameter");
+  kappa_tensor_container.addDoubleArray("row2", "Second row of the conductivity tensor parameter");
+  kappa_tensor_container.addDoubleArray("row3", "Third row of the conductivity tensor parameter");
```
white238, line 26: Leaving a comment here so that this is cleaned up after we have a better way of handling this in Inlet.

white238 (approved):
This is great! Thank you for sticking this PR through, this was way more involved than we originally thought.

white238:
It also shows some functionality needed in Inlet that we originally didn't think about.

jamiebramwell (approved):

PR #1002, opened 2023-09-21
support operations on specific parts of a mesh

samuelpmishLLNL:
Right now, if you want to specify that a material, traction or body force only apply to part of the domain, you have to fake it by using a conditional expression to mask out the part of the domain:

```cpp
  mechanics.setTraction([](const auto& x, const auto& n, const double) { 
    // tractions are calculated on all faces, but only those with x[1] > 0.99 are nonzero
    return -0.01 * n * (x[1] > 0.99);
  });
```

this is wasteful and undesirable in how it couples two independent concepts: the integrand and the domain of integration.

------

this PR adds a new type, `Domain`, to represent a specific part of a mesh. `Domain`s can describe exactly one of:
- A set of vertices (create with `Domain::OfVertices( ... )`)
- A set of edges (create with `Domain::OfEdges( ... )`)
- A set of faces (create with `Domain::OfFaces( ... )`)
- A set of elements (create with `Domain::OfElements( ... )`)

`Domain`s are created by providing a function with signature like[^1]

[^1]: see [here](https://github.com/LLNL/serac/blob/67bb5810d6f4218cfbd851c5fe81e5fb6e37977d/src/serac/numerics/functional/domain.hpp#L23-L33) for all signatures

$$
\texttt{bool(std::vector< tensor < double, dim > > vertices, int attr)}
$$

That function gets applied to the {vertices, edges, faces, elements} of the mesh, and the {vertices, edges, faces, elements} that return `true` are added to the `Domain`. e.g.

```cpp
    // create a collection of edges from information about their spatial position
    Domain d1 = Domain::ofEdges(mesh, std::function([](std::vector<vec2> x, int /* bdr_attr */){ 
        return (0.5 * (x[0][1] + x[1][1])) < 0.25; // y coordinate of edge midpoint 
    }));

    // create a collection of edges from their attribute
    Domain d2 = Domain::ofEdges(mesh, std::function([](std::vector<vec2> x, int bdr_attr){ 
        return (bdr_attr == 2);
    }));

    // succinct alternative to above
    // Domain d2 = Domain::ofEdges(mesh, by_attr(2));

    // create a collection of edges from some combination of position/attr
    Domain d3 = Domain::ofEdges(mesh, std::function([](std::vector<vec2> x, int bdr_attr){ 
        return (bdr_attr == 2) && (norm(x[0]) < 0.5);
    }));

    // define domains through set operations
    Domain d4 = ((d1 | d2) & d3);
```

----------

The next steps are to make `Domain`s usable in `Functional` to specify which parts of the mesh to integrate over and in physics modules to express domains affected by essential boundary conditions.


samuelpmishLLNL:
I'm still seeing some doxygen warnings on my local build (v1.9.1) from an unrelated part of the code:

```
/home/sam/code/serac/src/serac/physics/state/finite_element_vector.hpp:61: warning: argument 'name' of command @param is not found in the argument list of serac::FiniteElementDual::FiniteElementVector(typename FunctionSpace)
/home/sam/code/serac/src/serac/physics/state/finite_element_vector.hpp:61: warning: argument 'mesh' of command @param is not found in the argument list of serac::FiniteElementState::FiniteElementVector(typename FunctionSpace)
/home/sam/code/serac/src/serac/physics/state/finite_element_vector.hpp:61: warning: argument 'name' of command @param is not found in the argument list of serac::FiniteElementState::FiniteElementVector(typename FunctionSpace)
```

I played with this for a bit and couldn't figure out how to fix it-- I think maybe Doxygen is parsing the C++ sourcefile incorrectly and thinking that the template parameter is a function argument. 

With any luck, our CI jobs use against a different version of Doxygen that is okay with `finite_element_vector.hpp:61`.

jamiebramwell:
> I'm still seeing some doxygen warnings on my local build (v1.9.1) from an unrelated part of the code:
> 
> ```
> /home/sam/code/serac/src/serac/physics/state/finite_element_vector.hpp:61: warning: argument 'name' of command @param is not found in the argument list of serac::FiniteElementDual::FiniteElementVector(typename FunctionSpace)
> /home/sam/code/serac/src/serac/physics/state/finite_element_vector.hpp:61: warning: argument 'mesh' of command @param is not found in the argument list of serac::FiniteElementState::FiniteElementVector(typename FunctionSpace)
> /home/sam/code/serac/src/serac/physics/state/finite_element_vector.hpp:61: warning: argument 'name' of command @param is not found in the argument list of serac::FiniteElementState::FiniteElementVector(typename FunctionSpace)
> ```
> 
> I played with this for a bit and couldn't figure out how to fix it-- I think maybe Doxygen is parsing the C++ sourcefile incorrectly and thinking that the template parameter is a function argument.
> 
> With any luck, our CI jobs use against a different version of Doxygen that is okay with `finite_element_vector.hpp:61`.

This we filter this error out for CI as I couldn't figure out a way to fix it in doxygen. The regexes to ignore are listed in this file: https://github.com/LLNL/serac/blob/develop/scripts/azure-pipelines/docs_ignore_regexs.txt

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -157,8 +157,8 @@ template <uint32_t differentiation_index, int Q, mfem::Geometry::Type geom, type
           typename lambda_type, typename derivative_type, int... indices>
 void evaluation_kernel_impl(FunctionSignature<test(trials...)>, const std::vector<const double*>& inputs,
                             double* outputs, const double* positions, const double* jacobians, lambda_type qf,
-                            [[maybe_unused]] derivative_type* qf_derivatives, uint32_t num_elements,
-                            std::integer_sequence<int, indices...>)
+                            [[maybe_unused]] derivative_type* qf_derivatives, const int* elements,
```
btalamini, line 161: Summarizing a face-to-face conversation: I asked if raw pointers were a good idea. Paraphrasing Sam (crudely), the answer is that this is temporary and the container choice will get revisited when we port this to GPU.

src/serac/numerics/functional/domain.cpp
```diff
@@ -0,0 +1,481 @@
...
+  // [x1, x2, x3, ..., y1, y2, y3 ..., (z1, z2, z3, ...)]
+  mfem::Vector vertices;
+  mesh.GetVertices(vertices);
+
+  int tri_id  = 0;
+  int quad_id = 0;
+  int tet_id  = 0;
+  int hex_id  = 0;
+
+  // faces that satisfy the predicate are added to the domain
```
btalamini:
```suggestion
  // elements that satisfy the predicate are added to the domain
```

src/serac/numerics/functional/domain.cpp
```diff
@@ -0,0 +1,481 @@
...
+  for (int i = 0; i < num_elems; i++) {
+    mfem::Array<int> vertex_ids;
+    mesh.GetElementVertices(i, vertex_ids);
+
+    std::vector<tensor<double, d>> x(std::size_t(vertex_ids.Size()));
+    for (int v = 0; v < vertex_ids.Size(); v++) {
+      for (int j = 0; j < d; j++) {
+        x[uint32_t(v)][j] = vertices[j * num_vertices + vertex_ids[v]];
+      }
+    }
```
btalamini: This is repeated several times - is it worth factoring out? It's tiny any maybe not worth it. OTOH, it would make it a little easier to update in case MFEM changes their (undocumented!) layout.

src/serac/numerics/functional/domain.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+  /// @overload
+  static Domain ofVertices(const mfem::Mesh& mesh, std::function<bool(vec3)> func);
+
+  /**
+   * @brief create a domain from some subset of the edges in an mfem::Mesh
+   * @param mesh the entire mesh
+   * @param func predicate function for determining which edges will be included in this domain
+   */
+  static Domain ofEdges(const mfem::Mesh& mesh, std::function<bool(std::vector<vec2>, int)> func);
```
btalamini, line 75: Is there a good way to document the predicate signature?

src/serac/numerics/functional/functional.hpp
```diff
@@ -219,7 +221,7 @@ class Functional<test(trials...), exec> {
   {
     auto mem_type = mfem::Device::GetMemoryType();
 
-    for (auto type : {Integral::Type::Domain, Integral::Type::Boundary}) {
+    for (auto type : {Domain::Type::Elements, Domain::Type::BoundaryElements}) {
```
btalamini, line 224: The `for` loop over two things seems overly complicated, but Sam explained to me in person that this list may expand in the future.

src/serac/numerics/functional/functional_qoi.inl
```diff
@@ -160,21 +160,38 @@ public:
    * @tparam lambda the type of the integrand functor: must implement operator() with an appropriate function signature
    * @tparam qpt_data_type The type of the data to store for each quadrature point
    * @param[in] integrand The user-provided quadrature function, see @p Integral
-   * @param[in] domain The domain on which to evaluate the integral
+   * @param[in] mesh The domain on which to evaluate the integral
```
btalamini, line 154: good catch

src/serac/numerics/functional/tests/check_gradient.hpp
```diff
@@ -77,6 +77,8 @@ void check_gradient(serac::Functional<T>& f, const mfem::Vector& U, double epsil
   double e3 = df1_cd[0].DistanceTo(df_jvp1.GetData()) / denominator;
   double e4 = df1_cd[1].DistanceTo(df_jvp1.GetData()) / denominator;
   EXPECT_TRUE((fabs(e3 / e4 - 4.0) < 0.1) || fmin(e3, e4) < 1.0e-9);
+
+  // std::cout << e1 << " " << e2 << " " << e3 << " " << e4 << std::endl;
```
btalamini: Debug info

btalamini (approved):
Great work!

src/serac/numerics/functional/tests/geometric_factors_tests.cpp
```diff
@@ -0,0 +1,112 @@
...
+    GeometricFactors gf(d, q, mfem::Geometry::TRIANGLE);
+    int              components_per_J = 4;
+    int              qpts_per_elem    = (q * (q + 1)) / 2;
+    int              num_elems        = 1;
+    EXPECT_EQ(gf.J.Size(), components_per_J * qpts_per_elem * num_elems);
+
+    gf.X.Print(std::cout, 2);
+  }
+
+  std::cout << std::endl;
```
tupek2, line 52: remove prints?

tupek2 (approved):

src/serac/numerics/functional/domain.hpp
tupek2: please don't, its in the function arguments, which is the hard truth, whereas comments are lies (statistically speaking).  no need to duplicate

btalamini: It's not clear what the `int` is supposed to be. If I implement a predicate, this tells me what my signature of my predicate has to be, but not what the arguments mean.

jamiebramwell: For the physics modules, I typically document what the arguments and output of the function argument should be in an extra doxygen comment. The issue is these predicates are typically user-defined, so there's not really a good "source of truth" for the documentation. That said, I know first hand how easy these are to get out of date.

src/serac/numerics/functional/domain.hpp
```diff
@@ -0,0 +1,134 @@
...
+ * This region can be an entire mesh or some subset of its elements
+ */
+struct Domain {
+  /// @brief enum describing what kind of elements are included in a Domain
+  enum Type
+  {
+    Elements,
+    BoundaryElements
+  };
+
```
jamiebramwell, line 28: If we have only two types of domains, what is the purpose of `ofVertices`, `ofEdges`, `ofFaces` etc? Why not just have `ofElements` and `ofBoundaryElements`?

jamiebramwell: In other words, what operations could we do on a domain of vertices in 3D?

src/serac/numerics/functional/domain.cpp
```diff
@@ -0,0 +1,481 @@
...
+
+template <int d, typename T>
+static Domain domain_of_edges(const mfem::Mesh& mesh, std::function<T> predicate)
+{
+  assert(mesh.SpaceDimension() == d);
+
+  Domain output{mesh, 1 /* edges are 1-dimensional */};
+
+  // layout is undocumented, but it seems to be
+  // [x1, x2, x3, ..., y1, y2, y3 ..., (z1, z2, z3, ...)]
```
jamiebramwell, line 100: I believe this is hardcoded currently in MFEM: https://github.com/mfem/mfem/blob/ecfab8fe9f6f01dae8fc9fc59783bf9581eb3780/mesh/mesh.cpp#L8315

src/serac/numerics/functional/domain.hpp
```diff
@@ -0,0 +1,134 @@
...
+  /// @brief whether the elements in this domain are on the boundary or not
+  Type type_;
+
+  /// note: only lists with appropriate dimension (see dim_) will be populated
+  ///       for example, a 2D Domain may have `tris_` and `quads_` non-nonempty,
+  ///       but all other lists will be empty
+  ///
+  /// @cond
+  std::vector<int> vertices_;
```
jamiebramwell: When looking at this in other parts of the code, it wasn't immediately obvious that these are indicies into the returned appropriate mfem containers. I suggest documenting what methods are used to get the appropriate mfem containers as well as possibly renaming them `vertex_idxs_` or something similar.

src/serac/numerics/functional/domain.cpp
```diff
@@ -0,0 +1,481 @@
...
+    if (d == 2) {
+      attr = mesh.GetAttribute(i);
+    } else {
+      int bdr_id = face_id_to_bdr_id[i];
+      attr       = (bdr_id > 0) ? mesh.GetBdrAttribute(bdr_id) : -1;
+    }
+
+    if (predicate(x, attr)) {
+      if (x.size() == 3) {
+        output.tris_.push_back(tri_id);
```
jamiebramwell: I'm a bit confused how this works. The vertices and the edges store the MFEM index, but this just seems to be a list of 1...N for how many faces are in the domain. Am I missing some logic here?

src/serac/numerics/functional/domain.hpp
tupek2: OK, when I saw the words 'predicate signature', I interpreted it as documenting the type of inputs, and was seeing something like
* where the arguments to the function are a vector and an int. 


These kinds of comments drive me crazy.  But, I agree, that the intent of std::functions are not always clear, so documenting for external interfaces seems prudent.

src/serac/numerics/functional/geometric_factors.hpp
```diff
@@ -20,22 +21,22 @@ struct GeometricFactors {
    * @brief calculate positions and jacobians for quadrature points belonging to
    * elements with the specified geometry, belonging to the provided mesh.
    *
-   * @param mesh the mesh
+   * @param domain the domain of integration
    * @param q a parameter controlling the number of quadrature points per element
    * @param elem_geom which kind of element geometry to select
    */
-  GeometricFactors(const mfem::Mesh* mesh, int q, mfem::Geometry::Type elem_geom);
+  GeometricFactors(const Domain& domain, int q, mfem::Geometry::Type elem_geom);
```
jamiebramwell, line 28: Does this mean we can no longer upstream this to MFEM?

src/serac/numerics/functional/domain.cpp
samuelpmishLLNL: mfem::Mesh puts all of its elements (of any geometry) into a single list, but mfem's "E-vector" data layout only works with elements of one geometry. So, I'm splitting up mfem's element indices into separate lists for each element geometry. For example:

| mfem::Geometry | mfem element id | tri id | quad id |
| -------------- | --------------- | ------ | ------- |
| Triangle       | 0               | 0      |         |
| Triangle       | 1               | 1      |         |
| Square         | 2               |        | 0       |
| Triangle       | 3               | 2      |         |
| Square         | 4               |        | 1       |
| Square         | 5               |        | 2       |
| Square         | 6               |        | 3       |

Then, if an element satisfies the predicate it is added to the `Domain`.

Vertex and edge indices are treated slightly differently since there is only 1 kind of "element" for 0D and 1D, so no separation is required.

src/serac/numerics/functional/domain.hpp
samuelpmishLLNL: per the discussion [here](https://github.com/LLNL/serac/pull/1002#discussion_r1392842098), they're indices into their respective `mfem::Geometry`'s E-vector, I'll clarify the doxygen and change the member name.

src/serac/numerics/functional/domain.cpp
samuelpmishLLNL: I think it's very unlikely that they change this layout, but it makes me uncomfortable to have the correctness of our code depend on an undocumented implementation detail of some library.

src/serac/numerics/functional/domain.hpp
samuelpmishLLNL: `ofEdges`/`ofFaces` aren't just aliases for `ofBoundaryElements` in 2D/3D. `ofEdges`/`ofFaces` lets you include interior ones as well, so you could use them for:
- upcoming DG-type integration
- distributed loads over faces/edges
- custom materials (e.g. conductive wires in circuit/FEM simulations)

Re: vertex sets, I don't think our users have an immediate use for it but it was so similar to the other parts I added it in anyway. Some potential future applications could be:
- applying precomputed point loads to some part of a mesh
- tracking / constraining individual points (e.g. fabio's digital image correlation)
- prescribing essential boundary conditions (?)

I wanted to hook up `Domain` to the essential boundary condition stuff in serac, but the scope of this PR has already crept quite a bit!

src/serac/numerics/functional/geometric_factors.hpp
samuelpmishLLNL: In practice, it would have required some modification anyway to get it accepted into mfem, so a small change like this is easy to revert (or adapt to `mfem::SubMesh`?).

src/serac/numerics/functional/domain.cpp
jamiebramwell: Maybe add a reference to where this is used in MFEM so we know where to look if it breaks?

src/serac/numerics/functional/domain.cpp
jamiebramwell: Ahhh I missed that the quad/tri ids were always being incremented even if they didn't satisfy the predicate. Thank you for the explanation!

jamiebramwell (approved):

PR #1006, opened 2023-09-29
Improve precision of determinant of matrices near the identity matrix

### LINKED ISSUE #634: Implement determinants for identity perturbations of tensors
it might eventually be worth implementing a special function (with a better name) for `det_I_plus(du_dX) == det(I + du_dX)`, since there can be considerable loss of precision on this when `du_dX` is small

_Originally posted by @samuelpmishLLNL in https://github.com/LLNL/serac/pull/624#discussion_r797161609_

btalamini:
Implements a new function `detApIm1` that computes `det(A + I) - 1` while avoiding catastrophic cancellation for matrices that are close to the identity matrix. The formulas follow from the Cayley-Hamilton theorem applied to $A + I$.

Closes #634

### DIFF
--- a/src/serac/numerics/functional/tensor.hpp
+++ b/src/serac/numerics/functional/tensor.hpp
@@ -1219,11 +1219,11 @@ SERAC_HOST_DEVICE constexpr auto det(const tensor<T, 3, 3>& A)
 }
 
 /**
- * @brief Take determinant of a matrix near the identity without losing precision
+ * @brief computes det(A + I) - 1, where precision is not lost when the entries A_{ij} << 1
  *
  * detApIm1(A) = det(A + I) - 1
- * When the entries of A are small compared to unity, the straightforward way of
- * computing this will suffer from catasrophic cancellation.
+ * When the entries of A are small compared to unity, computing
+ * det(A + I) - 1 directly will suffer from catastrophic cancellation.
  *
  * @param A Input matrix
  * @return det(A + I) - 1, where I is the identity matrix
@@ -1239,20 +1239,18 @@ SERAC_HOST_DEVICE constexpr auto detApIm1(const tensor<T, 2, 2>& A)
 template <typename T>
 SERAC_HOST_DEVICE constexpr auto detApIm1(const tensor<T, 3, 3>& A)
 {
+  // clang-format off
   // equivalent to tr(A) + I2(A) + det(A)
-  return A(0, 0) 
-       + A(1, 1) 
-       + A(2, 2) 
-
+  return A(0, 0) + A(1, 1) + A(2, 2) 
        - A(0, 1) * A(1, 0) * (1 + A(2, 2))
        + A(0, 0) * A(1, 1) * (1 + A(2, 2))
        - A(0, 2) * A(2, 0) * (1 + A(1, 1))
        - A(1, 2) * A(2, 1) * (1 + A(0, 0))
        + A(0, 0) * A(2, 2) 
        + A(1, 1) * A(2, 2) 
-
        + A(0, 1) * A(1, 2) * A(2, 0) 
        + A(0, 2) * A(1, 0) * A(2, 1);
+  // clang-format on
 }
 
 /**
--- a/src/serac/numerics/functional/tensor.hpp
+++ b/src/serac/numerics/functional/tensor.hpp
@@ -1219,11 +1219,11 @@ SERAC_HOST_DEVICE constexpr auto det(const tensor<T, 3, 3>& A)
 }
 /**
  * @brief Take determinant of a matrix near the identity without losing precision
- * 
+ *
  * detApIm1(A) = det(A + I) - 1
- * When the entries of A are small compared to unity, the straightforward way of 
+ * When the entries of A are small compared to unity, the straightforward way of
  * computing this will suffer from catasrophic cancellation.
- * 
+ *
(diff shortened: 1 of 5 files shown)


src/serac/numerics/functional/tensor.hpp
```diff
@@ -1218,6 +1218,34 @@ SERAC_HOST_DEVICE constexpr auto det(const tensor<T, 3, 3>& A)
...
+
+/// @overload
+template <typename T>
+SERAC_HOST_DEVICE constexpr auto detApIm1(const tensor<T, 3, 3>& A)
+{
+  auto I1 = tr(A);
+  auto I2 = 0.5 * (I1 * I1 - inner(A, transpose(A)));
+  auto I3 = det(A);
+  return I1 + I2 + I3;
+}
```
samuelpmishLLNL: Out of curiosity, I just profiled an alternate implementation of these functions, where it explicitly inlines these invariant calculations, to get a sense of how much of a performance hit there is to use your nice, readable implementation.

```cpp
namespace alternate {
template <typename T>
constexpr auto detApIm1(const tensor<T, 2, 2>& A)
{
  return A(0, 0) - A(0, 1) * A(1, 0) + A(1, 1) + A(0, 0) * A(1, 1);
}

template <typename T> 
constexpr auto detApIm1(const tensor<T, 3, 3>& A)
{
  return A(0, 0) 
       + A(1, 1) 
       + A(2, 2) 
       - A(0, 1) * A(1, 0) 
       + A(0, 0) * A(1, 1) 
       - A(0, 2) * A(2, 0) 
       - A(1, 2) * A(2, 1) 
       + A(0, 0) * A(2, 2) 
       + A(1, 1) * A(2, 2) 
       - A(0, 2) * A(1, 1) * A(2, 0) 
       + A(0, 1) * A(1, 2) * A(2, 0) 
       + A(0, 2) * A(1, 0) * A(2, 1) 
       - A(0, 0) * A(1, 2) * A(2, 1) 
       - A(0, 1) * A(1, 0) * A(2, 2) 
       + A(0, 0) * A(1, 1) * A(2, 2);

}
```

The results on my two machines:

Mac M2
```
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
BM_detApIm1_2x2          0.509 ns        0.509 ns   1000000000
BM_detApIm1_2x2_alt      0.433 ns        0.433 ns   1000000000
BM_detApIm1_3x3           2.17 ns         2.17 ns    326024182
BM_detApIm1_3x3_alt       1.68 ns         1.68 ns    417606281
```


Ubuntu 22.04 w/ intel 13900k
```
--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
BM_detApIm1_2x2          0.397 ns        0.397 ns   1000000000
BM_detApIm1_2x2_alt      0.344 ns        0.344 ns   1000000000
BM_detApIm1_3x3           2.01 ns         2.01 ns    349186377
BM_detApIm1_3x3_alt       1.61 ns         1.61 ns    435050324
```

So, only ~20% faster. Is that enough of a difference to justify the less readable implementation-- I'm not sure.

samuelpmishLLNL (approved):

src/serac/numerics/functional/tensor.hpp
tupek2: seems worth it if well tested to do the faster one.  Its under a function call.  Also, there are some more terms that can be grouped, might be negligibly even faster if fma operations are used.

tupek2 (approved):

src/serac/numerics/functional/tensor.hpp
samuelpmishLLNL: fwiw, the arithmetic instructions for the optimized build (taken from M2 build) are pretty much all fma already:

```
auto alternate::detApIm1<double>(serac::tensor<double, 3, 3> const&):
        ldp     d0, d1, [x0]
        ldp     d2, d3, [x0, #32]
        fadd    d4, d0, d2
        ldp     d17, d5, [x0, #56]
        fadd    d4, d4, d5
        ldp     d7, d6, [x0, #16]
        fmsub   d4, d1, d6, d4
        fmadd   d4, d0, d2, d4
        ldr     d16, [x0, #48]
        fmsub   d4, d7, d16, d4
        fmsub   d4, d3, d17, d4
        fmadd   d4, d0, d5, d4
        fmadd   d4, d2, d5, d4
        fnmul   d18, d7, d2
        fmadd   d4, d18, d16, d4
        fmul    d18, d1, d3
        fmadd   d4, d18, d16, d4
        fmul    d7, d6, d7
        fmadd   d4, d7, d17, d4
        fnmul   d3, d0, d3
        fmadd   d3, d3, d17, d4
        fnmul   d1, d1, d6
        fmadd   d1, d1, d5, d3
        fmul    d0, d0, d2
        fmadd   d0, d0, d5, d1
        ret
```

btalamini: Sam, thanks for the profiling! I agree with Mike: if the functions are tested, might as well pick up the speed. You can always put the neat math in the docstring.

Should I merge as is, or should we switch to the inline versions?

samuelpmishLLNL: I think this one is good to go now, but I don't have the old version of clang-format that appeases our azure pipeline.

PR #1007, opened 2023-10-11
Feature/tupek/solid transient adj

tupek2:
Implementation of math for solid transient adjoints.  Tests cover a lot more situations than thermal, including varying timesteps, nonzero boundary and initial conditions, initial displacement, initial velocity, and shape sensitivities.  There is one HUGE caveat, the initial conditions on a dirichlet constrained node MUST be consistent with the time varying applied boundary conditions.  Without this, with the way we are currently enforcing time dependent boundary conditions, there is a real sensitivity to the initial disp/velo on constrained nodes that the adjoint math does not currently take into account.  We may consider erroring when users apply inconsistent initial and boundary conditions.

The 'interface' I currently have for the current implementation is sub-optimal, but hopefully it is building up another case which can be used to create a more elegant unified design for time integration, adjoint time integration and boundary condition enforcement.

Issues to add after this is through:

1. add warning when users specify initial conditions inconsistent with boundary conditions
2. override initial conditions with boundary conditions for nodes with a conflict
3. thermal parameters adjoint is wrong for nonlinear materials because it uses d_residual_d_ which bakes in the temperature and rate, which is being updated on the reverse pass.  should change d_residual_d_ to take arguments so we can pass the correct temperature
4. thermal needs to account for varying timesteps, and needs testing for nontrivial bcs, initial conditions, etc. also refactor tests a bit to match solid versions
5. add parameter sensitivity tests for solids.  there is a bug in solid_mechanics.hpp at the moment, so need to fix (same error as in thermal)


src/serac/physics/base_physics.cpp
```diff
@@ -279,6 +279,13 @@ void BasePhysics::saveSummary(axom::sidre::DataStore& datastore, const double t)
   }
 }
 
+FiniteElementState BasePhysics::loadCheckpointedState(const std::string& state_name, int /*cycle*/) const
+{
+  return state(state_name);
+}
+
+double BasePhysics::loadCheckpointedTimestep(int /*cycle*/) const { return 0.0; }
```
samuelpmishLLNL, line 287: Do we want these to be pure virtual functions (so you get a compilation error if the derived class doesn't implement)?

samuelpmishLLNL (approved):

src/serac/physics/base_physics.cpp
tupek2: ultimately, yes.  but I wanted to settle on an interface before going through the physics and implementing it for each

src/serac/physics/heat_transfer.hpp
```diff
@@ -695,7 +696,7 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
       // We store the previous timestep's temperature as the current temperature for use in the lambdas computing the
       // sensitivities.
 
-      auto [r, drdu] = (*residual_)(differentiate_wrt(temperature_), zero_, shape_displacement_,
+      auto [_, drdu] = (*residual_)(differentiate_wrt(temperature_), zero_, shape_displacement_,
```
btalamini, line 703: Is this a case for the `[[maybe_unused]]` attribute? I'll leave it to you C++ gurus to decide.

src/serac/physics/base_physics.hpp
```diff
@@ -93,30 +93,44 @@ class BasePhysics {
    * @param state_name The name of the Finite Element State primal solution to retrieve
    * @return The named primal Finite Element State
    */
-  virtual const FiniteElementState& state(const std::string& state_name) = 0;
+  virtual const FiniteElementState& state(const std::string& state_name) const = 0;
```
btalamini, line 96: 👍

btalamini (approved):

samuelpmishLLNL (approved):

src/serac/physics/heat_transfer.hpp
tupek2: it does not warn in this case, so it seems un-needed.  Unless anyone prefers that (not me!), but there clearly are outputs that are used here, so I think that might be why it doesn't warn.

PR #1011, opened 2023-10-12
Clean up the adjoint solve interface

jamiebramwell:
This PR is an effort to simplify the adjoint solver interface in an attempt to bring it closer to the new forward solver interface introduced in https://github.com/LLNL/serac/pull/977 .

Specifically, it:

- Removes the input arguments to the adjoint solve call in favor of a adjoint load set method similar to the forward pass.
- Adds initialization cycle and time information to the base physics constructor for use in restarts.
- Adds a way to set a generic primal solution field to an initial value
- Adds an index-based accessor for the parameter fields
- Removes some leftover dead code.


tupek2 (approved):
We are adding a bit more state, but I think unifying the interface is more important.  2nd thought: we probably don't need to have methods for getting the adjoint solutions.  I'm inclined to keep those internal for now, as I know of no use outside the physics modules, and their definitions could evolve.  Something for a later PR though.

jamiebramwell:
> We are adding a bit more state, but I think unifying the interface is more important. 2nd thought: we probably don't need to have methods for getting the adjoint solutions. I'm inclined to keep those internal for now, as I know of no use outside the physics modules, and their definitions could evolve. Something for a later PR though.

After talking to @kswartz92 , it sounds like direct access of the adjoints is useful for implementing refinement strategies.

jamiebramwell:
@samuelpmishLLNL @btalamini I believe this is aready to go. It should be a short one.

jamiebramwell:
This should contain all of the fixes needed for optimization problems. @sethwatts @kswartz92 @tepperly

samuelpmishLLNL (approved):

PR #1016, opened 2023-10-20
Forward openmp variant to strumpack

jamiebramwell:
Some downstream dependencies turn off openmp, so we should forward that variant to strumpack.


chapman39 (approved):

jamiebramwell:
@white238 , can you take a look at this?

white238 (approved):

PR #1018, opened 2023-10-25
feature: Prepare to introduce RAJA for all loops

johnbowen42:
First, here is some context on this PR.  This PR is the first step of introducing RAJA support to serac.  It is part of the larger PR here [https://github.com/LLNL/serac/pull/987](https://github.com/LLNL/serac/pull/987), and is a subset of that work.  This PR does the following
(1) Eliminates multiple variadic template parameters from the enclosing methods of where RAJA::forall loops will be called.  Introduces workarounds
(2) Adds static asserts on thermal integrands to prevent users from supplying extended lambdas as an integrand
(3) Adds SERAC_HOST_DEVICE annotations to necessary functions
(4) Adds a functor ThermalMaterialIntegrand instead of the generic extended lambda from before

This is a lightweight PR, and is a prerequisite for the RAJA work mentioned above.


jamiebramwell:
@white238 , can you take a look at this?

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -10,7 +10,10 @@
 #include "serac/numerics/functional/function_signature.hpp"
 #include "serac/numerics/functional/differentiate_wrt.hpp"
 
+#include <RAJA/index/RangeSegment.hpp>
+#include <RAJA/RAJA.hpp>
```
white238, line 14:
```suggestion
```

nitpick but RAJA isn't used in this PR

src/serac/numerics/functional/tests/CMakeLists.txt
```diff
@@ -58,4 +58,8 @@ if(ENABLE_CUDA)
     serac_add_tests( SOURCES ${functional_tests_cuda}
                      DEPENDS_ON gtest serac_functional serac_state ${functional_depends} cuda)
 
+            
+    #blt_add_executable(NAME functional_cuda
+    #                   SOURCES functional_cuda.cpp
```
white238: I assume this is remnant of the cherry picking?

src/serac/physics/heat_transfer.hpp
```diff
@@ -285,6 +285,64 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+  template <typename MaterialType>
+  struct ThermalMaterialIntegrand {
+    /**
+     * @brief Construct a ThermalMaterialIntegrand functor with material model of type `MaterialType`.
+     * @param[in] material A functor representing the material model.  Should be a functor, or a class/struct with
+     * public operator() method.  Must NOT be a generic lambda, or serac will not compile due to static asserts below.
+     */
+    ThermalMaterialIntegrand(MaterialType material) : material_(material) {}
+
+    // Due to nvcc's lack of support for generic lambdas (i.e. functions of the form
```
white238: 👍

white238 (approved):

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -153,18 +153,26 @@ auto batch_apply_qf(lambda qf, const tensor<double, 3, n>& positions, const tens
...
 }
 
-template <uint32_t differentiation_index, int Q, mfem::Geometry::Type geom, typename test, typename... trials,
-          typename lambda_type, typename derivative_type, int... indices>
-void evaluation_kernel_impl(FunctionSignature<test(trials...)>, const std::vector<const double*>& inputs,
-                            double* outputs, const double* positions, const double* jacobians, lambda_type qf,
-                            [[maybe_unused]] derivative_type* qf_derivatives, uint32_t num_elements,
-                            std::integer_sequence<int, indices...>)
+template <mfem::Geometry::Type geom, typename test, typename... trials>
+auto get_trial_elements(FunctionSignature<test(trials...)>)
```
white238: Please add a brief comment explaining this. Also is this the best name for this and the next function? you mentioned `get_empty_trial_elements` when we chatted.

src/serac/numerics/functional/tests/CMakeLists.txt
johnbowen42: Uh oh yes, will delete

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -134,44 +137,53 @@ auto batch_apply_qf(lambda qf, const tensor<double, dim, n> x, qpt_data_type* qp
...
+auto get_trial_elements(FunctionSignature<test(trials...)>)
+{
+  return tuple<finite_element<geom, trials>...>{};
+}
+
+template <mfem::Geometry::Type geom, typename test, typename... trials>
+auto get_test(FunctionSignature<test(trials...)>)
+{
+  return finite_element<geom, test>{};
+}
```
samuelpmishLLNL: it looks like these definitions are duplicated in the domain/boundary integral headers

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -134,44 +137,53 @@ auto batch_apply_qf(lambda qf, const tensor<double, dim, n> x, qpt_data_type* qp
...
       reinterpret_cast<const typename decltype(type<indices>(trial_elements))::dof_type*>(inputs[indices])...};
 
   // for each element in the domain
-  for (uint32_t e = 0; e < num_elements; e++) {
+  for (uint32_t e = 0; e < num_elements; ++e) {
     // load the jacobians and positions for each quadrature point in this element
     auto J_e = J[e];
     auto x_e = x[e];
 
+    [[maybe_unused]] static constexpr trial_element_type empty_trial_element{};
```
samuelpmishLLNL: the prefix "empty_" is potentially misleading here, making it sound as though the object is incomplete or uninitialized

Also, this `trial_element_type` is actually a tuple, so it would be clearer (to me) if its name was `trial_element_types` or `trial_elements_tuple` to convey that

src/serac/physics/heat_transfer.hpp
samuelpmishLLNL: this comment is misleading, as `nvcc` does support generic lambdas, and it also supports their use in cuda kernels. For example,

https://godbolt.org/z/TM8zacGcM

What it doesn't support is the corner case of generic, extended lambdas.

src/serac/physics/heat_transfer.hpp
```diff
@@ -285,6 +285,64 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+    // auto lambda = [](auto) {}; ), this cannot be allowed.  In order for this code
+    // to be portable to CUDA platforms, the asserts below prevent serac from compiling
+    // if such a lambda is supplied.
+  private:
+    class DummyArgumentType {};
+    static_assert(!std::is_invocable<MaterialType, DummyArgumentType&>::value);
+    static_assert(!std::is_invocable<MaterialType, DummyArgumentType*>::value);
+    static_assert(!std::is_invocable<MaterialType, DummyArgumentType>::value);
```
samuelpmishLLNL: It's not clear to me what this is checking. If we want to detect for generic extended lambdas, then we have a compiler specific intrinsic:

https://gist.github.com/samuelpmish/5679a488c8724b215e96e107569b4fe9#file-main-cu-L49

johnbowen42: this static assert will pass if given a functor but fail if given a generic extended lambda, you can try it out in godbolt if you want.  This is not compiler specific, which in my opinion is better in this case because we want the option to be able to compile exclusively with clang or gcc

src/serac/numerics/functional/domain_integral_kernels.hpp
johnbowen42: sure we can change the name to make it more clear that we are deducing a type of tuple

src/serac/numerics/functional/domain_integral_kernels.hpp
johnbowen42: Which header would you like to see this in?

src/serac/numerics/functional/boundary_integral_kernels.hpp
johnbowen42: Will do

src/serac/physics/heat_transfer.hpp
samuelpmishLLNL: > this static assert will pass if given a functor but fail if given a generic extended lambda, you can try it out in godbolt if you want. 

it incorrectly disallows functors with generic `operator()` definitions: https://godbolt.org/z/7nhEc1jYY

it incorrectly disallows non-extended lambdas: https://godbolt.org/z/rKrTcovcj

NVIDIA wrote a dedicated compiler intrinsic for checking exactly what we want to know, why not use it?

-----------

> This is not compiler specific, which in my opinion is better 

It's important to remember that the underlying limitation here is compiler specific, so imposing the restriction on all compilers isn't a benefit.

For instance, a user who wants to run their simulations on a CPU can (and probably should) use generic lambdas for their material / traction definitions, where appropriate. It doesn't seem right to take away a useful feature from _all_ users, because one compiler doesn't support it.

------------

> we want the option to be able to compile exclusively with clang or gcc

Then we can guard the assertion behind conditional compilation checks that are only enabled for the compiler that cares

```
    #if __NVCC__
    static_assert(!__nv_is_extended_device_lambda_closure_type(MaterialType), "error: serac doesn't support extended lambdas with nvcc");
    #endif
```

see: https://godbolt.org/z/9n9oxWbeb

johnbowen42: > NVIDIA wrote a dedicated compiler intrinsic for checking exactly what we want to know, why not use it?

If we guard the static assert behind an ifdef, users will be able to write generic lambdas for CPU versions of serac.  Then, if they want to switch to using GPU-enabled serac, their code will require refactoring.  In my opinion it is better to simplify the requirements to users, and force them to write templated functors.  I suppose the generic functor error is a limitation, but I feel that putting the static assert behind the if guard will only complicate things for users and make serac seem less portable. 

Ultimately, we want serac to be portable to CPU and GPU platforms.  We want user code to be portable to CPU and GPU platforms as well.  By forcing users to write templated functors, we are providing users with simplicity and guarantee of portability.

johnbowen42: I just spoke offline with @jamiebramwell and she said that still allowing generic lambdas for CPU generic lambdas would permit faster development/be more convenient for users.  I think this is a great point so we will go for @samuelpmishLLNL 's approach

white238: We should document this difference very thoroughly in Doxygen and in Sphinx. Maybe even providing an example on how to convert from generic lambdas to the functor.

We should also be very careful in our unit testing. Defaulting to the non-lambda version should be the normal, but also add a couple test cases to ensure the generic CPU lambdas continue to work.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -153,18 +153,14 @@ auto batch_apply_qf(lambda qf, const tensor<double, 3, n>& positions, const tens
...
-          typename lambda_type, typename derivative_type, int... indices>
-void evaluation_kernel_impl(FunctionSignature<test(trials...)>, const std::vector<const double*>& inputs,
+/// @trial_elements the element type for each trial space
+template <uint32_t differentiation_index, int Q, mfem::Geometry::Type geom, typename test_element,
+          typename trial_element_type, typename lambda_type, typename derivative_type, int... indices>
+void evaluation_kernel_impl(trial_element_type trial_elements, test_element, const std::vector<const double*>& inputs,
                             double* outputs, const double* positions, const double* jacobians, lambda_type qf,
                             [[maybe_unused]] derivative_type* qf_derivatives, uint32_t num_elements,
-                            std::integer_sequence<int, indices...>)
+                            camp::int_seq<int, indices...>)
```
jamiebramwell, line 162: Is `std::integer_sequence` not compatible with RAJA?

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -134,44 +137,42 @@ auto batch_apply_qf(lambda qf, const tensor<double, dim, n> x, qpt_data_type* qp
   return outputs;
 }
 
-template <uint32_t differentiation_index, int Q, mfem::Geometry::Type geom, typename test, typename... trials,
-          typename lambda_type, typename state_type, typename derivative_type, int... indices>
-void evaluation_kernel_impl(FunctionSignature<test(trials...)>, const std::vector<const double*>& inputs,
-                            double* outputs, const double* positions, const double* jacobians, lambda_type qf,
+template <uint32_t differentiation_index, int Q, mfem::Geometry::Type geom, typename test_element_type,
+          typename trial_element_tuple_type, typename lambda_type, typename state_type, typename derivative_type,
```
jamiebramwell, line 141: This is a nice improvement!

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -265,14 +268,14 @@ void action_of_gradient_kernel(const double* dU, double* dR, derivatives_type* q
   using test_element  = finite_element<g, test>;
   using trial_element = finite_element<g, trial>;
 
-  static constexpr bool is_QOI   = (test::family == Family::QOI);
-  static constexpr int  num_qpts = num_quadrature_points(g, Q);
+  const bool is_QOI   = (test::family == Family::QOI);
```
jamiebramwell: Why are these no longer `constexpr`?

src/serac/numerics/functional/function_signature.hpp
```diff
@@ -17,5 +19,34 @@ struct FunctionSignature<output_type(input_types...)> {
...
+ * Expansion of the template parameter trials must occur in a separate template
+ * function from evaluation_kernel_impl.  The __host__ __device__ marker on the
+ * lambda declared in evaluation_kernel_impl prevents use of multiple variadic
+ * template parameters in evaluation_kernel_impl.  See section 14.7.3 of the
+ * CUDA programming guide, item 9.
+ */
+template <mfem::Geometry::Type geom, typename test, typename... trials>
+auto trial_elements_tuple(FunctionSignature<test(trials...)>)
+{
+  return serac::tuple<serac::finite_element<geom, trials>...>{};
```
jamiebramwell, line 38: At some point we should revisit `camp::tuple` @white238

src/serac/physics/heat_transfer.hpp
jamiebramwell: I agree, this should be specified for extended lambdas.

jamiebramwell (commented):
Excellent work @johnbowen42 !

src/serac/physics/heat_transfer.hpp
jamiebramwell: Just commenting here that I agree we should allow generic lambdas for rapid prototyping on non-nvcc compilers. We should just make sure that the vast majority of our examples and tests use the functor syntax as @white238 states.

src/serac/numerics/functional/boundary_integral_kernels.hpp
johnbowen42: `camp::int_seq` works better with device code

src/serac/numerics/functional/domain_integral_kernels.hpp
johnbowen42: I changed them back.  They cannot be static and be captured by device lambda but can be constexpr

jamiebramwell (approved):

PR #1021, opened 2023-10-30
Additional contact examples

ebchin:
Adds more contact examples including
- beam bending with active set
- ironing
- twisting cubes
- twisting spheres with active set and NURBS sphere mesh

See the slides [here](https://mfem.org/pdf/workshop23/19_Chin_Tribol.pdf) to see these examples in action.


examples/contact/beam_bending.cpp
```diff
@@ -0,0 +1,112 @@
...
+  solid_solver.setParameter(0, K_field);
+
+  serac::FiniteElementState G_field(serac::StateManager::newState(serac::L2<0>{}, "shear_mod", "beam_mesh"));
+  mfem::Vector              G_values({0.25, 0.25, 2.5});
+  mfem::PWConstCoefficient  G_coeff(G_values);
+  G_field.project(G_coeff);
+  solid_solver.setParameter(1, G_field);
+
+  serac::solid_mechanics::ParameterizedNeoHookeanSolid<dim> mat{1.0, 0.0, 0.0};
+  solid_solver.setMaterial(serac::DependsOn<0, 1>{}, mat);
```
samuelpmishLLNL: can we add some description of these different coefficient values? Are there 3 bodies in contact or just 2 (and if so, which values correspond to which bodies)?

This is good motivation for us to adopt the `Domain` at the physics module level

examples/contact/beam_bending.cpp
```diff
@@ -0,0 +1,112 @@
...
+  });
+  double time = 0.0;
+  solid_solver.setDisplacementBCs({6}, [&time](const mfem::Vector&, mfem::Vector& u) {
+    u.SetSize(dim);
+    u    = 0.0;
+    u[2] = -0.05 * time;
+  });
+
+  // Add the contact interaction
+  solid_solver.addContactInteraction(0, {7}, {5}, contact_options);
```
samuelpmishLLNL: a lot of magic numbers here-- what are 0, {7} and {5} communicating?

examples/contact/ironing.cpp
```diff
@@ -0,0 +1,117 @@
...
+    if (time <= 2.0 + 1.0e-12) {
+      u[2] = -time * 0.15;
+    } else {
+      u[0] = -(time - 2.0) * 0.25;
+      u[2] = -0.3;
+    }
+  });
+
+  // Add the contact interaction
+  solid_solver.addContactInteraction(0, {6}, {11}, contact_options);
```
samuelpmishLLNL: what are 0, {6}, {11}?

examples/contact/ironing.cpp
```diff
@@ -0,0 +1,117 @@
...
+  solid_solver.setDisplacementBCs({12}, [&time](const mfem::Vector&, mfem::Vector& u) {
+    u.SetSize(dim);
+    u = 0.0;
+    if (time <= 2.0 + 1.0e-12) {
+      u[2] = -time * 0.15;
+    } else {
+      u[0] = -(time - 2.0) * 0.25;
+      u[2] = -0.3;
+    }
+  });
```
samuelpmishLLNL: we need to start using `time` through a function argument rather than a capture but I'm not sure if we hooked that up to the boundary conditions yet.

samuelpmishLLNL (approved):

examples/contact/ironing.cpp
ebchin: Looks like its there; I saw a version of `setDisplacementBCs` where the `std::function` takes time as an input parameter.  I switched the examples to use this.

src/serac/physics/contact/contact_data.cpp
```diff
@@ -323,10 +322,15 @@ void ContactData::updateDofOffsets() const
   }
   if (HYPRE_AssumedPartitionCheck()) {
     auto total_dofs = global_pressure_dof_offsets_[global_pressure_dof_offsets_.Size() - 1];
-    global_pressure_dof_offsets_.SetSize(3);
+    if (mesh_.GetNRanks() < 3) {
+      global_pressure_dof_offsets_.SetSize(3);
```
jamiebramwell, line 327: I'm confused what this check is for. Could you add a comment?

examples/contact/beam_bending.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "serac/mesh/mesh_utils.hpp"
+#include "serac/physics/state/state_manager.hpp"
+#include "serac/physics/materials/parameterized_solid_material.hpp"
+#include "serac/serac_config.hpp"
+
+int main(int argc, char* argv[])
+{
+  serac::initialize(argc, argv);
+
+  // NOTE: p must be equal to 1
```
jamiebramwell: I would add a comment that this restriction is due to Tribol capabilities (at least I think it is)

examples/contact/beam_bending.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+  // Construct the appropriate dimension mesh and give it to the data store
+  std::string filename = SERAC_REPO_DIR "/data/meshes/beam-hex-with-contact-block.mesh";
+
+  auto mesh = serac::mesh::refineAndDistribute(serac::buildMeshFromFile(filename), 2, 0);
+  serac::StateManager::setMesh(std::move(mesh), "beam_mesh");
+
+  serac::LinearSolverOptions linear_options{.linear_solver = serac::LinearSolver::Strumpack, .print_level = 1};
+#ifndef MFEM_USE_STRUMPACK
+  SLIC_INFO_ROOT("Contact requires MFEM built with strumpack.");
```
jamiebramwell, line 43: Why is this the case? Does SuperLU not work? I can imagine the iterative solvers don't work well.

examples/contact/beam_bending.cpp
```diff
@@ -0,0 +1,120 @@
...
+  serac::FiniteElementState G_field(serac::StateManager::newState(serac::L2<0>{}, "shear_mod", "beam_mesh"));
+  // each vector value corresponds to a different element attribute:
+  // [0] (element attribute 1) : the beam
+  // [1] (element attribute 2) : indenter block
+  mfem::Vector             G_values({0.25, 2.5});
+  mfem::PWConstCoefficient G_coeff(G_values);
+  G_field.project(G_coeff);
+  solid_solver.setParameter(1, G_field);
+
+  serac::solid_mechanics::ParameterizedNeoHookeanSolid<dim> mat{1.0, 0.0, 0.0};
```
jamiebramwell, line 80: Any reason you are using the parameterized material model? I don't think the sensitivity/adjoint analysis works with this yet.

ebchin: Not in particular.  It seemed like an easy way to get spatially dependent material properties.  The example works better with a stiffer indenter.  If there's a more compatible way to accomplish this, I'd be happy to make changes :)

examples/contact/beam_bending.cpp
ebchin: I had some issues with SuperLU in the tests where if the matrix was resized, then SuperLU wouldn't work.  This is an issue in `SolidMechanics::warmStartDisplacement()` where only the displacements are updated without the LM block.  I only tested with strumpack and SuperLU, but if MUMPS/Pardiso/etc. are added to serac in the future, this macro may need to be updated...

src/serac/physics/contact/contact_data.cpp
ebchin: Thanks for bringing attention to his, it looks like this should be 2 instead of 3.  This came up in single rank testing: the size of global_pressure_dof_offsets_ is two in this case and the code tries to access the third entry.  I added a comment to clarify what's going on here.

jamiebramwell: I don't see the change. Do you need to push?

examples/contact/beam_bending.cpp
jamiebramwell: Nope - that seems like a reasonable explanation! We have most of the stuff to have multiple materials now, so that should be a follow-on PR.

examples/contact/beam_bending.cpp
jamiebramwell: I'll add a issue to track this.

examples/contact/ironing.cpp
jamiebramwell: I believe these comments still stand. Maybe a picture or comments stating what these parameters are would help?

src/serac/physics/contact/contact_data.cpp
ebchin: Not sure why github isn't showing the changes, but it looks like they're in the branch?  https://github.com/LLNL/serac/blob/feature/chin23/new_contact_examples/src/serac/physics/contact/contact_data.cpp#L326-L335

examples/contact/ironing.cpp
ebchin: It looks like that code is outdated.  I believe it's been fixed: https://github.com/LLNL/serac/blob/feature/chin23/new_contact_examples/src/serac/physics/contact/contact_data.cpp#L326-L335

jamiebramwell (approved):
Great work @ebchin !

PR #1022, opened 2023-10-31
Additional base physics capability

jamiebramwell:
This adds two `BasePhysics` capabilities as needed by downstream users:

- Adds a way to set a generic primal solution field to an initial value
- Adds an index-based accessor for the parameter fields


jamiebramwell:
This should be ready to go.

jamiebramwell:
Combined with https://github.com/LLNL/serac/pull/1011

PR #1025, opened 2023-11-06
Update tribol

ebchin:
Update to the latest Tribol (bugfixes for contact examples and improved logging).

Also, this is the first time building on rzgenie with TOSS 4, so a new devtools build was required on quartz and rzgenie.  The python version is slightly changed, so the TOSS 4 spack config has been updated.


host-configs/docker/gcc@11.1.0.cmake
```diff
@@ -97,11 +97,13 @@ set(SUPERLUDIST_DIR "${TPL_ROOT}/superlu-dist-8.1.2-i66ldlr2l325korbopnlirbk53fz
 
 set(RAJA_DIR "${TPL_ROOT}/raja-2022.10.5-6hoezdnmcqqaghnhyqqfjq7kgtcbmxab" CACHE PATH "")
 
+# STRUMPACK not built
```
chapman39, line 100: Is strumpack not supposed to be built on blueos and docker containers? @jamiebramwell

chapman39 (approved):

host-configs/docker/gcc@11.1.0.cmake
jamiebramwell: We build and test it on LC.

jamiebramwell (approved):

PR #1026, opened 2023-11-06
feature: add RAJA kernel launches and basic CUDA support

johnbowen42:
As of 2/8, this the main execution loops in `domain_integral_kernels.hpp` are compiling, passing very basic unit tests, and using a reasonable amount of memory for hexahedron or cubic element types.  A couple main things remain to be done on this PR:

- [x] Refactor `boundary_integral_kernels.hpp` by copying the existing RAJA implementation inside `domain_integral_kernels.hpp`
- [x] Ensure existing CPU unit tests still pass


host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.cmake
samuelpmishLLNL, line 1: can the commented lines in this file be removed, or are they left there to show alternate options?

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,56 +164,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+
+  using qf_inputs_type = decltype(tuple{promote_each_to_dual_when<indices == differentiation_index>(
+      get<indices>(empty_trial_element).template interpolate_output_helper<Q>())...});
+
+#ifdef USE_CUDA
+  auto&           rm             = umpire::ResourceManager::getInstance();
+  auto            dest_allocator = rm.getAllocator("DEVICE");
+  qf_inputs_type* qf_inputs =
+      static_cast<qf_inputs_type*>(dest_allocator.allocate(sizeof(qf_inputs_type) * num_elements));
+  interpolate_out_type* interpolate_result =
```
samuelpmishLLNL: There's a lot of conditional compilation going on here that makes it harder to read these kernels-- and they were already pretty hard to read when I wrote them initially!

Can we investigate an alternate implementation that takes the device type as a template parameter, and use static polymorphism to specialize implementation details for each device instead of conditional compilation macros?

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,56 +164,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+      static_cast<interpolate_out_type*>(dest_allocator.allocate(sizeof(interpolate_out_type) * num_elements));
```
samuelpmishLLNL: It's not clear to me why there are additional memory allocations for these kernels, the output buffers should already have the appropriate size.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,56 +164,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+
+  // It's safer to copy the raw POD type using Umpire than J_Type.  const_cast is needed to be compatible with
+  // the Umpire's API.
+  auto device_jacobians = copy_data(const_cast<double*>(jacobians), sizeof(J_Type) * num_elements, "DEVICE");
+  auto device_positions = copy_data(const_cast<double*>(positions), sizeof(X_Type) * num_elements, "DEVICE");
```
samuelpmishLLNL: We don't want to be copying data back and forth between host and device here. Instead, just put the `jacobians` and `positions` on the device as soon as they're calculated and these copies can be omitted.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,56 +164,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+  // Reinterpret these pointers to enable simpler indexing etc.
+  auto device_J = reinterpret_cast<const J_Type*>(device_jacobians);
+  auto device_x = reinterpret_cast<const X_Type*>(device_jacobians);
+
+  auto device_r = copy_data(r, serac::size(*r) * sizeof(double), "DEVICE");
```
samuelpmishLLNL: similarly, it's preferable to make this kernel just take a pointer to device memory

CMakeLists.txt
```diff
@@ -131,6 +140,7 @@ include(${PROJECT_SOURCE_DIR}/cmake/thirdparty/SetupSeracThirdParty.cmake)
 
 set(CMAKE_C_FLAGS   ${SERAC_CMAKE_C_FLAGS})
 set(CMAKE_CXX_FLAGS ${SERAC_CMAKE_CXX_FLAGS})
+string(REPLACE " -Werror" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
```
white238: This should be undone before merging

CMakeLists.txt
```diff
@@ -22,8 +22,17 @@ project(serac LANGUAGES CXX C)
 # MPI is required in Serac.
 set(ENABLE_MPI ON CACHE BOOL "")
 
+
+if (ENABLE_CUDA)
```
white238: I dont think this is needed, if it is then probably only in the codevelop  build.

cmake/SeracMacros.cmake
```diff
@@ -148,13 +151,21 @@ macro(serac_add_tests)
...
+        endif()
 
         blt_add_executable(NAME        ${test_name}
                            SOURCES     ${filename}
                            OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
                            DEPENDS_ON  ${arg_DEPENDS_ON}
                            FOLDER      serac/tests )
 
+        if (DEFINED arg_USE_CUDA)
+            target_compile_definitions(${test_name} PUBLIC ENABLE_CUDA)
```
white238: What is the reason behind defining both of these?

johnbowen42: Only the second `USE_CUDA` is actually necessary.  This is the keyword I used to toggle using CUDA for execution vs not

johnbowen42: Good point

cmake/thirdparty/SetupSeracThirdParty.cmake
```diff
@@ -509,7 +509,7 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
         # This flag is empty due to us not enabling fortran but we need to strip it
         # so it doesn't propagate in our project
         if("${OpenMP_Fortran_FLAGS}" STREQUAL "")
-            set(OpenMP_Fortran_FLAGS "$<$<NOT:$<COMPILE_LANGUAGE:Fortran>>:-fopenmp=libomp>;$<$<COMPILE_LANGUAGE:Fortran>:-fopenmp>")
+            set(OpenMP_Fortran_FLAGS "-fopenmp")#"$<$<NOT:$<COMPILE_LANGUAGE:Fortran>>:-fopenmp=libomp>;$<$<COMPILE_LANGUAGE:Fortran>:-fopenmp>")
```
white238: Please try this without this. I am pretty sure this will break other cases but ultimately this will be removed in the new BLT+Axom+RAJA and friends update that is goin on now.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,56 +164,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+
+              (promote_each_to_dual_when<indices == differentiation_index>(
+                   get<indices>(interpolate_result[elements[e]]), &get<indices>(qf_inputs[elements[e]]), ctx),
+               ...);
+
+              ctx.teamSync();
+
+              // (batch) evalute the q-function at each quadrature point
+              auto qf_outputs =
+                  batch_apply_qf(qf, t, device_x[e], device_J[e], get<indices>(qf_inputs[elements[e]])...);
```
samuelpmishLLNL: If I'm reading this right, each CUDA thread is evaluating the q-function for every quadrature point in the element, which is not what we want. Instead, each CUDA thread should act on only one quadrature point, and write its outputs to a `__shared__` buffer, like the other steps. 

This also means that the launch parameters for the kernel need to match the number of quadrature points in the element (as opposed to just hardcoding it to BLOCK_SZ=128 right now).

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.cmake
```diff
@@ -69,29 +71,34 @@ set(BLT_MPI_COMMAND_APPEND "mpibind" CACHE STRING "")
 # Cuda
 #------------------------------------------------
 
-set(CUDAToolkit_ROOT "/usr/tce/packages/cuda/cuda-11.2.0" CACHE PATH "")
+set(CUDAToolkit_ROOT "/usr/tce/packages/cuda/cuda-12.0.0" CACHE PATH "")
```
white238: updating to CUDA 12 is now able to happen after Spack fixed a bug i reported. We should update to CUDA 12 in a separate PR.

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.cmake
```diff
@@ -69,29 +71,34 @@ set(BLT_MPI_COMMAND_APPEND "mpibind" CACHE STRING "")
...
+#set(CUDA_TOOLKIT_ROOT_DIR "/usr/tce/packages/cuda/cuda-10.1.105" CACHE PATH "")
 
 set(CMAKE_CUDA_ARCHITECTURES "70" CACHE STRING "")
 
 set(ENABLE_OPENMP ON CACHE BOOL "")
 
 set(ENABLE_CUDA ON CACHE BOOL "")
 
-set(CMAKE_CUDA_SEPARABLE_COMPILATION ON CACHE BOOL "")
+set(ENABLE_CLANG_CUDA OFF CACHE BOOL "")
```
white238: Please remove any setting of clang cuda

host-configs/lassen-blueos_3_ppc64le_ib_p9-clang@10.0.1.cmake
```diff
@@ -69,29 +71,34 @@ set(BLT_MPI_COMMAND_APPEND "mpibind" CACHE STRING "")
...
-set(CMAKE_CUDA_ARCHITECTURES "70" CACHE STRING "")
+set(CMAKE_CUDA_SEPARABLE_COMPILATION ON CACHE BOOL "")
+
+#set(CMAKE_CUDA_FLAGS " --expt-extended-lambda --expt-relaxed-constexpr " CACHE STRING "")
+#set(CMAKE_CUDA_FLAGS "-fopenmp" CACHE STRING "")
 
 # nvcc does not like gtest's 'pthreads' flag
 
-set(gtest_disable_pthreads ON CACHE BOOL "")
+set(gtest_disable_pthreads OFF CACHE BOOL "")
```
white238: Did this get fixed? historicaly nvcc hated gtests pthreads

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,56 +164,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
-    }
+  static constexpr trial_element_type empty_trial_element{};
+
+  using interpolate_out_type =
+      decltype(tuple{get<indices>(empty_trial_element).template interpolate_output_helper<Q>()...});
+
+  using qf_inputs_type = decltype(tuple{promote_each_to_dual_when<indices == differentiation_index>(
+      get<indices>(empty_trial_element).template interpolate_output_helper<Q>())...});
+
+#ifdef USE_CUDA
```
white238: There is a SERAC_USE_CUDA already defined in https://github.com/LLNL/serac/blob/7672d330192aec5f4eb46ef48e0f5ec713b203f7/src/serac/serac_config.hpp.in#L28C14-L28C28

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,56 +164,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+        RAJA::loop<teams_e>(
+            ctx, e_range,
+            [&ctx, t, device_J, device_x, u, qf, qpts_per_elem, rule, device_r, elements, qf_derivatives, qf_inputs,
+             interpolate_result](uint32_t e) {
+              static constexpr trial_element_type empty_trial_element{};
+              // batch-calculate values / derivatives of each trial space, at each quadrature point
+              (get<indices>(empty_trial_element)
+                   .interpolate(get<indices>(u)[elements[e]], rule, &get<indices>(interpolate_result[elements[e]]),
+                                ctx),
+               ...);
```
samuelpmishLLNL: `interpolate_result` needs to be in `__shared__` memory for performance reasons, so it should be either statically allocated in the kernel body, or dynamically allocated through the kernel launch parameters (and then aliasing into that buffer)

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -257,22 +346,53 @@ void action_of_gradient_kernel(const double* dU, double* dR, derivatives_type* q
...
+      static_cast<qf_inputs_type*>(host_allocator.allocate(sizeof(qf_inputs_type) * num_elements));
+#endif
+  rm.memset(qf_inputs, 0);
+
+#if defined(USE_CUDA)
+  using teams_e       = RAJA::LoopPolicy<RAJA::cuda_block_x_direct>;
+  using launch_policy = RAJA::LaunchPolicy<RAJA::cuda_launch_t<false>>;
+#else
+  using teams_e                    = RAJA::LoopPolicy<RAJA::seq_exec>;
+  using launch_policy              = RAJA::LaunchPolicy<RAJA::seq_launch_t>;
```
samuelpmishLLNL: same comments as above for the evaluation kernel apply here. 
I don't believe we don't need allocations / copies here

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -257,22 +346,53 @@ void action_of_gradient_kernel(const double* dU, double* dR, derivatives_type* q
...
-  }
+  RAJA::launch<launch_policy>(
+      RAJA::LaunchParams(RAJA::Teams(num_elements), RAJA::Threads(BLOCK_X, BLOCK_Y, BLOCK_Z)),
+      [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+        RAJA::loop<teams_e>(ctx, e_range, [du, rule, &ctx, qf_inputs, elements, qf_derivatives, dr](int e) {
+          // (batch) interpolate each quadrature point's value
+          trial_element::interpolate(du[elements[e]], rule, qf_inputs, ctx);
+
+          // (batch) evalute the q-function at each quadrature point
+          auto qf_outputs = batch_apply_chain_rule(qf_derivatives + e * nqp, *qf_inputs, ctx);
```
samuelpmishLLNL: see earlier comment re: `batch_apply_qf`

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -303,24 +423,47 @@ void element_gradient_kernel(ExecArrayView<double, 3, ExecutionSpace::CPU> dK, d
...
+
+          tensor<derivatives_type, nquad> derivatives{};
+          RAJA::RangeSegment              x_range(0, nquad);
+          RAJA::loop<threads_x>(ctx, x_range, [&](int q) { derivatives(q) = qf_derivatives[e * nquad + uint32_t(q)]; });
+
+          ctx.teamSync();
+
+          for (int J = 0; J < trial_element::ndof; J++) {
+            RAJA_TEAM_SHARED decltype(trial_element::batch_apply_shape_fn(J, derivatives, rule, ctx)) source_and_flux;
+            source_and_flux = trial_element::batch_apply_shape_fn(J, derivatives, rule, ctx);
```
samuelpmishLLNL: using a naked `operator=` to a shared memory buffer is a block-wide race condition in a CUDA kernel.

In this case, each thread is performing redundant calculations, so it should just be a local 128x performance hit rather than a correctness problem, but not what we want!

Instead, `batch_apply_shape_fn` needs to return by writing its output to a pointer to `__shared__` memory, like `batch_interpolate` and `batch_integrate`.

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -144,30 +147,42 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
+#ifdef USE_CUDA
+    using threads_x [[maybe_unused]] = RAJA::LoopPolicy<RAJA::cuda_thread_x_direct>;
+#else
+    using threads_x [[maybe_unused]] = RAJA::LoopPolicy<RAJA::seq_exec>;
+#endif
+    auto x_range = RAJA::RangeSegment(0, q * q * q);
+    RAJA::loop<threads_x>(ctx, x_range, [&](int tid) {
+      int qx = tid % q;
+      int qy = (tid / q) % q;
+      int qz = (tid / (q * q)) % q;
```
samuelpmishLLNL: the `%q` for qz isn't necessary

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -144,30 +147,42 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
+      double              phi_j      = B(qx, jx) * B(qy, jy) * B(qz, jz);
+      tensor<double, dim> dphi_j_dxi = {G(qx, jx) * B(qy, jy) * B(qz, jz), B(qx, jx) * G(qy, jy) * B(qz, jz),
+                                        B(qx, jx) * B(qy, jy) * G(qz, jz)};
 
-          int   Q   = (qz * q + qy) * q + qx;
-          auto& d00 = get<0>(get<0>(input(Q)));
-          auto& d01 = get<1>(get<0>(input(Q)));
-          auto& d10 = get<0>(get<1>(input(Q)));
-          auto& d11 = get<1>(get<1>(input(Q)));
+      int   Q   = (qz * q + qy) * q + qx;
```
samuelpmishLLNL: `Q` is the same as `tid` here

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -144,30 +147,42 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
 
     return output;
   }
 
   template <int q>
-  SERAC_HOST_DEVICE static auto interpolate(const dof_type& X, const TensorProductQuadratureRule<q>&)
+  static auto interpolate_output_helper()
+  {
+    return tensor<qf_input_type, q * q * q>{};
+  }
```
samuelpmishLLNL: if this is used for deducing types, then something like

```cpp
using batch_qf_input_type = tensor<qf_input_type, q * q * q>;
```

is cleaner

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -204,85 +228,195 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
+    RAJA_TEAM_SHARED tensor<double, c, q, q, q> value;
+    RAJA_TEAM_SHARED tensor<double, c, dim, q, q, q> gradient;
+    constexpr auto                                   B = calculate_B<apply_weights, q>();
+    constexpr auto                                   G = calculate_G<apply_weights, q>();
+    for (int i = 0; i < c; i++) {
+      RAJA_TEAM_SHARED decltype(deduce_contract_return_type<2, 1>(X[i], B)) A10;
+      RAJA_TEAM_SHARED decltype(deduce_contract_return_type<2, 1>(X[i], G)) A11;
+      RAJA_TEAM_SHARED decltype(deduce_contract_return_type<1, 1>(A10, B))  A20;
+      RAJA_TEAM_SHARED decltype(deduce_contract_return_type<1, 1>(A11, B))  A21;
+      RAJA_TEAM_SHARED decltype(deduce_contract_return_type<1, 1>(A10, G))  A22;
```
samuelpmishLLNL: This is considerably more shared memory than is needed. 
Since these arrays are not all in use at the same time, we can get away with a smaller allocation. See the snippets in https://github.com/LLNL/serac/files/12482282/mfem_dynamic_kernel_refactor.pdf section "An alternative implementation" for details.

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -204,85 +228,195 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
+        contract<2, 1>(X[i], G, &A11, qx, qy, qz);
+
+        ctx.teamSync();
+        contract<1, 1>(A10, B, &A20, qx, qy, qz);
+
+        ctx.teamSync();
+        contract<1, 1>(A11, B, &A21, qx, qy, qz);
+
+        ctx.teamSync();
+        contract<1, 1>(A10, G, &A22, qx, qy, qz);
```
samuelpmishLLNL: I think this should be refactored, there's a lot of implementation details leaking through the interface here (e.g. passing qx, qy, qz into the contract function, synchronization, etc).

Hardcoding BLOCK_X, BLOCK_Y and BLOCK_Z doesn't make sense to me, those values depend on the quadrature rule and polynomial order, and are different for each contraction. I believe I sent you some code for this earlier, maybe we can discuss this in more detail.

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -204,85 +228,195 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
+
+        ctx.teamSync();
+      });
+    }
+
+#endif
     // transpose the quadrature data into a flat tensor of tuples
+#ifdef USE_CUDA
+    RAJA_TEAM_SHARED
+#endif
```
samuelpmishLLNL: usually, these macros already include the cuda conditional compilation bits, so the user code doesn't have to have write them in their code. Does RAJA not do this for RAJA_TEAM_SHARED?

src/serac/numerics/functional/detail/hexahedron_H1.inl
samuelpmishLLNL, line 1: There's a lot of RAJA implementation leaking into the element library, which I don't believe is necessary. Also, the conditional compilation macros appearing in each element is unpleasant-- I think the only routines that potentially need it are the contractions (which could also switch based on a device template parameter mentioned earlier).

samuelpmishLLNL:
I only made it through the boundary integral kernels and H1_hexahedron element, but I imagine most of the comments in those two files apply to the domain integral kernels and other element types as well

johnbowen42:
> I only made it through the boundary integral kernels and H1_hexahedron element, but I imagine most of the comments in those two files apply to the domain integral kernels and other element types as well

This PR does not implement GPU versions of interpolate and integrate for the other element types

src/serac/numerics/functional/boundary_integral_kernels.hpp
jamiebramwell: I agree this is a good thing to do. However, in an effort to get this merged I say we table this for a follow on PR. Thoughts @white238 ?

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,62 +165,145 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+  auto&           rm             = umpire::ResourceManager::getInstance();
+  auto            dest_allocator = rm.getAllocator("DEVICE");
+  qf_inputs_type* qf_inputs =
+      static_cast<qf_inputs_type*>(dest_allocator.allocate(sizeof(qf_inputs_type) * num_elements));
+  interpolate_out_type* interpolate_result =
+      static_cast<interpolate_out_type*>(dest_allocator.allocate(sizeof(interpolate_out_type) * num_elements));
+
+  auto device_r = copy_data(r, serac::size(*r) * sizeof(double), "DEVICE");
+
+  printCUDAMemUsage();
```
white238: This should be removed before this gets merged.

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -204,85 +227,193 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
       }
     }
   }
 
 #if 0
 
   template <int q>
-  static SERAC_DEVICE auto interpolate(const dof_type& X, const tensor<double, dim, dim>& J,
+  static SERAC_DEVICE void interpolate(const dof_type& X, const tensor<double, dim, dim>& J,
```
white238, line 425: This has a return statement.

src/serac/numerics/functional/tests/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ blt_add_executable(NAME        tensor_unit_tests
 
 # Then add the examples/tests
 set(functional_tests_serial
-    functional_shape_derivatives.cpp
+    #functional_shape_derivatives.cpp
```
white238: can this be turned back on?

src/serac/numerics/functional/tests/functional_cuda.cpp
```diff
@@ -0,0 +1,200 @@
+#include <functional>
```
white238: remove this?

johnbowen42:
@jamiebramwell I think this PR is ready for benchmarking

jamiebramwell:
> @jamiebramwell I think this PR is ready for benchmarking

Thanks! I'm putting this together now...

jamiebramwell:
This is what I'm going to use to profile the `Functional` changes: https://github.com/LLNL/serac/pull/1092

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,62 +165,139 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
-    // (batch) integrate the material response against the test-space basis functions
-    test_element::integrate(get_value(qf_outputs), rule, &r[elements[e]]);
-  }
+  // for each element in the domain
+  RAJA::launch<launch_policy>(
+      RAJA::LaunchParams(RAJA::Teams(static_cast<int>(num_elements)), RAJA::Threads(BLOCK_SZ)),
+      [=] RAJA_HOST_DEVICE(RAJA::LaunchContext ctx) {
+        RAJA::loop<teams_e>(
+            ctx, e_range,
+            [&ctx, t, J, x, u, qf, trial_elements, qpts_per_elem, rule, device_r, elements, qf_derivatives, qf_inputs,
```
white238: Comment here to explain why the explicit captures were necessary.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -303,24 +411,48 @@ void element_gradient_kernel(ExecArrayView<double, 3, ExecutionSpace::CPU> dK, d
...
+  constexpr int                            nquad = num_quadrature_points(g, Q);
+  constexpr TensorProductQuadratureRule<Q> rule{};
+#ifdef USE_CUDA
+  using teams_e       = RAJA::LoopPolicy<RAJA::cuda_block_x_direct>;
+  using launch_policy = RAJA::LaunchPolicy<RAJA::cuda_launch_t<false>>;
+#else
+  using teams_e                    = RAJA::LoopPolicy<RAJA::seq_exec>;
+  using launch_policy              = RAJA::LaunchPolicy<RAJA::seq_launch_t>;
+#endif
+#ifdef USE_CUDA
```
white238: `USE_CUDA` is too generic but `SERAC_USE_CUDA` is set for the whole code when on blueos (maybe this needs to be rethought).

If we want to allow compiling Functional with and without cuda in the same build. Should this be `FUNCTIONAL_USE_CUDA`?

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -5,16 +5,38 @@
...
 #include "serac/numerics/functional/quadrature_data.hpp"
 #include "serac/numerics/functional/function_signature.hpp"
 #include "serac/numerics/functional/differentiate_wrt.hpp"
 
-#include <RAJA/index/RangeSegment.hpp>
-#include <RAJA/RAJA.hpp>
 #include <array>
 #include <cstdint>
 
+namespace {
```
white238: This should go into accelerator.hpp/cpp. We should also augment it to work with MPI and all cuda devices.

src/serac/numerics/functional/tests/functional_qoi.cpp
```diff
@@ -280,41 +339,40 @@ void qoi_test(mfem::ParMesh& mesh, H1<p1> trial1, H1<p2> trial2, Dimension<dim>)
...
-      DependsOn<0>{}, [&](double /*t*/, auto /*x*/, auto u) { return norm(serac::get<0>(u)); }, mesh);
-
-  double exact_answer   = 141.3333333333333;
-  double relative_error = (f(t, U) - exact_answer) / exact_answer;
-  EXPECT_NEAR(0.0, relative_error, 1.0e-10);
-
-  delete tmp;
-}
-
+// TEST(QoI, DependsOnVectorValuedInput)
```
white238: Why is this commented out?

src/serac/numerics/functional/tests/functional_qoi.cpp
```diff
@@ -337,239 +395,226 @@ TEST(QoI, AddAreaIntegral)
...
-  constexpr double expected_vol = 48.0;  // volume of 2 2x2x2 cubes == 16, so expected is 0.1 * 16
-  EXPECT_NEAR(volume, expected_vol, 3.0e-14);
-
-  constexpr double expected_val = 240.0;
-  EXPECT_NEAR(val, expected_val, 1.5e-13);
-
-  constexpr double expected_avg = 5.0;
-  EXPECT_NEAR(average, expected_avg, 1.0e-14);
-}
+// TEST(QoI, AddVolumeIntegral)
```
white238: Why is this commented out?

src/serac/physics/tests/parameterized_thermomechanics_example.cpp
```diff
@@ -37,12 +37,13 @@ struct ParameterizedThermoelasticMaterial {
...
 
   template <typename T1, typename T2, typename T3>
-  auto operator()(State& /*state*/, const tensor<T1, 3, 3>& grad_u, T2 temperature,
+  auto operator()(const State& /*state*/, const tensor<T1, 3, 3>& grad_u, T2 temperature,
                   T3 coefficient_of_thermal_expansion) const
   {
     auto theta = get<VALUE>(temperature);
     auto alpha = get<VALUE>(coefficient_of_thermal_expansion);
 
+    // std::cout<<"print e" << E << "print nu " << nu <<std::endl;
```
white238:
```suggestion
```

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -7,6 +7,7 @@
 
 #include <array>
 
+#include "serac/serac_config.hpp"
```
white238, line 12: Any file that calls RAJA should include their headers.

cmake/SeracMacros.cmake
```diff
@@ -148,13 +151,20 @@ macro(serac_add_tests)
...
+        endif()
 
         blt_add_executable(NAME        ${test_name}
                            SOURCES     ${filename}
                            OUTPUT_DIR  ${TEST_OUTPUT_DIRECTORY}
                            DEPENDS_ON  ${arg_DEPENDS_ON}
                            FOLDER      serac/tests )
 
+        if (DEFINED arg_USE_CUDA)
+            target_compile_definitions(${test_name} PUBLIC SERAC_USE_CUDA_KERNEL_EVALUATION)
```
white238: 👍

src/serac/infrastructure/accelerator.hpp
```diff
@@ -72,6 +72,24 @@ enum class ExecutionSpace
   Dynamic  // Corresponds to execution that can "legally" happen on either the host or device
 };
 
+#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
+
+/// @brief Alias for parallel threads policy on GPU
+using threads_x     = RAJA::LoopPolicy<RAJA::cuda_thread_x_direct>;
+using teams_e       = RAJA::LoopPolicy<RAJA::cuda_block_x_direct>;
```
white238: Should we use a more descriptive name than `x` and `e`?

src/serac/infrastructure/debug_print.hpp
```diff
@@ -121,3 +119,28 @@ void write_to_file(axom::Array<T, 3, axom::MemorySpace::Host> arr, std::string f
...
+/**
+ * @brief Helper function that prints usage of global device memory.  Useful for debugging potential
+ *        memory or register leaks
+ */
+#include <iostream>
+void printCUDAMemUsage()
+{
+  int deviceCount = 0;
+  cudaGetDeviceCount(&deviceCount);
+  int i = 0;
```
white238: We should add a TODO or have this loop over all the devices. Also it should say which MPI rank as well.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,62 +167,119 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+  using interpolate_out_type = decltype(tuple{get<indices>(trial_elements).template interpolate_output_helper<Q>()...});
 
-    // (batch) integrate the material response against the test-space basis functions
-    test_element::integrate(get_value(qf_outputs), rule, &r[elements[e]]);
-  }
+  using qf_inputs_type = decltype(tuple{promote_each_to_dual_when_output_helper<indices == differentiation_index>(
+      get<indices>(trial_elements).template interpolate_output_helper<Q>())...});
+
+#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
+  std::string device_name = "DEVICE";
```
white238: Is `device` the right descriptor here for `device_name` and `device_r`? `resource`?

My concern is it overlaps with the host/device paradigm.

scripts/llnl/common_build_functions.py
```diff
@@ -271,7 +271,7 @@ def build_and_test_host_config(test_root, host_config, report_to_stdout=False, e
     bld_output_file =  pjoin(build_dir,"output.log.make.txt")
     print("[starting build]")
     print("[log file: %s]" % bld_output_file)
-    res = sexe("cd %s && make -j 8 VERBOSE=1 " % build_dir,
+    res = sexe("cd %s && make -j 6 VERBOSE=1 " % build_dir,
```
tepperly: Should the degree of build parallelism be a constant like this? For some systems 6 seems to low and others perhaps too high.

src/serac/infrastructure/accelerator.hpp
```diff
@@ -12,7 +12,7 @@
  */
 
 #pragma once
-
+#include "RAJA/RAJA.hpp"
 #if defined(__CUDACC__)
 #define SERAC_HOST_DEVICE __host__ __device__
 #define SERAC_HOST __host__
```
tepperly, line 17:
```suggestion
#define SERAC_HOST_DEVICE RAJA_HOST_DEVICE
#define SERAC_HOST RAJA_HOST
#if defined(__CUDACC__)
```

src/serac/infrastructure/accelerator.hpp
```diff
@@ -72,6 +72,24 @@ enum class ExecutionSpace
...
+using launch_policy = RAJA::LaunchPolicy<RAJA::cuda_launch_t<false>>;
+
+#else
+
+/// @brief Alias for parallel threads policy on GPU.
+using threads_x = RAJA::LoopPolicy<RAJA::seq_exec>;
+/// @brief Alias for number of teams for GPU kernel launches.
+using teams_e = RAJA::LoopPolicy<RAJA::seq_exec>;
+/// @brief Alias for GPU kernel launch policy.
+using launch_policy = RAJA::LaunchPolicy<RAJA::seq_launch_t>;
```
tepperly:
```suggestion
/// @brief Alias for sequential policy on normal CPU.
using threads_x = RAJA::LoopPolicy<RAJA::seq_exec>;
/// @brief Alias for sequential policy on normal CPU.
using teams_e = RAJA::LoopPolicy<RAJA::seq_exec>;
/// @brief Alias for sequential policy on normal CPU.
using launch_policy = RAJA::LaunchPolicy<RAJA::seq_launch_t>;
```

src/serac/infrastructure/accelerator.hpp
```diff
@@ -72,6 +72,24 @@ enum class ExecutionSpace
...
+
```
tepperly, line 92: Have you thought about also defining OpenMP policies to take advantage of multi-threading on normal multi-core CPUs? Or enabling simd vectorizations with a RAJA::simd_exec policy? 

What about policies for El Capitain that doesn't use CUDA, right?

src/serac/infrastructure/accelerator.hpp
tepperly: It seems that `e` is about MFEM element iterations. I am guessing `x` is about iterating over vectors?

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,62 +167,119 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+              // (batch) integrate the material response against the test-space basis functions
+              test_element::integrate(get_value(qf_outputs), rule, &device_r[elements[e]], ctx);
+            });
+      });
+#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
+  rm.copy(r, device_r);
+  allocator.deallocate(device_r);
+#endif
+  allocator.deallocate(qf_inputs);
+  allocator.deallocate(interpolate_result);
```
tepperly, line 254: Shouldn't it be CUDA, El Cap, and normal CPU? Is there a way to do it without having so many #ifdef's. Is copying from host to host a no-op?

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -164,62 +167,119 @@ void evaluation_kernel_impl(trial_element_type trial_elements, test_element, dou
...
+                                              const RAJA::LaunchContext& ctx)
 {
   using return_type = decltype(chain_rule(derivative_type{}, T{}));
   tensor<tuple<return_type, zero>, n> outputs{};
-  for (int i = 0; i < n; i++) {
-    get<0>(outputs[i]) = chain_rule(qf_derivatives[i], inputs[i]);
-  }
+  RAJA::RangeSegment                  i_range(0, n);
+
+  RAJA::loop<threads_x>(ctx, i_range, [&](int i) { get<0>(outputs[i]) = chain_rule(qf_derivatives[i], inputs[i]); });
```
tepperly: Above it's `serac::get<0>`.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -257,22 +317,43 @@ void action_of_gradient_kernel(const double* dU, double* dR, derivatives_type* q
...
+  using qf_inputs_type = decltype(trial_element::template interpolate_output_helper<Q>());
 
-    // (batch) integrate the material response against the test-space basis functions
-    test_element::integrate(qf_outputs, rule, &dr[elements[e]]);
-  }
+#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
+  std::string device_name = "DEVICE";
+#else
+  std::string                      device_name = "HOST";
+#endif
```
tepperly: This `#ifdef` to get the device name should be done once in a single header file.
```c++
constexpr std::string deviceName() {
#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
   return "DEVICE";
#else
...
#endif
```

tepperly: Or maybe it's a `serac::getAllocator()` method that encapsulates several repeated lines.

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -257,22 +317,43 @@ void action_of_gradient_kernel(const double* dU, double* dR, derivatives_type* q
...
+
+  auto&           rm        = umpire::ResourceManager::getInstance();
+  auto            allocator = rm.getAllocator(device_name);
+  qf_inputs_type* qf_inputs = static_cast<qf_inputs_type*>(allocator.allocate(sizeof(qf_inputs_type) * num_elements));
```
tepperly, line 323: Can this allocate/deallocate be used with `std::unique_ptr<qf_inputs>`?

Maybe you could define:
```c++
template<typename MyType>
::std::unique_ptr<MyType>
allocateOnHardware(::std::size_t size) {
// #ifdef to get device name
// get Umpire instance
// get appropriate allocator
   return ::std::unique_ptr<MyType[]>(static_cast<MyType*>(allocator.allocate(size), [&](void *ptr) { rm.deallocate(ptr); });
}
```

src/serac/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -7,9 +7,12 @@
 
 #include <array>
 
+#include "serac/serac_config.hpp"
 #include "serac/numerics/functional/quadrature_data.hpp"
 #include "serac/numerics/functional/differentiate_wrt.hpp"
 
+#include "RAJA/RAJA.hpp"
```
tepperly, line 16: Since RAJA is a TPL, shouldn't it be
```c++
#include <RAJA/RAJA.hpp>
```
instead of double quotes?

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -10,14 +10,16 @@
  * @brief Specialization of finite_element for H1 on hexahedron geometry
  */
 
+#include "RAJA/RAJA.hpp"
```
tepperly, line 13: `#include <RAJA/RAJA.hpp>`

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -144,30 +147,36 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
     using flux_t   = decltype(get<0>(get<1>(in_t{})) + dot(get<1>(get<1>(in_t{})), tensor<double, dim>{}));
 
     tensor<tuple<source_t, flux_t>, q * q * q> output;
+    auto                                       x_range = RAJA::RangeSegment(0, q * q * q);
+    RAJA::loop<threads_x>(ctx, x_range, [&](int Q) {
+      int qx = Q % q;
+      int qy = (Q / q) % q;
+      int qz = (Q / (q * q));
```
tepperly, line 160: Wouldn't it be better and more flexible to use a [RAJA::kernel](https://raja.readthedocs.io/en/develop/sphinx/user_guide/feature/loop_basic.html#complex-loops-raja-kernel) to handle this multi-level for-loop.

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -204,85 +216,187 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
-              get<GRADIENT>(output.three_dimensional(qz, qy, qx))[i][j] = gradient(i, j, qz, qy, qx);
-            }
-          }
+    RAJA::loop<threads_x>(ctx, x_range, [&](int tid) {
+      if (tid >= q * q * q) {
+        return;
+      }
+      int qx = tid % q;
+      int qy = (tid / q) % q;
+      int qz = tid / (q * q);
```
tepperly, line 283: Should this be a `RAJA::kernel` explicit multi-level for loop?

src/serac/numerics/functional/detail/hexahedron_H1.inl
```diff
@@ -204,85 +216,187 @@ struct finite_element<mfem::Geometry::CUBE, H1<p, c> > {
...
-                source(qz, qy, qx) = reinterpret_cast<const double*>(&get<SOURCE>(qf_output[Q]))[i * ntrial + j];
-              }
-              if constexpr (!is_zero<flux_type>{}) {
-                for (int k = 0; k < dim; k++) {
-                  flux(k, qz, qy, qx) =
-                      reinterpret_cast<const double*>(&get<FLUX>(qf_output[Q]))[(i * dim + k) * ntrial + j];
-                }
-              }
+        RAJA::loop<threads_x>(ctx, x_range, [&](int tid) {
+          if (tid >= q * q * q) {
```
tepperly, line 342: RAJA::kernel?

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -149,62 +154,120 @@ void evaluation_kernel_impl(trial_element_tuple trial_elements, test_element, do
...
+  using qf_inputs_type = decltype(tuple{promote_each_to_dual_when_output_helper<indices == differentiation_index>(
+      get<indices>(trial_elements).template interpolate_output_helper<Q>())...});
+
+#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
+  std::string device_name = "DEVICE";
+  auto        device_r    = copy_data(r, serac::size(*r) * sizeof(double), device_name);
+#else
+  std::string                           device_name = "HOST";
+  typename test_element_type::dof_type* device_r    = r;
+#endif
```
tepperly: Similar concerns as stated above about lots of #ifdef's and repeated code.

src/serac/numerics/functional/domain_integral_kernels.hpp
```diff
@@ -149,62 +154,120 @@ void evaluation_kernel_impl(trial_element_tuple trial_elements, test_element, do
...
+              physical_to_parent<test_element_type::family>(qf_outputs, J, e, ctx);
+
+              // write out the q-function derivatives after applying the
+              // physical_to_parent transformation, so that those transformations
+              // won't need to be applied in the action_of_gradient and element_gradient kernels
+              if constexpr (differentiation_index != serac::NO_DIFFERENTIATION) {
+                RAJA::RangeSegment x_range(0, leading_dimension(qf_outputs));
+                RAJA::loop<threads_x>(ctx, x_range, [&](int q) {
+                  qf_derivatives[e * uint32_t(qpts_per_elem) + uint32_t(q)] = get_gradient(qf_outputs[q]);
+                });
```
tepperly, line 256: [RAJA documentation](https://raja.readthedocs.io/en/develop/sphinx/user_guide/feature/loop_basic.html#complex-loops-raja-kernel) suggests against using nested RAJA::forall. Is RAJA::loop the same thing?

src/serac/numerics/functional/function_signature.hpp
```diff
@@ -1,5 +1,13 @@
+// Copyright (c) 2019-2023, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
 #pragma once
-#include <camp/camp.hpp>
+#include "camp/camp.hpp"
```
tepperly, line 8: Don't third-party libraries use `<` and `>`  instead of double quotes?

src/serac/numerics/functional/integral.hpp
```diff
@@ -190,15 +200,25 @@ void generate_kernels(FunctionSignature<test(trials...)> s, Integral& integral,
...
+// Copy elements to device if using CUDA
+#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
+  int* device_elements =
+      copy_data(const_cast<int*>(elements), sizeof(int) * integral.domain_.get(geom).size(), "DEVICE");
+  integral.evaluation_[geom] = domain_integral::evaluation_kernel<NO_DIFFERENTIATION, Q, geom>(
+      s, qf, positions, jacobians, qdata, dummy_derivatives, device_elements, num_elements);
+#else
   integral.evaluation_[geom] = domain_integral::evaluation_kernel<NO_DIFFERENTIATION, Q, geom>(
       s, qf, positions, jacobians, qdata, dummy_derivatives, elements, num_elements);
+#endif
```
tepperly: It seems like it should be possible to have more code without `#ifdef`'s.

src/serac/numerics/functional/tests/functional_basic_h1_vector.cpp
```diff
@@ -138,20 +162,26 @@ void test_suite(std::string meshfile)
...
 
 int main(int argc, char* argv[])
 {
   int num_procs, myid;
 
   ::testing::InitGoogleTest(&argc, argv);
 
+#ifdef SERAC_USE_CUDA_KERNEL_EVALUATION
+  serac::accelerator::initializeDevice();
+#endif
```
tepperly: It would be better if there was just a call to:
```c++
  ::serac::initialize();
```
that happens all the time (regardless of CPU, GPU, APU, etc.) rather than having the #ifdef USE_CUDA, #ifdef USE_HIP, etc.

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -1,8 +1,16 @@
+// Copyright (c) 2019-2023, Lawrence Livermore National Security, LLC and
```
tepperly, line 1: 2024

tepperly (commented):
Why so much // clang-format: off and // clang-format: on?

In general, it seems that the RAJA abstractions should require fewer #ifdef/#endif or they should be isolated rather than reappearing in multiple locations.

scripts/llnl/common_build_functions.py
white238: This should probably be pulled out to a command line variable and default to something sensible otherwise. I'll make an issue and we can handle it later.

white238: #1107

src/serac/numerics/functional/boundary_integral_kernels.hpp
white238: Our general style is to only use `<>` for system libraries.

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
white238: The rest of the code hasn't been updated. I'll push a branch now. I was hoping to catch a point where there wasnt large PRs out but it's almost may.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -801,6 +801,55 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   * extended, generic lambda for compatibility with NVCC.
+   */
+  template <typename Material>
+  struct SetMaterialStressFunctor {
+    /// @brief Constructor for the functor
+    SetMaterialStressFunctor(Material material, GeometricNonlinearities gn) : material_(material), geom_nonlin(gn) {}
+    /// @brief Material model
+    Material material_;
+    /// @brief Enum value for geometric nonlinearities
+    GeometricNonlinearities geom_nonlin;
```
white238: Member variables should have a trailing underscore
```suggestion
    GeometricNonlinearities geom_nonlin_;
```

src/serac/physics/benchmarks/benchmark_functional.cpp
```diff
@@ -81,7 +81,7 @@ int main(int argc, char* argv[])
 {
   MPI_Init(&argc, &argv);
 
-  int parallel_refinement = 3;
+  int parallel_refinement = 2;
```
white238: Do you remember the reasons behind this change?

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -211,27 +219,50 @@ SERAC_HOST_DEVICE auto promote_to_dual_when(const T& x)
...
+    RAJA::RangeSegment x_range(0, n);
+    using return_type      = decltype(make_dual(T{}));
+    auto casted_output_ptr = static_cast<tensor<return_type, n>*>(output_ptr);
+    RAJA::loop<threads_x>(ctx, x_range, [&](int i) { (*casted_output_ptr)[i] = make_dual(x[i]); });
   }
   if constexpr (!dualify) {
-    return x;
+    RAJA::RangeSegment x_range(0, n);
+    auto               casted_output_ptr = static_cast<tensor<T, n>*>(output_ptr);
+    RAJA::loop<threads_x>(ctx, x_range, [&](int i) { (*casted_output_ptr)[i] = x[i]; });
```
white238: I assume I am not understanding something but is this copying the device or is it laying it out differently in memory? Why is it not just returning x anymore?

src/serac/numerics/functional/tests/CMakeLists.txt
```diff
@@ -48,17 +48,25 @@ foreach(filename ${functional_tests_mpi})
...
 
-    set(functional_tests_cuda 
-        tensor_unit_tests_cuda.cu 
-#        some of the GPU functionality is temporarily disabled to 
-#        help incrementally roll-out the variadic implementation of Functional
-#        TODO: re-enable GPU kernels in a follow-up PR
-#        functional_comparisons_cuda.cu
-       )
+    set(functional_tests_cuda
+    # functional_basic_hcurl.cpp
```
white238, line 63: Comment here explaining why these are turned off would be helpful.

src/serac/infrastructure/accelerator.hpp
tepperly: It would be better to not have to guess.

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
white238: This could use a comment explaining.

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -503,7 +535,7 @@ SERAC_HOST_DEVICE auto get_value(const tensor<dual<T>, n...>& arg)
  * @param[in] arg The tensor of dual numbers
  */
 template <int... n>
-SERAC_HOST_DEVICE constexpr auto get_gradient(const tensor<dual<double>, n...>& arg)
+SERAC_HOST_DEVICE /*constexpr*/ auto get_gradient(const tensor<dual<double>, n...>& arg)
```
tupek2: why no more constexpr

.gitlab/build_toss4.yml
```diff
@@ -40,7 +40,7 @@ toss4-clang_14_0_6-src:
     EXTRA_CMAKE_OPTIONS: "-DENABLE_BENCHMARKS=ON"
     DO_INTEGRATION_TESTS: "yes"
     ALLOC_NODES: "2"
-    ALLOC_TIME: "30"
+    ALLOC_TIME: "45"
```
chapman39, line 43: I would suggest also changing the deadline option from 1hour to 90minutes (on line 5). That way, jobs can sit on the queue a bit longer.

`SCHEDULER_PARAMETERS: "--res=ci --exclusive=user --deadline=now+1hour -N ${ALLOC_NODES} -t ${ALLOC_TIME} -A ${ALLOC_BANK}"`

to

`SCHEDULER_PARAMETERS: "--res=ci --exclusive=user --deadline=now+90minutes -N ${ALLOC_NODES} -t ${ALLOC_TIME} -A ${ALLOC_BANK}"`

PR #1028, opened 2023-11-09
Feature/tupek/thermal transient adj

tupek2:
Improve testing and some capabilities for thermal transient adjoints in support of necessary use cases.

Mostly testing nonlinear parametrized thermal gradients.  Also unifying testing with solid, and unifying a few things between solid_mechanics and heat_transfer.  Support varying timesteps, remove zero member variables.


src/serac/physics/heat_transfer.hpp
```diff
@@ -154,6 +154,7 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
         "EquationSolver argument is nullptr in HeatTransfer constructor. It is possible that it was previously moved.");
 
     states_.push_back(&temperature_);
+    // states_.push_back(&temperature_rate_);
```
jamiebramwell, line 157: Is there a reason this is commented out?

jamiebramwell (approved):

src/serac/physics/heat_transfer.hpp
tupek2: At the moment I'm not really saving the rate, but recomputing it in the reverse pass (which is easier for backward Euler).  Could be we eventually want to really save it off though.  Not sure.  And, if I uncomment just this line, there are some inconsistencies elsewhere which will break tests.

jamiebramwell: Got it. I would go ahead and delete it in that case.

btalamini: Drawing an analogy to solid mechanics, the temperature rate would be like the acceleration. Some integrators - like Newmark - need to carry this from one time step to the next as state. Other schemes - like Runge-Kutta and Euler - don't carry the highest order time derivative as state, they just compute it at each cycle and throw it away. consequently, I think we'll eventually want to save `temperature_rate_` at each cycle so that this code is usable for both kinds, but I don't mind leaving it out until we need it.

Compromise: Put a comment here that this is a reminder that we may want to save this in the future.

jamiebramwell:
@btalamini @samuelpmishLLNL  , could you take a look at this?

samuelpmishLLNL (approved):

PR #1032, opened 2023-11-16
Fix the finite element dual assignment operators

jamiebramwell:
This fixes a memory leak as reported by @tepperly .


jamiebramwell:
@tepperly , I added a few extras to your patch. Let me know if this works for you.

src/serac/physics/state/finite_element_state.hpp
```diff
@@ -77,13 +76,61 @@ class FiniteElementState : public FiniteElementVector {
...
+  }
+
   /**
    * @brief Fill a user-provided grid function based on the underlying true vector
    *
    * This distributes true vector dofs to the finite element (local) dofs  by multiplying the true dofs
    * by the prolongation operator.
    *
-   * @see <a href="https://mfem.org/pri-dual-vec/">MFEM documentation</a> for details
+   * @see <a href="https://mfem.org/pri-State-vec/">MFEM documentation</a> for details
```
jamiebramwell:
```suggestion
   * @see <a href="https://mfem.org/pri-dual-vec/">MFEM documentation</a> for details
```

src/serac/physics/state/finite_element_state.hpp
```diff
@@ -94,7 +141,7 @@ class FiniteElementState : public FiniteElementVector {
    * This distributes the grid function dofs to the true vector dofs by multiplying by the
    * restriction operator.
    *
-   * @see <a href="https://mfem.org/pri-dual-vec/">MFEM documentation</a> for details
+   * @see <a href="https://mfem.org/pri-State-vec/">MFEM documentation</a> for details
```
jamiebramwell:
```suggestion
   * @see <a href="https://mfem.org/pri-dual-vec/">MFEM documentation</a> for details
```

samuelpmishLLNL:
out of curiosity, what was the cause of the underlying leak?

jamiebramwell:
> out of curiosity, what was the cause of the underlying leak?

I defer to @tepperly as he developed and tested this patch.

tepperly:
> out of curiosity, what was the cause of the underlying leak?

I believe it's the operator=(&& rhs) variant. After an invocation of the move operator, the compiler assumes that the rhs has been completely pilfered (i.e., all of the allocated resources owned by rhs have been moved into this or freed). When rhs is an upcast, it doesn't properly deal with all the parts of rhs leaving some orphaned pointers. This is what my valgrind logs suggested at least. Applying this patch, plugged some leaks.

In particular, I believe it was when the right hand side was an instead of `FiniteElementDual`, the `rhs.linear_form_` was leaked.

tepperly:
The comments saying *this seems like the wrong way to be doing this assignment* drew my attention.

samuelpmishLLNL (approved):

tepperly:
It would be nice if this could get in soon, so LiDO's next TPL could have it. Without this patch, my uncertain bracket problem crashes in 10-20 MMA iterations do to the Linux OOM killler which sometimes crashes my desktop.

white238 (approved):

PR #1033, opened 2023-11-16
Feature/tupek/transient adj states

tupek2:
After discussing with Kenny.  The number of states for quasi-statics in solids went from 1 to 3 recently.  This slightly breaks the lido interface (though there is a work around).  Seems sensible for there to only be 1 (displacement) state on the stateNames list for quasiStatics.  It is expected by Lido that every stateName() will require an adjoint load to be passed in.  

Also, refactored so temperature_rate is a state and checkpointed for more consistency between the physics.


jamiebramwell (approved):
Thank you for cleaning this up!

samuelpmishLLNL (approved):

PR #1037, opened 2023-11-20
Update the time for quasi-static time-varying loads

jamiebramwell:
As quasi-static and dynamic problems use the same integrators, the ODE time point also needs to be updated in the quasi-static case as the ODE solver does not get called for that code path.


samuelpmish (approved):

samuelpmishLLNL (approved):

samuelpmishLLNL:
oops, I double-approve of this PR.

btalamini (approved):

PR #1039, opened 2023-11-27
add time to functional

samuelpmishLLNL:
This PR is an interface breaking change that aims to address a problem with the way serac's physics modules currently handle time in q-functions. 

Right now, the physics module has a member variable for the current simulation time. When a user registers a q-function for the residual calculation, the lambda function captures a reference to this time member variable and so that it can be used in the body of the q-function. The two main problems with this approach are:

1. Time is being passed around "implicitly"-- it doesn't appear as a function argument, and as a result, it's hard to reason about what time value is currently being used, or when it is modified.
2. Capturing a reference to a host variable will segfault when executing the q-function inside a CUDA kernel (preventing us from running our simulations on GPUs).

------------

Now, q-functions passed to `serac::Functional` all receive an additional argument of type `double` in the first position, and the `serac::Functional::operator()` member function also takes a `double` in the first position. For example, 

Before:
```cpp
  // create a serac::Functional object
  Functional<test_space(trial_space)> residual(&test_fespace, {&trial_fespace});

  // register a q-function
  residual.AddDomainIntegral(
      Dimension<dim>{}, DependsOn<0>{},
      [=](auto x, auto temperature) {
        auto [u, du_dx] = temperature;
        return serac::tuple{u, k * du_dx};
      }, *mesh);

  // invoke the integral kernels
  auto r = residual(u);
```

After:
```cpp
  // create a serac::Functional object (unchanged)
  Functional<test_space(trial_space)> residual(&test_fespace, {&trial_fespace});

  // register a q-function (changed)
  residual.AddDomainIntegral(
      Dimension<dim>{}, DependsOn<0>{},
      [=](double t, auto x, auto temperature) { // <--- the first q-function argument is time 
        auto [u, du_dx] = temperature;
        return serac::tuple{u, k * du_dx};
      }, *mesh);

    // invoke the integral kernels (changed)
  double time = ...;
  auto r = residual(time, u); // now the residual calculation requires time to be passed in
```


white238 (approved):

jamiebramwell:
@barreracruz @kswartz92 , this is a breaking change to anything that directly uses `Functional`. It's pretty straightforward to adopt, but it will require changing your interfaces. We could have a default `time = 0.0` function signature, but I think it's better to just deal with this breaking change. Any thoughts?

kswartz92:
> @barreracruz @kswartz92 , this is a breaking change to anything that directly uses `Functional`. It's pretty straightforward to adopt, but it will require changing your interfaces. We could have a default `time = 0.0` function signature, but I think it's better to just deal with this breaking change. Any thoughts?

Are there plans to include shape in functional as well? If so, would it make sense to introduce a new breaking functional interface with all of the intended changes?

samuelpmishLLNL:
> @barreracruz @kswartz92 , this is a breaking change to anything that directly uses `Functional`. It's pretty straightforward to adopt, but it will require changing your interfaces. We could have a default `time = 0.0` function signature, but I think it's better to just deal with this breaking change. Any thoughts?

For clarification: everything that goes through the physics modules (e.g. materials, tractions, boundary conditions) is unchanged, it's just the direct calls to `serac::Functional::operator()` that need an extra `double`s.

jamiebramwell:
> > @barreracruz @kswartz92 , this is a breaking change to anything that directly uses `Functional`. It's pretty straightforward to adopt, but it will require changing your interfaces. We could have a default `time = 0.0` function signature, but I think it's better to just deal with this breaking change. Any thoughts?
> 
> Are there plans to include shape in functional as well? If so, would it make sense to introduce a new breaking functional interface with all of the intended changes?

The shape interface will be built on top of the existing `Functional` infrastructure, so it's fundamentally a different development activity. That said, it is on the top of my to-do pile so I can get that done this week and you could bring them both over at the same time.

tupek2 (approved):

src/serac/physics/solid_mechanics.hpp
```diff
@@ -818,7 +819,8 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
   {
     residual_->AddDomainIntegral(
         Dimension<dim>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
-        [body_force, this](auto x, auto /* displacement */, auto /* acceleration */, auto shape, auto... params) {
+        [body_force, this](double /*t*/, auto x, auto /* displacement */, auto /* acceleration */, auto shape,
```
jamiebramwell: We should remove the `this` capture to ensure we aren't aliasing host memory.

jamiebramwell: For instance, the call on line 828 below should be the input `t` variable instead of `ode_time_point_`.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -720,7 +720,8 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
                                                              // fact that the displacement, acceleration, and shape
                                                              // fields are always-on and come first, so the `n`th
                                                              // parameter will actually be argument `n + NUM_STATE_VARS`
-        [this, material](auto /*x*/, auto& state, auto displacement, auto acceleration, auto shape, auto... params) {
+        [this, material](double /*t*/, auto /*x*/, auto& state, auto displacement, auto acceleration, auto shape,
```
jamiebramwell: I think not passing time into this lambda is OK as we should not allow time-dependent material properties.

src/serac/physics/heat_transfer.hpp
```diff
@@ -444,7 +444,8 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
   {
     residual_->AddDomainIntegral(
         Dimension<dim>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
-        [source_function, this](auto x, auto temperature, auto /* dtemp_dt */, auto shape, auto... params) {
+        [source_function, this](double /*t*/, auto x, auto temperature, auto /* dtemp_dt */, auto shape,
```
jamiebramwell: We should not capture `this`. `ode_time_point_` is used incorrectly on line 462 as a result.

src/serac/physics/heat_transfer.hpp
```diff
@@ -499,7 +500,7 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
   {
     residual_->AddBoundaryIntegral(
         Dimension<dim - 1>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
-        [this, flux_function](auto X, auto u, auto /* dtemp_dt */, auto shape, auto... params) {
+        [this, flux_function](double /*t*/, auto X, auto u, auto /* dtemp_dt */, auto shape, auto... params) {
```
jamiebramwell: Same comment re: `this`. `ode_time_point_` is used on line 517.

src/serac/physics/heat_transfer.hpp
```diff
@@ -871,7 +874,7 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
    */
   FiniteElementDual& computeTimestepSensitivity(size_t parameter_field) override
   {
-    auto drdparam     = serac::get<DERIVATIVE>(d_residual_d_[parameter_field]());
+    auto drdparam     = serac::get<DERIVATIVE>(d_residual_d_[parameter_field](ode_time_point_));
```
jamiebramwell, line 889: For this to work appropriately, `ode_time_point_` must be updated in the backward pass. Currently, only `time_` is updated.

src/serac/physics/solid_mechanics.hpp
samuelpmishLLNL: what about viscoelastic or other rate-dependent models?

src/serac/physics/solid_mechanics.hpp
samuelpmishLLNL: whoops, missed that one!

src/serac/physics/solid_mechanics.hpp
jamiebramwell: Rate is OK, but absolute time doesn't make sense. For example, we would want to have a bulk modulus that is a function of time.

jamiebramwell: Capturing an offline discussion here - passing time to calculate the timestep size (and therefore the rate quantities) is a possible path forward for rate-dependent materials. However, adding that capability is well outside the scope of this PR.

btalamini: FWIW, I agree with Jamie's points.

tupek2: One could envision aging models that have an explicit dependence on time.  But certainly not a use case we have at the moment.

jamiebramwell (approved):

src/serac/physics/heat_transfer.hpp
samuelpmishLLNL: I just put some big TODOs next to this line, since we don't have tests to confirm if this part of the code is functioning or not. We'll look into it more closely when refactoring the time integration procedures.

PR #1052, opened 2023-12-15
Update BLT and a lot of dependencies

white238:
This is a large update of libraries, basically all of the WSC libraries.

Highlights include:

* BLT: now uses the new `blt_install_tpl_setups` logic which installs the BLT targets to get recreated downstream instead of exporting them which causes early evaluation of generator expressions, all BLT targets now are namespaced
* Cleans up a lot of importing/exporting CMake target logic, including a new `serac_assert_find_succeeded` macro that checks for expected targets after a find
* MFEM: updated past the point Sam/Seth asked for


jamiebramwell (approved):
Thank you for your hard work @white238 !

scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
```diff
@@ -291,13 +291,19 @@ spack:
...
+    mfem:
+      require: "@4.6.2.4"
     raja:
-      require: "@2023.06.1~shared~examples~exercises"
+      require: "@2024.02.0~shared~examples~exercises"
+    tribol:
+      require: "@0.1.0.15~examples~tests"
     umpire:
-      require: "@2023.06.0~shared~examples"
+      require: "@2024.02.0~shared~examples~werror"
```
chapman39, line 309: Why not have this information in the Serac spack `package.py`?

scripts/spack/packages/tribol/package.py
```diff
@@ -30,26 +29,26 @@ def get_spec_path(spec, package_name, path_replacements={}, use_bin=False):
...
 
-class Tribol(CachedCMakePackage, CudaPackage):
+class Tribol(CachedCMakePackage, CudaPackage, ROCmPackage):
     """Tribol is an interface physics library."""
 
     homepage = "https://github.com/LLNL/Tribol"
     git      = "https://github.com/LLNL/Tribol.git"
 
-    version("develop", branch="develop", submodules=True)
+    version("develop", branch="develop", submodules=True, preferred=True)
```
chapman39: I don't think we would want to build with develop accidentally.

```suggestion
    version("develop", branch="develop", submodules=True)
```

src/tests/petsc_smoketest.cpp
```diff
@@ -28,6 +28,8 @@
 
 static char help[] = "Solves 2D Poisson equation using multigrid.\n\n";
 
+#include <algorithm>
```
chapman39, line 32: Minor nitpick but wrapping this or moving it to a preexisting SERAC_EDIT_START/ END would help track what we've changed to the example.

chapman39 (approved):
thank you @white238

scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
white238: I only put hard requirements in the `package.py` because it is used by other projects. I did put the version requirement in the `package.py` due to it being a requirement, but `~werror` was fine before and only failed due to a weird set of compilers.

chapman39:
While helping out Seth with his branch, we discovered that codevelop fails on TOSS4. Maybe it would be good to add a codevelop test on the gitlab ci as well in the future.

```bash
./config-build.py -hc host-configs/quartz-toss_4_x86_64_ib-clang\@14.0.6.cmake -DSERAC_ENABLE_CODEVELOP=ON
```

```
CMake Error at cmake/blt/cmake/BLTInstallableMacros.cmake:527 (target_link_libraries):
  Cannot specify link libraries for target "adiak::mpi" which is not built by
  this project.
  axom/src/cmake/thirdparty/SetupAxomThirdParty.cmake:208 (blt_patch_target)
  axom/src/cmake/CMakeBasics.cmake:20 (include)
  axom/src/CMakeLists.txt:125 (include)
```

white238:
Actually we may just want to not use caliper/adiak in Axom when doing the codevelop build. The output would complicate our timing output in normal cases and if we really wanted it we could turn it on for specific benchmarking cases.

PR #1054, opened 2023-12-20
Contact API updates (If no Tribol)

chapman39:
In the future we may want to test `TRIBOL_DIR=''` in Serac rather than Smith. Or always include tribol?

### DIFF
--- a/src/serac/physics/contact/contact_data.cpp
+++ b/src/serac/physics/contact/contact_data.cpp
@@ -12,8 +12,11 @@
 #ifdef SERAC_USE_TRIBOL
 #include "tribol/interface/tribol.hpp"
 #include "tribol/interface/mfem_tribol.hpp"
+#endif
 
 namespace serac {
+
+#ifdef SERAC_USE_TRIBOL
 
 ContactData::ContactData(const mfem::ParMesh& mesh)
     : mesh_{mesh},
@@ -328,11 +331,7 @@ void ContactData::updateDofOffsets() const
   offsets_up_to_date_ = true;
 }
 
-}  // namespace serac
-
 #else
-
-namespace serac {
 
 ContactData::ContactData([[maybe_unused]] const mfem::ParMesh& mesh)
     : have_lagrange_multipliers_{false}, num_pressure_dofs_{0}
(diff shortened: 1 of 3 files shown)


jamiebramwell:
@ebchin , can you take a quick look at this?

jamiebramwell:
> In the future we may want to test `TRIBOL_DIR=''` in Serac rather than Smith. Or always include tribol?

We should move the no-tribol tests here to catch issues like this.

jamiebramwell:
Per @chapman39 , the changes introduced in https://github.com/LLNL/serac/pull/982/files were not propagated to the no-tribol sections.

jamiebramwell (approved):
Thanks @chapman39 !

src/serac/physics/contact/contact_data.cpp
```diff
@@ -328,12 +328,14 @@ void ContactData::updateDofOffsets() const
   offsets_up_to_date_ = true;
 }
 
+}  // namespace serac
```
white238: Are we sure this is right? the `namespace serac` is inside the if of the ifdef then reopens in the else

white238: Pretty sure it wasnt right to begin with either... Can you move all namespacing outside of the ifdef?

chapman39: im pretty sure its right but it is weird to have namespace serac in both ifdef blocks.

chapman39: fixed

src/serac/physics/contact/contact_data.hpp
```diff
@@ -156,13 +156,22 @@ class ContactData {
...
 
+#ifdef SERAC_USE_TRIBOL
   /**
    * @brief Have there been contact interactions added?
    *
    * @return true if contact interactions have been added
    * @return false if there are no contact interactions
    */
   bool haveContactInteractions() const { return !interactions_.empty(); }
+#else
```
white238, line 169: would it be better to put the ifdef inside of the function?

white238 (approved):

src/serac/physics/contact/contact_data.hpp
chapman39: Yeah, I was doing this quickly. I'll update that

chapman39:
Fixes #1055

ebchin:
Thanks for fixing this @chapman39 !

PR #1057, opened 2023-12-20
Feature/tupek/reset states

tupek2:


src/serac/physics/heat_transfer.hpp
```diff
@@ -266,6 +258,38 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+  /**
+   * @brief Non virtual method to reset thermal states to zero.  This does not reset design parameters or shape.
+   *
+   * @param[in] cycle The simulation cycle (i.e. timestep iteration) to intialize the physics module to
+   * @param[in] time The simulation time to initialize the physics module to
+   */
+  void initializeThermalStates()
+  {
+    dt_          = 0.0;
+    previous_dt_ = -1.0;
```
btalamini, line 270: Not necessarily in this PR, but at some point, can we add a comment on why this value of `previous_dt_` is needed? Is it connected to the particular implementation of backward Euler?

src/serac/physics/base_physics.cpp
```diff
@@ -62,6 +57,25 @@ int BasePhysics::minCycle() const { return min_cycle_; }
...
+{
+  timesteps_.clear();
+
+  time_           = time;
+  max_time_       = time;
+  min_time_       = time;
+  cycle_          = cycle;
+  max_cycle_      = cycle;
+  min_cycle_      = cycle;
+  ode_time_point_ = time;
```
btalamini, line 70: I thought we removed `ode_time_point_` as part of the state.

src/serac/physics/tests/dynamic_solid_adjoint.cpp
```diff
@@ -243,54 +248,72 @@ struct SolidMechanicsSensitivityFixture : public ::testing::Test {
 
 TEST_F(SolidMechanicsSensitivityFixture, InitialDisplacementSensitivities)
 {
-  auto solid_solver_base = createNonlinearSolidMechanicsSolver(dataStore, nonlinear_opts, dyn_opts, mat);
-  auto [qoi_base, init_disp_sensitivity, _, __] = computeSolidMechanicsQoiSensitivity(*solid_solver_base, tsInfo);
+  auto solid_solver = createNonlinearSolidMechanicsSolver(dataStore, nonlinear_opts, dyn_opts, mat);
+  auto [qoi_base, init_disp_sensitivity, _, __] = computeSolidMechanicsQoiSensitivity(*solid_solver, tsInfo);
```
btalamini, line 252: Ok, I can see that these tests use the reset capability to compute the finite differences with one `solid_solver` instead of two. Seems like a good test.

btalamini (approved):

src/serac/physics/solid_mechanics.hpp
```diff
@@ -255,13 +241,11 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
     du_.SetSize(true_size);
-    du_ = 0.0;
-
     dr_.SetSize(true_size);
-    dr_ = 0.0;
-
     predicted_displacement_.SetSize(true_size);
-    predicted_displacement_ = 0.0;
+
+    shape_displacement_ = 0.0;
```
jamiebramwell, line 247: Do we consider resetting the shape displacement not part of the initialization step? In other words, do we expect this to be reset if you call the `reset` command? Any thoughts @kswartz92 ?

src/serac/physics/base_physics.hpp
```diff
@@ -107,6 +107,14 @@ class BasePhysics {
    */
   virtual std::vector<double> timesteps() const;
 
+  /**
+   * @brief Base method to reset physics states to zero.  This does not reset design parameters or shape.
```
jamiebramwell, line 119: I'm OK with this as it is documented. Any preference @barreracruz @kswartz92 ?

src/serac/physics/base_physics.cpp
jamiebramwell: This is part of the still-need-to-do ODE solver refactor. It's next on my list after the shape aware functional.

src/serac/physics/heat_transfer.hpp
jamiebramwell: I made an issue to track this: https://github.com/LLNL/serac/issues/1062

jamiebramwell (approved):
Thanks @tupek2 !

src/serac/physics/base_physics.hpp
kswartz92: I like having a reset method that does not change parameters/shape displacement, i.e., just prepares the solver for a new transient evaluation and sensitivity calculation

tupek2: also, naming thoughts for everyone?  I considered initialize, reset, reinitialize, clear, and zeroState.

jamiebramwell: I like `initialize` because we use it both for the first and subsequent calls, but I'd also be OK with `reset`.

src/serac/physics/solid_mechanics.hpp
kswartz92: I've gone back and forth on this, but I think it makes sense to separate resetting the design (i.e. parameters/shape) from resetting of the states (and related transient information e.g. time steps). If a user was solving an inverse problem for initial conditions, they may want to set parameters/shape and then leave them during their inverse problem. Similarly, I sometimes use "parameter" fields to define field quantities that are not actually changing in a design optimization. Having a reset states only method allows me to not place those fields on a lido graph. It does make a "reset" method slightly more confusing since it doesn't get the solver back to exactly the same place as a newly constructed solver, but I think the tradeoff is worth it.

src/serac/physics/base_physics.hpp
white238: If this is not resetting/initializing everything. I think it would be more understandable if it had something in the name to clarify that. Would having `state` in the name be enough? `clearEverythingButShapeAndParameters` is probably too verbose =)

I am not sure I personally like `initialize` because it seems weird to do this:

```
initialize()
// do lots of stuff
initialize()
// do more stuff
```

Would the user be calling initialize in the first case or is it hidden in a constructor?

But none of these are bad..

tupek2: I changes the public api on base_physics to be resetStates().  I think this makes more sense for external usage, as Chris mentions.  For the internal non-virtual methods, I kept as initialize because that is how they are used internally.

src/serac/physics/solid_mechanics.hpp
tupek2: keeping separate for now.  We have a resetStates, can add a resetDesign , or resetDesignFields (or whatever) when it is needed, which may be never as the user can already essentially do this, I think.

PR #1059, opened 2023-12-21
Shape aware functional

jamiebramwell:
This adds a wrapper around `Functional` to appropriately transform q-functions to reflect a shape displacement field. See documentation here: https://serac.readthedocs.io/en/feature-bramwell-shape_aware_functional/sphinx/dev_guide/functional.html#shape-aware-functional


src/serac/physics/heat_transfer.hpp
```diff
@@ -426,8 +412,25 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
   {
     ThermalMaterialIntegrand<MaterialType> integrand(material);
-    residual_->AddDomainIntegral(Dimension<dim>{}, DependsOn<0, 1, 2, active_parameters + NUM_STATE_VARS...>{},
-                                 integrand, mesh_);
+
+    // This does not compile
+    residual_->AddDomainIntegral(Dimension<dim>{}, DependsOn<0, 1, NUM_STATE_VARS + active_parameters...>{}, integrand,
+                                 mesh_);
+
+    // This does compile
```
jamiebramwell: @samuelpmishLLNL , this is where the problem is. I can compile this using the commented out generic lambda, but not using the integrand struct. The compiler says the appropriate paren operator cannot be found.

samuelpmishLLNL: Thanks for the tip, I'm pulling this branch now to take a look

samuelpmishLLNL: looks like it was just missing a `const` on the paren operator (since the functor being invoked was marked `const`, it wasn't allowed to use the non-const member function).

On my mac, all tests are passing except for `boundary_cond` when run in parallel (it passes when run without mpirun).

jamiebramwell: Ahhh that makes sense. Thank you for figuring this out!

jamiebramwell:
I still need to clean up this PR and add some docs, but the tests all pass with `HeatTransfer` and `SolidMechanics` using the new `ShapeAwareFunctional` objects.

jamiebramwell:
This is all working and (mostly) clean and ready for review. I would still like to write a sphinx page on the implemented transformations, but the code itself is ready for review.

src/serac/numerics/functional/shape_aware_functional.hpp
```diff
@@ -0,0 +1,485 @@
...
+                       std::array<const mfem::ParFiniteElementSpace*, num_trial_spaces> trial_fes)
+  {
+    static_assert(test_space.family != Family::HDIV, "Shape-aware functional not implemented for HDiv test functions");
+    static_assert(test_space.family != Family::HCURL,
+                  "Shape-aware functional not implemented for HCurl test functions");
+    static_assert(shape_space.family == Family::H1, "Only H1 spaces allowed for shape displacements");
+
+    for_constexpr<num_trial_spaces>([](auto i) {
+      static_assert(get<i>(trial_spaces).family != Family::HDIV,
+                    "Shape-aware functional not implemented for HDiv trial functions");
```
samuelpmishLLNL: Since no part of serac currently supports Hdiv, I think there are more appropriate places to put this kind of static assertion (e.g. either `serac::Functional` and/or `serac::FiniteElementVector`)

src/serac/numerics/functional/shape_aware_functional.hpp
```diff
@@ -0,0 +1,485 @@
...
+          typename = std::enable_if_t<space_type{}.family == Family::H1 || space_type{}.family == Family::L2>>
+SERAC_HOST_DEVICE auto modify_trial_argument(const jacobian_type& J, space_type /* space */, const trial_type& u)
+{
+  auto du_dx = get<DERIVATIVE>(u);
+
+  // Our updated spatial coordinate is x' = x + p. We want to calculate
+  // du/dx' = du/dx * dx/dx'
+  //        = du/dx * (dx'/dx)^-1
+  //        = du_dx * (J)^-1
+  auto trial_derivative = dot(du_dx, inv(J));
```
samuelpmishLLNL: it looks like this implementation will result in `inv(J)` being evaluated multiple times (once per argument).

src/serac/physics/tests/solid_robin_condition.cpp
```diff
@@ -73,13 +73,13 @@ void functional_solid_test_robin_condition()
...
-        auto f           = u * 3.0 * (X[0] < 0.01);
-        return f;  // define a displacement-proportional traction at the support
-      });
+  solid_solver.addCustomBoundaryIntegral(DependsOn<>{},
+                                         [](double /* t */, auto position, auto displacement, auto /*acceleration*/) {
+                                           auto [X, dX_dxi] = position;
+                                           auto [u, du_dxi] = displacement;
+                                           auto f           = u * 3.0 * (X[0] < 0.01);
+                                           return f;  // define a displacement-proportional traction at the support
+                                         });
```
samuelpmishLLNL: this one might be worth a `// clang-format off/on` block!

samuelpmishLLNL (approved):
looks good, I like the latex expressions for the correction terms in the readthedocs:

Screenshot 2024-01-19 at 8 31 06 AM

I think there are some ways to simplify the way that the qf inputs/outputs are transformed but we can revisit that later if you want.

jamiebramwell:
A few items in my last commit:

1. I moved the Jacobian calculations into a separate struct so they now only occur once per quadrature point (thanks to @samuelpmishLLNL  for pointing this out).
2. I removed the SFINAE in favor of constexpr if statements.
3. I added HCURL transformations and removed the static asserts.

src/docs/sphinx/dev_guide/functional.rst
```diff
@@ -405,3 +405,70 @@ inside a ``std::function`` object with the appropriate signature. This
...
+  r(p, u) &= \int_{\Gamma_p}  t(x, u) \cdot \psi\, dS_x \\
+  &= \int_{\Gamma_p} t(x, u)\cdot  \psi \left \Vert \frac{\partial x}{\partial \xi} \times  \frac{\partial x}{\partial \eta}  \right \Vert d\xi d\eta \\
+  &= \int_{\Gamma} t(X + p, u)\cdot \psi \left \Vert \frac{\partial (X + p)}{\partial \xi} \times  \frac{\partial (X + p)}{\partial \eta}  \right \Vert d\xi d\eta \\
+  \end{align*}
+
+While simple, these transformations can become repetitive and error-prone when writing Q-functions for parameter-free shape optimization problems. To 
+address this issue, we automated these transformations in ``ShapeAwareFunctional``. This object has the same interface and integral definition routines as 
+``Functional``, but it takes the shape displacement field :math:`p` as a required trial parameter. 
+
+The integrands caputured in the ``Functional::Add****Integral`` methods do not need any further modifications to become "shape-aware" and produce the shape derivatives
```
white238:
```suggestion
The integrands captured in the ``Functional::Add****Integral`` methods do not need any further modifications to become "shape-aware" and produce the shape derivatives
```

src/serac/numerics/functional/tests/functional_qoi.cpp
```diff
@@ -410,39 +411,137 @@ TEST(QoI, UsingL2)
...
-  using shape_space     = H1<2, dim>;
+
+  // The shape displacement must be vector-valued H1
+  using shape_space = H1<2, dim>;
+
+  // Shape-aware functional only supports H1 and L2 trial functions
   using parameter_space = H1<1>;
 
-  using qoi_type = serac::Functional<double(shape_space, parameter_space)>;
+  // Define the QOI type. Note that the shape aware functional has an extra template argument
```
white238:
```suggestion
  // Define the QOI type. Note that the shape-aware functional has an extra template argument
```

src/serac/numerics/functional/tests/functional_qoi.cpp
```diff
@@ -410,39 +411,137 @@ TEST(QoI, UsingL2)
...
+  // Define the compile-time finite element spaces for the shape displacement and parameter fields
+  static constexpr int dim{3};
+
+  // The shape displacement must be vector-valued H1
+  using shape_space = H1<2, dim>;
+
+  // Shape-aware functional only supports H1 and L2 trial functions
+  using parameter_space = H1<1, dim>;
+
+  // Define the QOI type. Note that the shape aware functional has an extra template argument
```
white238:
```suggestion
  // Define the QOI type. Note that the shape-aware functional has an extra template argument
```

src/serac/numerics/functional/shape_aware_functional.hpp
```diff
@@ -0,0 +1,469 @@
...
+  /**
+   * @brief Modify the trial argument using the correct physical to reference to shape-adjusted transformation for the
+   * underlying trial function space
+   *
+   * @tparam trial_type The trial function type (may be dual)
+   * @tparam space_type The trial function finite element space (e.g. H1, L2, Hcurl
+   *
+   * @param u The input trial function at a quadrature point
+   *
+   * @return The modified trial function adjusted for the underlying shape displacement field)
```
white238:
```suggestion
   * @return The modified trial function adjusted for the underlying shape displacement field
```

src/serac/numerics/functional/shape_aware_functional.hpp
```diff
@@ -0,0 +1,469 @@
...
+
+/**
+ * @brief Compute the boundary area correction term for boundary integrals with a shape displacement field
+ *
+ * @tparam position_type The position input argument type
+ * @tparam shape_type The shape displacement input argument type
+ * @param X The input position (value and gradient)
+ * @param shape The input shape displacement (value and gradient)
+ * @return The area correction factor transforming the boundary integral from the reference domain area into the
+ * shape-displaced area
```
white238:
```suggestion
 * @tparam position_type The position input argument type
 * @tparam shape_type The shape displacement input argument type
 *
 * @param X The input position (value and gradient)
 * @param shape The input shape displacement (value and gradient)
 *
 * @return The area correction factor transforming the boundary integral from the reference domain area into the
 * shape-displaced area
```

src/serac/numerics/functional/shape_aware_functional.hpp
```diff
@@ -0,0 +1,469 @@
...
+
+  /**
+   * @brief This function lets the user evaluate the serac::ShapeAwareFunctional with the given trial space values
+   *
+   * @note The first argument after time in the argument list is always the shape displacement field.
+   *
+   * @tparam T The types of the arguments passed in
+   * @tparam wrt The index of the argument to take the gradient with respect to. Index 0 is always the shape
+   * displacement
+   * @param t The time
```
white238, line 433:
```suggestion
   *
   * @param t The time
```

src/serac/numerics/functional/shape_aware_functional.hpp
```diff
@@ -0,0 +1,469 @@
...
+    return (*functional_)(DifferentiateWRT<wrt>{}, t, args...);
+  }
+
+  /**
+   * @brief This function lets the user evaluate the serac::ShapeAwareFunctional with the given trial space values
+   *
+   * @note The first argument after time in the argument list is always the shape displacement field.
+   *
+   * @tparam T The types of the arguments passed in
+   * @param t The time
```
white238:
```suggestion
   * @tparam T The types of the arguments passed in
   *
   * @param t The time
```

white238 (approved):

jamiebramwell:
Any other thoughts on this PR? Particularly @kswartz92 and @barreracruz since you will likely be the ones using it. There is an example of what the code would look like in the docs here: https://serac.readthedocs.io/en/feature-bramwell-shape_aware_functional/sphinx/dev_guide/functional.html#shape-aware-functional

barreracruz:
Thank you for including the write-up. I think you don't need a parameter argument when you compute `serac_volume` and `serac_area` in the examples. Is this correct? Also, is there a reason behind defining the shape space as order 2?

kswartz92:
> Any other thoughts on this PR? Particularly @kswartz92 and @barreracruz since you will likely be the ones using it. There is an example of what the code would look like in the docs here: https://serac.readthedocs.io/en/feature-bramwell-shape_aware_functional/sphinx/dev_guide/functional.html#shape-aware-functional

From a high level this looks good to me. I don't have an immediate need, but at some point it might make sense to allow different shape orders in the physics (hard-coded to linear at the moment).

jamiebramwell:
> Thank you for including the write-up. I think you don't need a parameter argument when you compute `serac_volume` and `serac_area` in the examples. Is this correct? Also, is there a reason behind defining the shape space as order 2?

Those examples are completely notional. I think it needs a parameter in the sense that the Functional interface requires some kind of trial function in the constructor.

There is no rationale behind a quadratic shape parameter. I just wanted something higher order to ensure that it works.

PR #1060, opened 2023-12-21
Fix bug and correct parameter getter routine

jamiebramwell:
1. Fixes a bug in the state name getter function for dynamic adjoint problems. While the previous code would compile, it would hit a runtime exception when called.
2. No longer requires the string-based parameter get function to also contain the base physics module name. This means that the strings input in the physics module constructor should be the same for the getter request.

### DIFF
--- a/src/serac/physics/base_physics.cpp
+++ b/src/serac/physics/base_physics.cpp
@@ -310,6 +310,22 @@ std::string addPrefix(const std::string& prefix, const std::string& target)
   }
   return prefix + "_" + target;
 }
+
+std::string removePrefix(const std::string& prefix, const std::string& target)
+{
+  std::string modified_target{target};
+  // Ensure the prefix isn't an empty string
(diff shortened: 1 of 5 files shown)


white238 (approved):

kswartz92:
is it feasible to make the names returned from `BasePhysics::parameterNames()` the same as the parameter names that go into the e.g. SolidMechanics constructor?

jamiebramwell:
> is it feasible to make the names returned from `BasePhysics::parameterNames()` the same as the parameter names that go into the e.g. SolidMechanics constructor?

Probably. I'll look into this.

jamiebramwell:
> > is it feasible to make the names returned from `BasePhysics::parameterNames()` the same as the parameter names that go into the e.g. SolidMechanics constructor?
> 
> Probably. I'll look into this.

done.

tupek2 (approved):

PR #1063, opened 2023-12-22
Clean up tribol codevelop cmake

chapman39:
My previous usage of two big if statements wasn't necessary. More importantly though, we needed to add the subdirectory of `"${PROJECT_SOURCE_DIR}/serac/tribol"` as opposed to `"${PROJECT_SOURCE_DIR}/serac/tribol/src"`. This was causing large parts of the Tribol CMake code to just not happen (e.g. SetupThirdParty.cmake).

Also, I found that we could just remove this section below entirely... originally @white238 and I added this to Tribol's `tribol_install` cmake macro but I've found that always creates a duplication error - even if this section is removed from Serac.

```cmake
        set(tribol_exported_targets tribol)
        blt_list_append(TO tribol_exported_targets ELEMENTS redecomp IF BUILD_REDECOMP)
```


white238 (approved):

chapman39:
Associated tribol PR https://github.com/LLNL/Tribol/pull/51

jamiebramwell (approved):

PR #1069, opened 2024-01-09
Add method to check if physics is quasi-static or not.

tupek2:
The lido wrapper handles quasi-static and dynamic a bit differently (mostly due to checkpointing, but other things).  As such, it is necessary to query which one a given base_physics is.


jamiebramwell (approved):

kswartz92 (approved):
Thanks for adding this!

samuelpmishLLNL (approved):

PR #1071, opened 2024-01-12
Test no-sundials in CI

### LINKED ISSUE #934: Sundials is no longer optional
If this was a mistake, Sundials should be guarded in the code. If it was not, the spack package variant should be removed.

chapman39:
* Adds `SERAC_USE_SUNDIALS` macro
* You can now disable sundials by setting `SUNDIALS_DIR=""` in the configure step
* Adds CI job testing no sundials
* Note that sundials will still be built by MFEM

Fixes #934


chapman39:
Just was missing an include in `equation_solver.cpp`. Should be good to go now @white238 @jamiebramwell

jamiebramwell (approved):

cmake/thirdparty/SetupSeracThirdParty.cmake
```diff
@@ -166,6 +166,20 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
         set(CALIPER_FOUND FALSE)
     endif()
 
+    #------------------------------------------------------------------------------
+    # Sundials
+    #------------------------------------------------------------------------------
+    if (SUNDIALS_DIR)
+        # Sundials is already built from MFEM
+        serac_assert_is_directory(VARIABLE_NAME SUNDIALS_DIR)
+        set(SERAC_USE_SUNDIALS ON CACHE BOOL "")
```
white238, line 175: This needs a comment about how it is brought in by MFEM, or we need to bring it in ourselves.

chapman39: Is `# Sundials is already built from MFEM` okay? Or should I explain something in more detail?

chapman39:
It was failing in docker because the path to sundials config uses `lib`, not `lib64`.

cmake/thirdparty/SetupSeracThirdParty.cmake
white238:
```suggestion
        # Note: Sundials is currently only used via MFEM and MFEM's target contains it's information
        set(SERAC_USE_SUNDIALS ON CACHE BOOL "")
```

Maybe something like this

chapman39:
It seems an azure job oom'ed: https://dev.azure.com/llnl-serac/serac/_build/results?buildId=10753&view=logs&j=b975a676-b456-5eef-7258-9f82d723b73d&t=e365c07f-9d5e-5f1f-caea-656e7a79fb12&l=210

> ##[error]Exit code 137 returned from process: file name '/home/vsts/agents/3.232.3/externals/node20_1/bin/node', arguments '"/home/vsts/work/_tasks/CmdLine_d9bafed4-0b18-4f58-968d-86655b4d2ce9/2.231.1/cmdline.js"'.

white238 (approved):

PR #1072, opened 2024-01-16
fix memory error in element integration

samuelpmishLLNL:
This PR fixes a bug found by @kswartz92 where some element geometries were trying to read floating point values from the {source/flux} outputs from the q-function, even when one of them was a `serac::zero` type.

Unfortunately, I already fixed this bug a year ago for the simplex elements https://github.com/LLNL/serac/pull/873 (https://github.com/LLNL/serac/pull/873/commits/03281a80e95e30a50ca43504177459026340f8c2) but neglected to apply the fix to the remaining element geometries. I'm sorry for any LIDO headaches caused by this.

### DIFF
--- a/src/serac/numerics/functional/detail/segment_L2.inl
+++ b/src/serac/numerics/functional/detail/segment_L2.inl
@@ -182,7 +182,7 @@ struct finite_element<mfem::Geometry::SEGMENT, L2<p, c> > {
             source(qx) = reinterpret_cast<const double*>(&get<SOURCE>(qf_output[qx]))[i * ntrial + j];
           }
           if constexpr (!is_zero<flux_type>{}) {
-            flux(qx)   = reinterpret_cast<const double*>(&get<FLUX>(qf_output[qx]))[i * ntrial + j];
+            flux(qx) = reinterpret_cast<const double*>(&get<FLUX>(qf_output[qx]))[i * ntrial + j];
           }
         }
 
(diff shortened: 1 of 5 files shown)


jamiebramwell (approved):
Thank you for the fix! I'm kind of surprised this didn't come up in our patch tests for various geometries.

src/serac/numerics/functional/detail/segment_L2.inl
```diff
@@ -178,8 +178,12 @@ struct finite_element<mfem::Geometry::SEGMENT, L2<p, c> > {
         f_buffer_type flux;
 
         for (int qx = 0; qx < q; qx++) {
-          source(qx) = reinterpret_cast<const double*>(&get<SOURCE>(qf_output[qx]))[i * ntrial + j];
-          flux(qx)   = reinterpret_cast<const double*>(&get<FLUX>(qf_output[qx]))[i * ntrial + j];
+          if constexpr (!is_zero<source_type>{}) {
```
white238, line 181: Should this be outside of the for loop?

samuelpmishLLNL: The behavior is equivalent both ways. I prefer this implementation slightly because it's a couple lines of code shorter.

white238: Is `q` zero when the `source_type` and/or `flux_type` is `is_zero`?

white238: Also is `source` and `flux` being initialized to a reasonable value since its skipping this part and then used on line 189?

samuelpmishLLNL: > Is q zero when the source_type and/or flux_type is is_zero?

q is always greater than zero

> Also is source and flux being initialized to a reasonable value since its skipping this part and then used on line 189?

the initial value doesn't matter since it's overwritten in the for loop, or it is a `serac::zero` type which has no data in the first place

white238 (approved):

PR #1078, opened 2024-02-13
disambiguate by_attr implementation

### LINKED ISSUE #1077: by_attr helper function won't compile
I noticed that the `by_attr` helper function provided to be used with `serac::Domain` won't compile. When I try to build an executable using it I get this error 
image

samuelpmishLLNL:
fixes https://github.com/LLNL/serac/issues/1077, adds more tests to verify

### DIFF
--- a/src/serac/numerics/functional/domain.hpp
+++ b/src/serac/numerics/functional/domain.hpp
@@ -141,9 +141,10 @@ Domain operator&(const Domain& a, const Domain& b);
 Domain operator-(const Domain& a, const Domain& b);
 
 /// @brief convenience predicate for creating domains by attribute
+template < int dim >
 inline auto by_attr(int value)
 {
-  return [value](auto, int attr) { return attr == value; };
+  return [value](std::vector< tensor<double, dim> >, int attr) { return attr == value; };
 }
 
 }  // namespace serac
--- a/src/serac/numerics/functional/domain.hpp
+++ b/src/serac/numerics/functional/domain.hpp
@@ -141,10 +141,10 @@ Domain operator&(const Domain& a, const Domain& b);
 Domain operator-(const Domain& a, const Domain& b);
 
 /// @brief convenience predicate for creating domains by attribute
-template < int dim >
+template <int dim>
 inline auto by_attr(int value)
 {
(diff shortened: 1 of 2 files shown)


jamiebramwell (approved):
Thanks for the quick fix @samuelpmishLLNL !

kswartz92:
When using a 3D mesh, would `ofBoundaryElements` work exactly like `ofFaces`? And in 2D would `ofBoundaryElements` work exactly as `ofEdges`?

samuelpmishLLNL:
> When using a 3D mesh, would `ofBoundaryElements` work exactly like `ofFaces`? And in 2D would `ofBoundaryElements` work exactly as `ofEdges`?

They're not the same thing. For example, a 3D mesh has face elements on the interior and the boundary, but `ofBoundaryFaces` will not visit interior faces when evaluating the predicate.

kswartz92:
> > When using a 3D mesh, would `ofBoundaryElements` work exactly like `ofFaces`? And in 2D would `ofBoundaryElements` work exactly as `ofEdges`?
> 
> They're not the same thing. For example, a 3D mesh has face elements on the interior and the boundary, but `ofBoundaryFaces` will not visit interior faces when evaluating the predicate.

Got it, thanks! Should we maybe add `ofBoundaryFaces` to the test? I primarily use `Domain` for boundary attributes. Unless it is tested somewhere else in the repo I didn't notice?

samuelpmishLLNL:
maybe the `ofFaces` and `ofEdges` should be removed in favor of `ofInteriorFaces` and `ofBoundaryFaces`. I wonder, is there a real use-case where one would want to include boundary and interior faces in the same domain?

kswartz92 (approved):
thanks @samuelpmishLLNL !

kswartz92:
> maybe the `ofFaces` and `ofEdges` should be removed in favor of `ofInteriorFaces` and `ofBoundaryFaces`. I wonder, is there a real use-case where one would want to include boundary and interior faces in the same domain?

I think it could be useful for solving physics on a partial domain. E.g. an interior material interface may become the domain boundary along with boundary faces.

PR #1081, opened 2024-02-15
adding domain to solid mechanics custom boundary integral

kswartz92:
I need to apply a Robin b.c. to a particular boundary attribute so it would be convenient to apply the custom boundary integral to a `serac::Domain`. This PR is very similar to the changes in #1075.


jamiebramwell (approved):
Sorry I left this one waiting. Thanks @kswartz92 !

jamiebramwell:
@samuelpmishLLNL @btalamini Any comments on this one?

src/serac/physics/tests/solid_robin_condition.cpp
```diff
@@ -74,13 +74,22 @@ void functional_solid_test_robin_condition()
...
   // clang-format off
+  Domain robinDomain = Domain::ofBoundaryElements(StateManager::mesh(mesh_tag), 
+    [](std::vector<serac::tensor<double, dim>> X, int) {
+      bool inDomain = false;
+      for (const auto& p : X) {
+        if (p[0]<0.01) {
+          inDomain = true;
+        }
+      }
+      return inDomain;
```
samuelpmishLLNL:
```suggestion
      for (const auto& point : X) {
        if (point[0]<0.01) { return true; }
      }
      return false;
```

samuelpmishLLNL (approved):
sorry I forgot about this one, it looks good to me

PR #1086, opened 2024-02-29
Dynamic history in memory

jamiebramwell:
This PR optionally saves the transient history information in memory instead of writing and reading from disk. The disk IO option was causing significant slowdowns on parallel file systems.


src/serac/physics/heat_transfer.hpp
```diff
@@ -986,6 +1027,12 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
   /// The temperature finite element state
   serac::FiniteElementState temperature_;
 
+  /// An optional vector of computed temperatures used for dynamic adjoints
+  std::vector<serac::FiniteElementState> temperatures_;
```
tupek2: temperature_ and temperatures_ look very similar.  Perhaps OK, but maybe consider making the name more distinct, e.g. checkpointed_temperatures_?

tupek2 (commented):
Do we have a test for the case when save_to_disk is true?  I may be missing something, but is the behavior different in the 2 cases now?  If !save_to_disk, we auto checkpoint after every nonlinear solve, but I'm not seeing if that is still happening if save_to_disk==true?  I may just be reading to quickly.  Previously, this required a call to 'outputStateToDisk()'?

jamiebramwell:
> Do we have a test for the case when save_to_disk is true? I may be missing something, but is the behavior different in the 2 cases now? If !save_to_disk, we auto checkpoint after every nonlinear solve, but I'm not seeing if that is still happening if save_to_disk==true? I may just be reading to quickly. Previously, this required a call to 'outputStateToDisk()'?

Good point. I'll auto-disk checkpoint if this option is on.

jamiebramwell:
While trying to clean up the names, I came up with this more standardized approach using `BasePhysics` methods. Let me know what you think!

tupek2 (approved):
I like the new strategy.  Better unification of disk and in memory usage.

src/serac/physics/base_physics.hpp
```diff
@@ -339,18 +342,17 @@ class BasePhysics {
    * @brief Accessor for getting named finite element state primal solution from the physics modules at a given
    * checkpointed cycle index
    *
-   * @param state_name The name of the Finite Element State primal solution to retrieve
    * @param cycle The cycle to retrieve state from
    * @return The named primal Finite Element State
    */
-  virtual FiniteElementState loadCheckpointedState(const std::string& state_name, int cycle) const;
+  virtual std::unordered_map<std::string, FiniteElementState> getCheckpointedStates(int cycle) const;
```
kswartz92: why is this signature changing?

tupek2: This should almost be a private function, but I think Kenny wanted it for error estimates, potentially.  Are you using this already Kenny?

tupek2: or maybe that was the get_adjoint function, sorry.

kswartz92: Lido calls this method to grab the checkpointed states and pass them to our qoi operator (and access solutions on the sensitivity pass). This will be a breaking change that we will have to address in Lido, which is fine, just wanted to understand why/if it's necessary.

jamiebramwell: There are two reasons:
1. Using the old disk-based method, this was causing the file to be opened once per field instead of a batch job. It's possible that this disk thrashing was part of our problem.
2. It makes the names slightly more consistent with other methods as this could just be reading from memory instead of loading it from disk.

I'm more concerned about 1, but either way should technically work.

jamiebramwell: any thoughts @tupek2 @kswartz92 ? I think I've found the memory errors and I want to close this one out. I can either revert this change or leave it as-is.

tupek2: I am fairly indifferent.  I think consistency is good, so if we are typically returning maps, we should make that consistent throughout.  Sadly creates churn for Kenny if that is the case!

kswartz92: I'd say I have a slight preference for returning the state. If the other concerns have been addressed, it seems more natural to not require the caller get every state at once. I'm not seeing many other places maps are returned

jamiebramwell: I'll revert this change and we can revisit it once checkpointing by disk is more standard.

jamiebramwell:
@kswartz92 , I reverted the API change. Internally, we are now group caching the requested cycle and just pulling from that if a different field from the same cycle is requested. This should reduce disk thrashing. Let me know if this works for you.

kswartz92 (approved):

jamiebramwell:
> Do we have a test for the case when save_to_disk is true? I may be missing something, but is the behavior different in the 2 cases now? If !save_to_disk, we auto checkpoint after every nonlinear solve, but I'm not seeing if that is still happening if save_to_disk==true? I may just be reading to quickly. Previously, this required a call to 'outputStateToDisk()'?

I made two of the thermal dynamic adjoint tests use disk-based checkpointing and now automatically output to disk every timestep if this option is active.

PR #1089, opened 2024-03-14
Fix mfem failing to include link location for mpi_mpifh

btalamini:
~~Temporary fix.~~ Seems that under some situations, mfem does not properly include the location of all libraries it is linking against. We're not sure exactly what causes this, but Chris has an inkling that it may be when mfem is built with strumpack.

Worked with Chris White.

Note from @white238 : I built this on TOSS4 and verified the very first thing on the link line is the directory I added and the one that is found by Spack is still there. It is correct on mine due to the lack of the subdirectory that @btalamini was running into but I am pretty sure this fixes the issue.


btalamini:
This compiles on my system, but some CI platforms are failing to build due to (more) MPI linking problems. Chris, I'l need more help to get this resolved.

white238 (approved):

cmake/thirdparty/SetupSeracThirdParty.cmake
```diff
@@ -580,4 +580,24 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
...
         endif()
     endforeach()
+
+    # When both MFEM and Strumpack are on, MFEM adds the Fortran
+    # MPI library/directory but in some cases Spack cannot determine
+    # the correct MPI lib directory. This guards against that.
+    # https://github.com/spack/spack/issues/24685
+    if(STRUMPACK_DIR)
+        list(GET MPI_C_LIBRARIES 0 _first_mpi_lib)
+        get_filename_component(_mpi_lib_dir ${_first_mpi_lib} DIRECTORY)
```
jamiebramwell, line 590: This is gross. Thank you for figuring it out!

jamiebramwell (approved):

btalamini:
We eventually got Serac to build without any changes to the Serac repo (other than pointing to a newer spack commit, though we don't know if this was neccesary). I allowed spack to build openMPI v4.1.6. Is the purpose of this PR to allow one to use an existing mpi installation? (I'm sorry, we tried several things, and I forget where this change fit in).

white238:
> We eventually got Serac to build without any changes to the Serac repo (other than pointing to a newer spack commit, though we don't know if this was neccesary). I allowed spack to build openMPI v4.1.6. Is the purpose of this PR to allow one to use an existing mpi installation? (I'm sorry, we tried several things, and I forget where this change fit in).

This will guard against the MPI fortran library not having the correct MPI lib library. It also removes the defaulting to the wrong spack.yaml.

PR #1091, opened 2024-03-21
Move to `tee` for build scripts

white238:
The current way was to run all the commands piped to a file, then read the file to the screen. This will allow more both to be written at the same time, helping with hangs.


chapman39 (approved):

jamiebramwell (approved):
This will be very helpful!

johnbowen42 (approved):
LGTM

PR #1093, opened 2024-03-25
allowing Functional QoI without trial spaces

kswartz92:
This small change allows creating a functional without trial spaces. I was trying to calculate the volume of a mesh and the surface area of a particular boundary attribute and was getting compilation errors when I tried to construct a functional object without a trial space. @samuelpmish suggested this fix.


jamiebramwell (approved):
I would add a comment explaining the `+0` as it's a bit non-intuitive. Otherwise, looks good!

samuelpmishLLNL (approved):

PR #1100, opened 2024-03-29
Fix FindPETSc.cmake path to use cmake current list dir

chapman39:
I was using `CMAKE_MODULE_PATH`, which could contain multiple paths, instead of something like `CMAKE_CURRENT_LIST_DIR`.

Related to: https://github.com/LLNL/serac/pull/1087


white238 (approved):
Thanks for the quick fix @chapman39 !

jamiebramwell (approved):

PR #1106, opened 2024-04-17
Add isoparametric derivatives to domain integral

samuelpmishLLNL:
This PR introduces an interface breaking change that modifies the qfunction signature for `DomainIntegral`s.

Specifically, before this PR the second qfunction argument for `DomainIntegral` was a vector pointing to the quadrature point's spatial position. In `BoundaryIntegral`, the second argument is a tuple of 
1. the spatial position of the quadrature point
2. the derivative of the quadrature point position w.r.t. isoparametric coordinates

----

This was problematic for 2 main reasons:
1. The signatures for qfunctions were inconsistent between `DomainIntegral` and `BoundaryIntegral`, which is confusing for users.
2. some applications need $\displaystyle \frac{\partial X}{\partial \xi}$ (e.g. TMOP / element quality, SUPG ... ) and there was no way to get that info before this PR.

So, these changes make the second argument of `DomainIntegral` qfunctions a tuple that includes $\displaystyle \frac{\partial X}{\partial \xi}$, addressing the issues listed above.

------

The new qfunction signature for `DomainIntegral`s looks like

```cpp
auto qf = [](
  // time
  double t, 

  // a tuple of the quadrature point position and dX_dxi
  tuple< tensor< double, dim >, tensor< double, dim, dim > > position,

  // optional struct of internal variables
  StateType & state,

  // any parameter fields requested
  auto ... params
) {
  // ...
}



```


src/serac/physics/tests/thermal_dynamics_patch.cpp
```diff
@@ -208,8 +208,8 @@ class LinearSolution {
     thermal.setFluxBCs(flux_function, EntireBoundary(thermal.mesh()));
 
     // volumetric source
-    auto source_function = [temp_rate_grad](auto X, auto /* time */, auto /* u */, auto /* du_dx */) {
-      return dot(X, temp_rate_grad);
+    auto source_function = [temp_rate_grad](auto position, auto /* time */, auto /* u */, auto /* du_dx */) {
```
tupek2, line 211: why change the name to position?  Is it x or X?  Aren't thermal things typically defined in terms of x?  But maybe there is no distinction here because its single physics?

tupek2 (approved):

btalamini (approved):

src/serac/physics/tests/thermal_dynamics_patch.cpp
samuelpmishLLNL: > why change the name to position?

I wanted to reserve $X$ for the vector of reference configuration coordinates of the quadrature point. Prior to this PR, that was the only information that was contained in that argument, so it was called $X$. But after this PR, that argument becomes a tuple of { $X$, $\\frac{dX}{d\xi}$ }, so I gave the tuple a new name so $X$ could still be used in the body of the qfunction.

> Is it x or X?

it refers to the reference configuration, so $X$ is clearer to me. For context, I think a lot of the initial qfunctions that were written used $x$, which was potentially confusing. So, I've been trying to go back and change them to $X$ when I touch a relevant part of the code.

> Aren't thermal things typically defined in terms of x? But maybe there is no distinction here because its single physics?

Yes-- I think physics only ever cares about the current configuration, but for the sake of implementation it's more convenient to just provide the reference configuration quantities and let the user apply whatever displacements are appropriate.

samuelpmishLLNL:
I don't know why the lassen build is failing, the error message points to a syntax error isn't present in the actual source code.

I remember seeing this kind of problem on our lassen CI jobs in the past, but I don't remember the cause / solution.

PR #1110, opened 2024-04-29
Python subprocess logging improvement + add `make` job option to build scripts

### LINKED ISSUE #1107: Add command line option for make job parallism
Should the degree of build parallelism be a constant like this? For some systems 6 seems to low and others perhaps too high.

_Originally posted by @tepperly in https://github.com/LLNL/serac/pull/1026#discussion_r1581265178_

chapman39:
This PR aims to improve the `sexe` python function so that subprocess output can be written to both stdout and file simultaneously, line by line. This is useful for when CI jobs crash before creating any log.

Also add `make` job option to build scripts. This is useful because in CI, docker containers should run `make` with a smaller job count (like 4), whereas on LLNL systems the job count can be much greater (like 16).

Fixes #1107


scripts/llnl/common_build_functions.py
```diff
@@ -25,34 +25,42 @@
...
             res = res.decode()
         return p.returncode,res
     elif output_file != None:
-        ofile = open(output_file,"w")
+        # Write to stdout and output file
         p = subprocess.Popen(cmd,
                              shell=True,
-                             stdout= ofile,
+                             stdout=subprocess.PIPE,
                              stderr=subprocess.STDOUT)
```
white238: Can you verify that stderr is going to both the file and the screen?

chapman39: Yes, it does. I created a small script as a test:

```python
sexe("echo hello!", output_file="log0.txt")
sexe("echo bye!", output_file="log1.txt")
sexe("cho oops!", output_file="log2.txt")
```

```bash
[meemee@ruby963:repo]$ ../shell_exec.py 
hello!
bye!
/bin/sh: cho: command not found
[meemee@ruby963:repo]$ cat log?.txt 
hello!
bye!
/bin/sh: cho: command not found
```

chapman39:
With this new change, there is only one subprocess call - making the function a bit cleaner. The only noticeable difference from before is now in the event of an error, it will always print an error message. Before, it would only print an error message if either output file was `None` or return output was `False`.

A set of examples testing the script:

```py
shell_exec("echo hello file", output_file="/usr/workspace/meemee/serac/log0.txt")
shell_exec("echo hello screen", print_output=True)
shell_exec("echo hello to both", output_file="/usr/workspace/meemee/serac/log3.txt", print_output=True)
shell_exec("cho oops!", output_file="/usr/workspace/meemee/serac/log2.txt", print_output=True)
shell_exec("echo cant see me!")
shell_exec("ech mistake with no output!", error_prefix="WARNING:")

rcode, output = shell_exec("echo returning output", return_output=True)
print("printing the output in test.py: {0}".format(output))
```

```
[meemee@ruby968:repo]$ ./scripts/llnl/test.py 
hello screen
hello to both
/bin/sh: cho: command not found
[ERROR: [return code: 127] from command: cho oops!]
[WARNING: [return code: 127] from command: ech mistake with no output!]
printing the output in test.py: returning output

[meemee@ruby968:repo]$ cat ../log?.txt
hello file
/bin/sh: cho: command not found
hello to both
```

chapman39:
Azure using 4 jobs: https://dev.azure.com/llnl-serac/serac/_build/results?buildId=11313&view=logs&j=d86163eb-53f2-5090-1bd9-76480b1ac38b&t=67a51bbe-48ca-58cf-58a4-5e9dff7df8a8&l=304

LC using 16 jobs (CTRL+F `make -j 16`): https://lc.llnl.gov/gitlab/smith/serac/-/jobs/1836806/raw , which I think saves ~2 minutes per job.

scripts/llnl/common_build_functions.py
```diff
@@ -71,11 +93,11 @@ def build_info():
     res["built_from_branch"] = "unknown"
     res["built_from_sha1"]   = "unknown"
     res["platform"] = get_platform()
-    rc, out = sexe('git branch -a | grep \"*\"',ret_output=True,error_prefix="WARNING:")
+    rc, out = shell_exec('git branch -a | grep \"*\"',ret_output=True,error_prefix="WARNING:")
```
white238: `ret_output` was changed to `return_output`

scripts/llnl/common_build_functions.py
```diff
@@ -71,11 +93,11 @@ def build_info():
...
     out = out.strip()
     if rc == 0 and out != "":
         res["built_from_branch"]  = out.split()[1]
-    rc,out = sexe('git rev-parse --verify HEAD',ret_output=True,error_prefix="WARNING:")
+    rc,out = shell_exec('git rev-parse --verify HEAD',ret_output=True,error_prefix="WARNING:")
```
white238: Same

kennyweiss (approved):
Thanks @chapman39 -- looks a lot cleaner and more useful.
In the past, we've had some issues w/ this function and non-ascii text/output, e.g. `utf-8`. It might make sense to explicitly check against those, if possible. (I don't see any changes that would trigger different behavior, but just putting it out there).

scripts/llnl/common_build_functions.py
chapman39: thanks for catching this. fixed.

scripts/llnl/build_src.py
```diff
@@ -52,6 +52,10 @@ def parse_args():
                       dest="verbose",
                       default=False,
                       help="Output logs to screen as well as to files")
+    parser.add_option("-j", "--jobs",
+                      dest="jobs",
+                      default=4,
```
white238: I think we should default to just `-j` which will use the max system amount instead of picking an arbitrary amount.

chapman39: agreed. fixed.

scripts/llnl/common_build_functions.py
```diff
@@ -288,19 +295,15 @@ def build_and_test_host_config(test_root, host_config, report_to_stdout=False, e
...
-    tst_cmd = "cd %s && make CTEST_OUTPUT_ON_FAILURE=1 test ARGS=\"--no-compress-output -T Test -VV -j8\"" % build_dir
-
-    res = sexe(tst_cmd,
-               output_file = tst_output_file,
-               echo=True)
-
-    if report_to_stdout:
-        with open(tst_output_file, 'r') as test_out:
-            print(test_out.read())
+    tst_cmd = "cd %s && make CTEST_OUTPUT_ON_FAILURE=1 test ARGS=\"--no-compress-output -T Test -VV -j %s\"" % (build_dir, job_count)
```
white238: This can't go this high due to the parallelism we are using at the test level with MPI and OpenMP.

I think the formula is actually `(total CPU cores) / (highest level of OpenMP or MPI per test [pretty sure it's 4]) = allowable job slots on the test line`

8 has never caused a problem for us because usually people have 32 cores.

chapman39: How would you suggest handling this?

I could set it to 8 if `jobs` is an empty string. Or I can create a second `jobs-test` argument?

white238: I would either leave it at 8 or honor the option if it is less than 8.

white238 (approved):
Thanks @chapman39

scripts/llnl/common_build_functions.py
```diff
@@ -288,19 +295,20 @@ def build_and_test_host_config(test_root, host_config, report_to_stdout=False, e
...
 
-    tst_cmd = "cd %s && make CTEST_OUTPUT_ON_FAILURE=1 test ARGS=\"--no-compress-output -T Test -VV -j8\"" % build_dir
-
-    res = sexe(tst_cmd,
-               output_file = tst_output_file,
-               echo=True)
+    if job_count == "":
+        test_job_count=8
+    else:
+        test_job_count=min(job_count, 8)
```
white238:
```suggestion
    # Use a maximum of 8 job slots for unit tests due to extra parallelism from OpenMP/MPI
    test_job_count = 8
    if job_count != ""::
        test_job_count = min(job_count, 8)
```

white238: It weirds me out that Python allows variables from inside if blocks to be used....

chapman39:
I think there's some issues checking if "not empty string" vs "empty string" in python. ill push a fix soon

edit: should be fixed now

btalamini (approved):

PR #1112, opened 2024-05-01
newton line search

tupek2:


samuelpmishLLNL (commented):
I see that there is one integration test for each of the new solver options, could we also add at least one unit test for those classes as well (e.g. some toy problem like rosenbrock), and maybe an manufactured test that results in a NaN.

Also, I'm not familiar with the trust region implementation, do you have a theory reference document that describes the algorithm?

kswartz92 (approved):

src/serac/numerics/equation_solver.cpp
```diff
@@ -5,20 +5,593 @@
...
+      x0.Add(1.0, x);
+
+      real_t stepScale = 1.0;
+      add(x0, -stepScale, c, x);
+      norm = evaluate_norm(x, r);
+
+      const int               max_ls_iters = nonlinear_options.max_line_search_iterations;
+      static constexpr real_t reduction    = 0.5;
+
+      const double sufficientDecreaseParam = 0.0;  // 1e-15;
```
btalamini, line 109: Nitpick: the style is inconsistent. Camel case and underscores are mixed. Could you make a note to clean this up later? I know some of this is bound to change soon, anyway.

btalamini (approved):
Looks good. Nice addition.

src/serac/numerics/equation_solver.cpp
tupek2: We have a mix throughout Serac, and its different than MFEM.  I agree we should evolve toward a consistent one.  I've been trying to do _ for vars, and lower-camel for functions.  But that goes against my years of habits, which is the opposite, so I swap those often.  Is there a way to automate?

PR #1113, opened 2024-05-07
Add `SERAC_ENABLE_BENCHMARKS` cmake option. Test profiling/ benchmarks in CI.

chapman39:
This PR...

* Adds an explicit `SERAC_ENABLE_BENCHMARKS` cmake option. It will automatically be set to `ON` if `ENABLE_BENCHMARKS=ON`. It can still optionally be set to `OFF`.
* Automatically sets `SERAC_ENABLE_PROFILING=ON` if either benchmarking variable is `ON`.
* Turns on benchmarking/ profiling in CI
* Fixes issue in the case `EXTRA_BUILD_OPTIONS` has spaces in `build_and_test.sh`


white238:
We have talked about this vaguely in two meetings but to be clear about what we should do. We should have two CMake options:

* `SERAC_ENABLE_BENCHMARKS` - this should enable the actual benchmark executables themselves
* `SERAC_ENABLE_PROFILING` - this should enable the caliper C++ macros and their use inside of Serac's libraries

Benchmarks are reliant on Profiling being on but we should be able to turn on Profiling without having the benchmarks built.

They should also be guarded for our project so downstream projects can turn off functionalities they don't need.

chapman39:
Thanks for the clarification @white238. I updated this PR's description to adjust for recent changes.

cmake/SeracBasics.cmake
```diff
@@ -28,6 +28,9 @@ endif()
 option(SERAC_ENABLE_CODE_CHECKS "Enable Serac's code checks" ${_enable_serac_code_checks})
 option(SERAC_ENABLE_PROFILING "Enable profiling functionality" OFF)
 
+cmake_dependent_option(SERAC_ENABLE_BENCHMARKS "Enable benchmark executables" ON
```
white238: Can you add a check here that does something like this:

```
if ((SERAC_ENABLE_BENCHMARKS OR ENABLE_BENCHMARKS) AND NOT SERAC_ENABLE_PROFILING)
    message(FATAL_ERROR "SERAC_ENABLE_BENCHMARKS requires SERAC_ENABLE_PROFILING to be turned on")
endif()
```

So we have a better user experience.

white238 (approved):
Thanks @chapman39 ! Especially dealing with my everchanging wants!

chapman39:
@btalamini @samuelpmish @jamiebramwell Just need one more review.

btalamini (approved):

PR #1114, opened 2024-05-08
fix off-by-one error when looking up boundary element attributes

samuelpmishLLNL:
thanks to @kswartz92 for noticing an error in a surface area calculation and providing a minimal reproducer


white238 (approved):

kswartz92:
Would it make sense to add the example I provided as a test?

chapman39 (approved):

tupek2 (approved):
I wouldn't mind Kenny's test, if its a unit test.  Could be in a follow on commit though.  Up to you Sam.

btalamini:
I don't want to pile on, but I agree with Kenny and Mike, this should be covered with a unit test. Sam, if you're busy with the Functional refactor, I can take that up.

PR #1118, opened 2024-05-14
Give jobs a deadline on Quartz

chapman39:
With this PR, CI jobs now have a deadline on Quartz. This way, if jobs get stuck on the queue after gitlab fails, it won't clog up the system. I choose an hour because that's the same deadline as Gitlab.

If `--deadline=now+1hour` and `-t25`, the job will dequeue after 35 minutes (60-25).

I couldn't find a similar argument on `lalloc`/ `bsub` commands (Lassen).

Also, I snuck in some more copyright updates.

### DIFF
--- a/.gitlab/build_quartz.yml
+++ b/.gitlab/build_quartz.yml
@@ -2,7 +2,7 @@
 # This is the shared configuration of jobs for quartz
 .on_quartz:
   variables:
-    ALLOC_OPTIONS: "--res=ci --exclusive=user -A ${ALLOC_BANK}"
+    ALLOC_OPTIONS: "--res=ci --exclusive=user -A ${ALLOC_BANK} --deadline=now+1hour"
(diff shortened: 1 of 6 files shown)


chapman39:
We can change the deadline time to whatever makes sense. We could also use the `immediate` option instead.

Sort of a funny way to test this argument is by allocating 60 minutes and setting the deadline to an hour from now - requiring you immediately get an allocation.

```
[meemee@ruby965:repo]$ s1 -t60 --deadline=now+1hour echo "hello"
srun -N1 -Aeng -t60 --deadline=now+1hour echo hello
srun: job 2630692 queued and waiting for resources
srun: Job has been cancelled
srun: error: Unable to allocate resources: Job/step already completing or completed
```

white238 (approved):

btalamini (approved):
Great idea.

PR #1132, opened 2024-06-05
Tupek/feature/reaction sensitivities

tupek2:


src/serac/numerics/functional/domain.cpp
```diff
@@ -118,10 +118,12 @@ static Domain domain_of_edges(const mfem::Mesh& mesh, std::function<T> predicate
       int attr   = (bdr_id > 0) ? mesh.GetBdrAttribute(bdr_id) : -1;
       if (predicate(x, attr)) {
         output.edge_ids_.push_back(i);
+        output.mfem_edge_ids_.push_back(i);
```
white238, line 121: Reminder to document the difference and/or reason behind keeping these two mappings.

samuelpmishLLNL: apparently, the table at the top of domain.cpp already explained the two sets of indices (although it doesn't appear in the diff, since that was added in an earlier PR)

samuelpmishLLNL (approved):

btalamini (approved):

PR #1135, opened 2024-06-11
Fix bug in BC helper function for vector spaces

kswartz92:
A while back I merged a helper function for grabbing the tdofs associated with element attribute (MFEM only provides a method for boundary attributes). At the time I had only used the function with scalar FE spaces. I was recently using this function with a vector FE space and realized it would only grab the first component of the field when I was asking for all components due to using the wrong MFEM method to grab the element dofs.


samuelpmishLLNL (approved):
can we also add a small test that verifies the new implementation is working as intended?

kswartz92:
> can we also add a small test that verifies the new implementation is working as intended?

Yeah I'll add this

tupek2 (approved):

kswartz92:
> > can we also add a small test that verifies the new implementation is working as intended?
> 
> Yeah I'll add this

Added the test, ready to merge if @samuelpmishLLNL and @tupek2 are happy with the current state

samuelpmishLLNL (approved):

samuelpmishLLNL:
> Added the test, ready to merge if @samuelpmishLLNL and @tupek2 are happy with the current state

Thanks for the reminder!

PR #1138, opened 2024-06-12
Utilize MFEM's petsc/ slepc wrappers

chapman39:
* Let MFEM manage petsc/ slepc TPLs
* Build MFEM with petsc/ slepc wrappers
* Change smoke tests to use MFEM's examples instead


src/tests/CMakeLists.txt
```diff
@@ -38,14 +38,10 @@ endif()
...
-                    DEPENDS_ON    PkgConfig::PETSC gtest axom::slic
-                    NUM_MPI_TASKS 4)
-
-    if(SLEPC_FOUND)
-        serac_add_tests(SOURCES       slepc_smoketest.cpp
-                        DEPENDS_ON    PkgConfig::SLEPC gtest axom::slic
-                        NUM_MPI_TASKS 4)
-    endif()
-endif()
+serac_add_tests(SOURCES       mfem_petsc_smoketest.cpp
```
white238: Was there a discussion that PETSC and SLEPC were now a hard requirement?

chapman39: My understanding is that when mfem is built with `+petsc+slepc`, we will have full access to those libraries, so I figured there would be no point to guard whether we're building with petsc/ slepc. I could re-add them though.

That is, assuming this PR goes through. I'm having issues with slepc.

The motivation behind this PR as a whole was a discussion I had with @tupek2 and his summer student (I'm not sure what his github account is) during the bug fix day. They mentioned it would be much easier to use petsc if they could use the mfem wrappers.

white238: If we are allowing MFEM to handle the petsc/slepc build options (which is fine imo), we need to check if they are enabled in the MFEM build or make them a hard requirement in our builds.

They are currently variants in our Spack package recipe:

https://github.com/LLNL/serac/blob/242c2b7d402fa4e774b16f70d91224892d6189c9/scripts/spack/packages/serac/package.py#L60-L62

If someone turns off these variants then these unit tests would fail to build.

chapman39: That's a good point. I'll update this.

chapman39:
At the moment, BlueOS builds get this warning many times during the serac build:

`clang-10: warning: -Wl,-rpath=/usr/WS2/smithdev/libs/serac/blueos_3_ppc64le_ib_p9/2024_06_24_13_38_57/clang-10.0.1/arpack-ng-3.9.0-jf6qgpron7zyy466nlme6v3fdr6vhp7u/lib64: 'linker' input unused [-Wunused-command-line-argument]`

This is due to the fact I had to manually add the arpack lib directories in the MFEM module. I could check if the `SYS_TYPE` is BlueOS, and in that case not add the `-Wl,-rpath`?

MFEM team is aware of this and are considering adding support to find arpack within MFEM itself. I have an issue on this here: https://github.com/mfem/mfem/issues/4364

chapman39:
I'm having some issues getting the mfem smoke tests to work with cuda, which is why I've commented them out. I get the following error:

```
lalloc 1 -W30 ctest -R mfem_petsc_smoketest --output-on-failure

[3]PETSC ERROR: PETSc is configured with GPU support, but your MPI is not GPU-aware. For better performance, please use a GPU-aware MPI.
[3]PETSC ERROR: If you do not care, add option -use_gpu_aware_mpi 0. To not see the message again, add the option to your .petscrc, OR add it to the env var PETSC_OPTIONS.
[3]PETSC ERROR: If you do care, for IBM Spectrum MPI on OLCF Summit, you may need jsrun --smpiargs=-gpu.
[3]PETSC ERROR: For Open MPI, you need to configure it --with-cuda (https://www.open-mpi.org/faq/?category=buildcuda)
[3]PETSC ERROR: For MVAPICH2-GDR, you need to set MV2_USE_CUDA=1 (http://mvapich.cse.ohio-state.edu/userguide/gdr/)
[3]PETSC ERROR: For Cray-MPICH, you need to set MPICH_GPU_SUPPORT_ENABLED=1 (man mpi to see manual of cray-mpich)
```

> your MPI is not GPU-aware

Is this true? I don't really understand. @white238 

EDIT: This is probably because I haven't configured petsc with cuda... I just found out petsc is a `CudaPackage`, which I'm guessing means it has a `cuda` variant? https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/petsc/package.py#L10

DOUBLE EDIT: Going back to this, turns out petsc was build with `+cuda`, so I'm back to not knowing what's going on.

scripts/spack/packages/serac/package.py
```diff
@@ -98,6 +98,8 @@ class Serac(CachedCMakePackage, CudaPackage):
     depends_on("mfem+amgx", when="+cuda")
     depends_on("mfem+asan", when="+asan")
     depends_on("mfem+strumpack", when="+strumpack")
+    depends_on("mfem+petsc", when="+petsc")
```
zatkins-work, line 101: There may be some additional variant flags we want to set for PETSc by default -- e.g. `^petsc+scalapack+strumpack+suite-sparse+mumps` to ensure better interoperability with the things MFEM is built with.

src/tests/mfem_petsc_smoketest.cpp
```diff
@@ -0,0 +1,453 @@
...
+      // options unless forcewrap is true.
+      bool wrap = forcewrap ? true : (pa ? true : !strlen(petscrc_file));
+      if (wrap)
+      {
+         pcg = new PetscPCGSolver(MPI_COMM_WORLD);
+         pcg->SetOperator(*A);
+         if (useh2)
+         {
+            delete prec;
+            prec = new PetscH2Solver(*A.Ptr(),fespace);
```
zatkins-work, line 319: This line will error if PETSc isn't configured with `--download-h2opus`, which doesn't seem to be an option exposed to Spack. Maybe we should remove the `useh2` flag?

src/tests/mfem_petsc_smoketest.cpp
```diff
@@ -0,0 +1,453 @@
...
+      pcg->SetPrintLevel(1);
+
+      UserMonitor mymon(a,b);
+      if (visualization && petscmonitor)
+      {
+         pcg->SetMonitor(&mymon);
+         pcg->iterative_mode = true;
+         X.Randomize();
+      }
+      pcg->Mult(B, X);
```
zatkins-work, line 352:
```suggestion
      pcg->Mult(B, X);
// SERAC_EDIT_BEGIN
      KSPConvergedReason reason;
      KSPGetConvergedReason(pcg, &reason);
      EXPECT_GT(reason, 0);
// SERAC_EDIT_END
```
We should probably check that we actually solved the system, given that the check is pretty free.

src/tests/mfem_petsc_smoketest.cpp
```diff
@@ -0,0 +1,453 @@
...
+  // #endif
+  int fake_argc = sizeof(fake_argv) / sizeof(fake_argv[0]);
+  ex1_main(fake_argc, const_cast<char**>(fake_argv));
+  std::string output = ::testing::internal::GetCapturedStdout();
+
+  int rank = 0;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+  if (rank == 0) {
+    EXPECT_NE(output.find("--order 1"), std::string::npos);
+    EXPECT_NE(output.find("0 KSP Residual norm"), std::string::npos);
```
zatkins-work: Alternative way to check if we converged: If we add the command line flag `-ksp_converged_reason`, we can then search the output for a string like `KSP_CONVERGED_` to make sure that the solver didn't diverge. See [PETSc docs](https://petsc.org/release/manualpages/KSP/KSPConvergedReason/) for a full list of the converged reasons if you're curious. I think this is worse than the other option I suggested.

scripts/spack/packages/serac/package.py
zatkins-work: Not sure if spack has a built in way to say that those dependencies should propagate.

zatkins-work (approved):
Looks good to me! I'm building now to make sure it all builds locally.

scripts/spack/packages/serac/package.py
chapman39: I don't think there is, but I can add these in the `depends_on("petsc...` section

chapman39:
I am off today but I wanted to make sure CI passed. Looks like only Codevelop is failing, which I'll be able to fix without much issues Monday.

cmake/thirdparty/FindARPACK.cmake
```diff
@@ -0,0 +1,31 @@
...
+# ARPACK_LIBRARIES
+
+# Set ARPACK_DIR to specify the prefix of ARPACK installation
+
+if(ARPACK_DIR)
+    set(ARPACK_INCLUDE_DIRS "${ARPACK_DIR}/include")
+
+    foreach(_lib_dir ${ARPACK_DIR}/lib;${ARPACK_DIR}/lib64)
+      if (EXISTS ${_lib_dir})
+        set(ARPACK_LIBRARIES "${_lib_dir}/libparpack.so")
```
zatkins-work: This line is an issue for MacOS builds -- I need `set(ARPACK_LIBRARIES "${_lib_dir}/libparpack.dylib")`

zatkins-work: Can we not use the `pkgconfig` files generated by ARPACK?

zatkins-work: Other than this minor issue, this builds fine locally!

btalamini (approved):

cmake/thirdparty/FindARPACK.cmake
chapman39: Ah, good point.

scripts/spack/packages/serac/package.py
chapman39: Okay I looked into this a bit more. This is from the latest TPL build spec (TOSS4):

```
 -   l3f6io5f4k3u4gsgaqowqwxdx6n5bm25      ^mfem@4.7.0.0%gcc@10.3.1~amgx~asan~conduit~cuda~debug~examples~exceptions~fms~ginkgo~gnutls~gslib~hiop+lapack~libceed~libunwind+metis~miniapps~mpfr+mpi~mumps+netcdf~occa~openmp+petsc~pumi~raja~rocm~shared+slepc+static+strumpack~suite-sparse+sundials+superlu-dist~threadsafe~umpire+zlib build_system=generic patches=8d01394 timer=auto arch=linux-rhel8-ivybridge
...
 -   zblarmtekggmgfqfj6cw2z7wttmwt7yt      ^petsc@3.21.0%gcc@10.3.1~X~batch~cgns~complex~cuda~debug+double~exodusii~fftw+fortran~giflib~hdf5~hpddm~hwloc+hypre~int64~jpeg~knl~kokkos~libpng~libyaml~memkind+metis~mkl-pardiso~mmg~moab~mpfr+mpi~mumps+openmp~p4est~parmmg~ptscotch~random123~rocm~saws~scalapack~shared+strumpack~suite-sparse+superlu-dist~sycl~tetgen~trilinos~valgrind~zoltan build_system=generic clanguage=C memalign=none arch=linux-rhel8-ivybridge
[e]  cxr37yod2jojy2bqaehlfohnhpkww62e          ^diffutils@3.6%gcc@10.3.1 build_system=autotools arch=linux-rhel8-ivybridge
 -   ix44yrtjbpgkmioueo6k5xrgkn7ert27          ^netlib-scalapack@2.2.0%gcc@10.3.1~ipo~pic+shared build_system=cmake build_type=Release generator=make patches=072b006,1c9ce5f,244a9aa arch=linux-rhel8-ivybridge
 ```

* strumpack - This variant is handled already for petsc.
* suite-sparse and mumps - These variants appear to be disabled in mfem. Should they be turned on?
* scalapack - This is sort of weird, but petsc seems to turn this on if `strumpack` is turned on [here](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/petsc/package.py#L322). I could add `+scalapack` to be explicit, though.

File: cmake/thirdparty/FindARPACK.cmake (hunk 6 above)
Comment (chapman39) [reply cmake/thirdparty/FindARPACK.cmake]: @zatkins-work Let me know if my recent changes resolves your build issues on mac

File: src/tests/mfem_petsc_smoketest.cpp (hunk 5 above)
Comment (chapman39) [reply src/tests/mfem_petsc_smoketest.cpp]: I agree we should check if it has converged. I have noticed it prints this: `Linear  solve converged due to CONVERGED_ATOL iterations 17`. I could search for `Linear  solve converged due to CONVERGED_`?

I was originally doing something like this, but when I was trying to test the cuda builds they weren't converging, or something like that.

File: cmake/thirdparty/FindARPACK.cmake (hunk 6 above)
Comment (white238) [reply cmake/thirdparty/FindARPACK.cmake]: You should be using `find_library`:

https://cmake.org/cmake/help/latest/command/find_library.html

This will toggle `lib` and `lib64` and the file extension as well:

```
find_library(_arpack_library parpack PATHS ${ARPACK_DIR}/lib)
```

Comment (white238) [reply cmake/thirdparty/FindARPACK.cmake]: You should also search for a specific header so you know that the installation works.  Here is an example:

https://github.com/LLNL/axom/blob/develop/src/cmake/thirdparty/FindC2C.cmake

Comment (chapman39) [reply cmake/thirdparty/FindARPACK.cmake]: just fixed that, thanks

File: scripts/spack/packages/serac/package.py (hunk 2 above)
Comment (chapman39) [reply scripts/spack/packages/serac/package.py:101]: Setting as resolved. Created issue here for possible follow up PR: https://github.com/LLNL/serac/issues/1161

File: cmake/thirdparty/SetupSeracThirdParty.cmake [hunk 7]
```diff
@@ -136,6 +136,49 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
     endif()
     message(STATUS "Sundials support is ${SERAC_USE_SUNDIALS}")
 
+    #------------------------------------------------------------------------------
+    # PETSc
+    #------------------------------------------------------------------------------
+    if (PETSC_DIR)
+        serac_assert_is_directory(DIR_VARIABLE PETSC_DIR)
+        set(SERAC_USE_PETSC ON CACHE BOOL "")
```
white238, line 145: We should have a comment explaining the anti-pattern here. Just something simple that says we are using PETSC/SLEPC through MFEM.

cmake/thirdparty/SetupSeracThirdParty.cmake
```diff
@@ -136,6 +136,51 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
...
+    else()
+        set(SERAC_USE_PETSC OFF CACHE BOOL "")
+        set(PETSC_FOUND FALSE)
+    endif()
+    message(STATUS "PETSc support is ${SERAC_USE_PETSC}")
+
+    #------------------------------------------------------------------------------
+    # SLEPc
+    #------------------------------------------------------------------------------
+    if (SLEPC_DIR AND SERAC_USE_PETSC)
```
white238, line 158: Should this also be inside of the `if(PETSC_DIR)` section like below?

scripts/spack/packages/mfem/fix-dep-ordering.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/makefile b/makefile
+index a4b6296da..b8c62eb21 100644
+--- a/makefile
++++ b/makefile
+@@ -285,7 +285,7 @@ endif
+ 
+ # List of MFEM dependencies, that require the *_LIB variable to be non-empty
+ MFEM_REQ_LIB_DEPS = ENZYME SUPERLU MUMPS METIS FMS CONDUIT SIDRE LAPACK SUNDIALS\
+- SUITESPARSE STRUMPACK GINKGO GNUTLS NETCDF PETSC SLEPC MPFR PUMI HIOP\
++ SUITESPARSE STRUMPACK GINKGO GNUTLS NETCDF SLEPC PETSC MPFR PUMI HIOP\
```
white238, line 10: Does this need to go into spack or mfem?

scripts/spack/packages/serac/package.py
```diff
@@ -204,6 +206,8 @@ class Serac(CachedCMakePackage, CudaPackage):
     # Conflicts
     #
 
+    conflicts("~petsc", when="+slepc", msg="PETSc must be built when building with SLEPc!")
```
white238, line 209: 👍

cmake/thirdparty/SetupSeracThirdParty.cmake
white238: I guess that is what the `SERAC_USE_PETSC` does...

white238 (approved):

scripts/spack/packages/mfem/fix-dep-ordering.patch
chapman39: MFEM. I have a PR for it here https://github.com/mfem/mfem/pull/4363

PR #1143, opened 2024-06-12
Tupek/checkpoint unification

tupek2:


btalamini (approved):
This looks good. It will be nice when there isn't so much repetition between the physics modules. Maybe as you refactor them, getting some of this stuff out of the physics modules will provide opportunities to de-duplicate.

tupek2:
An upcoming refactor is all about consolidating this stuff and much more.  May have to wait until after summer though.

kswartz92 (approved):

PR #1144, opened 2024-06-13
Eigendecomposition of 3x3 symmetric tensors

btalamini:
Implements the Scherzinger-Dohrmann fast and accurate eigendecomposition of 3x3 symmetric matrices. I am intentionally not implementing derivatives, and avoiding the non-differentiability of the eigenvectors at eigenvalue degeneracies. The envisioned use case is for computing isotropic tensor functions, which themselves are differentiable even when repeated eigenvalues occur (where the underlying scalar function is differentiable, of course). I will add some of the commonly used ones in a successive PR.


tupek2 (approved):
We can speed this up with the optimism version if/when it becomes a bottleneck!

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -754,4 +754,163 @@ auto eigenvalues(const serac::tensor<T, size, size>& A)
...
+ *
+ * @param A Matrix for which the eigendecomposition will be computed. Must be
+ * symmetric, this is not checked.
+ * @return tuple with the eigenvalues in the first element, and the matrix of
+ * eigenvectors (columnwise) in the second element.
+ * 
+ * @note based on "A robust algorithm for finding the eigenvalues and
+ * eigenvectors of 3x3 symmetric matrices", by Scherzinger & Dohrmann
+ */
+SERAC_HOST_DEVICE tuple<vec3, mat3> eig_symm(const mat3 & A) {
```
tupek2, line 799: inline

btalamini:
@samuelpmishLLNL, this is ready to review. Would you mind taking a look?

src/serac/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -754,4 +754,161 @@ auto eigenvalues(const serac::tensor<T, size, size>& A)
...
+    // angle used to find eigenvalues
+    double tmp   = (0.5 * J3) * std::pow(3.0 / J2, 1.5);
+    double alpha = std::acos(fmin(fmax(tmp, -1.0), 1.0)) / 3.0;
+
+    // consider the most distinct eigenvalue first
+    if (6.0 * alpha < M_PI) {
+      eta[0] = 2 * std::sqrt(J2 / 3.0) * std::cos(alpha);
+    } else {
+      eta[0] = 2 * std::sqrt(J2 / 3.0) * std::cos(alpha + 2.0 * M_PI / 3.0);
+    }
```
samuelpmishLLNL, line 824: If anyone else was surprised to see trigonometry and angles show up when solving a cubic equation, here's an animation of the underlying idea:
cubic_roots

The cubic roots can be expressed as x-coordinates of 3 points (120 degrees apart) on a circle with center `tr(A) / 3.0` and radius `2 * std::sqrt(J2 / 3.0)`.

samuelpmishLLNL (approved):

PR #1149, opened 2024-06-17
Implement finite deformation log strain J2 plasticity

btalamini:
Implements a J2 plasticity model for finite deformation. The model is based on the multiplicative decomposition $F = F^e F^p$, and the elastic response is hyperelastic and isotropic using the logarithmic elastic strain. 

Implementation details: The update of the plastic distortion tensor $F^p$ is through the first order exponential integrator, $F^p_{n+1} = \exp(\Delta \bar{\epsilon}^p N^p) F^p_{n} $. This technique, from [Weber and Anand](https://doi.org/10.1016/0045-7825(90)90131-5), has the special property that the plastic flow is exactly volume preserving. Another nice feature is that the most of the algorithm coincides with the small strain one, except for a small amount of pre- and post-processing to account for the finite deformation kinematics. We could easily implement this model and the linear version in the same struct, with just a few helper functions. For now, I'm leaving it separate.


btalamini:
@samuelpmishLLNL and @tupek2, this is ready for review.

src/serac/physics/materials/tests/nonlinear_J2_material.cpp
```diff
@@ -77,6 +77,166 @@ TEST(NonlinearJ2Material, Uniaxial)
...
+  double nu      = 0.25;
+  double sigma_y = 350e6;
+  double eps0    = sigma_y / E;
+  double n       = 3;
+
+  // hardening model
+  solid_mechanics::PowerLawHardening hardening{.sigma_y = sigma_y, .n = n, .eps0 = eps0};
+
+  // material model
+  solid_mechanics::J2FiniteDeformationNonlinear<decltype(hardening)> material{
```
samuelpmishLLNL: most of the `decltype(hardening)` in these tests can be omitted if we add a Class Template Argument Deduction guide (see https://godbolt.org/z/P1q1v3hPG, although CTAD doesn't work with the designated initializer syntax `{.foo = foo, .bar = bar, ...}`).

Alternatively, some `using` statements like

```cpp
using hardening_t = solid_mechanics::PowerLawHardening;
using material_model_t = solid_mechanics::J2FiniteDeformationNonlinear< hardening_t >;
```

at the top of the file could make some of the declarations in it a little cleaner

samuelpmishLLNL (approved):

tupek2 (approved):

src/serac/physics/materials/tests/nonlinear_J2_material.cpp
btalamini: Thanks, CTAD was new to me, so I learned something. For this case, I chose the second route, because we want the materials to be simple for C++ laypeople to implement, and I don't want to give users the impression that CTAD is a necessary ingredient.

PR #1150, opened 2024-06-18
revert axom submodule

white238:


white238:
I did notice this is out of sync with the commit in the Spack package. We should prioritize rebuilding the TPLs soon. I gather this was included to fix the Axom bug that was recently introduced and fix that would have caused issues in the codevelop build.

chapman39 (approved):
thanks @white238

tupek2 (approved):

PR #1151, opened 2024-06-19
Fix lido bug relating to slepc

chapman39:
Fixes a lido bug in the case they build codeveloping serac and petsc/ slepc disabled.

MFEM will set SLEPC_DIR to something wrong if it's not included in that list. Then, (since SLEPC_DIR has been set) Serac will assume we want slepc enabled and will look for it in the wrong location.


chapman39:
The commit [00add9f](https://github.com/LLNL/serac/pull/1151/commits/00add9f4b443cab6985ed228775edbea0fb54961) will be temporarily merged into lido. Since serac develop already guards against petsc/ slepc not being found (main lido does not have this yet), only a minor change is needed for serac.

white238 (approved):

white238:
I would love to figure out where and why MFEM is messing with those variables....

tupek2 (approved):

PR #1153, opened 2024-06-20
Add support for using PETSc preconditioners and solvers

zatkins-work:
This PR extends the MFEM PETSc wrappers to allow PETSc solvers to be used interchangeably with the existing MFEM solvers with minimal conversion of matrix types (at most once per Newton step). Most importantly, this allows the use of PETSc's [geometric algebraic multigrid](https://petsc.org/release/manualpages/PC/PCGAMG/) with a properly set near null space for elasticity and PETSc's [hierarchical multigrid](https://petsc.org/release/manualpages/PC/PCHMG/) which uses GAMG on each of the components of a multi-component PDE. 

Additionally, this allows for using the nonlinear and linear solvers available in PETSc alongside existing MFEM solvers and preconditioners.

### DIFF
--- a/src/serac/numerics/equation_solver.cpp
+++ b/src/serac/numerics/equation_solver.cpp
@@ -796,12 +796,14 @@ std::unique_ptr<mfem::NewtonSolver> buildNonlinearSolver(const NonlinearSolverOp
     nonlinear_solver = std::make_unique<NewtonSolver>(comm, nonlinear_opts);
   } else if (nonlinear_opts.nonlin_solver == NonlinearSolver::TrustRegion) {
     nonlinear_solver = std::make_unique<TrustRegion>(comm, nonlinear_opts, linear_opts, prec);
+#if defined(MFEM_USE_PETSC) && defined(SERAC_USE_PETSC)
   } else if (nonlinear_opts.nonlin_solver == NonlinearSolver::PetscNewton) {
     nonlinear_solver = std::make_unique<mfem_ext::PetscNewtonSolver>(comm, SNESNEWTONLS, SNESLINESEARCHBASIC);
   } else if (nonlinear_opts.nonlin_solver == NonlinearSolver::PetscNewtonBacktracking) {
     nonlinear_solver = std::make_unique<mfem_ext::PetscNewtonSolver>(comm, SNESNEWTONLS, SNESLINESEARCHBT);
   } else if (nonlinear_opts.nonlin_solver == NonlinearSolver::PetscNewtonCriticalPoint) {
     nonlinear_solver = std::make_unique<mfem_ext::PetscNewtonSolver>(comm, SNESNEWTONLS, SNESLINESEARCHCP);
+#endif
   }
   // KINSOL
   else {
@@ -886,6 +888,7 @@ std::pair<std::unique_ptr<mfem::Solver>, std::unique_ptr<mfem::Solver>> buildLin
     case LinearSolver::PetscGMRES:
       SLIC_ERROR_ROOT("PETSc linear solver requested for non-PETSc build.");
       exitGracefully(true);
+      break;
 #endif
     default:
       SLIC_ERROR_ROOT("Linear solver type not recognized.");
@@ -991,10 +994,6 @@ std::unique_ptr<mfem_ext::PetscPCSolver> buildPetscPreconditioner(PetscPCType pc
       preconditioner = std::make_unique<mfem_ext::PetscPCSolver>(comm, PCLU);
       // Automatically shift the LU factorization to ensure positive definiteness
       PetscCallAbort(comm, PCFactorSetShiftType(*preconditioner, MAT_SHIFT_POSITIVE_DEFINITE));
-#ifdef PETSC_HAVE_SUPERLU_DIST
-      // Use the SuperLU_dist solvers, if available
-      PetscCallAbort(comm, PCFactorSetMatSolverType(*preconditioner, MATSOLVERSUPERLU_DIST));
-#endif
       break;
     case PetscPCType::ILU:
       preconditioner = std::make_unique<mfem_ext::PetscPCSolver>(comm, PCILU);
@@ -1003,11 +1002,6 @@ std::unique_ptr<mfem_ext::PetscPCSolver> buildPetscPreconditioner(PetscPCType pc
       break;
     case PetscPCType::CHOLESKY:
       preconditioner = std::make_unique<mfem_ext::PetscPCSolver>(comm, PCCHOLESKY);
-      // Automatically shift the ILU factorization to ensure positive definiteness
-      PetscCallAbort(comm, PCFactorSetShiftType(*preconditioner, MAT_SHIFT_POSITIVE_DEFINITE));
-      break;
-    case PetscPCType::ICC:
-      preconditioner = std::make_unique<mfem_ext::PetscPCSolver>(comm, PCICC);
       // Automatically shift the ILU factorization to ensure positive definiteness
       PetscCallAbort(comm, PCFactorSetShiftType(*preconditioner, MAT_SHIFT_POSITIVE_DEFINITE));
       break;
@@ -1035,10 +1029,6 @@ std::unique_ptr<mfem_ext::PetscPCSolver> buildPetscPreconditioner(PetscPCType pc
       preconditioner = std::make_unique<mfem_ext::PetscGAMGSolver>(comm);
       // Automatically shift the LU factorization to ensure positive definiteness
       PetscOptionsInsertString(nullptr, "-mg_coarse_sub_pc_factor_shift_type positive_definite");
-#ifdef PETSC_HAVE_SUPERLU_DIST
-      // Use the SuperLU_dist solvers, if available
-      PetscOptionsInsertString(nullptr, "-mg_coarse_sub_pc_factor_mat_solver_type superlu_dist");
-#endif
       break;
     }
     case PetscPCType::NONE:
@@ -1062,7 +1052,6 @@ PetscPCType stringToPetscPCType(const std::string& type_str)
       {"lu", PetscPCType::LU},
       {"ilu", PetscPCType::ILU},
       {"cholesky", PetscPCType::CHOLESKY},
-      {"icc", PetscPCType::ICC},
       {"svd", PetscPCType::SVD},
       {"asm", PetscPCType::ASM},
       {"gasm", PetscPCType::GASM},
--- a/src/serac/numerics/equation_solver.cpp
+++ b/src/serac/numerics/equation_solver.cpp
@@ -843,7 +843,7 @@ std::unique_ptr<mfem::NewtonSolver> buildNonlinearSolver(const NonlinearSolverOp
 std::pair<std::unique_ptr<mfem::Solver>, std::unique_ptr<mfem::Solver>> buildLinearSolverAndPreconditioner(
     LinearSolverOptions linear_opts, MPI_Comm comm)
 {
-  auto preconditioner = buildPreconditioner(linear_opts.preconditioner, linear_opts.preconditioner_print_level, comm);
+  auto preconditioner = buildPreconditioner(linear_opts, comm);
 
   if (linear_opts.linear_solver == LinearSolver::SuperLU) {
     auto lin_solver = std::make_unique<SuperLUSolver>(linear_opts.print_level, comm);
@@ -868,6 +868,19 @@ std::pair<std::unique_ptr<mfem::Solver>, std::unique_ptr<mfem::Solver>> buildLin
     case LinearSolver::GMRES:
       iter_lin_solver = std::make_unique<mfem::GMRESSolver>(comm);
       break;
+#ifdef MFEM_USE_PETSC
+    case LinearSolver::PetscCG:
+      iter_lin_solver = std::make_unique<serac::mfem_ext::PetscKSPSolver>(comm, KSPCG, std::string());
+      break;
+    case LinearSolver::PetscGMRES:
+      iter_lin_solver = std::make_unique<serac::mfem_ext::PetscKSPSolver>(comm, KSPGMRES, std::string());
+      break;
+#else
+    case LinearSolver::PetscCG:
+    case LinearSolver::PetscGMRES:
+      SLIC_ERROR_ROOT("PETSc linear solver requested for non-PETSc build.");
+      exitGracefully(true);
+#endif
     default:
       SLIC_ERROR_ROOT("Linear solver type not recognized.");
       exitGracefully(true);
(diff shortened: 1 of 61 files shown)


src/serac/numerics/equation_solver.hpp
```diff
@@ -17,9 +17,11 @@
 #include <variant>
 
 #include "mfem.hpp"
+#include "mfem/linalg/petsc.hpp"
```
chapman39: I think once https://github.com/LLNL/serac/pull/1138 merges, this line will no longer be needed?

chapman39: And where-ever else this file is included.

zatkins-work: actually, I think that they can already be removed -- I didn't realize that `mfem.hpp` already included `mfem/linalg/petsc.hpp`.

tupek2 (approved):

src/serac/numerics/equation_solver.cpp
```diff
@@ -868,6 +878,20 @@ std::pair<std::unique_ptr<mfem::Solver>, std::unique_ptr<mfem::Solver>> buildLin
...
       break;
+#ifdef SERAC_USE_PETSC
+    case LinearSolver::PetscCG:
+      iter_lin_solver = std::make_unique<serac::mfem_ext::PetscKSPSolver>(comm, KSPCG, std::string());
+      break;
+    case LinearSolver::PetscGMRES:
+      iter_lin_solver = std::make_unique<serac::mfem_ext::PetscKSPSolver>(comm, KSPGMRES, std::string());
+      break;
+#else
+    case LinearSolver::PetscCG:
```
white238, line 889: 👍 

Thanks for the helpful error message!

src/serac/numerics/equation_solver.cpp
```diff
@@ -868,6 +878,20 @@ std::pair<std::unique_ptr<mfem::Solver>, std::unique_ptr<mfem::Solver>> buildLin
     case LinearSolver::GMRES:
       iter_lin_solver = std::make_unique<mfem::GMRESSolver>(comm);
       break;
+#ifdef SERAC_USE_PETSC
+    case LinearSolver::PetscCG:
```
white238, line 882: Should this be `LinearSolver::PETScCG`?

src/serac/numerics/equation_solver.cpp
```diff
@@ -1003,8 +1034,9 @@ void EquationSolver::defineInputFileSchema(axom::inlet::Container& container)
   iterative_container.addInt("max_iter", "Maximum iterations for the linear solve.").defaultValue(5000);
   iterative_container.addInt("print_level", "Linear print level.").defaultValue(0);
   iterative_container.addString("solver_type", "Solver type (gmres|minres|cg).").defaultValue("gmres");
-  iterative_container.addString("prec_type", "Preconditioner type (JacobiSmoother|L1JacobiSmoother|AMG|ILU).")
+  iterative_container.addString("prec_type", "Preconditioner type (JacobiSmoother|L1JacobiSmoother|AMG|ILU|Petsc).")
```
white238, line 1037:
```suggestion
  iterative_container.addString("prec_type", "Preconditioner type (JacobiSmoother|L1JacobiSmoother|AMG|ILU|PETSc).")
```

src/serac/numerics/equation_solver.cpp
```diff
@@ -1064,6 +1096,12 @@ serac::LinearSolverOptions FromInlet<serac::LinearSolverOptions>::operator()(con
 #endif
   } else if (prec_type == "GaussSeidel") {
     options.preconditioner = serac::Preconditioner::HypreGaussSeidel;
+#ifdef SERAC_USE_PETSC
+  } else if (prec_type == "Petsc") {
```
white238, line 1100: Can you review the capitalization of PETSc? We should be consistent.

src/serac/numerics/petsc_solvers.cpp
```diff
@@ -0,0 +1,1031 @@
...
+static PetscErrorCode solverWrapperView(PC pc, PetscViewer viewer)
+{
+  SolverWrapperCtx* ctx;
+
+  PetscFunctionBeginUser;
+  auto* void_ctx = static_cast<void*>(&ctx);
+  PetscCall(PCShellGetContext(pc, &void_ctx));
+  ctx = static_cast<SolverWrapperCtx*>(void_ctx);
+  if (ctx->solver) {
+    mfem::PetscPreconditioner* ppc = dynamic_cast<mfem::PetscPreconditioner*>(ctx->solver);
```
white238, line 46: Stream of conciousness..... so MFEM does `Petsc` and it looks like below this PETSc itself does `Petsc` as well in its function calls???

src/serac/numerics/petsc_solvers.cpp
```diff
@@ -0,0 +1,1031 @@
...
+  mfem::PetscPreconditioner::SetOperator(pA11);
+  Mat A22;
+  Vec zero;
+  PetscCallAbort(GetComm(), MatNestGetSubMat(*pA, 1, 1, &A22));
+  // Make sure all diagonal elements are set
+  PetscCallAbort(GetComm(), MatSetOption(A22, MAT_NEW_NONZERO_ALLOCATION_ERR, PETSC_FALSE));
+  PetscCallAbort(GetComm(), MatCreateVecs(A22, &zero, NULL));
+  PetscCallAbort(GetComm(), VecSet(zero, 0));
+  PetscCallAbort(GetComm(), MatDiagonalSet(A22, zero, ADD_VALUES));
+  if (delete_pA) delete pA;
```
white238: I would prefer not to have single statement if's w/o brackets. It helps prevent common errors that can occur when adding additional lines to a control statement.  With the new clang-format version we recent moved to we can fix these with style...

src/serac/numerics/petsc_solvers.cpp
```diff
@@ -0,0 +1,1031 @@
...
+  // This is faster, but will fail if the step leads to a nan or inf value
+  while (num_failures++ < max_failures) {
+    PetscCall(VecWAXPY(W, -lambda_orig, Y, X));
+    if (linesearch->ops->viproject) PetscCall((*linesearch->ops->viproject)(snes, W));
+    PetscCall((*linesearch->ops->snesfunc)(snes, W, Ftemp));
+    PetscCall(VecDot(Ftemp, Y, &fty));
+    if (!PetscIsInfOrNanScalar(fty)) {
+      if (monitor) {
+        PetscCall(PetscViewerASCIIAddTab(monitor, linesearch_obj->tablevel));
+        auto msg = axom::fmt::format("    Line search: dot(F,Y) = {}, no back-off steps needed", fty, lambda);
```
white238: This has a mismatch in formatting parameters.

white238:
Just to be clear after making several comments. I only care about consistency in the capitalization of PETSc. I would be fine to have one that matches MFEM's use and then have comments and strings be the correct one, or swapping over to the correct one entirely. If anyone else has strong opinions please say so now.

src/serac/numerics/petsc_solvers.hpp
```diff
@@ -0,0 +1,516 @@
...
+  }
+
+  /**
+   * @brief Construct a PetscPreconditionerSpaceDependent from a generic operator
+   * @param comm The MPI communicator used by the vectors and matrices in the solve
+   * @param op mfem::Operator used to construct the preconditioner
+   * @param pc_type The PETSc PCType of the preconditioner
+   * @param prefix Prefix for all PETSc command line options options
+   *
+   * @note Additional arguments for the preconditioner can be set via the command line, e.g. `-pc_jacobi_type rowsum`.
```
white238, line 141: Was this tested with the driver? I am concerned that the CLI11 command line parsing might mess this up.

src/serac/numerics/solver_config.hpp
```diff
@@ -100,31 +100,64 @@ struct TimesteppingOptions {
...
+  Strumpack, /**< Strumpack MPI-enabled direct frontal solver*/
+  PetscCG,   /**< PETSc MPI-enabled conjugate gradient solver */
+  PetscGMRES /**< PETSc MPI-enabled generalize minimal residual solver */
 };
 // _linear_solvers_end
 
+/// Convert linear solver enums to their string names
+inline std::string linearName(const LinearSolver& s)
+{
+  if (s == LinearSolver::CG) return "CG";
```
white238: This should be a case statement.

src/serac/numerics/solver_config.hpp
```diff
@@ -100,31 +100,64 @@ struct TimesteppingOptions {
...
+  PetscNewtonBacktracking,   /**< PETSc Newton with backtracking line search */
+  PetscNewtonCriticalPoint,  /**< PETSc Newton with critical point line search */
+  PetscTrustRegion           /**< PETSc trust region solver */
 };
 // _nonlinear_solvers_end
 
+/// Convert nonlinear linear solver enums to their string names
+inline std::string nonlinearName(const NonlinearSolver& s)
+{
+  if (s == NonlinearSolver::Newton) return "Newton";
```
white238: Case statement.

Should this error or is it caught by the calling function? Is empty string enought? Should it be something like "UNKNOWN"?

src/serac/numerics/tests/equationsolver_petsc.cpp
```diff
@@ -0,0 +1,171 @@
...
+                                              .print_level    = 1};
+
+  EquationSolver eq_solver(nonlin_opts, lin_opts);
+
+  eq_solver.setOperator(residual_opr);
+
+  eq_solver.solve(x_computed);
+
+  for (int i = 0; i < x_computed.Size(); ++i) {
+    EXPECT_LT(std::abs((x_computed(i) - x_exact(i))) / x_exact(i), 1.0e-6);
```
white238, line 117: Is there anyway for `x_computed` and `x_exact` to be different sizes at this point?

src/serac/numerics/tests/equationsolver_petsc.cpp
```diff
@@ -0,0 +1,171 @@
...
+    AllEquationSolverTests, EquationSolverSuite,
+    testing::Combine(
+        testing::Values(NonlinearSolver::Newton, NonlinearSolver::PetscNewton, NonlinearSolver::PetscNewtonBacktracking,
+                        NonlinearSolver::PetscNewtonCriticalPoint, NonlinearSolver::PetscTrustRegion),
+        testing::Values(LinearSolver::CG, LinearSolver::GMRES, LinearSolver::PetscCG, LinearSolver::PetscGMRES),
+        testing::Values(Preconditioner::Petsc),
+        testing::Values(PetscPCType::JACOBI, PetscPCType::JACOBI_L1, PetscPCType::JACOBI_ROWSUM,
+                        PetscPCType::JACOBI_ROWMAX, PetscPCType::PBJACOBI, PetscPCType::BJACOBI, PetscPCType::LU,
+                        PetscPCType::ILU, PetscPCType::CHOLESKY, PetscPCType::SVD, PetscPCType::ASM, PetscPCType::GASM,
+                        PetscPCType::GAMG)),  //, PetscPCType::HMG)));
```
white238, line 153: Can you add a comment why this is commented out?

src/serac/physics/tests/solid_shape.cpp
```diff
@@ -191,12 +198,10 @@ TEST(SolidMechanics, MoveShapeNonlinear) { shape_test(GeometricNonlinearities::O
 int main(int argc, char* argv[])
 {
   ::testing::InitGoogleTest(&argc, argv);
-  MPI_Init(&argc, &argv);
+  serac::initialize(argc, argv);
 
   axom::slic::SimpleLogger logger;
```
white238: After the change to `serac::initialize` you dont need the creation of the `SimpleLogger` since the serac function creates one.

zatkins-work:
> Just to be clear after making several comments. I only care about consistency in the capitalization of PETSc. I would be fine to have one that matches MFEM's use and then have comments and strings be the correct one, or swapping over to the correct one entirely. If anyone else has strong opinions please say so now.

PETSc typically uses the correct capitalization (PETSc) in documentation and comments, while using `Petsc` in function names for clarity.

src/serac/numerics/petsc_solvers.hpp
zatkins-work: Ah, we'll have to use `allow_extras()` on the CLI11 app. Luckily, none of the PETSc options are valid CLI11 flags (they only use one `-`), so there shouldn't ever be conflicts.

src/serac/numerics/equation_solver.cpp
zatkins-work: I vote no, since that would be different than the MFEM/PETSc style.

src/serac/numerics/solver_config.hpp
zatkins-work: We don't need the empty string case if we use a switch statement

src/serac/numerics/tests/equationsolver_petsc.cpp
zatkins-work: I don't think so, but I can add an assertion for that as well if you want

src/serac/numerics/tests/equationsolver_petsc.cpp
zatkins-work: Added

src/serac/numerics/petsc_solvers.cpp
zatkins-work: Sounds good, I think this should definitely be enforced by style

src/serac/numerics/solver_config.hpp
zatkins-work: Just kidding, GCC isn't smart enough to figure that out. I'm setting it to "UNKNOWN"

src/serac/numerics/equation_solver.cpp
zatkins-work: personally, I think that for input files/CLI interface, it's not great to force users to remember the proper capitalization for their arguments. "PETSc" in particular is a pretty weird one that I expect will cause issues/annoyance. I think that all lowercase or all uppercase (or using case-independent conversion) would be a better choice.

src/serac/numerics/petsc_solvers.cpp
```diff
@@ -0,0 +1,1047 @@
...
+  Mat               A;
+  PetscBool         is_hypre;
+  SolverWrapperCtx* ctx;
+
+  PetscFunctionBeginUser;
+  auto* void_ctx = static_cast<void*>(&ctx);
+  PetscCall(PCShellGetContext(pc, &void_ctx));
+  ctx = static_cast<SolverWrapperCtx*>(void_ctx);
+  mfem::PetscParVector xx(x, true);
+  mfem::PetscParVector yy(y, true);
```
samuelpmishLLNL, line 74: what's the meaning of the `true` in the second argument?

samuelpmishLLNL (approved):
This looks great, I look forward to seeing some timing studies to see which options perform well for different kinds of problems.

src/serac/numerics/petsc_solvers.cpp
zatkins-work: The `true` means to add `1` to the reference count of the underlying vector, i.e., the vector `x` should not be destroyed when `xx` goes out of scope. This is kind of opposite mfem's convention elsewhere, where this true would refer to ownership.

PR #1155, opened 2024-06-24
Tupek/newton profiling

tupek2:
Break out lambdas into functions so they can be caliper'd with CALI_CXX_MARK_FUNCTION;  profiling must be turned on in the configuration, and a call somewhere to profiling::initialize and profling::finalize must be added above in the call stack in order for the caliper data to be output.

### DIFF
--- a/src/serac/numerics/equation_solver.cpp
+++ b/src/serac/numerics/equation_solver.cpp
@@ -51,18 +51,21 @@ public:
     return normEval;
   }
 
+  /// assemble the jacobian
   void assembleJacobian(const mfem::Vector& x) const
   {
     CALI_CXX_MARK_FUNCTION;
     grad = &oper->GetGradient(x);
   }
 
+  /// set the preconditioner for the linear solver
   void setPreconditioner() const
   {
     CALI_CXX_MARK_FUNCTION;
     prec->SetOperator(*grad);
   }
 
+  /// solve the linear system
   void solveLinearSystem(const mfem::Vector& r, mfem::Vector& c) const
   {
     CALI_CXX_MARK_FUNCTION;
@@ -322,6 +325,7 @@ public:
   /// take a dogleg step in direction s, solution norm must be within trSize
   void dogleg_step(const mfem::Vector& cp, const mfem::Vector& newtonP, double trSize, mfem::Vector& s) const
   {
+    CALI_CXX_MARK_FUNCTION;
     // MRT, could optimize some of these eventually, compute on the outside and save
     double cc = Dot(cp, cp);
     double nn = Dot(newtonP, newtonP);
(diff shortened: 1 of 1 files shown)


src/serac/numerics/equation_solver.cpp
btalamini, line 1: Not part of this PR, but a question for later: is it reasonable to catch the exception base class here, or should we be only catching specific exceptions?

src/serac/numerics/equation_solver.cpp
```diff
@@ -89,10 +113,9 @@ class NewtonSolver : public mfem::NewtonSolver {
 
       real_t norm_nm1 = norm;
 
-      grad = &oper->GetGradient(x);
-      prec->SetOperator(*grad);
-
-      prec->Mult(r, c);  // c = [DF(x_i)]^{-1} [F(x_i)-b]
+      assembleJacobian(x);
```
btalamini, line 115: I like how this reads now!

src/serac/numerics/equation_solver.cpp
```diff
@@ -452,19 +507,20 @@ class TrustRegion : public mfem::NewtonSolver {
...
-        trPrecond.SetOperator(*K);
+        trPrecond.SetOperator(*grad);
         cumulativeCgIters = 0;
         if (print_options.iterations) {
           // currently it will always be updated
           // mfem::out << "Updating trust region preconditioner." << std::endl;
         }
       }
 
-      auto hess_vec_func = [=](const mfem::Vector& x_, mfem::Vector& v_) { K->Mult(x_, v_); };
```
btalamini, line 466: Ooh, were there significant deep copies here?

btalamini (approved):

src/serac/numerics/equation_solver.cpp
tupek2: My theory is any exception lower down likely implies we dont want to take that solution step.  Unless there is a case you can think of otherwise?

src/serac/numerics/equation_solver.cpp
tupek2: Yes, potentially.

src/serac/numerics/equation_solver.cpp
jamiebramwell: Agreed! This is much nicer!

jamiebramwell (approved):
This will be extremely useful. Thanks @tupek2 !

src/serac/numerics/equation_solver.cpp
btalamini: There could be an exception for non-numerical reasons, like the disk is full. I think we'd ideally like to catch only the ones that represent problems the solver can handle. But this is a story for another day.

src/serac/numerics/equation_solver.cpp
```diff
@@ -49,6 +52,27 @@ class NewtonSolver : public mfem::NewtonSolver {
     return normEval;
   }
 
+  /// assemble the jacobian
+  void assembleJacobian(const mfem::Vector& x) const
+  {
+    CALI_CXX_MARK_FUNCTION;
+    grad = &oper->GetGradient(x);
```
samuelpmishLLNL, line 59: While we're at it, could we instrument this GetGradient call with some caliper annotations as well? This is such an expensive operation, it would be beneficial to have more information about where that time is being spent.

samuelpmishLLNL (approved):

PR #1160, opened 2024-06-29
Feature/linear hardening

btalamini:
We didn't have linear hardening as a standalone option. You could get it as a special case of power law or Voce by choosing the parameters in a particular way, but it's easier for users to have the linear case explicitly available. 

This is a first step towards reducing code repetition. We have several J2 model implementations now, including one that *only* has linear hardening (but also has linear kinematic hardening). I want to have one implementation with options that are easily configurable.

There are a few other small items in this PR:
* I used the new linear hardening struct in tests that were using linear hardening behavior
* I improved the readability of an old test (behavior is unmodified)
* I switched one of the existing tests from power law to Voce, because this hardening option was not used in any test. Increases test coverage one iota.


src/serac/physics/materials/tests/nonlinear_J2_material.cpp
```diff
@@ -30,12 +30,20 @@ TEST(NonlinearJ2Material, PowerLawHardeningWorksWithDuals)
 
 TEST(NonlinearJ2Material, SatisfiesConsistency)
 {
+  // clang-format off
```
btalamini, line 33: The changes to this test are only for readability, the behavior is unmodified.

src/serac/physics/materials/tests/nonlinear_J2_material.cpp
```diff
@@ -172,10 +182,10 @@ TEST(FiniteDeformationNonlinearJ2Material, DerivativeCorrectness)
 
 TEST(FiniteDeformationNonlinearJ2Material, FrameIndifference)
 {
-  using Hardening = solid_mechanics::PowerLawHardening;
+  using Hardening = solid_mechanics::VoceHardening;
```
btalamini, line 185: This change is for test coverage, so that there is now one test using the Voce hardening rule.

tupek2 (approved):

btalamini:
@samuelpmishLLNL , would you mind taking a look at this?

samuelpmish (approved):

samuelpmishLLNL (approved):

PR #1165, opened 2024-07-03
Check if no-op cali macros are defined first

### LINKED ISSUE #1164: Caliper macros doubly defined
We now define a bunch of no-op Caliper macros when Serac is built without Caliper. It seems like LiDO does the same thing, and so when LiDO is built in codevelop mode, the compilation fails because the macros are doubly defined. Not sure if this is properly a Serac issue or a Caliper issue. I kindly request advice from @white238 or @chapman39 .

chapman39:
Haven't tested with lido, but it does pass tests when benchmarking is disabled.

Fixes #1164

### DIFF
--- a/src/serac/infrastructure/profiling.hpp
+++ b/src/serac/infrastructure/profiling.hpp
@@ -75,13 +75,27 @@
  */
 
 #ifndef SERAC_USE_CALIPER
+#ifndef CALI_CXX_MARK_FUNCTION
 #define CALI_CXX_MARK_FUNCTION
+#endif
+#ifndef CALI_CXX_MARK_LOOP_BEGIN
 #define CALI_CXX_MARK_LOOP_BEGIN(id, name)
+#endif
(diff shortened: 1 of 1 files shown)


btalamini (approved):
Thanks for fixing this. This fix is fine with me, but I can't comment on whether it's good software engineering. I'll leave that to Chris.

white238:
I spoke with @chapman39 about this yesterday. This is my fault. I shouldn't have advised him to remove our Serac specific macros. I forgot the original reason we did this. We should put our project specific ones back so that it does the following:

* Allows our downstream users to selectively turn off timings in Serac so it doesn't pollute their timings.
* Doesn't redefine macros that will conflict with others, namely Caliper.

This should not be merged and @chapman39 knows to just undo the important part of the last Caliper macro PR. Luckily it's just copy/pasting and search/replace. We should put a comment in `profiling.hpp` to explain why we map these defines so I don't forget in the future and repeat this cycle.

PR #1167, opened 2024-07-04
Re-add Serac Cali macros

### LINKED ISSUE #1164: Caliper macros doubly defined

chapman39:
This PR re-adds the serac caliper wrapper macros.

The motivation is handling the case where benchmarking is disabled. In this case, we defined no-op versions of caliper macros, so marked code can still compile. However, if a Serac user (e.g. LiDO) does the same thing, we have a double define of macros, which results in a compiler error.

TODO
- [x] Re-add macros
- [x] Add comment mentioning motivation in profiling.hpp

Fixes #1164

### DIFF
--- a/src/serac/infrastructure/profiling.hpp
+++ b/src/serac/infrastructure/profiling.hpp
@@ -74,6 +74,9 @@
  * Marks a particular scope for Caliper profiling
  */
 
+// NOTE: The motivation behind wrapping Caliper macros to avoid conflicting macro definitions in the no-op case, and
+// give downstream users the option to disable profiling Serac if it pollutes their timings.
+
 #ifdef SERAC_USE_CALIPER
 #define SERAC_MARK_FUNCTION CALI_CXX_MARK_FUNCTION
 #define SERAC_MARK_LOOP_BEGIN(id, name) CALI_CXX_MARK_LOOP_BEGIN(id, name)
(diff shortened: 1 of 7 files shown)


tupek2 (approved):

btalamini (approved):

src/serac/infrastructure/profiling.hpp
```diff
@@ -40,48 +40,60 @@
...
- * No-op macro in case Serac is not built with Caliper
+ * @def SERAC_MARK_END(id)
+ * Marks the end of a region Caliper profiling
  */
 
 /**
- * @def CALI_CXX_MARK_SCOPE(name)
- * No-op macro in case Serac is not built with Caliper
+ * @def SERAC_MARK_SCOPE(name)
+ * Marks a particular scope for Caliper profiling
```
white238: Can you edit these to say "No-op macro when ENABLE_PROFILING is off"?

white238 (approved):

PR #1168, opened 2024-07-05
Refactor/unify small strain plasticity

btalamini:
There were 2 flavors of small-strain J2 models. This combines them into one.

When I first created the nonlinear hardening models, I didn't immediately fold them into the existing J2 material model because it had a feature (kinematic hardening) that I didn't want to deal with while making the other generalizations to the return map algorithm. This PR implements kinematic hardening in the general J2 model, making the linear-only J2 model implementation redundant.

In more detail, this PR will do the following:

* Add linear kinematic hardening to the small-strain J2 plasticity model, making the old implementation redundant.
* Remove the redundant small-strain J2 plasticity model
* Turn Sam's implementation of the J2 verification problem (from Brannon and Leelavanichkul, 2009) into an actual gtest case, using the unified J2 model
* Rename the single small-strain J2 model to `J2SmallStrain`
* Rename the finite-strain J2 model to `J2`

I'm open to discussion on the renames.


src/serac/physics/materials/tests/J2_material.cpp
```diff
@@ -95,12 +101,244 @@ int main()
...
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+
+  axom::slic::SimpleLogger logger;
+
+  int result = RUN_ALL_TESTS();
+
+  return result;
+}
```
white238:
```suggestion
}

```

tupek2 (approved):

white238 (approved):

PR #1173, opened 2024-07-11
Switch TOSS4 CI to ruby, remove eng bank

chapman39:
Quartz is retiring next week, so we need to switch to ruby.

Until all users have the `eng` bank on ruby, we cannot use it. As a result CI queue times may be longer.

### DIFF
--- a/.gitlab/build_quartz.yml
+++ b/.gitlab/build_quartz.yml
@@ -2,10 +2,12 @@
 # This is the shared configuration of jobs for quartz
 .on_quartz:
   variables:
-    SCHEDULER_PARAMETERS: "--res=ci --exclusive=user -A ${ALLOC_BANK} --deadline=now+1hour -N ${ALLOC_NODES} -t ${ALLOC_TIME}"
+    # TODO Re-add eng bank to scheduler parameters once all users have the bank on ruby
+    # -A ${ALLOC_BANK}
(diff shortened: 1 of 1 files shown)


.gitlab/build_quartz.yml
```diff
@@ -2,10 +2,12 @@
 # This is the shared configuration of jobs for quartz
 .on_quartz:
   variables:
-    SCHEDULER_PARAMETERS: "--res=ci --exclusive=user -A ${ALLOC_BANK} --deadline=now+1hour -N ${ALLOC_NODES} -t ${ALLOC_TIME}"
+    # TODO Re-add eng bank to scheduler parameters once all users have the bank on ruby
```
white238, line 5: Should probably rename this file.

white238 (approved):

tupek2 (approved):
Bummer

btalamini:
Which bank will the CI run on with this?

chapman39:
> Which bank will the CI run on with this?

@btalamini 

@white238 may have a better answer, but I believe it is whatever is the default for the user running the CI job. My guess is `cbronze`.

.gitlab/build_quartz.yml
chapman39: agreed. ill do that in another pr

white238:
It uses the persons default bank. Most people have one but I have seen cases where people weren't assigned one when given access to the machine.  The CI reservation does help with how quickly jobs start so it is not too bad actually.

PR #1177, opened 2024-07-15
Fix issues with warm start. 

tupek2:
It was using current time jacobians in a few places and linearizing about the new parameters.  This broke Jorge-Luis' lqe problems as the converged state looked to be unstable due the using the wrong Jacobian in the warm start.  The linear solver would then essentially have to choice but to fail.

### DIFF
--- a/src/serac/physics/solid_mechanics.hpp
+++ b/src/serac/physics/solid_mechanics.hpp
@@ -1627,24 +1627,24 @@ protected:
                             displacement_, acceleration_, *parameters_[parameter_indices].state...);
       }...};
 
-
-  /// @brief Array functions computing the derivative of the residual with respect to each given parameter evaluated at the previous value of state
+  /// @brief Array functions computing the derivative of the residual with respect to each given parameter evaluated at
+  /// the previous value of state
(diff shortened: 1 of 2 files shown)


src/serac/physics/solid_mechanics.hpp
```diff
@@ -1734,14 +1749,12 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
-      auto drdparam        = serac::get<DERIVATIVE>(d_residual_d_[parameter_index](ode_time_point_));
+      auto drdparam        = serac::get<DERIVATIVE>(d_residual_d_previous_[parameter_index](time_));
       auto residual_update = drdparam(parameter_difference);
 
       // Flip the sign to get the RHS of the Newton update system
       // J^-1 du = - residual
-      residual_update *= -1.0;
-
-      dr_ += residual_update;
+      dr_ -= residual_update;
```
white238, line 1757: 👍

white238 (approved):

btalamini (approved):

PR #1179, opened 2024-07-19
stress extrapolation

samuelpmishLLNL:


samuelpmishLLNL:
Finally revisiting this PR.

Here's an example of how to fit a field to the J2 stress resulting from a displacement field (in 2D), the user writes:

```cpp
FiniteElementState sigma_J2 = fit< dim, output_space(input_space) >([&](double /*t*/, [[maybe_unused]] auto position, [[maybe_unused]] auto displacement_){ 
    auto du_dx = get_value(get<1>(displacement_));
    auto stress = mat(du_dx);
    return tuple{I2(dev(stress)), zero{}}; // compute the 2nd invariant of the deviatoric part of the stress
  }, pmesh, u);
```
where `input_space` is the function space for the displacement, and `output_space` is the function space of the fitted approximation, and `u` is the displacement field to interpolate for the supplied q-function.

----------

Here are examples of the J2 stress on a fracture specimen in plane strain, being approximated by both a piecewise constant field (`L2<0>`) and a continuous, piecewise linear field (`H1<1>`):
notched_plate

And here are some closeups of the `L2<0>` and `H1<1>` approximations of the J2 stress near the crack tip:
notched_plate_crack_tip

samuelpmishLLNL:
I am seeing a numerical inconsistency in parallel:

Screenshot 2024-08-05 at 8 58 47 PM

I'll investigate tomorrow

---------

edit: this has been fixed

src/serac/physics/fit.hpp
```diff
@@ -0,0 +1,65 @@
...
+  }, mesh);
+  auto M = get<1>(phi_phi(DifferentiateWRT<0>{}, 0.0 /* t */, fitted_field));
+
+  // rhs
+  std::array< const mfem::ParFiniteElementSpace *, sizeof ... (T) > trial_spaces = {&solution_fields.space() ...};
+  serac::Functional< signature > phi_f(&fitted_field.space(), trial_spaces);
+  phi_f.AddDomainIntegral(Dimension<dim>{}, DependsOn<i...>{}, f, mesh);
+  mfem::Vector b = phi_f(0.0, solution_fields ...);
+
+  mfem::CGSolver cg(MPI_COMM_WORLD);
```
samuelpmishLLNL, line 45:
```
mfem::CGSolver cg(MPI_COMM_WORLD);
```

This was the offending line of code that caused the incorrect results in parallel. Originally, I created the `CGSolver` without the parallel communicator which made each processor solve its own problem in isolation, rather than together.

samuelpmishLLNL:
I believe the parallel issue is fixed now, and this is ready for review.

kswartz92:
Is the L2 field equivalent to the volume averaging trick I've been using for a while in LiDO? That is, integrating the stress in each element and dividing by element volume

samuelpmishLLNL:
> Is the L2 field equivalent to the volume averaging trick I've been using for a while in LiDO? That is, integrating the stress in each element and dividing by element volume

Yes, the volume averaging trick is equivalent to specifying `L2<0>` for the approximation space. However, now you can try higher order fits too, if you like (it might also work for fitting vector-valued quantities as well, I can't remember). In theory, the same feature also supports fitting functions to `Hcurl` spaces as well, although I don't have a test for that.

kswartz92:
> > Is the L2 field equivalent to the volume averaging trick I've been using for a while in LiDO? That is, integrating the stress in each element and dividing by element volume
> 
> Yes, the volume averaging trick is equivalent to specifying `L2<0>` for the approximation space. However, now you can try higher order fits too, if you like (it might also work for fitting vector-valued quantities as well, I can't remember). In theory, the same feature also supports fitting functions to `Hcurl` spaces as well, although I don't have a test for that.

Awesome, once this propagates to LiDO I'll update our serac visualization operator to use it so we can plot smoother fields. Thanks @samuelpmish!

tupek2 (approved):

src/serac/physics/tests/fit_test.cpp
```diff
@@ -0,0 +1,124 @@
...
+  solid_solver.outputStateToDisk("paraview" + std::to_string(n));
+
+  double dt = 1.0;
+  solid_solver.advanceTimestep(dt);
+
+  auto u = solid_solver.displacement();
+
+  Empty internal_variables{};
+
+  sigma_J2 = fit< dim, output_space(input_space) >([&](double /*t*/, [[maybe_unused]] auto position, [[maybe_unused]] auto displacement_){
```
jamiebramwell: Should this be added as a default output to the solid mechanics solver or should we leave it up to users to define their favorite stress measure of choice?

jamiebramwell (approved):
This is a great feature! Thanks @samuelpmishLLNL !

src/serac/physics/tests/fit_test.cpp
samuelpmishLLNL: My vote is to limit the scope of this PR to just adding the feature and discuss potential changes to the physics modules later on.

PR #1180, opened 2024-07-20
Re-add eng bank, switch toss4 CI machine to dane

chapman39:
My commit says `ruby` accidentally, since I'm so used to writing "switch to ruby", I guess. But this PR will move CI to test on `dane` with the `eng` bank.

Does everyone have dane+eng access? Or should we stick with ruby?


white238 (approved):
I dunno but we will see!!!

tupek2 (approved):

PR #1183, opened 2024-07-20
Tupek/test quasi static checkpoint

tupek2:
The checkpointing was not being done for quasi-static problems.  Now, it is written a bit more generically with a bit more responsibility on base_physics and loops over states so that it will work for any physics and dynamic or quasi-static.  In addition, we were not decrementing the time correctly, so any time dependent load was not being accounted for in the reverse pass correctly.  This tries to clear up some of the logic and naming, and fix these know failing situations.  Tests were added to check time varying loads and quasi-static evolution.

Additional things in this: remove ode_time_point_, as it was very confusingly intertwined with  time_.  If there is a case to keep both, we can undo this.  Also, the checkpoint to disk capability is a bit subtle.  Now it is always writing to disk.  If the user externally also writes state to disk, it will mess up the checkpointing.  I think we can considered deleting the writing to disk checkpointing altogether at some point.  For now, it is fairly hard to use.

### DIFF
--- a/src/serac/physics/base_physics.cpp
+++ b/src/serac/physics/base_physics.cpp
@@ -356,26 +356,26 @@ FiniteElementState BasePhysics::loadCheckpointedState(const std::string& state_n
   return checkpoint_states_.at(state_name)[static_cast<size_t>(cycle)];
 }
 
-  std::unordered_map<std::string, FiniteElementState> BasePhysics::getCheckpointedStates(int cycle_to_load) const
-  {
-    std::unordered_map<std::string, FiniteElementState> previous_states_map;
-    std::vector<FiniteElementState*> previous_states_ptrs;
+std::unordered_map<std::string, FiniteElementState> BasePhysics::getCheckpointedStates(int cycle_to_load) const
+{
+  std::unordered_map<std::string, FiniteElementState> previous_states_map;
+  std::vector<FiniteElementState*>                    previous_states_ptrs;
 
-    if (checkpoint_to_disk_) {
-      for (const auto& state_name : stateNames()) {
-        previous_states_map.emplace(state_name, state(state_name));
-        previous_states_ptrs.emplace_back(const_cast<FiniteElementState*>(&state(state_name)));
-      }
-      StateManager::loadCheckpointedStates(cycle_to_load, previous_states_ptrs);
-      return previous_states_map;
-    } else {
-      for (const auto& state_name : stateNames()) {
-        previous_states_map.emplace(state_name, checkpoint_states_.at(state_name)[static_cast<size_t>(cycle_to_load)]);
-      }
+  if (checkpoint_to_disk_) {
+    for (const auto& state_name : stateNames()) {
+      previous_states_map.emplace(state_name, state(state_name));
+      previous_states_ptrs.emplace_back(const_cast<FiniteElementState*>(&state(state_name)));
     }
-
+    StateManager::loadCheckpointedStates(cycle_to_load, previous_states_ptrs);
     return previous_states_map;
+  } else {
+    for (const auto& state_name : stateNames()) {
+      previous_states_map.emplace(state_name, checkpoint_states_.at(state_name)[static_cast<size_t>(cycle_to_load)]);
+    }
   }
+
+  return previous_states_map;
+}
 
 double BasePhysics::getCheckpointedTimestep(int cycle) const
 {
--- a/src/serac/physics/base_physics.cpp
+++ b/src/serac/physics/base_physics.cpp
@@ -57,7 +57,7 @@ double BasePhysics::minTime() const { return min_time_; }
 
 int BasePhysics::minCycle() const { return min_cycle_; }
 
-std::vector<double> BasePhysics::timesteps() const { return timesteps_; }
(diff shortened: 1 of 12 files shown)


src/serac/physics/heat_transfer.hpp
```diff
@@ -346,7 +343,8 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
     } else {
       // Step the time integrator
       // Note that the ODE solver handles the essential boundary condition application itself
-      ode_.Step(temperature_, time_, dt);
+      double time_tmp = time_;
+      ode_.Step(temperature_, time_tmp, dt);
```
kswartz92, line 355: why do you make a copy of `time_` before passing to `ode_`?

kswartz92:
are you planning to add the test with a `Functional` qoi to this PR?

src/serac/physics/heat_transfer.hpp
tupek2: time_ is already passed to the constructor of the time integrator.  It is already incremented there.  This value would also be incremented.  I want to make sure the incrementing wasnt happening twice.

tupek2:
> are you planning to add the test with a `Functional` qoi to this PR?

added

kswartz92:
Update- Irvin pulled the latest changes from this branch into his lido build and his problem has correct sensitivities and a working inverse problem with time dependent b.c.s and time dependent QoI. Thanks for all the help @tupek2!

src/serac/physics/solid_mechanics.hpp
```diff
@@ -1352,9 +1354,21 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     mfem::HypreParVector adjoint_essential(displacement_adjoint_load_);
     adjoint_essential = 0.0;
 
+    SLIC_ERROR_ROOT_IF(cycle_ <= min_cycle_,
+                       "Maximum number of adjoint timesteps exceeded! The number of adjoint timesteps must equal the "
+                       "number of forward timesteps");
```
kswartz92, line 1365: it might be helpful here to mention to users that this message can be caused by two backwards passes attempted without a forward pass or a clearAdjointStates() in between

samuelpmishLLNL (approved):
Sorry for the delay reviewing this, I was hoping to grok it more thoroughly but it sounds like it's holding other things up.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -137,10 +137,11 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
   SolidMechanics(const NonlinearSolverOptions nonlinear_opts, const LinearSolverOptions lin_opts,
                  const serac::TimesteppingOptions timestepping_opts, const GeometricNonlinearities geom_nonlin,
                  const std::string& physics_name, std::string mesh_tag, std::vector<std::string> parameter_names = {},
-                 int cycle = 0, double time = 0.0, bool checkpoint_to_disk = false)
+                 int cycle = 0, double time = 0.0, bool checkpoint_to_disk = false, bool use_warm_start = true)
```
white238, line 142: This needs to be documented... its documented below with this:

```
   * @param use_warm_start A flag to turn on or off the displacement warm start predictor which helps robustness for
   * large deformation problems
```
I would submit a code suggestion but its too far away from a changed line =/

src/serac/physics/base_physics.cpp
```diff
@@ -57,7 +57,7 @@ double BasePhysics::minTime() const { return min_time_; }
 
 int BasePhysics::minCycle() const { return min_cycle_; }
 
-std::vector<double> BasePhysics::timesteps() const { return timesteps_; }
+const std::vector<double>& BasePhysics::timesteps() const { return timesteps_; }
```
white238, line 60: 👍

src/serac/physics/base_physics.cpp
```diff
@@ -356,21 +356,34 @@ FiniteElementState BasePhysics::loadCheckpointedState(const std::string& state_n
...
+  }
+
+  return previous_states_map;
 }
 
 double BasePhysics::getCheckpointedTimestep(int cycle) const
 {
   SLIC_ERROR_ROOT_IF(cycle < 0, axom::fmt::format("Negative cycle number requested for physics module {}.", name_));
-  SLIC_ERROR_ROOT_IF(cycle > max_cycle_,
+  SLIC_ERROR_ROOT_IF(cycle > static_cast<int>(timesteps_.size()),
```
white238, line 383: Thanks for making this error check consistent with the message but should we have a check that we don't go over `max_cycle_`?

src/serac/physics/heat_transfer.hpp
white238: Can you put a comment of why you are doing this here and below?

tupek2: I added words.

src/serac/physics/solid_mechanics.hpp
tupek2: Done!

white238 (approved):

PR #1187, opened 2024-07-30
WIP: Add small comments in Mac TPL build documentation

chapman39:
I'm going through the Mac build documentation for my machine. So far I just added one comment.

I was getting thrown off by the documentation mentioning `/opt/homebrew` for the installation prefix, when it's `/usr/local` for Intel-based Macs. I'm new to using `brew`, so maybe I'm just a noob, but it did take me a bit to figure out.

### DIFF
--- a/src/docs/sphinx/quickstart.rst
+++ b/src/docs/sphinx/quickstart.rst
@@ -290,7 +290,7 @@ If you plan to install the developer tools, you should also run
    $ brew install cppcheck doxygen llvm@14
    $ ln -s /opt/homebrew/opt/llvm@14/bin/clang-format /opt/homebrew/bin/clang-format
 
-If you have installed Homebrew using the default installation prefix, most packages will be accessible through the prefix ``/opt/homebrew``.
+If you have installed Homebrew using the default installation prefix, most packages will be accessible through the prefix ``/opt/homebrew`` (or ``/usr/local`` if you're using an Intel Mac)
 Some packages are not linked into this prefix to prevent conflicts with MacOS-provided versions.
 These will only be accessible via the prefix ``/opt/homebrew/opt/[package-name]``.
 Homebrew will warn about such packages after installing them.
--- a/src/docs/sphinx/quickstart.rst
+++ b/src/docs/sphinx/quickstart.rst
@@ -290,7 +290,7 @@ If you plan to install the developer tools, you should also run
    $ brew install cppcheck doxygen llvm@14
    $ ln -s /opt/homebrew/opt/llvm@14/bin/clang-format /opt/homebrew/bin/clang-format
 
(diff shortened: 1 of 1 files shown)


btalamini:
I think there is a command to query homebrew's prefix. Googling turns up `brew --prefix`. I don't have homebrew, so I can't check - can one of you brew users try? This could make the instructions more generic, as well as handle the case that a user installed homebrew in a non-default location

chapman39:
@btalamini I added a comment about `brew --prefix`, thanks.

Rather referring to the assumed prefix `/opt/homebrew`, do we want something generic like `${BREW_PREFIX}` in the documentation? I don't have a strong preference either way.

chapman39:
I'm very confused why this is happening when I have `+devtools`. Has `+devtools` been tested on macos?

```
[exe: /Users/chapman39/dev/serac/serac_libs1/spack/bin/spack -D /Users/chapman39/dev/serac/serac_libs1/spack_env concretize --fresh ]
==> Error: concretization failed for the following reasons:

   1. Cannot select a single "version" for package "py-numpy"
[spack version: 0.22.0.dev0 (6d244b3f6776b3697861c6b5cebbee376c30bff3)
]
```

Only when I remove the python section from my spack.yaml, I'm able to concretize with `+devtools`.

tupek2 (approved):

chapman39:
Looks like boundary condition `boundary_cond` test fails. I can look into this a bit more but wanted to report.

```
***********************************
[WARNING in line 67 of file /Users/chapman39/dev/serac/repo/src/serac/infrastructure/../../serac/physics/boundary_conditions/boundary_condition.hpp]
MESSAGE=Attempting to compare tags of two different enum types (always false)
***********************************

...

[  FAILED  ] 1 test, listed below:
[  FAILED  ] BoundaryCond.DirectTrueDofs
```

src/docs/sphinx/quickstart.rst
```diff
@@ -291,6 +291,8 @@ If you plan to install the developer tools, you should also run
    $ ln -s /opt/homebrew/opt/llvm@14/bin/clang-format /opt/homebrew/bin/clang-format
 
 If you have installed Homebrew using the default installation prefix, most packages will be accessible through the prefix ``/opt/homebrew``.
+Note for Intel-based Macs, the installation prefix is ``/usr/local``. If you set a custom prefix or aren't sure what the prefix is, run ``brew --prefix``.
```
btalamini, line 296: @chapman39, since you asked, my preference would be to replace `/opt/homebrew` everywhere in this section with `$(brew --prefix)`  since that will always do the right thing. But we can do that later. Let's just get this in.

btalamini (approved):

white238 (approved):

PR #1188, opened 2024-07-30
Small updates to ubuntu build instructions based on feedback from col…

tupek2:
…laborators.

### DIFF
--- a/scripts/spack/configs/linux_ubuntu_20/spack.yaml
+++ b/scripts/spack/configs/linux_ubuntu_20/spack.yaml
@@ -1,7 +1,7 @@
 spack:
   compilers::
   - compiler:
-      spec: gcc@9.3.0
+      spec: gcc@9.4.0
       paths:
         cc: /usr/bin/gcc
         cxx: /usr/bin/g++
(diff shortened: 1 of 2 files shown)


scripts/spack/configs/linux_ubuntu_20/spack.yaml
```diff
@@ -116,7 +116,7 @@ spack:
       - spec: cmake@3.20.4
         prefix: /usr
     cppcheck:
-      version: [1.90]
+      version: [1.90']
```
white238:
```suggestion
      version: [1.90]
```

white238 (approved):

scripts/spack/configs/linux_ubuntu_20/spack.yaml
tupek2: apparently the 1.90 converts to 1.9 and gives an error for external users?  The ' is supposed to make it a string.  I have no real way to test this, so thought I'd put it up to see if anyone had thoughts.

white238: weird I've never seen a string there but there are examples in the Spack repo but it requires the whole string to be quoted like such:

```
version: ['1.90']
```

btalamini (approved):

PR #1198, opened 2024-08-05
Add cylinder example and contact force monitoring

zatkins-work:
Adds monitoring of contact force by default in `SolidMechanicsContact` and adds a good example of Serac's contact capabilities.

output


src/serac/physics/solid_mechanics_contact.hpp
```diff
@@ -240,14 +249,26 @@ class SolidMechanicsContact<order, dim, Parameters<parameter_space...>,
...
     contact_.setPressures(mfem::Vector(augmented_solution, displacement_.Size(), contact_.numPressureDofs()));
+    contact_.update(cycle_, ode_time_point_, dt);
+    forces_.SetVector(contact_.forces(), 0);
+  }
+
+  void warmStartDisplacement()
+  {
+    SolidMechanicsBase::warmStartDisplacement();
+    auto dt = 1.0;
+    contact_.update(cycle_, time_, dt);
```
tupek2: I think the contact warmStart needs to include the contact forces and stiffnessess.  But this can wait for another day.

tupek2 (approved):

src/serac/physics/solid_mechanics_contact.hpp
zatkins-work: Agreed -- this actually wasn't supposed to be here

zatkins-work: These commits were cherry-picked from a much messier working tree and this apparently slipped through

data/meshes/hollow-cylinder.mesh
ebchin, line 1: This mesh file is quite a bit bigger than the others.  This is probably fine as is for now, but we might be able to use the NURBS representation + refinement in MFEM to create the same mesh more compactly?  See this file for an example: https://github.com/LLNL/serac/blob/atkins12/contact-bug/data/meshes/ball-nurbs.mesh

ebchin (approved):
Nice work @zatkins-work !

PR #1203, opened 2024-08-08
Remove excess perfomance warnings in PETSc solver wrappers

zatkins-work:
Removes warnings about the cost of matrix conversions which appear to have no meaningful performance impact.


src/serac/numerics/petsc_solvers.cpp
```diff
@@ -598,7 +598,7 @@ PetscErrorCode convertKSPPreSolve(KSP ksp, [[maybe_unused]] Vec rhs, [[maybe_unu
   if (!solver->checked_for_convert_ || solver->needs_hypre_wrapping_) {
     PetscBool is_hypre;
     PetscCall(PetscObjectTypeCompare(reinterpret_cast<PetscObject>(A), MATHYPRE, &is_hypre));
-    SLIC_WARNING_ROOT_IF(
+    SLIC_DEBUG_ROOT_IF(
```
white238, line 601: Is this a user setting that we are changing? If so we should either find a spot to emit it once or error out and let them fix it.

zatkins-work: Not really, no. We're just converting the internal matrix format. I don't think there is a way for a user to really fix it, this would only happen because MFEM or PETSc decided to use MATHYPRE instead of MATAIJ. The conversion is cheap and doesn't affect the original (user-provided) matrix, so I don't think they need to know about it.

white238: If that is the case this is a good change then.

white238 (approved):

tupek2 (approved):

Issue #1207, opened 2024-08-09
`boundary_cond` test fails on macos


chapman39:
Looks like boundary condition `boundary_cond` test fails. I can look into this a bit more but wanted to report.

```
***********************************
[WARNING in line 67 of file /Users/chapman39/dev/serac/repo/src/serac/infrastructure/../../serac/physics/boundary_conditions/boundary_condition.hpp]
MESSAGE=Attempting to compare tags of two different enum types (always false)
***********************************

...

[  FAILED  ] 1 test, listed below:
[  FAILED  ] BoundaryCond.DirectTrueDofs
```

_Originally posted by @chapman39 in https://github.com/LLNL/serac/issues/1187#issuecomment-2259403916_

PR #1209, opened 2024-08-12
remove dead boundary condition code

samuelpmishLLNL:
a lot (if not most) of the code in the boundary condition class appears to be unused and/or untested (at least in serac), this is just a PR that removes some of that old code.


chapman39:
i think this will fix https://github.com/LLNL/serac/issues/1207

samuelpmish:
> i think this will fix https://github.com/LLNL/serac/issues/1207

Oops, maybe I misspoke in the meeting earlier, but this PR was originally motivated by looking at that issue, but doesn't fix it yet. The failing test is built on a part of the `BoundaryCondition` class that is actually used, so I can't remove that part yet. 

Although, it may be worth just deleting that failing test, since it doesn't appear to be testing the correctness of anything.

tupek2 (approved):

chapman39:
is this ready? i can approve.

chapman39 (approved):

PR #1210, opened 2024-08-12
Guard missed test and document how to style

### LINKED ISSUE #1208: save clang-format git patch as artifact on failure
Right now, if the CI job for checking style fails, it dumps text to the log that can be made into a patch. However, that text is interleaved with some other console output, and seemingly prefixed by timestamps, which means it takes manual effort to produce the actual patch file that can be `git apply`-ed. On my machine/browser, the Azure devops UI doesn't play nicely with text selections (e.g. vertically scrolling the log changes the selected text) which makes it annoying to successfully select the relevant patch info.

It would be more ergonomic if instead, on a style failure the patch information would be written to a file that gets uploaded as an artifact. That way, the user only has to click a button to download the file and we're sure that no clerical errors were made when manually creating the patch.

white238:
https://serac.readthedocs.io/en/task-white238-guard_missed_test/sphinx/dev_guide/style_guide.html#how-to-style-your-code

Fixes #1208

### DIFF
--- a/src/docs/sphinx/dev_guide/style_guide.rst
+++ b/src/docs/sphinx/dev_guide/style_guide.rst
@@ -35,9 +35,10 @@ is not required because the lifetime of ``B`` is entirely dependent on the lifet
 How to style your code
 ----------------------
 
-We have two methods of enabling styling your code via ClangFormat. The first is to use the `style` build
-target on a configuration with the correct version of ClangFormat enabled. Here is an example on how to do
-it on LC's Ruby machine:
+We have two methods of enabling styling your code via ClangFormat. 
+
+The first method is to use the `style` build target on a configuration with the correct version of
+ClangFormat enabled. Here is an example on how to do it on LC's Ruby machine:
 
 .. code-block:: bash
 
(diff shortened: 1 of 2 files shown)


tupek2 (approved):

white238:
The Spack team also adds a comment to the GitHub Pull Request Template to tell you all the automated things their SpackBot can do. We only have `/style` but we could put a comment there to remind people. @samuelpmishLLNL do you think that would be helpful?

samuelpmishLLNL:
> We only have /style but we could put a comment there to remind people. @samuelpmishLLNL do you think that would be helpful?

I think the note in the readthedocs suffices

src/docs/sphinx/dev_guide/style_guide.rst
```diff
@@ -32,6 +32,24 @@ to one of ``A``'s instance variables ``Foo f``, ``B`` should store a non-owning
...
+
+.. code-block:: bash
+
+    $ ./config-build.py host-configs/ruby-toss_4_x86_64_ib-clang@14.0.6.cmake
+    $ cd build-ruby-toss_4_x86_64_ib-clang@14.0.6-debug
+    $ make style
+
+The second method is to make a comment of ``/style`` on your open GitHub pull request. This will trigger
+a GitHub Action that will automatically style your code and commit it to your branch. You will need to
+`git pull` after it is finished to continue to work on that branch.
```
samuelpmishLLNL, line 51: Who is allowed to make use of this `/style` command?

samuelpmishLLNL (approved):

src/docs/sphinx/dev_guide/style_guide.rst
white238: Definitely anyone who has write access to our repo can trigger it. I would need to try what happens if someone outside our organization does it. They certainly cannot change the behavior or the GitHub Action.

PR #1212, opened 2024-08-21
Move creation of Quadrature Data to `StateManager`

white238:
In preparation to adding quadrature data to the Sidre Data Store, I moved the creation to the `StateManager` class.

I also added some "has" functions to improve readability...


tupek2 (commented):
I like the quadrature data living off the state manager (eventually maybe for output, checkpointing, restarting, etc.), but I like the syntax and idea that the physics module knows how to create a QuadratureData.  Could we have the user syntax be the same, but in the physics call through to the new StateManager::newQuadratureData() function?

tupek2 (approved):
Seems to make sense to me

samuelpmishLLNL (approved):

PR #1213, opened 2024-08-21
add mesh partitioning tool

samuelpmishLLNL:
This PR comes from a request by @sethwatts on behalf of the digital twins project. It adds a new command line program that reads in an mfem mesh, partitions it into some number of pieces, and writes out those pieces to separate files.


src/serac/mesh/CMakeLists.txt
```diff
@@ -22,6 +22,9 @@ blt_add_library(
     DEPENDS_ON  ${mesh_depends}
     )
 
+add_executable(partitioner partitioner.cpp)
+target_link_libraries(partitioner PUBLIC serac_mesh)
```
white238:
```suggestion
blt_add_executable(NAME       partitioner
                   SOURCES    partitioner.cpp
                   DEPENDS_ON serac_mesh)
```

white238:
Should we have a "tools" directory for things of this flavor? I think this is the first one.

white238:
wrong button sorry.

src/serac/mesh/CMakeLists.txt
samuelpmishLLNL: what's the difference?

samuelpmishLLNL:
I don't understand why, but one of the CI jobs is concluding that this code is unsafe because of a null pointer dereference(?):

```cpp
  std::vector< uint32_t > blocks(num_blocks + 1);
  blocks[0] = 0;
```

Any idea how to work around this? It seems like a false positive to me.

src/serac/mesh/CMakeLists.txt
white238: In this case, not much but it keeps it consistent with the rest of the code.

white238:
> I don't understand why, but one of the CI jobs is concluding that this code is unsafe because of a null pointer dereference(?):
> 
> ```c++
>   std::vector< uint32_t > blocks(num_blocks + 1);
>   blocks[0] = 0;
> ```
> 
> Any idea how to work around this? It seems like a false positive to me.

I agree that seems to be a false positive. Have you tried converting num_blocks to `std::size_t`? Possibly even adding an error check against `num_blocks` being less than 0 even though that cannot happen.

tupek2:
Where does this executable live?  How does one know where to get it.  I do thing something like an tool, or utilities, or apps directory for these kinds of things might make sense.

samuelpmish:
> Where does this executable live? How does one know where to get it.

I believe CMake puts executables in `CMAKE_RUNTIME_OUTPUT_DIRECTORY`, so it's currently written to `/path/to/serac/build_dir/bin`, but I think a user could override that to specify a different location.

> I do think something like an tool, or utilities, or apps directory for these kinds of things might make sense.

I agree, although I think it makes more sense to worry about organization if/when we have more than 1.

tupek2 (approved):

btalamini (approved):

PR #1215, opened 2024-08-21
Turn vdim ordering on by default.

tupek2:


white238 (approved):

chapman39 (approved):

chapman39:
one of the login nodes for dane is down which is probably why that clang job is failing.

PR #1217, opened 2024-08-22
Guard buckling example against no-petsc build

### LINKED ISSUE #1216: Guard buckling example
I found this error when building LiDO codeveloping Serac. LiDO does not enable PETSc by default (though perhaps we should), which caused this error.

```
/usr/workspace/meemee/lido-2.0/repo/smith/serac/examples/buckling/cylinder.cpp:46:52: error: no member named 'stringToPetscPCType' in namespace 'serac::mfem_ext'
  return std::to_string(static_cast<int>(mfem_ext::stringToPetscPCType(in)));
                                         ~~~~~~~~~~^
/usr/workspace/meemee/lido-2.0/repo/smith/serac/examples/buckling/cylinder.cpp:103:62: error: no member named 'stringToPetscPCType' in namespace 'serac::mfem_ext'
            return std::to_string(static_cast<int>(mfem_ext::stringToPetscPCType(in)));
                                                   ~~~~~~~~~~^
```

The buckling example should also be guarded with PETSC_FOUND. https://github.com/LLNL/serac/blob/develop/examples/buckling/CMakeLists.txt#L7

```cmake
if(TRIBOL_FOUND AND PETSC_FOUND)
    blt_add_executable( NAME        buckling_cylinder
                        SOURCES     cylinder.cpp
                        OUTPUT_DIR ${EXAMPLE_OUTPUT_DIRECTORY}
                        DEPENDS_ON  serac_physics serac_mesh
                        )
endif()
```

To prevent things like this, it may be beneficial to add a test that disables PETSc, similar to what we do with Tribol and Sundials. And to minimize the number of jobs, maybe we could also mash the disable-X tests together? (i.e. test no petsc, sundials, tribol all together?)

chapman39:
Checks if petsc is found for the new-ish buckling example.

Fixes #1216


tupek2 (commented):
Is this going to add an additional build to our tests?

chapman39:
> Is this going to add an additional build to our tests?

Yes. The other option would be to combine all the disable-X (tribol, sundials, petsc) into one job, but that wouldn't be as robust.

tupek2:
> > Is this going to add an additional build to our tests?
> 
> Yes. The other option would be to combine all the disable-X (tribol, sundials, petsc) into one job, but that wouldn't be as robust.

tupek2:
I would vote to try and reduce our testing overhead.  I think its becoming way too much of a bottleneck.  Combining them would be OK with me, its not clear what scenario that would be missing.

chapman39:
im going to remove the extra job from this PR for now and create a separate issue. since in any case we should resolve the buckling example issue.

tupek2 (approved):

white238 (approved):

PR #1223, opened 2024-08-26
add support for interior face integrals

samuelpmishLLNL:
This PR aims to implement a feature requested by @tupek2 to be able to integrate over interior faces. The main challenge is in handling DG/L2 spaces for this type of integral, as they have separate values on each side.

----------------

The the new kind of integral behaves similarly to the existing boundary integral, except that for L2 spaces it represents:

Screenshot 2024-11-15 at 12 01 29 PM

where the domain of integration $\Omega$ represents the specified collection of interior faces, $\phi_1, \phi_2$ are the basis functions from side 1 and side 2, and $u_1, u_2$ are the values of the DG field evaluated from side 1 and side 2. For example, consider a mesh made of two adjacent quadrilateral elements, with a `L2<1, 1>` field defined on it. The red disks represent the nodes of the DG field (with the positions of the nodes shifted inward slightly to exaggerate which side of the interior face they belong to), and the blue edge represents the domain $\Omega$.

dg_example_diagram

In this example, the basis functions that appear in the integrand are:

Screenshot 2024-11-15 at 3 57 10 PM

where $N_0, N_1$ are the usual linear interpolation shape functions in 1D.

--------

Here is an example of a residual calculation involving a vector-valued DG space and interior faces:

```cpp

  ...

  constexpr int p = 2; // polynomial order
  using test_space  = L2<p, dim>;
  using trial_space = L2<p, dim>;

  constexpr int DERIVATIVE = 1;

  Domain interior_faces = InteriorFaces(*mesh);

  // Construct the new functional object using the specified test and trial spaces
  Functional<test_space(trial_space)> residual(&fespace, {&fespace});

  // register a new integral calculation with this residual object
  residual.AddInteriorFaceIntegral(
      Dimension<dim-1>{}, DependsOn<0>{},
      [=](double /*t*/, auto X, auto velocity) {

        // compute the (unit) surface normal
        auto dX_dxi = get<DERIVATIVE>(X);
        auto n = normalize(cross(dX_dxi));

        // extract the velocity values from each side of the interface
        // note: the orientation convention is such that the normal 
        //       computed as above will point from from side 1->2
        auto [u_1, u_2] = velocity; 

        // carry out whatever calculations are necessary to compute f_1, f_2
        auto a = dot(u_2 - u_1, n);
        auto f_1 = u_1 * a;
        auto f_2 = u_2 * a;

        // note: the values returned in the tuple are integrated 
        // against the basis functions from their respective sides
        return serac::tuple{f_1, f_2};

      }, interior_faces);

  ...

  // evaluate the residual with nodal values specified by U
  mfem::Vector U = ...; 
  mfem::Vector r = residual(t, U);
```

--------------

Here's a table showing which argument types are passed to qfunction and which weight functions are applied to the output of qfunctions, depending on the type of integral and trial/test space (respectively):

Screenshot 2024-11-19 at 8 34 25 AM


src/drivers/CMakeLists.txt
samuelpmishLLNL, line 1: I've temporarily disabled these serac "drivers" since:
- to my knowledge, no one uses them
- we don't really have an input-file interface for a lot of things (`Domain` included)
- it's caused confusion for new users (understandably, but incorrectly) thinking that this is the primary way to use serac

Keeping them enabled would add extra burden to this PR which I don't feel is justified in the near future.

samuelpmishLLNL:
This is ready for review, so please take a look! My apologies to the reviewers for the number of files touched.

src/drivers/CMakeLists.txt
white238: There is a user that is currently using it for testing the memory mapped file library.

src/serac/numerics/functional/domain.hpp
```diff
@@ -156,4 +226,34 @@ inline auto by_attr(int value)
...
+    counts[uint32_t(geom)] = uint32_t(domain.get(geom).size());
+  }
+  return counts;
+}
+
+/**
+ * @brief convenience function for computing the arithmetic mean of some list of vectors
+ */
+template <int dim>
+inline tensor<double, dim> average(std::vector<tensor<double, dim> >& positions)
```
tupek2, line 295: I think this has been reimplemented locally in a few places.  Maybe already removed duplication, or can be something for the future?

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -387,29 +459,188 @@ axom::Array<DoF, 2, axom::MemorySpace::Host> GetFaceDofs(const mfem::FiniteEleme
...
+
+        // 3. get the dofs for the entire element
+        mfem::Array<int> elem_dof_ids;
+        fes->GetElementDofs(elem, elem_dof_ids);
+
+        mfem::Geometry::Type elem_geom = mesh->GetElementGeometry(elem);
+
+        // mfem uses different conventions for boundary element orientations in 2D and 3D.
+        // In 2D, mfem's official edge orientations on the boundary will always be a mix of
+        // CW and CCW, so we have to discard mfem's orientation information in order
```
tupek2, line 551: what is CW, CCW?

white238: clockwise/counter clockwise?

src/serac/numerics/functional/domain.hpp
samuelpmishLLNL: this is the deduplication of those other functions-- I've removed all of them (that I found) and put the common implementation here

cmake/SeracBasics.cmake
```diff
@@ -43,7 +43,9 @@ endif()
 
 option(SERAC_ENABLE_PROFILING "Enable profiling functionality" OFF)
 
-cmake_dependent_option(SERAC_ENABLE_BENCHMARKS "Enable benchmark executables" ON "ENABLE_BENCHMARKS" OFF)
+if (ENABLE_BENCHMARKS)
+    set(SERAC_ENABLE_BENCHMARKS ON)
+endif()
```
chapman39, line 48: While `cmake_dependent_option` is admittedly weird I think it is okay as is. `SERAC_ENABLE_BENCHMARKS` generally shouldn't be set explicitly unless you want to enable benchmarks yet disable Serac's. Simply set `ENABLE_BENCHMARKS`.

`SERAC_ENABLE_BENCHMARKS` can only be set if `ENABLE_BENCHMARKS` is `ON`. If `ENABLE_BENCHMARKS` is `ON`, it will automatically set `SERAC_ENABLE_BENCHMARKS` to `ON`. Otherwise `SERAC_ENABLE_BENCHMARKS` is forced to `OFF`.

src/serac/numerics/functional/functional.hpp
```diff
@@ -450,29 +461,60 @@ class Functional<test(trials...), exec> {
...
-    // this is used to mark when operations have been performed,
-    // to avoid doing them more than once
-    bool already_computed[Domain::num_types][num_trial_spaces]{};  // default initializes to `false`
-
     for (auto& integral : integrals_) {
-      auto type = integral.domain_.type_;
+      Domain& dom = integral.domain_;
+
+      const serac::BlockElementRestriction& G_test = dom.get_restriction(test_function_space_);
+      mpi::out << "G_test.ESize() " << G_test.ESize() << std::endl;
```
tupek2: keep prints?

src/serac/numerics/functional/functional.hpp
```diff
@@ -565,6 +608,179 @@ class Functional<test(trials...), exec> {
       return df_;
     }
 
+#if 1
```
tupek2: should remove?

src/serac/numerics/functional/geometric_factors.cpp
```diff
@@ -59,84 +59,157 @@ void compute_geometric_factors(mfem::Vector& positions_q, mfem::Vector& jacobian
...
+  DISPATCH_KERNEL(CUBE, 3, 2, 0);
+  DISPATCH_KERNEL(CUBE, 3, 3, 0);
+  DISPATCH_KERNEL(CUBE, 3, 4, 0);
 
 #undef DISPATCH_KERNEL
 
   std::cout << "should never be reached " << std::endl;
 }
 
+#if 0
```
tupek2: delete?

src/serac/numerics/functional/integral.hpp
```diff
@@ -16,14 +16,18 @@
 #include "serac/numerics/functional/function_signature.hpp"
 #include "serac/numerics/functional/domain_integral_kernels.hpp"
 #include "serac/numerics/functional/boundary_integral_kernels.hpp"
+#include "serac/numerics/functional/interior_face_integral_kernels.hpp"
 #include "serac/numerics/functional/differentiate_wrt.hpp"
 
+// TODO REMOVE AFTER DEBUGGING
```
tupek2: Is this the time to remove?

src/serac/numerics/functional/tests/bug_residual.cpp
```diff
@@ -0,0 +1,102 @@
...
+using namespace serac;
+using namespace serac::profiling;
+
+double t = 0.0;
+
+int num_procs, myid;
+
+#define L2_SCALAR_SPACE
+
+// this is an attempt to reproduce an error message described by Mike
```
tupek2: should we delete this test?

src/serac/numerics/functional/tests/cuda_sandbox.cu
```diff
@@ -0,0 +1,244 @@
...
+                (-5 * xi[0] * (2 + (-3 + sqrt5) * xi[0] - 2 * (3 + sqrt5) * xi[1])) / 2.};
+      if (i == 9)
+        return {-(xi[1] * (-1 + 2 * xi[0] + xi[1])),
+                1 + xi[0] - xi[0] * xi[0] - 2 * (5 + xi[0]) * xi[1] + 15 * xi[1] * xi[1]};
+    }
+
+    return {};
+  }
+
+#if 0
```
tupek2: delete?

src/serac/numerics/functional/tests/dg_restriction_operators.cpp
```diff
@@ -0,0 +1,559 @@
...
+  mfem::VectorFunctionCoefficient vfunc(dim, vector_func_mfem<dim>);
+
+  H1_gf.ProjectCoefficient(sfunc);
+  Hcurl_gf.ProjectCoefficient(vfunc);
+  L2_gf.ProjectCoefficient(sfunc);
+
+  auto H1_dofs    = GetFaceDofs(H1_fes.get(), face_geom, FaceType::INTERIOR);
+  auto Hcurl_dofs = GetFaceDofs(Hcurl_fes.get(), face_geom, FaceType::INTERIOR);
+
+#if 0
```
tupek2: delete?

src/serac/numerics/functional/tests/dg_restriction_operators.cpp
```diff
@@ -0,0 +1,559 @@
...
+  // TODO: check that the actual values match their respective functions
+  //       evaluated directly at the nodes (for H1, Hcurl, and L2 gfs)
+}
+
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+TEST(DomainInterior, TriMesh10) { parametrized_test<mfem::Geometry::TRIANGLE, 1>(0); }
+#if 1
```
tupek2: keep?

src/serac/numerics/functional/tests/functional_basic_dg.cpp
```diff
@@ -0,0 +1,285 @@
...
+//   |      o ----- o      |
+//   |   *            *    |
+//   | *          1      * |
+//   o---------------------o
+//
+
+// [   ---   L    ---   |  --  FND --  ]
+//                         ^^
+
+#if 0
```
tupek2: ?

src/serac/numerics/functional/tests/functional_basic_dg.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+  double t = 0.0;
+
+  auto value = residual(t, U);
+  // check_gradient(residual, t, U);
+}
+
+TEST(basic, L2_test_tris_and_quads_linear) { L2_test<2, 1>(SERAC_REPO_DIR "/data/meshes/patch2D_tris_and_quads.mesh"); }
+
+#if 0
```
tupek2: turn back on?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -263,6 +263,7 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     initializeSolidMechanicsStates();
   }
 
+#if 0
```
tupek2, line 266: ?

tupek2 (approved):
I'm personally OK with disabling the .lua for now.  We may be able to re-prioritize in the spring to revive, but we cannot keep this PR up forever, the scope is already massive.

src/serac/physics/solid_mechanics.hpp
samuelpmishLLNL: this is related to the disabled lua input decks

cmake/SeracBasics.cmake
chapman39: please re-add `cmake_dependent_option(SERAC_ENABLE_BENCHMARKS "Enable benchmark executables" ON "ENABLE_BENCHMARKS" OFF)`

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -276,8 +278,74 @@ axom::Array<DoF, 2, axom::MemorySpace::Host> GetElementRestriction(const mfem::F
...
+    // to apply the native-to-lexicographic permutation
+    if (isH1(*fes)) {
+      for (int k = 0; k < dofs.Size(); k++) {
+        elem_dofs.push_back({uint64_t(dofs[lex_perm[uint32_t(geom)][uint32_t(k)]])});
+      }
+    }
+
+    // the dofs mfem returns for Hcurl include information about
+    // dof orientation, but not for triangle faces on 3D elements.
+    // So, we need to manually
```
btalamini, line 314: If this is unsupported for now, should we abort or warn?

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -387,29 +459,188 @@ axom::Array<DoF, 2, axom::MemorySpace::Host> GetFaceDofs(const mfem::FiniteEleme
...
+axom::Array<DoF, 2, axom::MemorySpace::Host> GetFaceDofs(const serac::fes_t* fes, mfem::Geometry::Type face_geom,
+                                                         const std::vector<int>& mfem_face_ids)
 {
-  dof_info = GetElementRestriction(fes, elem_geom);
+  std::vector<DoF> face_dofs;
+  mfem::Mesh*      mesh         = fes->GetMesh();
+  mfem::Table*     face_to_elem = mesh->GetFaceToElementTable();
 
-  ordering = fes->GetOrdering();
+  // note: this assumes that all the elements are the same polynomial order
```
btalamini, line 469: Do you have an idea how we can enforce this? Check on the `Domain` creation? Doesn't need to be in this PR, but maybe open as issue.

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -387,29 +459,188 @@ axom::Array<DoF, 2, axom::MemorySpace::Host> GetFaceDofs(const mfem::FiniteEleme
...
+  std::vector<Array2D<int> >     local_face_dofs = geom_local_face_dofs(p);
+  std::vector<std::vector<int> > lex_perm        = lexicographic_permutations(p);
 
-  lsize          = uint64_t(fes->GetVSize());
-  components     = uint64_t(fes->GetVDim());
-  num_nodes      = lsize / components;
-  num_elements   = uint64_t(dof_info.shape()[0]);
-  nodes_per_elem = uint64_t(dof_info.shape()[1]);
-  esize          = num_elements * nodes_per_elem * components;
+  // int components_per_node = fes->GetVDim();
```
btalamini, line 479: I think this is left here as an example of what we'd need to do if we can't get mfem to support the custom restriction operator. Can you leave a few words for context on why this dead code should be preserved?

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -468,39 +699,40 @@ void ElementRestriction::ScatterAdd(const mfem::Vector& E_vector, mfem::Vector&
...
+/// create a BlockElementRestriction for the elements in a given domain
+BlockElementRestriction::BlockElementRestriction(const fes_t* fes, const Domain& domain)
 {
-  int dim = fes->GetMesh()->Dimension();
-
-  if (dim == 2) {
-    for (auto geom : {mfem::Geometry::TRIANGLE, mfem::Geometry::SQUARE}) {
-      restrictions[geom] = ElementRestriction(fes, geom);
-    }
+  // TODO: changing the mfem_XXX_ids arrays to mfem_ids[XXX] would simplify this
```
btalamini, line 707: Thanks for the hint - I made #1285 to track this. I'll fix it soon since I'm working on `Domain` anyway.

src/serac/numerics/functional/element_restriction.hpp
```diff
@@ -194,6 +196,9 @@ struct ElementRestriction {
   /// the number of nodes in each element
   uint64_t nodes_per_elem;
 
+  /// an array mapping from domain element ids [0, num_elements) to
```
btalamini: this comment seems half-finished. Was it going to continue "global element ids"?

src/serac/numerics/functional/functional.hpp
```diff
@@ -20,13 +20,14 @@
...
 #include "serac/numerics/functional/finite_element.hpp"
 #include "serac/numerics/functional/integral.hpp"
-#include "serac/numerics/functional/dof_numbering.hpp"
 #include "serac/numerics/functional/differentiate_wrt.hpp"
 
 #include "serac/numerics/functional/element_restriction.hpp"
 
 #include "serac/numerics/functional/domain.hpp"
 
+#include "serac/infrastructure/mpi_fstream.hpp"
```
btalamini: I think Mike pointed these out already, but just please remember to remove

src/serac/numerics/functional/interior_face_integral_kernels.hpp
```diff
@@ -0,0 +1,384 @@
...
+#pragma once
+
+#include <array>
+
+#include "serac/serac_config.hpp"
+#include "serac/numerics/functional/quadrature_data.hpp"
+#include "serac/numerics/functional/differentiate_wrt.hpp"
+
+// TODO REMOVE AFTER DEBUGGING
+#include "serac/infrastructure/mpi_fstream.hpp"
```
btalamini: Another one to remove ... sorry, you probably don't need a reminder on each one.

btalamini (approved):

src/serac/numerics/functional/element_restriction.cpp
samuelpmishLLNL: good question, I'll defer that decision for now (it appears that this is new code in the PR, but it's just a copy of existing code)

src/serac/numerics/functional/element_restriction.cpp
samuelpmishLLNL: I'm not sure. We could just iterate through the elements and check that they agree, but that doesn't catch the (unlikely) case that the ParMesh was partitioned into chunks that each have constant (but different) p.

src/serac/numerics/functional/element_restriction.hpp
samuelpmishLLNL: I think this member variable is unnecessary now, so I'll just delete it

src/serac/numerics/functional/interior_face_integral_kernels.hpp
samuelpmishLLNL: these reminders are helpful

PR #1224, opened 2024-08-26
Switch CI to ruby

chapman39:
Dane has had login nodes down since Thursday, so probably makes sense to switch CI to ruby until that's resolved.


tupek2 (approved):

white238 (approved):

samuelpmishLLNL (approved):
Is there some way for us to just specify which pools of potential runners to use? With github runners, multiple machines can be part of the same pool (so that if a few machines go down, the jobs are just picked up by whatever machines are still active). 

It seems troubling to need to make PRs to a repo to mitigate downtime of specific CI runner machines.

PR #1230, opened 2024-09-11
Easier initial guess for acceleration in dynamics.

tupek2:
Sometimes the initial acceleration guess results in element inversion.  This change defaults the guess to 0.0, which is more robust in a challenging problem, but could potentially be slower for some kinds of problems.

### DIFF
--- a/src/serac/numerics/odes.cpp
+++ b/src/serac/numerics/odes.cpp
@@ -238,7 +238,6 @@ void SecondOrderODE::Solve(const double time, const double c0, const double c1, 
   // use 0 as our starting guess for unconstrained dofs
   d2u_dt2 = state_.d2u_dt2;
   d2u_dt2.SetSubVector(constrained_dofs, 0.0);
-  // d2U_dt2_.SetSubVectorComplement(constrained_dofs, 0.0);
   d2u_dt2 += d2U_dt2_;
   d2u_dt2.SetSubVectorComplement(constrained_dofs, 0.0);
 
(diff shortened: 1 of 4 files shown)


src/serac/numerics/odes.cpp
```diff
@@ -235,11 +235,12 @@ void SecondOrderODE::Solve(const double time, const double c0, const double c1,
   dU_dt_.SetSubVectorComplement(constrained_dofs, 0.0);
   state_.du_dt += dU_dt_;
 
-  // use the previous solution as our starting guess
+  // use 0 as our starting guess for unconstrained dofs
   d2u_dt2 = state_.d2u_dt2;
   d2u_dt2.SetSubVector(constrained_dofs, 0.0);
-  d2U_dt2_.SetSubVectorComplement(constrained_dofs, 0.0);
+  // d2U_dt2_.SetSubVectorComplement(constrained_dofs, 0.0);
```
btalamini: Looks good. Why not remove the commented line?

btalamini (approved):

tests
tupek2, line 1: @chapman39  Is there a reason why these tests need to be in a submodule?  I find it nice to always have code and tests update together in the same repo, when possible.  Also, I have no idea if I even did this submodule thing correctly.

white238: The tests are in a submodule so that if the repo gets large with all the data that builds up, it can be cleaned out w/o having to alter the history of the serac repo. This has happened on all simulation codes I've been on <shrugs>

samuelpmishLLNL (approved):

PR #1234, opened 2024-09-17
Change git strategy to start from a clean slate each time

### LINKED ISSUE #1233: Occasional CI failures on LC
Sometimes, the following error occurs in CI:

```
fatal: not a git repository: axom/../.git/modules/axom
Synchronizing submodule url for 'axom'
logout
```

For example: https://lc.llnl.gov/gitlab/smith/serac/-/jobs/2119259

Re-running the failed job resolves the error, but we should look into the root of the issue.

Comment from @white238 :

> i fear there is something wrong with the deletion cronjob we have. my suspicion is its deleting subdirectories in the .git directory and not the whole pipeline directory but i haven't been able to catch a failure so i could investigate the state

According to @jamiebramwell , the cronjob which deletes old gitlab builds occurs every day at 1 am, but this issue can occur at any time. crontab: `0 1 * * * ~/bin/cleanup_gitlab_builds.sh -d 7`

white238:
The default `GIT_STRATEGY` for Gitlab is `fetch`, which reuses the git repo if it is present. This changes it to `clone` which will start from a clean slate.

Hopefully this fixes #1233 but we won't know for a while due to this being intermittent.

https://docs.gitlab.com/ee/ci/runners/configure_runners.html#git-strategy

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,6 +6,7 @@
 variables:
   #TODO: uncomment this when everyone has service user access across all the machines
   #LLNL_SERVICE_USER: asmith
+  GIT_STRATEGY: clone
   GIT_SUBMODULE_STRATEGY: recursive
   PROJECT_ALLOC_NAME: ${CI_PROJECT_NAME}_ci_${CI_PIPELINE_ID}
   BUILD_ROOT: ${CI_PROJECT_DIR}


chapman39 (approved):

tupek2 (approved):
Is this going to increase ci time much. or maybe not significantly?

white238:
> Is this going to increase ci time much. or maybe not significantly?

I doubt it, our repo isn't very large.

PR #1238, opened 2024-09-23
Add adjoint load for the case of adjoints associated with resultant/dual forces.

tupek2:


tupek2:
This enables a simpler and more consistent interface for reaction force sensitivities.  Also, add function to fetch a previous reaction force vector.  This is really recomputing it behind the scenes.

src/serac/physics/solid_mechanics.hpp
```diff
@@ -1201,7 +1229,9 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
             std::unique_ptr<mfem::HypreParMatrix> m_mat(assemble(M));
 
             // J = M + c0 * K
+            J_.reset();
```
btalamini, line 1232: Is this fixing a bug? Because the heat transfer module still does it the old way. If so, can you open an issue?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -1316,9 +1346,11 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
 
     SLIC_ERROR_ROOT_IF(disp_adjoint_load == loads.end(), "Adjoint load for \"displacement\" not found.");
 
-    displacement_adjoint_load_ = disp_adjoint_load->second;
-    // Add the sign correction to move the term to the RHS
-    displacement_adjoint_load_ *= -1.0;
+    if (disp_adjoint_load != loads.end()) {
```
btalamini, line 1349: This check and the `SLIC_ERROR_ROOT_IF` above look like they're guarding the same problem. Can we combine these so that they don't get out of sync? Doesn't need to happen in this PR.

btalamini (approved):

kswartz92 (approved):

PR #1239, opened 2024-09-23
No Optional Solvers CI Job

### LINKED ISSUE #1221: Questions about requiring currently optional packages (petsc, tribol, sundials) and testing them in CI
This issue is a discussion place about optional packages (tribol, sundials, petsc, etc.).

* Should we require Tribol? PETSc? Sundials?
* Which optional packages should we test in CI?
* Ideally we should test all combinations of optional packages enabled/ disabled, but this isn't feasible.
* Requiring some more packages will reduce the amount of CI jobs required, since we won't have as many combination of options to cover.

Right now, we do not test the no-petsc case in CI, which lead to this issue for example https://github.com/LLNL/serac/pull/1217.

chapman39:
Additionally disables petsc with the no-sundials job, and renames the job to "no solvers".

The motivation is to test as many options as we can with as few jobs as possible.

Fixes #1221

### DIFF
--- a/.gitlab/build_toss4.yml
+++ b/.gitlab/build_toss4.yml
@@ -61,11 +61,11 @@ toss4-gcc_10_3_1-src-no-tribol:
     ALLOC_TIME: "30"
   extends: .src_build_on_toss4
 
-toss4-gcc_10_3_1-src-no-sundials:
+toss4-gcc_10_3_1-src-no-optional-solvers:
   variables:
     COMPILER: "gcc@10.3.1"
     HOST_CONFIG: "ruby-toss_4_x86_64_ib-${COMPILER}.cmake"
-    EXTRA_CMAKE_OPTIONS: "-DENABLE_BENCHMARKS=ON -USUNDIALS_DIR"
+    EXTRA_CMAKE_OPTIONS: "-DENABLE_BENCHMARKS=ON -USUNDIALS_DIR -UPETSC_DIR"
     ALLOC_NODES: "1"
     ALLOC_TIME: "20"
   extends: .src_build_on_toss4


.gitlab/build_toss4.yml
```diff
@@ -61,11 +61,11 @@ toss4-gcc_10_3_1-src-no-tribol:
     ALLOC_TIME: "30"
   extends: .src_build_on_toss4
 
-toss4-gcc_10_3_1-src-no-sundials:
+toss4-gcc_10_3_1-src-no-optional-solvers:
   variables:
     COMPILER: "gcc@10.3.1"
     HOST_CONFIG: "ruby-toss_4_x86_64_ib-${COMPILER}.cmake"
-    EXTRA_CMAKE_OPTIONS: "-DENABLE_BENCHMARKS=ON -USUNDIALS_DIR"
+    EXTRA_CMAKE_OPTIONS: "-DENABLE_BENCHMARKS=ON -USUNDIALS_DIR -UPETSC_DIR"
```
white238, line 68: Not really connected to this PR but do we want benchmarks enabled for these jobs?

white238 (approved):

tupek2 (approved):
Yes!

.gitlab/build_toss4.yml
chapman39: i think it makes sense to at least build them here, so make sure they compile. at the moment they do not run in ci. at the moment they aren't too lengthy, but assuming the benchmarks take longer, it would make sense to run them on merge or weekly, rather than on every push.

white238: Thats a good point.

PR #1245, opened 2024-10-09
Run benchmarks weekly in CI

chapman39:
This PR...

* Adds `run_benchmarks.py` script that builds Serac, runs benchmarks, and moves resulting Caliper files in shared location on LC.
* Sets up CI such that this script is run on all platforms (ruby, lassen, tioga)
* Setup [weekly Pipeline Schedule on Gitlab](https://lc.llnl.gov/gitlab/smith/serac/-/pipelines/472258)

Also,

* Disables benchmarks on full builds (since benchmarks are being compiled on src already)
* Fixes cray full CI builds? (it was not using the spec in `specs.json`)
* Fixes `physics_benchmark_solid_nonlinear_solve` benchmark to handle petsc being disabled
* Disables all cray CI jobs until all developers have access (only on-push was disabled previously)


.gitlab-ci.yml
```diff
@@ -19,11 +19,15 @@ variables:
...
+    - if: '$FULL_BUILD != "ON" && $BENCHMARKS_BUILD != "ON"'
 
 .full_workflow:
   rules:
-    - if: '$FULL_BUILD == "ON"'
+    - if: '$FULL_BUILD == "ON" && $BENCHMARKS_BUILD != "ON"'
+
+.benchmarks_workflow:
+  rules:
+    - if: '$FULL_BUILD != "ON" && $BENCHMARKS_BUILD == "ON"'
```
chapman39: We could alternatively use some sort of `CI_WORKFLOW_TYPE` var which could be src, full, or benchmarks. That way we don't have to handle the two variables. I just don't have access to nightly build settings.

chapman39:
In a future PR I'll update the docs to explain how to properly view this data on SPOT, but it's more or less the following:

* Go to https://lc.llnl.gov/spot2/?sf=/usr/WS2/smithdev/califiles
* Click the check mark button on the top right
* Ensure "cluster" is ticked on
* Click on "ruby" under the cluster pie chart (of whatever machine you want to view)
* Click on the benchmark you want to view under the "executable" pie chart
* Scroll down to the table and and select the "compare" tab
* Above the charts set "Group by" to "mpi.world.size"

You'll notice at the moment each chart is practically a straight line, because there haven't been any changes to Serac yet. If you want to view another cluster or executable, click on the previous one to deselect it before selecting a new one. That way to you don't view both at the same time.

tupek2 (approved):

chapman39:
Weekly pipeline passing https://lc.llnl.gov/gitlab/smith/serac/-/pipelines/472258

scripts/llnl/run_benchmarks.py
```diff
@@ -0,0 +1,107 @@
...
+    # Build Serac
+    os.chdir(repo_dir)
+    os.makedirs(test_root, exist_ok=True)
+    build_and_test_host_config(test_root=test_root, host_config=host_config_path,
+                               report_to_stdout=True, extra_cmake_options=cmake_options,
+                               skip_install=True, skip_tests=True)
+
+    # Go to build location
+    build_dir=""
+    dirs = glob.glob(pjoin(test_root, "*"))
```
white238: this may get the wrong build directory if there is an existing one.

src/serac/physics/benchmarks/CMakeLists.txt
```diff
@@ -13,20 +13,24 @@ set(physics_benchmark_targets
...
                        OUTPUT_DIR ${PROJECT_BINARY_DIR}/benchmarks
                        FOLDER serac/benchmarks
                        )
 
     # Add benchmarks with various task counts
     foreach(task_count 1 4 16)
-        blt_add_benchmark(NAME ${physics_benchmark}_${task_count}_task_count
-                          COMMAND ${physics_benchmark}
+        set(physics_benchmark ${physics_benchmark_exec}_${task_count}_task_count)
+        blt_add_benchmark(NAME ${physics_benchmark}
```
white238, line 27: This ends up being empty due to only changing part of the variables. i would recommend just putting it back to `physics_benchmark`

src/serac/physics/benchmarks/CMakeLists.txt
```diff
@@ -13,20 +13,24 @@ set(physics_benchmark_targets
...
+                          COMMAND ${physics_benchmark_exec}
                           NUM_MPI_TASKS ${task_count}
                           WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
                           )
+
+        # Increase benchmark time limit
+        set_tests_properties(${physics_benchmark} PROPERTIES TIMEOUT 4800)
```
white238:
```suggestion
        set(physics_benchmark_name ${physics_benchmark_exec}_${task_count}_task_count)
        blt_add_benchmark(NAME ${physics_benchmark_name}
                          COMMAND ${physics_benchmark_exec}
                          NUM_MPI_TASKS ${task_count}
                          WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
                          )

        # Increase benchmark time limit
        set_tests_properties(${physics_benchmark_name} PROPERTIES TIMEOUT 4800)
```

The multiple variable names is getting a bit confusing in this file. I think this helps. Possibly also changing the above `physics_benchmark_targets` to just `physics_benchmarks` because they aren't targets anymore

.gitlab-ci.yml
white238: You now can!

chapman39:
on-push ci: https://lc.llnl.gov/gitlab/smith/serac/-/pipelines/474176
manual nightly full: https://lc.llnl.gov/gitlab/smith/serac/-/pipelines/474187
manuall weekly benchmarks: https://lc.llnl.gov/gitlab/smith/serac/-/pipelines/474186

white238 (approved):

PR #1248, opened 2024-10-17
Quick fix to make sure trust region scaling is parallel consistent.

tupek2:


src/serac/numerics/equation_solver.cpp
```diff
@@ -502,10 +502,12 @@ class TrustRegion : public mfem::NewtonSolver {
...
-    settings.min_cg_iterations                          = static_cast<size_t>(nonlinear_options.min_iterations);
-    settings.max_cg_iterations                          = static_cast<size_t>(linear_options.max_iterations);
-    settings.cg_tol                                     = 0.5 * norm_goal;
-    double tr_size                                      = nonlinear_options.trust_region_scaling * std::sqrt(X.Size());
+    settings.min_cg_iterations = static_cast<size_t>(nonlinear_options.min_iterations);
+    settings.max_cg_iterations = static_cast<size_t>(linear_options.max_iterations);
+    settings.cg_tol            = 0.5 * norm_goal;
+
+    scratch        = 1.0;
+    double tr_size = nonlinear_options.trust_region_scaling * std::sqrt(Dot(scratch, scratch));
```
kswartz92, line 510: If `scratch` is of length `X.size()` then setting each element to 1.0 and taking the inner product with itself is would give the same value as `X.size()` which is still parallel inconsistent?

tupek2: scratch is X.size() on each processor.  Dot() does a parallel reduction over procs to give the total size.  Its kind of an mfem shortcut trick.

kswartz92: ahh okay, I was assuming `Dot` was a local operation

kswartz92 (approved):

samuelpmishLLNL (approved):

PR #1249, opened 2024-10-19
Multi-material solid mechanics

btalamini:
Enable solid mechanics to work with materials defined on different domains. Includes materials with internal state variables.

For now, I added this without breaking changes. This means I overloaded the `setMaterial` several times and left in place the old ones that do not take a domain. Eventually, I'd prefer the domain to be required. We can map this out later.

### DIFF
--- a/src/serac/numerics/functional/functional.hpp
+++ b/src/serac/numerics/functional/functional.hpp
@@ -301,7 +301,7 @@ public:
 
   /// @overload
   template <int dim, int... args, typename lambda, typename qpt_data_type = Nothing>
-  void AddDomainIntegral(Dimension<dim>, DependsOn<args...>, const lambda& integrand, Domain& domain,
+  void AddDomainIntegral(Dimension<dim>, DependsOn<args...>, const lambda& integrand, const Domain& domain,
                          std::shared_ptr<QuadratureData<qpt_data_type>> qdata = NoQData)
   {
     if (domain.mesh_.GetNE() == 0) return;
--- a/src/serac/physics/state/state_manager.hpp
+++ b/src/serac/physics/state/state_manager.hpp
@@ -130,6 +130,30 @@ public:
     return std::make_shared<QuadratureData<T>>(elems, qpts_per_elem, initial_state);
   }
 
+  // make qdata by domain
+  template <typename T>
+  static std::shared_ptr<QuadratureData<T>> newQuadratureDataBuffer(const Domain& domain, int order, int dim,
+                                                                    T initial_state)
+  {
+    int Q = order + 1;
+
+    std::array<uint32_t, mfem::Geometry::NUM_GEOMETRIES> elems = geometry_counts(domain);
+    std::array<uint32_t, mfem::Geometry::NUM_GEOMETRIES> qpts_per_elem{};
+
+    std::vector<mfem::Geometry::Type> geometries;
+    if (dim == 2) {
+      geometries = {mfem::Geometry::TRIANGLE, mfem::Geometry::SQUARE};
+    } else {
+      geometries = {mfem::Geometry::TETRAHEDRON, mfem::Geometry::CUBE};
+    }
+
+    for (auto geom : geometries) {
+      qpts_per_elem[size_t(geom)] = uint32_t(num_quadrature_points(geom, Q));
+    }
+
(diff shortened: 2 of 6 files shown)


src/serac/physics/state/state_manager.hpp
```diff
@@ -130,6 +130,39 @@ class StateManager {
...
+   * @param initial_state the value to be broadcast to each quadrature point
+   * @return shared pointer to quadrature data buffer
+   */
+  template <typename T>
+  static std::shared_ptr<QuadratureData<T>> newQuadratureDataBuffer(const Domain& domain, int order, int dim,
+                                                                    T initial_state)
+  {
+    int Q = order + 1;
+
+    std::array<uint32_t, mfem::Geometry::NUM_GEOMETRIES> elems = geometry_counts(domain);
```
white238, line 114: There is only two lines different in this vs the `mesh_tag` `newQuadratureDataBuffer` function. Can you reduce the duplication between them?

src/serac/physics/tests/solid_multi_material.cpp
```diff
@@ -0,0 +1,280 @@
...
+  constexpr double H      = 1.0;
+  constexpr double VOLUME = L * W * H;
+
+  auto mesh = mesh::refineAndDistribute(buildCuboidMesh(8, 1, 1, L, W, H), serial_refinement, parallel_refinement);
+
+  const std::string mesh_tag{"mesh"};
+
+  auto& pmesh = serac::StateManager::setMesh(std::move(mesh), mesh_tag);
+
+  // _solver_params_start
```
white238, line 63: These tags are used in the sphinx documentation and are not needed here. Looking at the code these tags have bled into other files and should only be in `serac/physics/tests/solid.cpp`.

src/serac/physics/tests/solid_multi_material.cpp
```diff
@@ -0,0 +1,280 @@
+// Copyright (c) 2019-2024, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+#include "serac/physics/solid_mechanics.hpp"
+
+#include "axom/slic/core/SimpleLogger.hpp"
```
white238:
```suggestion
```
This isn't needed due to using the `serac::initiliaze` call which uses the Serac SLIC logger.

src/serac/numerics/functional/functional.hpp
```diff
@@ -301,7 +301,7 @@ class Functional<test(trials...), exec> {
 
   /// @overload
   template <int dim, int... args, typename lambda, typename qpt_data_type = Nothing>
-  void AddDomainIntegral(Dimension<dim>, DependsOn<args...>, const lambda& integrand, Domain& domain,
+  void AddDomainIntegral(Dimension<dim>, DependsOn<args...>, const lambda& integrand, const Domain& domain,
```
samuelpmishLLNL, line 304: the DG PR requires making changes to the domain class when defining integrals, so we might have to revert this back to non-const `Domain & domain` soon (or use `mutable` stuff)

samuelpmishLLNL (approved):

src/serac/physics/state/state_manager.hpp
btalamini: Yes. I removed the `mesh_tag` version. It wasn't being used. The solid mechanics module handles the logic of whether the variables are allocated over the entire mesh or a single `Domain`.

src/serac/physics/tests/solid_multi_material.cpp
btalamini: Good catch, I didn't even see those.

src/serac/physics/tests/solid_multi_material.cpp
btalamini: Done

src/serac/numerics/functional/functional.hpp
btalamini: Noted. There's the usual virality: If we take away the `const` here, it will have to propagate down to `Functional` (which is why I added it here). I'll let you pick which.

white238 (approved):

PR #1255, opened 2024-10-28
Tupek/checkpoint to disk again

tupek2:
The reading of checkpointed states was off by an iteration due to order in which we were constructing the map from state name to FiniteElementState.  The map was constructed before the states were read from disk.  A test has been changed to test this restart to disk capability, and also to be a bit more robust when a finer mesh is used.


btalamini (approved):

samuelpmishLLNL (approved):

src/serac/physics/base_physics.cpp
```diff
@@ -363,10 +363,12 @@ std::unordered_map<std::string, FiniteElementState> BasePhysics::getCheckpointed
 
   if (checkpoint_to_disk_) {
     for (const auto& state_name : stateNames()) {
-      previous_states_map.emplace(state_name, state(state_name));
       previous_states_ptrs.emplace_back(const_cast<FiniteElementState*>(&state(state_name)));
     }
     StateManager::loadCheckpointedStates(cycle_to_load, previous_states_ptrs);
+    for (const auto& state_name : stateNames()) {
```
white238, line 369: Does `loadCheckpointedStates` alter the order of `stateNames`? If so can you put a comment here?

white238 (approved):

src/serac/physics/base_physics.cpp
tupek2: No, it doesn't change the order.  But, it does copy the values from state(state_name).  The issue was we were copying those values when making that map, but that was happening before we read from disk.  So the map fields were out of date.

white238: I can see me copy/pasting this line back into the upper loop in the future.....

tupek2: good thing the test will fail and yell at you!

PR #1258, opened 2024-10-29
Add Quadrature Data to Sidre

### LINKED ISSUE #819: Include quadrature point data in checkpoints
We need to include quadrature data in restart files. We can't use the `mfem::QuadratureFunction` interface because this only works for struct-of-arrays, while serac uses array-of-structs. Instead, we should use a serialize the quadrature data object and use named buffers in Axom.

https://github.com/LLNL/serac/blob/b3485f17fc2fb69de078fe413bab5c358ca9f920/src/serac/numerics/functional/quadrature_data.hpp#L105-L146 

https://github.com/LLNL/axom/blob/09ab17b68b67eacf45a9624e1909a1d700b1985a/src/axom/sidre/core/MFEMSidreDataCollection.hpp#L280-L301

https://github.com/LLNL/axom/blob/09ab17b68b67eacf45a9624e1909a1d700b1985a/src/axom/sidre/core/MFEMSidreDataCollection.hpp#L507

white238:
Fixes #819


tupek2 (approved):

src/serac/physics/state/state_manager.cpp
```diff
@@ -275,16 +277,18 @@ double StateManager::time(std::string mesh_tag)
   return datacolls_.at(mesh_tag).GetTime();
 }
 
-void checkMesh(const mfem::ParMesh& pmesh)
+void checkMesh(const mfem::ParMesh& pmesh, bool is_restart)
```
btalamini, line 280: Is this ever called with the `is_restart` argument set true? I thought we put that in yesterday, but I don't see any calls with the argument used. Sorry if I'm just missing it.

src/serac/physics/state/state_manager.hpp
```diff
@@ -95,19 +95,119 @@ class StateManager {
...
+                             "Loaded Quadrature Data geometry Sidre view did not have 'states'");
+          axom::sidre::View* states_view = geom_group->getView("states");
+          states_view->setExternalDataPtr(states.data());
+        }
+      }
+    }
+
+    if (is_restart_) {
+      // NOTE: This call will reload all external buffers from file stored in the DataStore
+      // TODO: This should be changed to load only the current material quadrature data after
```
chapman39, line 194: might be good to point to an issue either in serac or another proj to track this

src/serac/physics/tests/solid_robin_condition.cpp
```diff
@@ -47,7 +47,6 @@ void functional_solid_test_robin_condition()
   auto mesh = mesh::refineAndDistribute(buildMeshFromFile(filename), serial_refinement, parallel_refinement);
   auto& pmesh = serac::StateManager::setMesh(std::move(mesh), mesh_tag);
 
-  // _solver_params_start
```
chapman39, line 50: is this referenced in any docs?

src/serac/physics/state/tests/test_state_manager.cpp
```diff
@@ -17,9 +17,138 @@
...
+}  // namespace detail
+
+TEST(state_manager, QuadratureData_Restart)
+{
+  // This test checks that the state manager can save and load
+  // a quadrature data object. It does this by creating a
+  // quadrature data object, populating it with some data,
+  // saving it to disk, and then loading it back from disk.
+  // It then checks that the loaded data matches the original
+  // data.
```
chapman39, line 70: 👍

chapman39 (approved):

src/serac/physics/tests/solid_robin_condition.cpp
white238: These were copy/pasted from the actual place used in the documentation, which is the only one left now.

src/serac/physics/state/state_manager.hpp
white238: oh man.. thanks for reminding me of this. This is no longer true

src/serac/physics/state/state_manager.hpp
```diff
@@ -95,19 +95,119 @@ class StateManager {
...
    * @param domain The spatial domain over which to allocate the quadrature data
    * @param order The order of the discretization of the primal fields
    * @param dim The spatial dimension of the mesh
    * @param initial_state the value to be broadcast to each quadrature point
    * @return shared pointer to quadrature data buffer
    */
   template <typename T>
-  static std::shared_ptr<QuadratureData<T>> newQuadratureDataBuffer(const Domain& domain, int order, int dim,
-                                                                    T initial_state)
+  static std::shared_ptr<QuadratureData<T>> newQuadratureDataBuffer(const std::string& mesh_tag, const Domain& domain,
```
btalamini, line 214: The mesh holds domains now. We will want to eventually pass the string identifying the domain instead of passing the domain object, but I don't think we're ready to do that yet. We need to flesh out our mesh class more fully. Nothing to do here, just a note.

btalamini (approved):
This looks great. Nice test additions.

src/serac/physics/state/state_manager.hpp
white238: In the many PRs and months since I forgot a piece in Axom that needed to be addressed. I created a PR in axom:

https://github.com/LLNL/axom/pull/1555

and also added better comments and the new lines that need to be swapped out when that PR goes through and our TPLs need to rebuilt.

src/serac/physics/state/state_manager.cpp
white238: Thanks for pointing this out. I forgot to pass in the bool on restart and fixed this. It is now used.

PR #1261, opened 2024-11-05
Switch Convention to First Piola stress and remove GeometricNonlinearities flag

btalamini:
Shift our convention for the material models in solid mechanics from using Cauchy stress to the first Piola stress (aka first Piola-Kirchhoff stress). There are several motivations for this:

- **Removes the** `GeometricNonlinearities` **flag and the possible inconsistent states this allowed.** Now, if you set a geometrically linear material model, you get linear behavior, and if you set a geometrically nonlinear material model, you get the nonlinear behavior. Before, you set a material, linear or nonlinear, and had to set the `GeometricNonlinearities` flag correspondingly, or else nonsensical results would follow. I'm not aware of any meaningful use of having the flag set differently than the material model. This was just a big footgun and a common point of confusion from our users.
- **Brings consistency between the material model and the other input data.** Our methods for setting tractions and body forces expected reference configuration (Piola) forces. Our boundary conditions functors also pass reference configuration coordinates. The old constitutive model convention bucked this pattern.
- **The output of the constitutitve model (stress) is now conjugate to the main input to the constitutive model (displacement gradient).** This is just a nice symmetry to have. This means that the output stress is the derivative of the strain energy density with respect to its argument, so if we ever extend our AD capabilities to multiple derivatives, we're in a better position to use minimization solvers directly.

I did not update the documentation in this PR to limit the breadth of changes. I'm doing that in a sister PR, #1264.

### DIFF
--- a/data/input_files/tests/solid/qs_linear.lua
+++ b/data/input_files/tests/solid/qs_linear.lua
@@ -39,12 +39,6 @@ solid = {
     -- neo-Hookean material parameters
     materials = { { model = "NeoHookean", mu = 0.25, K = 10.0, density = 1.0 }, },
 
-    -- Turn the material nonlinearities off
-    -- TODO: this should be replaced with a proper material definition
-    -- BT 11/5/24: I think this option is outdated and no longer used, can an inlet
-    -- expert confirm?
-    material_nonlin = false,
-
     -- boundary condition parameters
     boundary_conds = {
         ['displacement'] = {
--- a/data/input_files/tests/solid/qs_linear.lua
+++ b/data/input_files/tests/solid/qs_linear.lua
@@ -39,11 +39,10 @@ solid = {
     -- neo-Hookean material parameters
     materials = { { model = "NeoHookean", mu = 0.25, K = 10.0, density = 1.0 }, },
 
-    -- Turn the geometric nonlinearities off
-    geometric_nonlin = false,
-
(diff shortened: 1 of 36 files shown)


tupek2 (approved):

data/input_files/tests/solid/qs_linear.lua
```diff
@@ -39,11 +39,10 @@ solid = {
...
     materials = { { model = "NeoHookean", mu = 0.25, K = 10.0, density = 1.0 }, },
 
-    -- Turn the geometric nonlinearities off
-    geometric_nonlin = false,
-
     -- Turn the material nonlinearities off
     -- TODO: this should be replaced with a proper material definition
+    -- BT 11/5/24: I think this option is outdated and no longer used, can an inlet
+    -- expert confirm?
     material_nonlin = false,
```
samuelpmish: there are no other uses of "material_nonlin" in the repo, so it's probably safe to remove?

btalamini: That's what I was thinking. @white238, I don't know this part well. Can you confirm?

white238: `geometric_nonlin` and `material_nonlin` where input options for the pre-functional `Solid` class. They are not used anymore and can be deleted.

samuelpmishLLNL (approved):

PR #1267, opened 2024-11-07
Update radiuss submodule

chapman39:
Newer Spack versions have [changed the way the function `cache_extra_test_sources` is called](https://github.com/spack/spack/commit/10b6d7282ae086a7d03361da205499fd1229977a#diff-8018c28dbd7ebe14804238fa6d4b867ce0b418774cac17c44e68e7e003ac9c60R367), requiring all packages that use it to change. [RAJA's Spack package was affected by this](https://github.com/LLNL/radiuss-spack-configs/blob/main/packages/raja/package.py#L465), so we need to update the radiuss-spack-configs, since that's where the RAJA Spack package is located.


chapman39:
it's affecting the lido tpl builds which is how i encountered this issue. serac probably hasn't hit it yet since the spack version lido's using is slightly newer

chapman39:
im just going to match the same spack version between lido and serac. i wanted spack a little farther in lido since i had a spack pr that merged somewhat recently, but we'll just get that once serac updates spack.

PR #1270, opened 2024-11-12
fixes for builds with no raja or umpire

ebchin:
I tried building serac without RAJA or Umpire and found some code that needs some `#ifdef` macros around it.  This PR adds them.

### DIFF
--- a/src/serac/infrastructure/debug_print.hpp
+++ b/src/serac/infrastructure/debug_print.hpp
@@ -74,7 +74,7 @@ std::ostream& operator<<(std::ostream& out, serac::SignedIndex i)
  * @param filename the name of the output file
  */
 template <typename T>
-#ifdef SERAC_USE_RAJA
+#ifdef SERAC_USE_UMPIRE
 void write_to_file(axom::Array<T, 2, axom::MemorySpace::Host> arr, std::string filename)
 #else
 void write_to_file(axom::Array<T, 2> arr, std::string filename)
@@ -101,7 +101,7 @@ void write_to_file(axom::Array<T, 2> arr, std::string filename)
  * @param filename the name of the output file
  */
 template <typename T>
-#ifdef SERAC_USE_RAJA
+#ifdef SERAC_USE_UMPIRE
 void write_to_file(axom::Array<T, 3, axom::MemorySpace::Host> arr, std::string filename)
 #else
 void write_to_file(axom::Array<T, 3> arr, std::string filename)
--- a/src/serac/infrastructure/debug_print.hpp
+++ b/src/serac/infrastructure/debug_print.hpp
@@ -68,53 +68,48 @@ std::ostream& operator<<(std::ostream& out, serac::SignedIndex i)
 }
 
 /**
- * @brief write a 2D array of values out to file, in a space-separated format
+ * @brief write an array of values out to file, in a space-separated format
  * @tparam T the type of each value in the array
  * @param v the values to write to file
  * @param filename the name of the output file
  */
-template <typename T>
-void write_to_file(axom::Array<T, 2> arr, std::string filename)
+template <typename ArrayT>
+void write_to_file(const ArrayT& arr_orig, std::string filename)
 {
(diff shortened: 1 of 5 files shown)


src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -267,17 +271,29 @@ axom::Array<DoF, 2, axom::MemorySpace::Host> GetElementRestriction(const mfem::F
   }
 
   if (n == 0) {
+#ifdef SERAC_USE_RAJA
```
white238: I think this should be `SERAC_USE_UMPIRE`

ebchin: Yeah makes sense, basically all of the `axom::MemorySpace::Host` stuff needs to be fixed.

src/serac/infrastructure/debug_print.hpp
```diff
@@ -74,7 +74,11 @@ std::ostream& operator<<(std::ostream& out, serac::SignedIndex i)
  * @param filename the name of the output file
  */
 template <typename T>
+#ifdef SERAC_USE_UMPIRE
 void write_to_file(axom::Array<T, 2, axom::MemorySpace::Host> arr, std::string filename)
+#else
```
white238: You would think that `Host` would always be there but its guarded in axom and `Dynamic` is the one unguarded....

https://github.com/LLNL/axom/blob/2244835b3389c427c337b5df782982d2ae609bfe/src/axom/core/memory_management.hpp#L37-L47

ebchin: Seems like host should be safe to unguard in addition to dynamic?  Regardless, we could use dynamic + umpire allocator IDs to achieve the same without the clunky #ifdef blocks.  Though we would lose the explicit typing of the memory.  Thoughts?

chapman39:
Should we test this in CI? Maybe in Azure?

chapman39 (approved):
thanks @ebchin !

ebchin:
> Should we test this in CI? Maybe in Azure?

We probably don't want to get into all of the combinatorics, but I see the value in adding a CI test with minimal TPLs, just to check stuff like this.

src/serac/infrastructure/debug_print.hpp
white238: I say go with `Dynamic` for now to avoid the ifdef's... but this really needs to be revisited. Might need to chat with Axom about this.

kennyweiss: @white238 and I spoke -- it makes sense to go with (the default) Dynamic

ebchin: It's not clear to me Dynamic will give the desired result when the original `axom::Array` is on device.  It seems like this copy constructor will be called: https://github.com/LLNL/axom/blob/develop/src/axom/core/Array.hpp#L1180-L1190 which copies the allocator ID from the existing array.  The result will be an array on device and the operator() call on L88 will fail.  Is there a different copy constructor being called by this function?

ebchin: Implemented a fix suggested by @kennyweiss

src/serac/infrastructure/debug_print.hpp
```diff
@@ -68,53 +68,48 @@ std::ostream& operator<<(std::ostream& out, serac::SignedIndex i)
...
+  {
+    for (axom::IndexType i = 0; i < arr.shape()[0]; i++) {
+      outfile << "{";
+      for (axom::IndexType j = 0; j < arr.shape()[1]; j++) {
+        outfile << arr(i, j);
+        if (j < arr.shape()[1] - 1) outfile << ", ";
+      }
+      outfile << "}\n";
+    }
+  } else if constexpr(ArrayT::Dims == 3) {
```
white238: We got compiler errors when we have `else if constexpr` before. Can you remove the `else` and make it just two `if constexpr` blocks?

white238 (approved):

Issue #1272, opened 2024-11-17
Some factory methods for `Domain' don't fully construct valid objects


btalamini:
The Boolean set operation methods that construct instances of `Domain` don't fill out all the member variables. The missing ones (the `mfem_*_ids_`) are needed for getting indices into `FiniteElementState` instances.

This bug is also present in the `EntireDomain` and `EntireBoundary` methods, but I'm about to submit a simple fix for those.

samuelpmishLLNL:
I could fix this in its own PR or roll it into the DG one (which is already making some improvements to the Domain class)-- do you have any preference?

PR #1273, opened 2024-11-17
Fix bug in domains created by Entire{Domain,Boundary} functions

btalamini:
The bug was that the mfem entity id vectors (eg, `mfem_edge_ids_`) were not populated, which are needed to fetch the indices of the dofs in the domain (through the `dof_list()` function). When calling `dof_list()`, no error was thrown, you would just get an empty list. The simulation would continue, but with the dofs left out of whatever operations you used the `Domain`s for.

This fixes the bug by removing the repeated (though incomplete) code in the Entire* functions. Now they call through to the standard factory functions (e.g., `ofElements`) used for domain creation. It now just sends a predicate that is always true. 

The main changes are in `domain.{hpp,cpp}`. My change added a template parameter (for the spatial dimension) to the `Entire*` functions, so the rest of the changes are just adding the template argument to the call sites throughout the code.

This partially address Issue #1272 . There is still a similar problem with the boolean set operators (`&`, `|`, and `-`). I could add in the lines to populate the missing entity id vectors in the operator implementations, but I'd prefer a fix that didn't amount to repeating more code. I'm open to suggestions.


white238 (approved):

src/serac/numerics/functional/domain.hpp
```diff
@@ -135,10 +135,20 @@ struct Domain {
...
+}
 
 /// @brief constructs a domain from all the boundary elements in a mesh
-Domain EntireBoundary(const mfem::Mesh& mesh);
+template <int dim>
+Domain EntireBoundary(const mfem::Mesh& mesh)
+{
+  auto tautological_predicate = [](std::vector<tensor<double, dim>>, int) -> bool { return true; };
+  return Domain::ofBoundaryElements(mesh, tautological_predicate);
+}
```
samuelpmishLLNL, line 151: I think these can be implemented in a way that doesn't propagate the template parameter `dim` up to the user, which would make this PR avoid the breaking interface change:

https://godbolt.org/z/c5YeqPrrW

samuelpmishLLNL (approved):
Thanks for fixing my bugs!

PR #1275, opened 2024-11-22
Fix acquisition of dof lists from Domains

### LINKED ISSUE #1272: Some factory methods for `Domain' don't fully construct valid objects

btalamini:
There are two sets of books in `Domain`:

- Element index lists that are broken out by element geometry, answering questions like "for this tetrahedron in the Domain, where does it go in the list of all tetrahedra in the mesh?". This is used by `Functional` to identify entities that should be integrated over.
- Global element indices, for answering "which element is this in the `mfem::Mesh` data structure?" (which does not segregate by element geometry). These are used to be able to grab dofs that live on a `Domain`.

As noted in #1272, some of the factory methods and operators on `Domain` did not populate all the required data. The omissions were the second group, the global element data, so grabbing dofs were broken in some cases. 

Fundamentally, this class should enforce that every time an entry is added to the first group, a corresponding entry is also added to the second. That's an interface-breaking change I don't want to undertake now. I fixed the bug and moved the class one step closer to the ideal by adding `addElement(...)` and `addElements(...)` methods that require both indices to be supplied at once. In sum this PR does the following:

- I changed all the factory methods to use the `addElement(...)` and `addElements(...)` methods, which fixes the bug, and sets a pattern that makes it less likely a modifier of this class will do the wrong thing.
- I added documentation to discourage a future modifier from manipulating the lists individually
- I added tests that would have caught the missing functionality.

Fixes #1272.

### DIFF
--- a/src/serac/numerics/functional/domain.hpp
+++ b/src/serac/numerics/functional/domain.hpp
@@ -132,6 +132,8 @@ struct Domain {
 
   void addElement(int geom_id, int elem_id, mfem::Geometry::Type element_geometry);
 
+  void addElements(const std::vector<int>& geom_id, const std::vector<int>& elem_id, mfem::Geometry::Type element_geometry);
+
   /// @brief get mfem degree of freedom list for a given FiniteElementSpace
   mfem::Array<int> dof_list(mfem::FiniteElementSpace* fes) const;
 };
--- a/src/serac/numerics/functional/domain.hpp
+++ b/src/serac/numerics/functional/domain.hpp
@@ -130,8 +130,6 @@ struct Domain {
     exit(1);
   }
 
-  void addBoundaryElement(int geom_id, int elem_id, mfem::Geometry::Type element_geometry);
-
   void addElement(int geom_id, int elem_id, mfem::Geometry::Type element_geometry);
 
   /// @brief get mfem degree of freedom list for a given FiniteElementSpace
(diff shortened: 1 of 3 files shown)


src/serac/numerics/functional/domain.hpp
```diff
@@ -132,6 +155,21 @@ struct Domain {
 
   /// @brief get mfem degree of freedom list for a given FiniteElementSpace
   mfem::Array<int> dof_list(mfem::FiniteElementSpace* fes) const;
+
+  /// @brief Add an element to the domain
+  ///
+  /// This is meant for internal use on the class. Prefer to use the factory
```
tupek2, line 162: if this is for internal use, is it marked private?

tupek2 (approved):

src/serac/numerics/functional/domain.hpp
samuelpmishLLNL: @tupek2 and I briefly talked about this offline and we decided to keep it public for now (mostly out of convenience)

samuelpmishLLNL (approved):

PR #1277, opened 2024-11-29
Hide data in `Domain`

btalamini:
Following up on Mike's comment: this changes `Domain` from a struct to a class, with the expected hiding of internal data.

### DIFF
--- a/src/serac/numerics/functional/domain.cpp
+++ b/src/serac/numerics/functional/domain.cpp
@@ -475,8 +475,18 @@ Domain EntireBoundary(const mfem::Mesh& mesh)
   }
 }
 
+/// @cond
+using c_iter = std::vector<int>::const_iterator;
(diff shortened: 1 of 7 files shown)


tupek2 (commented):
Looks good to me, seems like we need to get that new formatting in.  Its going crazy on these files.

btalamini:
> Looks good to me, seems like we need to get that new formatting in. Its going crazy on these files.

I agree. In particular, the forced alignment of the right side of `=` assignments triggers changes to lines that weren't touched by the PR, making the PR bigger, and more difficult to review, since you have to read closely to see if the change is substantial or style.

samuelpmishLLNL (approved):

btalamini:
Let's hold this back until I get the boundary condition PR in. I'll handle the conflicts that emerge in this PR later.

PR #1279, opened 2024-12-03
Fix Smith CI

chapman39:
Smith CI has been running only Serac tests due to an issue with using symlinks! This MR attempts to resolve this.

EDIT: Turns out the fix was within Smith. The remaining changes here are minor cleanups.

### DIFF
--- a/scripts/llnl/common_build_functions.py
+++ b/scripts/llnl/common_build_functions.py
@@ -637,26 +637,8 @@ def get_build_dir(prefix, host_config):
     return pjoin(prefix, "build-" + host_config_root)
 
 
-_repo_dir = ""
 def get_repo_dir():
-    """ Return absolute path to Smith or Serac repo """
-    # Only set var once
-    global _repo_dir
-    if not _repo_dir:
-        # Set repo dir to be relative to this script
-        _repo_dir = os.path.abspath(pjoin(get_script_dir(), "../.."))
-
-        # Check the parent dir for an uberenv config with a package_name "smith".
-        # This means Serac is a submodule. Change the repo dir to be parent dir (i.e.
-        # Smith repo dir)
-        smith_uberenv_config_path = pjoin(_repo_dir, "../.uberenv_config.json")
-        if os.path.exists(smith_uberenv_config_path):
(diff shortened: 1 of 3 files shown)


tupek2 (approved):

white238 (approved):

white238:
This being broken in this way actually has more implications than we think. Anywhere that `get_repo_dir` is called needs to be reviewed or that function needs to be fixed.

white238:
Good catch btw....

chapman39:
The previous solution didn't solve the problem all the way, since there were issues with `get_project_name()` returning serac instead of smith.

This new solution modified `get_repo_dir()` to find the smith dir if it's a serac submodule. The only catch with this solution is that you won't be able to build and test only serac if it's a submodule of smith.

scripts/llnl/common_build_functions.py
```diff
@@ -637,9 +637,26 @@ def get_build_dir(prefix, host_config):
...
+    global _repo_dir
+    if not _repo_dir:
+        # Set repo dir to be relative to this script
+        _repo_dir = os.path.abspath(pjoin(get_script_dir(), "../.."))
+
+        # Check the parent dir for an uberenv config with a package_name "smith".
+        # This means Serac is a submodule. Change the repo dir to be parent dir (i.e.
+        # Smith repo dir)
+        smith_uberenv_config_path = pjoin(_repo_dir, "../.uberenv_config.json")
+        if os.path.exists(smith_uberenv_config_path):
```
white238, line 653: Should use `get_project_name()` here...

scripts/llnl/common_build_functions.py
```diff
@@ -637,9 +637,26 @@ def get_build_dir(prefix, host_config):
...
+            with open(smith_uberenv_config_path) as json_file:
```
white238: Should use get_project_name() here...

chapman39: The problem with that is `get_project_name()` depends on `get_repo_dir()`. I could move the `.uberenv_config.json` query to a separate function though.

white238 (approved):

PR #1287, opened 2024-12-06
Transition Azure Pipelines to Github Actions

### LINKED ISSUE #1265: Switch Azure CI to GitHub Actions
Azure has some problems with

* container size (CUDA containers take 97% of allowed space) as well as
* queuing jobs (too many jobs on the queue from branches, not PRs, causes 4+ hour wait times).

So moving to GitHub Actions may resolve those issues.

chapman39:
Migrates Azure Pipelines to GitHub Actions.

Some benefits:
* You can now cancel a job at any point.
* More storage (14GB instead of Azure's 10GB) (sources: [azure](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#capabilities-and-limitations), [github actions](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)).
* Jobs will no longer run on push. They will only run if there is an update to a pull request.
* Previous of a PR jobs are automatically canceled if an update has been made.
* Fails fast. If one job fails, the entire pipeline will cancel.
* You can re-try failed jobs.

Things todo that I can't do:
- [ ] Make GitHub Actions required to pass before merging into develop https://stackoverflow.com/a/60779334
- [ ] Remove Azure pipeline `LLNL.serac`

fixes #1265


chapman39:
alternative to hacky solution for setting docker image name vars: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#creating-configuration-variables-for-a-repository

but this would require users to access admin settings every time docker images are updated.

chapman39:
I'm not sure how to remove the azure pipeline `LLNL.serac` from the ci checks.

tupek2 (approved):
Chris should weigh in.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,105 @@
+name: Unit Tests
+
+on:
+  pull_request:
+  push:
+    branches:
+    - main
```
white238:
```suggestion
    - develop
```

Our `main` branch isn't really used atm.

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+
+# Cancel previous jobs if an update has been made to the pull request
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
white238:
```suggestion
  cancel-in-progress: ${{ github.ref != 'refs/heads/develop' }}
```

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,105 @@
+name: Unit Tests
```
white238: Better name here... "Build and Test", "CI Tests"??

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+env:
+  CLANG_DOCKER_IMAGE: seracllnl/tpls:clang-14_08-15-24_21h-49m
+  CUDA_DOCKER_IMAGE: seracllnl/tpls:cuda-12_08-15-24_21h-49m
+  GCC_DOCKER_IMAGE: seracllnl/tpls:gcc-13_08-15-24_21h-49m
+
+jobs:
+  # Hacky solution to reference env variables outside of `run` steps https://stackoverflow.com/a/74217028
```
white238: Good job figuring this out but man is this wonky... The fact they had to write this page to explain contexts and didn't fix it... sigh.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#context-availability

.github/workflows/unit-tests.yml
```diff
@@ -0,0 +1,105 @@
...
+            --verbose \
+            --host-config ${{ matrix.config.host_config }} \
+            --extra-cmake-options "${{ matrix.config.cmake_opts }} -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}" \
+            --jobs 4
+      - name: Upload Test Results
+        uses: actions/upload-artifact@v4
+        with:
+          name: Test Results ${{ matrix.config.host_config }} ${{ matrix.build_type }} ${{ matrix.config.cmake_opts }}
+          path: "**/Test.xml"
+          if-no-files-found: ignore # Missing test file is expected on CUDA builds
```
white238: What happens if the non-CUDA builds don't generate files?

.github/workflows/unit-tests.yml
chapman39: there is a better way using variable contexts but it requires admin access to set those variables. and they would be somewhere in github settings, rather than in the yml.

.github/workflows/unit-tests.yml
chapman39: it would be ignored. we might be able to use if statements or revert to using the dummy xml file

.github/workflows/unit-tests.yml
chapman39: i decided not to name it "build and test", since that's the name of one of the jobs (build_and_test and check_code). ill switch to CI Tests if that works with you

.github/workflows/unit-tests.yml
chapman39: ah right thanks for catching that

.github/workflows/unit-tests.yml
white238: This is fine. If we put them in the github settings then the updates cannot be reviewed in PRs

.github/workflows/unit-tests.yml
white238: i would prefer either of those. This makes the base case allow no tests to be run without reporting.

btalamini (approved):

PR #1288, opened 2024-12-10
Tupek/trust region subspace solver

tupek2:


src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +438,93 @@ class TrustRegion : public mfem::NewtonSolver {
...
-        return;
-      } else if (zzNp1 > (trSize * trSize)) {
-        // mfem::out << "step outside trust region.\n";
+      const bool go_to_boundary = curvature <= 0 || zzNp1 > trSize * trSize;
+      if (go_to_boundary) {
         projectToBoundaryWithCoefs(z, d, trSize, zz, zd, dd);
-        results.interior_status = TrustRegionResults::Status::OnBoundary;
+        if (curvature <= 0) {
+          if (print_options.iterations || print_options.warnings) {
+            mfem::out << "negative curvature direction found\n";
```
tupek2: remove

src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +438,93 @@ class TrustRegion : public mfem::NewtonSolver {
...
     z          = 0.0;
     double zz  = 0.;
     double rPr = Dot(rCurrent, Pr);
     double zd  = 0.0;
     double dd  = Dot(d, d);
 
+    // std::cout << "initial energy = " << computeEnergy(r0, hess_vec_func, z) << std::endl;
+
     for (cgIter = 1; cgIter <= settings.max_cg_iterations; ++cgIter) {
+      // check if this is a decent direction
```
tupek2: spelling

src/serac/numerics/solver_config.hpp
```diff
@@ -396,6 +396,16 @@ struct NonlinearSolverOptions {
   /// Scaling for the initial trust region size
   double trust_region_scaling = 0.1;
 
+  /// Specify subspace option
+  /// 0: no subspace solve
+  /// 1: subspace solve only when indefinite system found
+  /// 2: subspace solve only when indefinite system of step to boundary
```
tupek2: 'or' step to boundary

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+  /// @brief copy constructor
+  /// @param a matrix
+  DenseMat(const Mat& a) : A(a) {}
+
+  /// @brief constructor
+  /// @param a matrix
+  DenseMat(const DenseMat& a)
+  {
+    MatDuplicate(a.A, MAT_COPY_VALUES, &A);
+    MatCopy(a.A, A, SAME_NONZERO_PATTERN);
```
btalamini: I'm sure you looked into this, but just to check: is this redundant? Glancing at the PetsC docs, it seems like `MatDuplicate` with `MAT_COPY_VALUES` ought to copy the matrix. Why the `MatCopy`?

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  /// @brief assignment
+  /// @param a matrix
+  DenseMat& operator=(const DenseMat& a)
+  {
+    MatCopy(a.A, A, SAME_NONZERO_PATTERN);
```
btalamini: According to https://petsc.org/release/manualpages/Mat/MatCopy/, `A` must already have the same nonzero pattern as `a.A` or this will crash. I think this will be a bug if `*this` already has a matrix and it's a different nonzero pattern.

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+  {
+    double val;
+    MatGetValue(A, i, j, &val);
+    return val;
+  }
+
+  /// @brief set value
+  void setValue(int i, int j, double val) { MatSetValues(A, 1, &i, 1, &j, &val, INSERT_VALUES); }
+
+  /// @brief scalar multiply
```
btalamini:
```suggestion
  /// @brief matrix-vector multiply
```

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+/// Dense Vector class which wraps petsc vector for the case of a SeqDense vector (on 1 processor)
+struct DenseVec {
+  /// @brief constructor
+  DenseVec(const Vec& vin) : v(vin) {}
+
+  /// @brief constructor
+  DenseVec(const DenseVec& vin)
+  {
+    VecDuplicate(vin.v, &v);
+    VecCopy(vin.v, v);
```
btalamini, line 103: Same question as above: are both these calls necessary?

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+    return val;
+  }
+
+  /// @brief index into
+  double operator[](size_t i) const
+  {
+    int    i_int = static_cast<int>(i);
+    double val;
+    VecGetValues(v, 1, &i_int, &val);
+    return val;
```
btalamini:
```suggestion
    return (*this)[int(i)];
```

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  /// @brief set value
+  void setValue(int i, double val) { VecSetValues(v, 1, &i, &val, INSERT_VALUES); }
+
+  /// @brief set value
+  void setValue(size_t i, double val)
+  {
+    int i_int = static_cast<int>(i);
+    VecSetValues(v, 1, &i_int, &val, INSERT_VALUES);
```
btalamini:
```suggestion
    setValue(int(i), val);
```

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+
+  /// @brief convert to standard vector
+  std::vector<double> getValues() const
+  {
+    size_t              sz = static_cast<size_t>(size());
+    std::vector<double> vout(sz);
+    std::vector<int>    allints(sz);
+    for (size_t i = 0; i < sz; ++i) {
+      allints[i] = static_cast<int>(i);
+    }
```
btalamini, line 198: I see this in a few places. Maybe make a `range` function?

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+
+  /// petsc vector
+  Vec v;
+};
+
+/// @brief matrix vector multiply
+DenseVec DenseMat::operator*(const DenseVec& v) const
+{
+  Vec out;
+  VecDuplicate(v.v, &out);
```
btalamini: Get the size of `out` from the number of rows in the matrix (or restrict your `DenseMat` to square matrices, if that's what you want).

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+
+/// @brief add a scalar to a vector
+DenseVec operator+(const DenseVec& a, double b)
+{
+  Vec c;
+  VecDuplicate(a.v, &c);
+  VecSet(c, b);
+  VecAXPY(c, 1.0, a.v);
+  return c;
+}
```
btalamini, line 262: I think it would be good to also implement the symmetric operation. Here you go:

```c++
DenseVec operator+(double b, const DenseVec& a) {
    return a + b;
}
```

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,367 @@
...
+  EPSSetDimensions(eps, isize, PETSC_DETERMINE, PETSC_DETERMINE);
+  EPSSetFromOptions(eps);
+
+  EPSSolve(eps);
+
+  EPSType type;
+  EPSGetType(eps, &type);
+  // PetscPrintf(PETSC_COMM_SELF," Solution method: %s\n\n",type);
+  EPSGetDimensions(eps, &jsize, NULL, NULL);
+  SLIC_WARNING_IF(isize != jsize, "The requested and achieved number of eigenvalues do not match eigh call");
```
btalamini: You checked this already, right? Though I suppose this isn't hurting anyone.

src/serac/numerics/equation_solver.cpp
```diff
@@ -245,23 +252,32 @@ struct TrustRegionResults {
   }
 
   /// enumerates the possible final status of the trust region steps
-  enum class Status
```
btalamini: I'm curious - did the fact that it was an `enum class` cause problems somewhere?

src/serac/numerics/equation_solver.cpp
```diff
@@ -317,13 +333,74 @@ class TrustRegion : public mfem::NewtonSolver {
...
+
+    std::vector<const mfem::Vector*> H_directions;
+    for (auto& Hd : Hds) {
+      H_directions.emplace_back(Hd);
+    }
+    for (auto& H_left : H_left_mosts) {
+      H_directions.emplace_back(H_left.get());
+    }
+
+    std::tie(directions, H_directions) = removeDependantDirections(directions, H_directions);
```
btalamini: typo 
```suggestion
    std::tie(directions, H_directions) = removeDependentDirections(directions, H_directions);
```

src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +438,93 @@ class TrustRegion : public mfem::NewtonSolver {
     }
   }
 
+  /// compute the energy of the linearized system for a given solution vector z
+  template <typename HessVecFunc>
+  double computeEnergy(const mfem::Vector& r_local, const HessVecFunc& H, const mfem::Vector& z) const
+  {
+    double       rz = Dot(r_local, z);
+    mfem::Vector tmp(r_local);
+    tmp *= 0.0;
```
btalamini: I'm unsure of the intent here. Is this essentially trying to do `tmp = zeros_like(r_local)`? If so, you could write 
```c++
mfem::Vector tmp(r_local.size());
tmp = 0.0;
```
or

```c++
mfem::Vector tmp;
tmp.SetDataAndSize(0.0, r_local.Size());
```

src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +438,93 @@ class TrustRegion : public mfem::NewtonSolver {
...
     auto& cgIter = results.cg_iterations_count;
     auto& d      = results.d;
     auto& Pr     = results.Pr;
-    auto& Hd     = results.Hd;
+    auto& Hd     = results.H_d;
 
     const double cg_tol_squared = settings.cg_tol * settings.cg_tol;
 
-    if (Dot(r0, r0) <= cg_tol_squared && cgIter >= settings.min_cg_iterations) {
+    if (Dot(r0, r0) <= cg_tol_squared && settings.min_cg_iterations == 0) {
```
btalamini, line 479: My preference: I would like a message when this somewhat unusual exit condition is reached. But you know this best, use your own judgment.

src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +438,93 @@ class TrustRegion : public mfem::NewtonSolver {
...
     }
 
     rCurrent = r0;
     precond(rCurrent, Pr);
-    d = 0.0;
-    add(d, -1.0, Pr, d);  // d = -Pr
+
+    // d = -Pr
+    d = Pr;
+    d *= -1.0;
```
btalamini, line 490: nice, much clearer

src/serac/numerics/equation_solver.cpp
btalamini: A descent direction *is* a decent direction :)

src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +438,93 @@ class TrustRegion : public mfem::NewtonSolver {
...
+        d *= -1;
+        results.interior_status = TrustRegionResults::Status::NonDescentDirection;
+      }
+
       hess_vec_func(d, Hd);
       const double curvature = Dot(d, Hd);
       const double alphaCg   = curvature != 0.0 ? rPr / curvature : 0.0;
 
-      auto& zPred = Hd;  // re-use Hd, this is where bugs come from
+      auto& zPred = Pr;  // re-use Pr, carefully
```
btalamini: Why alias Pr like this? I don't follow the purpose.

src/serac/numerics/equation_solver.cpp
```diff
@@ -558,8 +658,8 @@ class TrustRegion : public mfem::NewtonSolver {
       double cauchyPointNormSquared = tr_size * tr_size;
       trResults.reset();
 
-      hess_vec_func(r, trResults.Hd);
-      const double gKg = Dot(r, trResults.Hd);
+      hess_vec_func(r, trResults.H_d);
+      const double gKg = Dot(r, trResults.H_d);
```
btalamini, line 681: This question is probably stupid, but it's been a while since I've waded through this algorithm. Is `H_d` being re-used to hold the Hessian times the residual here?

src/serac/numerics/tests/test_trust_region_solver.cpp
```diff
@@ -0,0 +1,166 @@
...
+    AstatePtrs.push_back(&Astates[i]);
+  }
+
+  double delta                           = 0.001;
+  auto [sol, leftvecs, leftvals, energy] = serac::solveSubspaceProblem(states, AstatePtrs, b, delta, 1);
+
+  serac::FiniteElementState serac_sol(b);
+  serac_sol = sol;
+
+  ASSERT_NEAR(std::sqrt(serac::innerProduct(serac_sol, serac_sol)), delta, 1e-12);
```
btalamini: You usually want to use EXPECT_ (instead of ASSERT_) for tests, since it's not fatal and lets later checks happen if one fails. The ASSERT_ ones are for when you need to end the program. It doesn't really matter here since there are no further checks, but people tend to copy and paste tests, so best not to introduce the pattern.

src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +438,93 @@ class TrustRegion : public mfem::NewtonSolver {
...
     for (cgIter = 1; cgIter <= settings.max_cg_iterations; ++cgIter) {
+      // check if this is a decent direction
+      if (Dot(d, rCurrent) > 0) {
+        d *= -1;
+        results.interior_status = TrustRegionResults::Status::NonDescentDirection;
+      }
+
       hess_vec_func(d, Hd);
       const double curvature = Dot(d, Hd);
       const double alphaCg   = curvature != 0.0 ? rPr / curvature : 0.0;
```
btalamini, line 509: Ok, this prevents getting infinity or nan, but setting `alphaCg` in that case to zero means that the solver will stagnate at this point. Should we set it to 1, and then the "go to the boundary" step at least tries to move to a better point?

btalamini (approved):

src/serac/numerics/dense_petsc.hpp
tupek2: I added this because the VectorDuplicate does not behave this way. I tested, you are correct, the MatCopy is not needed here.

src/serac/numerics/dense_petsc.hpp
tupek2: This was not needed yet, deleted.

src/serac/numerics/dense_petsc.hpp
tupek2: yes

src/serac/numerics/dense_petsc.hpp
tupek2: not today!

src/serac/numerics/equation_solver.cpp
tupek2: yes, this is a potentially dangerous reuse of memory.

src/serac/numerics/equation_solver.cpp
tupek2: reusing memory.  it looks safe enough to me, but confusing potentially.

src/serac/numerics/equation_solver.cpp
tupek2: its harder to print enum classes.  I dont have prints anymore, so I can put it back.

src/serac/numerics/equation_solver.cpp
tupek2: following the logic down a few lines, this is indeed what actually happens already.  the alphaCg is used for a trial solution.  If the curvature is <= 0, we then do the project to boundary stuff.

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,348 @@
...
+    double eigenvalue;
+    EPSGetEigenpair(eps, i, &eigenvalue, PETSC_NULLPTR, eigenvectors[static_cast<size_t>(i)].v, PETSC_NULLPTR);
+    eigenvalues.setValue(i, eigenvalue);
+  }
+
+  EPSDestroy(&eps);
+  return std::make_pair(std::move(eigenvalues), std::move(eigenvectors));
+}
+
+#endif  // SERAC_USE_SLEPC
```
chapman39:
```suggestion
#endif  // SERAC_USE_SLEPC

```

src/serac/numerics/equation_solver.cpp
```diff
@@ -361,64 +446,93 @@ class TrustRegion : public mfem::NewtonSolver {
...
 
-      auto& zPred = Hd;  // re-use Hd, this is where bugs come from
+      auto& zPred = Pr;  // re-use Pr, carefully
```
chapman39: may be beneficial to explain carefully how or at least why

src/serac/numerics/solver_config.hpp
```diff
@@ -396,6 +396,16 @@ struct NonlinearSolverOptions {
   /// Scaling for the initial trust region size
   double trust_region_scaling = 0.1;
 
+  /// Specify subspace option
+  /// 0: no subspace solve
+  /// 1: subspace solve only when indefinite system found
+  /// 2: subspace solve only when indefinite system or step to boundary
+  /// 3: subspace solve always
+  int subspace_option = 0;
```
chapman39: would it be worth making this a enum?

chapman39:
im not too familiar with petsc/ mfem, but could a lot of the petsc wrapper stuff use mfem's petsc wrapper classes? https://github.com/mfem/mfem/blob/master/linalg/petsc.hpp ?

also https://github.com/mfem/mfem/blob/master/linalg/slepc.hpp

src/serac/numerics/trust_region_solver.hpp
```diff
@@ -0,0 +1,45 @@
...
+std::tuple<mfem::Vector, std::vector<std::shared_ptr<mfem::Vector>>, std::vector<double>, double> solveSubspaceProblem(
+    const std::vector<const mfem::Vector*>& directions, const std::vector<const mfem::Vector*>& A_directions,
+    const mfem::Vector& b, double delta, int num_leftmost);
+
+std::pair<std::vector<const mfem::Vector*>, std::vector<const mfem::Vector*>> removeDependentDirections(
+    std::vector<const mfem::Vector*> directions, std::vector<const mfem::Vector*> A_directions);
+
+}  // namespace serac
+
+#endif  // SERAC_USE_SLEPC
```
chapman39:
```suggestion
#endif  // SERAC_USE_SLEPC

```

src/serac/numerics/trust_region_solver.cpp
```diff
@@ -0,0 +1,481 @@
...
+    directions_new.push_back(directions[kept_and_j.second]);
+    A_directions_new.push_back(A_directions[kept_and_j.second]);
+  }
+
+  return std::make_pair(directions_new, A_directions_new);
+}
+
+}  // namespace serac
+
+#endif  // SERAC_USE_SLEPC
```
chapman39:
```suggestion
#endif  // SERAC_USE_SLEPC

```

tupek2:
The mfem wrappers are intended for 'parallel' across processor things.  The new wrapper is for local, small, dense things.  In general, many coding experts recommend creating internal abstractions to external library wrappers anyways, but in this case, the mfem one doesn't even really fit the purpose.

src/serac/numerics/solver_config.hpp
tupek2: I think I like that.  Self documenting, less comments.

src/serac/numerics/equation_solver.cpp
tupek2: Yes

chapman39 (approved):
thanks @tupek2

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,348 @@
...
+  EPSSetProblemType(eps, EPS_HEP);
+  EPSSetWhichEigenpairs(eps, EPS_SMALLEST_REAL);
+  EPSSetDimensions(eps, isize, PETSC_DETERMINE, PETSC_DETERMINE);
+  EPSSetFromOptions(eps);
+
+  EPSSolve(eps);
+
+  EPSType type;
+  EPSGetType(eps, &type);
+  // PetscPrintf(PETSC_COMM_SELF," Solution method: %s\n\n",type);
```
white238: delete?

src/serac/numerics/dense_petsc.hpp
```diff
@@ -0,0 +1,348 @@
...
+  double n;
+  VecNorm(a.v, NORM_2, &n);
+  return n;
+}
+
+/// @brief computes the eigenvectors and eigenvalues of a dense symmetric matrix
+auto eigh(const DenseMat& Adense)
+{
+  auto [isize, jsize] = Adense.size();
+  SLIC_ERROR_IF(isize != jsize, "Eig must be called for symmetric matrices");
```
white238, line 314:
```suggestion
  SLIC_ERROR_IF(isize != jsize, "'eigh' must be called for symmetric matrices");
```
?

src/serac/numerics/equation_solver.cpp
```diff
@@ -317,13 +344,76 @@ class TrustRegion : public mfem::NewtonSolver {
...
+
+    left_mosts.clear();
+    for (auto& lv : leftvecs) {
+      left_mosts.emplace_back(std::move(lv));
+    }
+
+    double base_energy     = computeEnergy(g, hess_vec_func, z);
+    double subspace_energy = computeEnergy(g, hess_vec_func, sol);
+
+    // check that leftmost approximates its eigenvalue
```
white238: dead code or needs a note why it staying commented out

src/serac/numerics/tests/test_trust_region_solver.cpp
```diff
@@ -0,0 +1,166 @@
+#include <gtest/gtest.h>
```
white238, line 7: Copyright missing

src/serac/numerics/trust_region_solver.cpp
```diff
@@ -0,0 +1,481 @@
...
+  BVOrthogonalize(Q, R);
+
+  return std::make_pair(Q, DenseMat(R));
+}
+
+/**
+ * @brief compute the quadratic energy from small dense matrices and vectors
+ * @param A The stiffness matrix
+ * @param b The rhs vector
+ * @param x The current solution vector
```
white238, line 197: Does the doxygen check not throw a warning/error on missing the return parameter documentation?

src/serac/numerics/dense_petsc.hpp
tupek2: yep

src/serac/numerics/equation_solver.cpp
tupek2: deleted

src/serac/numerics/trust_region_solver.cpp
tupek2: it does not seem to error.  Presumably our functions should be named such that the return is often obvious.

src/serac/numerics/tests/test_trust_region_solver.cpp
tupek2: Added

PR #1291, opened 2024-12-11
Tupek/cutback bcs when solve fails

tupek2:


src/serac/physics/solid_mechanics.hpp
```diff
@@ -27,6 +27,16 @@
 
 namespace serac {
 
+/// @brief simple utility for computing matrix norm
```
ebchin, line 30: Can we put this somewhere else?  Though not sure if we have a good place for it now.

ebchin: It looks like this isn't being used anywhere.  Maybe we can take it out?

ebchin (approved):
Looks good @tupek2

btalamini (approved):

Issue #1282, opened 2024-12-04
Create a CMake macro to handle project specific options


white238:
`cmake_dependent_option` just doesn't work like I always think it should so we should create our own called `serac_project_specific_option` that encapsulates the behavior we want.

For example:

* `ENABLE_BENCHMARKS` is a global BLT created option, it is used for multiple projects. Allow a user to turn that on and honor it
* Unless the project specific one is set `SERAC_ENABLE_BENCHMARKS` then honor that one.
* Always create a `SERAC_ENABLE_BENCHMARKS` because the rest of our build system is guarded on that

chapman39:
personally i like the way `cmake_dependent_option` works. pretty much, you can't set the project-specific option without ensuring the general one is true. so if you want to enable benchmarks, we should just set `ENABLE_BENCHMARKS`. it will automatically enable `SERAC_ENABLE_BENCHMARKS`. 

let me know your thoughts @white238 @samuelpmish

chapman39:
from a discussion with @tupek2 : it would be nice to somehow set `-DSMITH_ENABLE_BENCHMARKS=ON` and only set that project's benchmarks... im not sure if that would work though, since we need to also set `-DENABLE_BENCHMARKS=ON` for BLT to setup benchmarking properly.

another thing to consider is these options need to be set properly before BLT is called, which would mean moving this call https://github.com/LLNL/serac/blob/develop/CMakeLists.txt#L127 which may complicate things in the case BLT changes/ adds options, we need to match BLT.


```cmake

include(${BLT_SOURCE_DIR}/SetupBLT.cmake)

#------------------------------------------------------------------------------
# Setup Macros/Basics
#------------------------------------------------------------------------------

include(${PROJECT_SOURCE_DIR}/cmake/SeracMacros.cmake)
include(${PROJECT_SOURCE_DIR}/cmake/SeracBasics.cmake) # this would need to be called before setupBLT
include(${PROJECT_SOURCE_DIR}/cmake/SeracCompilerFlags.cmake)
```

PR #1292, opened 2024-12-11
Provide a way to disable Serac benchmarks

chapman39:
@tupek2 is adding Smith benchmarks. Smith needs a way to disable Serac benchmarks while running Smith's.

Serac benchmarks should not be run within Smith, since because Smith doesn't update Serac very often, so they won't be as accurate.

Eventually we should look into better cmake options https://github.com/LLNL/serac/issues/1282, but for now `cmake_dependent_option` should be re-added, imo.

### DIFF
--- a/cmake/SeracBasics.cmake
+++ b/cmake/SeracBasics.cmake
@@ -43,9 +43,10 @@ endif()
 
 option(SERAC_ENABLE_PROFILING "Enable profiling functionality" OFF)
 
-if (ENABLE_BENCHMARKS)
-    set(SERAC_ENABLE_BENCHMARKS ON)
-endif()
+# ENABLE_BENCHMARKS must be ON in order to modify SERAC_ENABLE_BENCHMARKS.
+# SERAC_ENABLE_BENCHMARKS will automatically be set to ON if ENABLE_BENCHMARKS is ON.
+# SERAC_ENABLE_BENCHMARKS is an option to allow external projects to disable Serac benchmarks while enabling theirs.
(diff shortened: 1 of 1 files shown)


tupek2:
This seems reasonable to me.  There is a clear pattern.  Ideally, all our options can be like this, so I don't even have to specify project specific ones (DENABLE_CODEVELOP, for example doesnt exist, so we have to do it per project).

tupek2 (approved):

tupek2:
Thinking about this a bit... I think people get confused because
ENABLE_SERAC_BENCHMARKS=ON does not turn on ENABLE_BENCHMARKS.
ENABLE_BENCHMARKS is like the "house master value" in this design.  It must be turned on for the local values to turn on/off.
We can either flip the logic, so ENABLE_BENCHMARKS just changes the local defaults, which can then always be turned on/off independently.  Or we can keep the current design, but if we do, it might make it clearer if we change the local names to:
DISABLE_SERAC_BENCHMARKS=ON  (all these disables would default to false) to limit the confusion potential a bit.

Then its clearer that ENABLE_BENCHMARKS turns them on for all projects, and per project we can disable them if needed.

chapman39:
> DENABLE_CODEVELOP, for example doesnt exist, so we have to do it per project

this variable doesn't exist, because it's not an option in BLT. here is a list of all blt options https://github.com/LLNL/blt/blob/develop/cmake/BLTOptions.cmake

white238 (approved):

PR #1301, opened 2024-12-20
Fix a bunch of small errors.

tupek2:
Petsc was printing lots of errors.  This makes it a bit more quiet.  Also, a possible indexing error is fixed if no leftmost eigenvectors are asked for.

### DIFF
--- a/src/serac/numerics/dense_petsc.hpp
+++ b/src/serac/numerics/dense_petsc.hpp
@@ -78,6 +78,7 @@ struct DenseMat {
     return false;
   }
 
+  /// @brief  reassemble petsc dense matrix after values have been modified
   void reassemble() {
     MatAssemblyBegin(A, MAT_FINAL_ASSEMBLY);
     MatAssemblyEnd(A, MAT_FINAL_ASSEMBLY);
--- a/src/serac/numerics/dense_petsc.hpp
+++ b/src/serac/numerics/dense_petsc.hpp
@@ -16,7 +16,13 @@ struct DenseMat {
 
   /// @brief constructor
   /// @param a matrix
-  DenseMat(const DenseMat& a) { MatDuplicate(a.A, MAT_COPY_VALUES, &A); }
+  DenseMat(const DenseMat& a) 
+  {
+    MatDuplicate(a.A, MAT_COPY_VALUES, &A);
+    MatCopy(a.A, A, SAME_NONZERO_PATTERN); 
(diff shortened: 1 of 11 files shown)


white238 (approved):

src/serac/numerics/equation_solver.cpp
```diff
@@ -398,8 +418,9 @@ class TrustRegion : public mfem::NewtonSolver {
     double subspace_energy = computeEnergy(g, hess_vec_func, sol);
 
     if (print_options.iterations || print_options.warnings) {
+      double leftval = leftvals.size() ? leftvals[0] : 1.0;
       mfem::out << "Energy using subspace solver from: " << base_energy << ", to: " << subspace_energy << " / "
-                << energy_change << ".  Min eig: " << leftvals[0] << std::endl;
+                << energy_change << ".  Min eig: " << leftval << std::endl;
```
btalamini, line 423: It looks like the minimum eigenvalue estimate will be reported as 1.0 when no min eigenvalues estimates have been requested.  I'm fine with this for now, since it's really an inconsistent request from the user, but it is misleading. An option to consider is to check if `leftvals.size() == 0` once at the beginning of the step and echo a warning saying that the min eigenvalue will be reported as 1.0 in this case.

src/serac/numerics/dense_petsc.hpp
```diff
@@ -16,7 +16,13 @@ struct DenseMat {
...
   /// @brief constructor
   /// @param a matrix
-  DenseMat(const DenseMat& a) { MatDuplicate(a.A, MAT_COPY_VALUES, &A); }
+  DenseMat(const DenseMat& a)
+  {
+    MatDuplicate(a.A, MAT_COPY_VALUES, &A);
+    MatCopy(a.A, A, SAME_NONZERO_PATTERN);
+    MatAssemblyBegin(A, MAT_FINAL_ASSEMBLY);
+    MatAssemblyEnd(A, MAT_FINAL_ASSEMBLY);
+  }
```
btalamini, line 25: You've done the testing, so I trust you on this. If you actually wanted me to double check, let me know. Otherwise mark as resolved.

btalamini (approved):

src/serac/numerics/dense_petsc.hpp
tupek2: The assembly are definitely needed.  The Copy, maybe not.  But I'm sick of trying to guess what PETSc is going to do, best be safe.  It costs nothing in practice.

PR #1302, opened 2024-12-23
Only error when invalid maps are passed to base physics methods

tupek2:
Allow some base physics interfaces to be used in cases where maps are empty to allow end users to use the common interface in more cases.

### DIFF
--- a/src/serac/physics/base_physics.hpp
+++ b/src/serac/physics/base_physics.hpp
@@ -368,7 +368,7 @@ public:
    */
   virtual void setAdjointLoad(std::unordered_map<std::string, const serac::FiniteElementDual&> string_to_dual)
   {
-    if (string_to_dual.size()) {
+    if (!string_to_dual.empty()) {
       SLIC_ERROR_ROOT(
           axom::fmt::format("Failed to setAdjointLoad.  Adjoint analysis not defined for physics module {}", name_));
     }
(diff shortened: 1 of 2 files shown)


src/serac/physics/base_physics.hpp
```diff
@@ -363,19 +363,29 @@ class BasePhysics {
...
   /**
    * @brief Set the loads for the adjoint reverse timestep solve
+   * @param string_to_dual An unorder map from the state field name string to the finite element adjoint/dual load
+   * The adjoint load is d(qoi)/d(state)
    */
-  virtual void setAdjointLoad(std::unordered_map<std::string, const serac::FiniteElementDual&>)
+  virtual void setAdjointLoad(std::unordered_map<std::string, const serac::FiniteElementDual&> string_to_dual)
   {
-    SLIC_ERROR_ROOT(axom::fmt::format("Adjoint analysis not defined for physics module {}", name_));
+    if (string_to_dual.size()) {
```
btalamini: If you're checking whether the map is populated here, I suggest using the `empty()` method on STL containers, because the intent is clearer.

src/serac/physics/base_physics.hpp
```diff
@@ -363,19 +363,29 @@ class BasePhysics {
...
    * @brief Set the dual loads (dirichlet values) for the adjoint reverse timestep solve
    * This must be called after setAdjointLoad
+   * @param string_to_bc An unorder map from dual name string to finite element adjoint/state boundary condition
+   * The adjoint bc is d(qoi)/d(dual)
    */
-  virtual void setDualAdjointBcs(std::unordered_map<std::string, const serac::FiniteElementState&>)
+  virtual void setDualAdjointBcs(std::unordered_map<std::string, const serac::FiniteElementState&> string_to_bc)
   {
-    SLIC_ERROR_ROOT(axom::fmt::format("Adjoint analysis not defined for physics module {}", name_));
+    if (string_to_bc.size()) {
```
btalamini: Same comment.

btalamini (approved):

kswartz92 (approved):

PR #1303, opened 2024-12-27
Move git sha header to better location

chapman39:
I've been testing a serac-only lido build and noticed something weird about the git sha header... in serac, `git_sha.hpp` is being configured into `${PROJECT_BINARY_DIR}/include/serac/infrastructure`, but in smith (which is what lido has been testing against) configures it in an entirely different location: `${CMAKE_BINARY_DIR}/include/serac/infrastructure`.

Smith shouldn't need to configure this file in the first place, so it has been removed in the associated Smith MR.

I've updated Serac and Smith so that this header is being configured to a consistent location: `${CMAKE_BINARY_DIR}/include/serac`

### DIFF
--- a/src/serac/infrastructure/CMakeLists.txt
+++ b/src/serac/infrastructure/CMakeLists.txt
@@ -9,7 +9,7 @@
 # a full rebuild.
 serac_configure_file(
   git_sha.hpp.in
-  ${PROJECT_BINARY_DIR}/include/serac/infrastructure/git_sha.hpp
+  ${CMAKE_BINARY_DIR}/include/serac/git_sha.hpp
  )
 
 set(infrastructure_headers
@@ -17,7 +17,7 @@ set(infrastructure_headers
     accelerator.hpp
     cli.hpp
     debug_print.hpp
-    ${PROJECT_BINARY_DIR}/include/serac/infrastructure/git_sha.hpp
+    ${CMAKE_BINARY_DIR}/include/serac/git_sha.hpp
     initialize.hpp
     input.hpp
     logger.hpp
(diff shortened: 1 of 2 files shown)


chapman39:
I'm not sure if this conflicts with what you were trying to do in this PR @white238 https://github.com/LLNL/serac/pull/637 , but we can talk about it after break.

chapman39:
Please do not update this branch with develop until it's finalized, since Id like to get this change into lido without doing a tpl rebuild.

white238:
We should be using `PROJECT_BINARY_DIR` instead of `CMAKE_BINARY_DIR` in every place we can because it guarantees there will never be a conflict with downstream projects.

This is should also be fixed:

https://github.com/LLNL/serac/blob/a44317babcca0bfabb3b20cb8bd8697b3d9d9ee8/src/serac/infrastructure/CMakeLists.txt#L61

Were there conflicts you seeing or was it inconsistency?

chapman39:
> Were there conflicts you seeing or was it inconsistency?

This is the error I was seeing, since the LiDO build was relying on Smith to configure the git sha header again in a different location.

```
/usr/workspace/meemee/lido-2.0/repo/smith/serac/src/serac/infrastructure/about.cpp:47:10: fatal error: 'serac/infrastructure/git_sha.hpp' file not found
#include "serac/infrastructure/git_sha.hpp"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

chapman39:
> This is should also be fixed:

I think that one fix is all that was needed after all

chapman39:
see https://github.com/LLNL/serac/pull/1304

PR #1305, opened 2025-01-02
Fix compilation error on mac sequoia.

tupek2:


white238 (approved):

samuelpmishLLNL (approved):

src/serac/numerics/functional/functional_qoi.inl
```diff
@@ -396,7 +396,7 @@ private:
     {
     }
 
-    void Mult(const mfem::Vector& x, mfem::Vector& y) const { form_.GradientMult(x, y); }
+    void Mult(const mfem::Vector& x, mfem::Vector& y) const { form_.ActionOfGradient(x, y, which_argument); }
```
samuelpmishLLNL: Does this work?
```suggestion
    void Mult(const mfem::Vector& x, mfem::Vector& y) const { y[0] = form_.ActionOfGradient(x, which_argument); }
```

samuelpmishLLNL: or maybe just delete this Mult method, since this Gradient type isn't an mfem::Operator anyway

tupek2: I went with delete.  Seems it was unused.  Interesting that the newer clang catches that as a warning/error.

btalamini (approved):

PR #1307, opened 2025-01-03
Fixes for builds with no raja or umpire

ebchin:
This PR is in place of the (now closed) #1270.

Instead of `#ifdef`s on the function declaration, I added a `memory.hpp` which holds different host `axom::MemorySpace`s depending on if serac is built with Umpire or not.  This variable is then used in template parameters.


src/serac/infrastructure/CMakeLists.txt
```diff
@@ -41,9 +41,10 @@ set(infrastructure_sources
     terminator.cpp
     )
 
-set(infrastructure_depends axom::inlet axom::fmt axom::cli11 mfem ${serac_device_depends})
+set(infrastructure_depends axom::inlet axom::fmt axom::cli11 camp mfem ${serac_device_depends})
 blt_list_append(TO infrastructure_depends ELEMENTS tribol IF TRIBOL_FOUND)
 blt_list_append(TO infrastructure_depends ELEMENTS caliper adiak::adiak IF SERAC_ENABLE_PROFILING)
+blt_list_append(TO infrastructure_depends ELEMENTS blt::openmp IF SERAC_ENABLE_OPENMP)
```
ebchin, line 50: I don't think this is needed, but the openmp dependency for mfem wasn't picked up through `FindMFEM.cmake`.  Any ideas on how to fix this @white238 @chapman39 ?  This is with clang 16.0.6 on Ubuntu 24.04

white238 (approved):

tupek2 (approved):

src/serac/infrastructure/CMakeLists.txt
chapman39: what happens when you call `blt_print_target_properties(TARGET mfem)`? also can you check to see if mfem was built with `+openmp` in the concretizer tree?

chapman39 (approved):

PR #1308, opened 2025-01-07
Remove deprecation warning. 

tupek2:
Its warnings to errors, so this is essentially disabling the capability.  Unfortunately, we do now have a replacement for it yet, nor a design planned to replace it, as far as I'm aware.  We could consider making it more consistent with the new BC interface, e.g. taking Components an using tensor, instead of mfem::Vector.

### DIFF
--- a/src/serac/physics/solid_mechanics.hpp
+++ b/src/serac/physics/solid_mechanics.hpp
@@ -507,9 +507,8 @@ public:
    *
    * @note This method must be called prior to completeSetup()
    */
-  void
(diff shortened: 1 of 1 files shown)


btalamini:
I would prefer to fix this in a more direct way: change the *compiler* flags so that we don't promote deprecation warnings to errors (but still promote other warnings to errors). I think this is possible. I don't want to remove the deprecation.

btalamini (approved):

white238 (approved):

tupek2:
OK, I think we can consider turn off this warning in the future, and can maybe start doing deprecations at that time?  For now, its going in!

PR #1309, opened 2025-01-08
Update clang style to avoid changing lines abore and below.

tupek2:


white238 (approved):

btalamini (approved):

.clang-format
```diff
@@ -1,7 +1,5 @@
 BasedOnStyle: Google
-AlignConsecutiveAssignments: true
-AlignConsecutiveDeclarations: true
-AccessModifierOffset: -2
```
white238, line 4: I do mildly prefer this aligned how we have it but I will allow myself to be overruled if others feel strongly.  It helps not move large blocks of code inside a class even more.

PR #1310, opened 2025-01-08
Change displacement and velocity setters to use serac tensors

btalamini:
Make the interface consistent with that used for boundary conditions, using callables that take and return `tensor` objects instead of `mfem::Vector`s.

- Setting a `FiniteElementState` from a field function $$f(X)$$ is a general capability, so I moved this code out of the `SolidMechanics` class and into the `FiniteElementState` class.
- I wrote unit tests for this new capability. It turns out there were no unit tests in the `serac/physics/state/` directory, so I had to add build files for this.

### DIFF
--- a/src/serac/physics/solid_mechanics.hpp
+++ b/src/serac/physics/solid_mechanics.hpp
@@ -831,10 +831,10 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
    *
    *   returns: u, the displacement at X.
    */
-  template <typename Callable>
-  void setDisplacement(Callable applied_displacement)
+  template <typename AppliedDisplacementFunction>
+  void setDisplacement(AppliedDisplacementFunction applied_displacement)
   {
-    displacement_.setFromField(applied_displacement);
+    displacement_.setFromFieldFunction(applied_displacement);
   }
 
   /// @overload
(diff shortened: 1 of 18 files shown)


btalamini:
@white238, requesting help with a test. I had to disable a test that verifies an error check works as expected. It uses the `EXPECT_DEATH` gtest macro. It works fine when I launch the executable myself, but fails when the ctest launcher is executing the entire suite. Is there a way to get this unit test of the error check to operate?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -820,30 +820,42 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   *   returns: u, the displacement at X.
    */
-  void setDisplacement(std::function<void(const mfem::Vector& x, mfem::Vector& disp)> disp)
+  template <typename Callable>
+  void setDisplacement(Callable applied_displacement)
   {
-    // Project the coefficient onto the grid function
-    mfem::VectorFunctionCoefficient disp_coef(dim, disp);
-    displacement_.project(disp_coef);
+    displacement_.setFromField(applied_displacement);
```
tupek2: I think field in setFromField will eventually be confusing, as we are like to change the name of FiniteElementState to Field.  Which this is not taking as an argument.  Also, Callable is a bit inconsistent with template naming through Serac so far.   For example, SolidMechanics.hpp has AppliedDisplacementFunction as the template name, which is more descriptive.  Elsewhere, internally, its sometimes single letters.  I feel like Callable in introducing a new concept, and maybe a bit python.

tupek2 (approved):

src/serac/physics/solid_mechanics.hpp
btalamini: Changed to `setFromFieldFunction`.

samuelpmishLLNL (approved):

PR #1319, opened 2025-01-24
Tupek/set from field function ref

tupek2:


btalamini (approved):
Thanks for finding and patching the hole I left.

src/serac/physics/state/finite_element_state.hpp
```diff
@@ -73,7 +74,8 @@ decltype(first_argument_helper(&F::operator())) first_argument_helper(F);
  * Extract type of first argument of a free callable
  */
 template <typename T>
-using first_argument = decltype(first_argument_helper(std::declval<T>()));
+using first_argument = typename std::remove_const<
+    typename std::remove_reference<decltype(first_argument_helper(std::declval<T>()))>::type>::type;
```
chapman39, line 78: what is this doing?

tupek2: finding the type of the first argument to a function, with constness and reference removed if necessary.  This is the fix, the changed test documents the newly supported usage.

tupek2: as an aside, some personal experience suggests the compile time impacts of this kind of template meta-programming is non-trivial.

chapman39: thanks for the clarification

chapman39 (approved):

white238 (approved):

PR #1321, opened 2025-01-29
Put in a recursive strategy to slowly ramp up new boundary conditions…

tupek2:
This should only happen when the solver is failing, especially around the warm-start.


tupek2:
Maybe we will do this in the future.

PR #1328, opened 2025-02-12
Address a few more missing features in base physics.

tupek2:
Add const correctness, fix dual value initialization.

### DIFF
--- a/src/serac/physics/base_physics.hpp
+++ b/src/serac/physics/base_physics.hpp
@@ -1,4 +1,4 @@
-  // Copyright (c) 2019-2024, Lawrence Livermore National Security, LLC and
+// Copyright (c) 2019-2024, Lawrence Livermore National Security, LLC and
 // other Serac Project Developers. See the top-level LICENSE file for
 // details.
 //
@@ -346,7 +346,8 @@ class BasePhysics {
    * @pre `reverseAdjointTimestep` with an appropriate adjoint load must be called prior to this method as many times as
    * the forward advance is called.
    */
-  virtual const std::unordered_map<std::string, const serac::FiniteElementDual&> computeInitialConditionSensitivity() const
+  virtual const std::unordered_map<std::string, const serac::FiniteElementDual&> computeInitialConditionSensitivity()
+      const
   {
     SLIC_WARNING_ROOT(axom::fmt::format("Initial condition sensitivities not enabled in physics module {}", name_));
     return {};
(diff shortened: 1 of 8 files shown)


src/serac/physics/state/state_manager.cpp
```diff
@@ -193,7 +193,7 @@ void StateManager::storeDual(FiniteElementDual& dual)
     grid_function = new mfem::ParGridFunction(&dual.space(), static_cast<double*>(nullptr));
     datacoll.RegisterField(name, grid_function);
     std::unique_ptr<mfem::HypreParVector> true_dofs(grid_function->GetTrueDofs());
-    dual = *true_dofs;
+    *true_dofs = dual;
```
btalamini: why isn't this the same as the store state, with a call to fillFromGridFunction?

tupek2: because duals are not grid functions

tupek2: or maybe, duals do not have gridFunctions

tupek2: I suppose this whole bit doesn't make a ton of sense...

src/serac/physics/state/state_manager.cpp
```diff
@@ -192,8 +192,6 @@ void StateManager::storeDual(FiniteElementDual& dual)
     // Create a new grid function with unallocated data. This will be managed by sidre.
     grid_function = new mfem::ParGridFunction(&dual.space(), static_cast<double*>(nullptr));
     datacoll.RegisterField(name, grid_function);
```
btalamini, line 194: We discussed this via chat while you were working on this, so this is just for reference: the grid function should be set from the dual values. Just mark this as resolved once you get that change pushed in.

btalamini (approved):

src/serac/physics/base_physics.hpp
```diff
@@ -237,7 +237,7 @@ class BasePhysics {
    * @note The input parameter name should not contain the base physics name. It should be identical to what
    * is in the physics module constructor argument list.
    */
-  const FiniteElementState& parameter(const std::string& parameter_name) const
+  virtual const FiniteElementState& parameter(const std::string& parameter_name) const
```
ebchin, line 240: Not seeing why this is needed for this PR.  Is this to set up some more changes in the future?

ebchin (approved):
Looks good @tupek2!  Just one minor question.

src/serac/physics/base_physics.hpp
tupek2: Yeah, this is needed in some things I'm adding to smith.

PR #1332, opened 2025-03-05
Tupek/parameterized solid sensitivity test

tupek2:


btalamini (approved):

ebchin (approved):
Looks good @tupek2.  Does it make sense to move the new test fixture (`BucklingSensitivityFixture`) to a new source file?

tupek2:
> Looks good @tupek2. Does it make sense to move the new test fixture (`BucklingSensitivityFixture`) to a new source file?

I could go either way.  I'm OK with having all the transient tests in 1 file, but I understand the usual 1 class per file ideal.  Maybe I care about that less with unit tests, slightly.

PR #1334, opened 2025-03-10
Fix Smith symlink issues

chapman39:
This PR fixes an issue in Smith (...that I thought I fixed already but I guess not). This PR doesn't affect Serac.

In Smith, we have a symlink that points to the `scripts/llnl` directory. This is nice because then we don't have to repeat ourselves, but has been causing problems when attempting to differentiate Smith from Serac.

When working on enabling benchmarks CI in Smith, I noticed that `__file__` was relative to the script `scripts/llnl/common_build_functions.py` in *Serac*. This was causing `get_project_name` was returning `serac` when it should be `smith`, for example.

LivChat introduced me to `sys.argv[0]` as opposed to `__file__`. A major difference is that `sys.argv[0]` will return the path to the symlink (rather than the actual file it points to) if a symlink is being used, which is exactly what we need in Smith.

### DIFF
--- a/scripts/llnl/common_build_functions.py
+++ b/scripts/llnl/common_build_functions.py
@@ -691,7 +691,7 @@ def on_rz():
 
 
 def get_script_dir():
-    return os.path.dirname(os.path.abspath(__file__))
+    return os.path.dirname(os.path.abspath(sys.argv[0]))
 
 
 _project_name = ""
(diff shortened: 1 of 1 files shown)


white238:
Is this because `__file__` refers to the `common_build_functions.py` and `argv[0]` refers to the script that calls it ie. `build_tpls.py`?

If so please put a comment...

white238 (approved):

tupek2 (approved):

PR #1336, opened 2025-03-11
Initial migration to use new design features.

tupek2:
Add residual interface, part of the mesh abstraction, and an existance test.


tupek2:
Added much more complete testing of the residual functionality, and testing (through its use) some parts of the newly proposed Mesh layer which stores domains.

src/serac/physics/mesh.cpp
```diff
@@ -0,0 +1,66 @@
...
+                axom::fmt::format("Could not find domain named {0} in mesh with tag {1}", domain_name, mesh_tag));
+  return domains_.at(domain_name);
+}
+
+serac::Domain& Mesh::addDomainOfBoundaryElements(const std::string& domain_name,
+                                                 std::function<bool(std::vector<vec2>, int)> func)
+{
+  SLIC_ERROR_IF(domains_.find(domain_name) != domains_.end(),
+                axom::fmt::format("A domain named {0} already exists in mesh with tag {1}", domain_name, mesh_tag));
+  domains_.insert({domain_name, Domain::ofBoundaryElements(*mfem_mesh, func)});
```
btalamini: Should we prefer `emplace()` to `insert()` here? As it stands, the returned reference will not be the same object that the mesh holds on to. Could be a problem if with the restriction operators that the domains store, right?

src/serac/physics/residual.hpp
```diff
@@ -0,0 +1,90 @@
...
+class Residual {
+ public:
+  /// @brief base constructor takes the name of the physics
+  Residual(std::string name) : name_(name) {}
+
+  /// @brief destructor
+  virtual ~Residual() {}
+
+  /// @brief using
+  using FieldPtr = FiniteElementState*;
```
btalamini, line 33: This is a matter of taste, but I don't like putting type info like `Ptr` in the names. I find it cluttering in code.

src/serac/physics/residual.hpp
```diff
@@ -0,0 +1,90 @@
...
+  using DualFieldPtr = FiniteElementDual*;
+
+  /** @brief Virtual interface for computing residual from a vector of serac::FiniteElementState*
+   *
+   * @param time time
+   * @param fields vector of serac::FiniteElementState* as arguments to the residual
+   * @param block_row integer which specifies which row of a block system to get the residual for, defaults to 0
+   * @return mfem::Vector
+   */
+  virtual mfem::Vector residual(double time, const std::vector<FieldPtr>& fields, int block_row = 0) const = 0;
```
btalamini, line 45: Should the return type be more specific to make intent clear? `FiniteElementDual`?

btalamini, line 45: You're deferring the internal state variable inputs. This is reasonable, I just want to point out that we need to consider what that will look like eventually.

src/serac/physics/solid_residual.hpp
```diff
@@ -0,0 +1,629 @@
...
+   *
+   * @note The actual types of these arguments passed will be `double`, `tensor<double, ... >` or tuples thereof
+   *    when doing direct evaluation. When differentiating with respect to one of the inputs, its stored
+   *    values will change to `dual` numbers rather than `double`. (e.g. `tensor<double,3>` becomes `tensor<dual<...>,
+   * 3>`)
+   *
+   * @param qdata the buffer of material internal variables at each quadrature point
+   *
+   * @pre MaterialType must have a public member variable `density`
+   * @pre MaterialType must define operator() that returns the Cauchy stress
```
btalamini: There are a few errors in the comment:

- The material must have a *method* to return the density (not a public member)
- The stress operator is called `pkStress`, not `operator()`
- The material returns PK, not Cauchy

src/serac/physics/solid_residual.hpp
```diff
@@ -0,0 +1,629 @@
...
```
btalamini: Same point as above.

btalamini (approved):

src/serac/physics/residual.hpp
tupek2: I agree.  I didn't want to go down that path, as I'm not sure if we've even settled on the data structure for state variable fields yet. maybe they are eventually also considered a 'field', maybe something else... Fun to figure that out, but not yet.

src/serac/physics/mesh.cpp
tupek2: Maybe, I never knew of map emplace until I just looked it up.  Probably the above does the same thing, otherwise this probably wouldn't be working.  Maybe its doing some extra copy at worse.  I dont understand why the returned reference would be different than the object the mesh hold onto?

src/serac/physics/residual.hpp
tupek2: I can try changing in later PRs, if you dont mind, a changing now has merge conflict potential.  This residual class in particular is rapidly evolving.

src/serac/physics/residual.hpp
tupek2: I'm 50/50.  So, functional takes mfem::Vector and returns mfem::Vector.  I presume dfem is similar.  If we return FiniteElementDual, its clearer, and nicer is some ways, but it is forcing an extra vector copy on the user... so, I dont really care either way, but that is the trade-off in my mind.

src/serac/physics/residual.hpp
tupek2: Also, its a reasonable position, but I want to mention that I used Ptr in typedef names in the past when I suspect that the type of that pointer might change (maybe *, maybe shared_ptr, maybe unique_ptr).  In fact, this exact typedef was a std::shared_ptr just a week ago!

src/serac/physics/mesh.hpp
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2019-2024, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file common.hpp
```
chapman39:
```suggestion
 * @file mesh.hpp
```

chapman39: the brief could probably be updated as well

src/serac/physics/residual.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  /// @brief name
+  std::string name() const { return name_; }
+
+ private:
+  /// name
+  std::string name_;
+};
+
+}  // namespace serac
```
chapman39:
```suggestion
}  // namespace serac

```

src/serac/physics/solid_residual.hpp
```diff
@@ -0,0 +1,629 @@
...
+    for_constexpr<sizeof...(parameter_space)>([&](auto) { parameter_fe_spaces.push_back(&params.back()->space()); });
+  }
+
+  using ResidualT = SolidResidual<order, dim, Parameters<parameter_space...>>;
+
+  return std::make_shared<ResidualT>(physics_name, mesh.tag(), states[3]->space(), states[0]->space(),
+                                     parameter_fe_spaces, parameter_names);
+}
+
+}  // namespace serac
```
chapman39:
```suggestion
}  // namespace serac

```

src/serac/physics/residual.hpp
btalamini: No, I don't mind. It was just a thought.

src/serac/physics/mesh.cpp
btalamini: I didn't read closely enough. You're right, you're returning a reference to the one in the map, so no sync problems. Just the extra copy. But these things are lightweight, it won't really matter.

src/serac/physics/solid_residual.hpp
tupek2: Fixed, please double check!

src/serac/physics/mesh.hpp
tupek2: Fixed for both

src/serac/physics/residual.hpp
tupek2: I think I fixed this, but I cannot tell for some reason.

src/serac/physics/mesh.cpp
tupek2: I changed to emplace.

btalamini:
@ebchin , you were an active participant in the design discussions. Your viewpoint is valuable. Could you take a look at this one?

src/serac/physics/mesh.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+namespace serac {
+
+// Forward declare
+struct Domain;
+
+/**
+ * @brief Helper class for constructing a mesh consistent with serac
+ */
+class Mesh {
```
ebchin, line 26: This is great.  I added something similar in Tribol that has static members for creating mesh primitives and does elementary mesh operations too (translate, rotate, scale, union, etc.).  It might be worth adding that to this class or a related class in the future, as we could probably get rid of all of the contact test mesh files (and probably other mesh files) with this additional functionality.

src/serac/physics/mesh.hpp
```diff
@@ -0,0 +1,89 @@
...
+  /// @brief Returns parallel communicator
+  MPI_Comm getComm() const;
+
+  /// @brief Returns domain corresponding to the entire mesh
+  serac::Domain& entireDomain() const;
+
+  /// @brief Returns registered domain with specified name
+  serac::Domain& domain(const std::string& domain_name) const;
+
+  /// @brief create domain of 3D boundary elements with specified name
```
ebchin, line 62: There's only 2 options, but we could template this on dim?  Then do explicit template instantiations in the source file to get rid of the duplication and not have to move the source to the header.

src/serac/physics/materials/solid_material.hpp
```diff
@@ -128,6 +128,48 @@ struct NeoHookean {
...
+   */
+  template <typename T, int dim, typename Density>
+  SERAC_HOST_DEVICE auto pkStress(State& /* state */, const tensor<T, dim, dim>& du_dX, const Density&) const
+  {
+    using std::log1p;
+    constexpr auto I = Identity<dim>();
+    auto lambda = K - (2.0 / 3.0) * G;
+    auto B_minus_I = dot(du_dX, transpose(du_dX)) + transpose(du_dX) + du_dX;
+
+    auto logJ = log1p(detApIm1(du_dX));
```
white238, line 153:
```suggestion
    constexpr auto I = Identity<dim>();
    auto lambda = K - (2.0 / 3.0) * G;
    auto B_minus_I = dot(du_dX, transpose(du_dX)) + transpose(du_dX) + du_dX;

    auto logJ = std::log1p(detApIm1(du_dX));
```
Nitpick: personally adverse to `using` statements unless they make readability better due to being used multiple times.

src/serac/physics/mesh.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2019-2025, Lawrence Livermore National Security, LLC and
```
white238, line 1: Does `mesh.cpp` and `mesh.hpp` belong better in `src/serac/mesh`? I assume this is because of the `state_manager` usage but maybe this means we should rethink the repo structure/naming.

src/serac/physics/materials/solid_material.hpp
tupek2: unfortunately, that 'using' is required to resolve the somewhat ambiguous scope of overloaded functions in serac.  Due to the templates, the types of the argument to log1p will be different in different contexts, so your change wont actually compile when gradients are being computed.  This is something I've wanted to fix for some time, but haven't had the bandwidth.  I think all math functions need to be in the serac:: namespace and reimplemented to take double as argument.  Then we replace all these calls with, eg. serac::log1p or serac::sqrt, etc.

white238: Ah, while unfortunate that makes sense. A comment would be helpful then something like "resolve ambigous log1p overload"

src/serac/physics/residual.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+  /**
+   * @brief Jacobian-vector product, will overwrite any existing values in jvpReactions
+   * @param time time
+   * @param fields vector of serac::FiniteElementState* as arguments to the residual
+   * @param vFields right hand side 'v' fields
+   * @param jvpReactions output vjps, 1 per row of a block system: d{r}_i / d{fields}_j * fieldsV[j]
+   * nullptr fieldsV are assumed to be all zero to avoid extra calculations
+   */
+  virtual void jvp(double time, const std::vector<FieldPtr>& fields, const std::vector<FieldPtr>& vFields,
```
white238, line 69: `jvp` and `vjp` are very terse. We have avoided being that terse in the past. Is this a very common acronym that is easily understood by users? I am also concerned about how easy it is to swap two letters.

src/serac/physics/mesh.hpp
```diff
@@ -0,0 +1,89 @@
...
+  serac::Domain& addDomainOfBoundaryElements(const std::string& domain_name,
+                                             std::function<bool(std::vector<vec2>, int)> func);
+
+ private:
+  /// @brief Sets up some initial domains, for now just the 'entire_domain', but eventually we can read of
+  /// names/blocks/attributes from the mesh and create default domains
+  void createDomains();
+
+  /// @brief
+  std::string mesh_tag;
```
white238, line 80: These two are missing their trailing underscore to denote they are member variables and also their brief's are empty.

src/serac/physics/materials/solid_material.hpp
white238: Made an issue #1349

src/serac/physics/materials/solid_material.hpp
```diff
@@ -128,6 +128,48 @@ struct NeoHookean {
...
+
+  /**
+   * @brief stress calculation for a NeoHookean material model
+   *
+   * When applied to 2D displacement gradients, the stress is computed in plane strain,
+   * returning only the in-plane components.
+   * @return The first Piola stress
+   */
+  template <typename T, int dim, typename Density>
+  SERAC_HOST_DEVICE auto pkStress(State& /* state */, const tensor<T, dim, dim>& du_dX, const Density&) const
```
ebchin, line 146: Why is this called `pkStress` vs other materials which just use `operator()` to return PK stress?  I like the clarity, but is the plan to make this change elsewhere to match this new material?

src/serac/physics/materials/solid_material.hpp
```diff
@@ -128,6 +128,48 @@ struct NeoHookean {
...
+ */
+struct NeoHookeanWithFieldDensity {
+  using State = Empty;  ///< this material has no internal variables
+
+  /**
+   * @brief stress calculation for a NeoHookean material model
+   *
+   * When applied to 2D displacement gradients, the stress is computed in plane strain,
+   * returning only the in-plane components.
+   * @return The first Piola stress
```
ebchin, line 143: I know the `@param` and `@tparam` Doxygen comments can be mostly useless at times, but I think they are pretty helpful for this function.  Can they be added here?

src/serac/physics/residual.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace serac {
+
+class FiniteElementState;
+class FiniteElementDual;
+
+/// @brief Abstract residual class
+class Residual {
+ public:
+  /// @brief base constructor takes the name of the physics
```
ebchin, line 26: I think this needs an `@param`.  Also, should the constructor and destructor be `protected`?

src/serac/physics/residual.hpp
```diff
@@ -0,0 +1,90 @@
...
+  virtual void vjp(double time, const std::vector<FieldPtr>& fields, const std::vector<DualFieldPtr>& vReactions,
+                   std::vector<FieldPtr>& vjpFields) const = 0;
+
+  /// @brief name
+  std::string name() const { return name_; }
+
+ private:
+  /// name
+  std::string name_;
+};
```
ebchin, line 88: Nice, I like this interface.  Excited to try it out on some problems.

src/serac/physics/solid_residual.hpp
```diff
@@ -0,0 +1,629 @@
...
+   *     auto [u, du_dxi] = displacement;
+   *     auto f           = u * 3.0 (X[0] < 0.01);
+   *     return f;  // define a displacement-proportional traction at a given support
+   *  });
+   *
+   * ~~~
+   *
+   * @note This method must be called prior to completeSetup()
+   */
+  template <int... active_parameters, typename callable>
```
ebchin, line 103: Should additional terms in the residual be a part of `SolidResidual`?  Maybe each term should be its own class?

src/serac/physics/solid_residual.hpp
```diff
@@ -0,0 +1,629 @@
...
+  void setRateMaterial(const MaterialType& material, std::shared_ptr<QuadratureData<StateType>> qdata = EmptyQData)
+  {
+    setRateMaterial(DependsOn<>{}, material, qdata);
+  }
+
+  /**
+   * @brief Functor representing a body force integrand.  A functor is necessary instead
+   * of an extended, generic lambda for compatibility with NVCC.
+   */
+  template <typename BodyForceType>
```
ebchin, line 203: This is another residual term that might benefit from being its own class.  Otherwise, we might risk bloating `solid_residual` with more and more terms that may or may not be used.  Same with traction, etc. below.

ebchin (approved):
@tupek2 great additions, looking forward to see how this evolves.  My main comment is to consider a class that sits above `SolidResidual` to keep each term from cluttering up that class.  Maybe `SolidWeakForm` or something?  Or change `SolidResidual` to `SolidResidualTerm`?

src/serac/physics/materials/solid_material.hpp
tupek2: sure

src/serac/physics/materials/solid_material.hpp
tupek2: added comments to clarify this is a new material design for the new residual.  eventually I think we will evolve to this throughout.

src/serac/physics/mesh.cpp
tupek2: Interesting observation.  I think for now I will leave in physics, as maybe I think of it as the 'mesh' object used by the physics models.  But yeah, otherwise we'd have to start moving state_manager and finite_element_state, etc. into a different folder that mesh would then depend on.

src/serac/physics/mesh.hpp
tupek2: This would be wonderful.  At the moment, the main reason for this mesh class is to save off domains so they have the correct lifetimes, but these other benefits would be wonderful.  I'd like to push this mesh class's usage out a bit more first.

src/serac/physics/mesh.hpp
tupek2: I considered this actually... the moment there is a third instantiation I completely agree.  2 is just on the boarderline where I'm happy to duplicate the signature.

src/serac/physics/mesh.hpp
tupek2: fixing...

src/serac/physics/residual.hpp
tupek2: It been while since I've thought alot about destructors and public vs protected... I always just do public virtual ~.  Here is some rule people on the internet think make sense:

C.35: A base class destructor should be either public and virtual, or protected and nonvirtual

https://stackoverflow.com/questions/8970466/is-there-a-use-for-making-a-protected-destructor-virtual

PR #1338, opened 2025-03-11
Dont stop github action workflow on style failure

chapman39:
Using fail-fast, other tests in a workflow will continue to run even if one test (e.g. style) fails.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -32,6 +32,7 @@ jobs:
     needs:
     - set_image_vars
     strategy:
+      fail-fast: false
       matrix:
         build_type: [ Debug, Release ]
         config:
@@ -88,6 +89,7 @@ jobs:
     needs:
     - set_image_vars
     strategy:
+      fail-fast: false
       matrix:
         check_type: [coverage, docs, style, header]
     container:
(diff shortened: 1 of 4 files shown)


src/tools/partitioner.cpp
```diff
@@ -20,7 +20,7 @@ using timer = axom::utilities::Timer;
 // partition the range {0, 1, 2, ... , n - 1} into
 // `num_blocks` roughly equal-sized contiguous chunks
 std::vector<uint32_t> partition_range(uint32_t n, uint32_t num_blocks)
-{
+      {
```
white238: ?

chapman39: just testing to see if it works

.github/workflows/ci-tests.yml
```diff
@@ -32,6 +32,7 @@ jobs:
     needs:
     - set_image_vars
     strategy:
+      fail-fast: false
```
tupek2: I'm slightly inclined to leave this one as true.  A test failure essentially means you have a real bug.  A docs/style failure kind of means you just need to be a bit more careful with non-code-behavior altering wordings, but you'd like to know you haven't broke any tests.  Does this seem reasonable?

chapman39: i was thinking that as well. makes sense, ill change it.

tupek2 (approved):

white238 (approved):

btalamini:
I like this idea.

PR #1340, opened 2025-03-17
spack v1.0 support: `%foo +bar` -> `+bar %foo`

haampie:
Spack v1.0 will parse `pkg %gcc +foo` as "pkg with direct dependency gcc
with variant foo enabled" instead of "pkg with variant foo enabled and compiler
gcc". Reorder as `pkg +foo %gcc` to be compatible with Spack v0.x and
v1.0.

### DIFF
--- a/scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
+++ b/scripts/spack/configs/toss_4_x86_64_ib/spack.yaml
@@ -78,9 +78,9 @@ spack:
     mvapich2:
       buildable: false
       externals:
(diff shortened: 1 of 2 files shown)


chapman39:
Thanks @haampie , this is reminding me that Uberenv also needs to be updated as well to handle 1.0 changes.

chapman39 (approved):
@white238 i can look into what changes need to be made in uberenv, but one major thing im thinking of is uberenv automatically adding the `%` at the beginning of a spec, which is going to not work for 1.0

PR #1344, opened 2025-03-21
thermomech_static_adjoint

lihanyu97:
Add static adjoint capability to thermomechanics.


btalamini (approved):
This looks good. Really nice work!

tupek2 (approved):
I somewhat (very) concerned about the amount of duplication between our physics.  But, I suppose the grand refactor is supposed to address that.  In the meantime, I guess its ok?

btalamini:
> I somewhat (very) concerned about the amount of duplication between our physics. But, I suppose the grand refactor is supposed to address that. In the meantime, I guess its ok?

I am too. I think it's good to have a coupled physics example in place (with tests) and then factor out the duplication.

PR #1346, opened 2025-03-24
Tupek/allow for more functional arguments

tupek2:
Users are running up against this limit on the number of arguments to functional.


btalamini (approved):
11 entries ought to be enough for anybody.

white238 (approved):
Would love to not have our own tuple. We tried at one point to use the one in Camp but had compiler issues if i remember correctly. Maybe with the upcoming dfem change we can remove this.

PR #1351, opened 2025-04-01
Move hypre version to config yaml (for lido)

chapman39:
This moves the hypre version locking to the config yaml. This allows lido to use a different version of hypre than Serac. Eventually, Serac/ Smith may be interested in moving to a newer hypre version as well.


tupek2:
are you aware of any downside for us moving to a more recent hypre? did we not update hypre in the last tpl update step?

chapman39:
@tupek2 the downside is that it might break some of our tests, which we are dealing with on the lido side. besides that, im not aware of a reason not to upgrade. it was not updated in the last tpl update.

tupek2 (approved):

chapman39:
merging into https://github.com/LLNL/serac/pull/1353

PR #1353, opened 2025-04-03
lido fix: newer hypre requires init call

chapman39:
In lido codevelop builds, we test Smith and Serac's tests. When lido updated to hypre 2.32.0, some of the tests broke due to hypre now requiring an init call. In the future it might be better to make these tests call serac's init function, but this MR quickly resolves the issues that were occuring in lido.

Some of the history from my previous lido fix MR is in this one, but it doesn't include all of the commits required, so both should be merged. Preferably this one should merge last.

### DIFF
--- a/scripts/spack/packages/serac/package.py
+++ b/scripts/spack/packages/serac/package.py
@@ -111,7 +111,7 @@ class Serac(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("netcdf-c@4.7.4")
 
-    depends_on("hypre~superlu-dist+mpi")
+    depends_on("hypre@2.26.0:~superlu-dist+mpi")
 
     depends_on("petsc", when="+petsc")
     depends_on("petsc+strumpack", when="+petsc+strumpack")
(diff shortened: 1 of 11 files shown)


scripts/spack/packages/serac/package.py
```diff
@@ -111,7 +111,7 @@ class Serac(CachedCMakePackage, CudaPackage, ROCmPackage):
 
     depends_on("netcdf-c@4.7.4")
 
-    depends_on("hypre@2.26.0~superlu-dist+mpi")
+    depends_on("hypre~superlu-dist+mpi")
```
white238: Should this maybe be a lower bound? I assume at some point a version won't work...

white238 (approved):
Approved to unblock user. This will be cleaned up in a follow up PR to use the `initialize` function. Thanks for the quick turn around @chapman39 !

tupek2 (approved):
Looks good, RAII next I suppose.

scripts/spack/packages/serac/package.py
chapman39: agreed, i changed it

PR #1354, opened 2025-04-04
Tupek/residual redesign

tupek2:
This completes some of the follow ons from the previous review (naming, _, etc.), but also extends and simplifies the use by including a serac::FunctionalResidual, which will eliminate a TON of duplication in the downstream smith new physics.  Also, now SolidResidual derives from FunctionalResidual.

### DIFF
--- a/src/serac/physics/functional_residual.hpp
+++ b/src/serac/physics/functional_residual.hpp
@@ -33,7 +33,6 @@ template <int spatial_dim, typename ShapeSpace, typename OutputSpace, typename..
 class FunctionalResidual<spatial_dim, ShapeSpace, OutputSpace, Parameters<InputSpaces...>,
                          std::integer_sequence<int, input_indices...>> : public Residual {
  public:
-
   using SpacesT = std::vector<const mfem::ParFiniteElementSpace*>;  ///< typedef
 
   /**
@@ -54,7 +53,7 @@ class FunctionalResidual<spatial_dim, ShapeSpace, OutputSpace, Parameters<InputS
 
     SLIC_ERROR_ROOT_IF(
         sizeof...(InputSpaces) != input_mfem_spaces.size(),
-        axom::fmt::format("{} parameter spaces given in the template argument but {} parameter names were supplied.",
+        axom::fmt::format("{} parameter spaces given in the template argument but {} input mfem spaces were supplied.",
                           sizeof...(InputSpaces), input_mfem_spaces.size()));
 
     if constexpr (sizeof...(InputSpaces) > 0) {
@@ -252,5 +251,16 @@ class FunctionalResidual<spatial_dim, ShapeSpace, OutputSpace, Parameters<InputS
   /// @brief functional residual evaluator, shape aware
   std::unique_ptr<ShapeAwareFunctional<ShapeSpace, OutputSpace(InputSpaces...)>> residual_;
 };
+
+/// @brief Helper function to construct vector of spaces from an existing vector of FiniteElementState.
+/// @param states vector of FiniteElementState
+inline std::vector<const mfem::ParFiniteElementSpace*> getSpaces(const std::vector<serac::FiniteElementState>& states)
+{
+  std::vector<const mfem::ParFiniteElementSpace*> spaces;
+  for (auto& f : states) {
+    spaces.push_back(&f.space());
+  }
+  return spaces;
+}
 
 }  // namespace serac
--- a/src/serac/physics/functional_residual.hpp
+++ b/src/serac/physics/functional_residual.hpp
@@ -87,7 +87,8 @@ class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>
   template <int... active_parameters, typename BodyIntegralType>
   void addBodyIntegral(DependsOn<active_parameters...>, std::string domain_name, BodyIntegralType body_integral)
   {
-    residual_->AddDomainIntegral(Dimension<dim>{}, DependsOn<active_parameters...>{}, body_integral, mesh_->domain(domain_name));
+    residual_->AddDomainIntegral(Dimension<dim>{}, DependsOn<active_parameters...>{}, body_integral,
+                                 mesh_->domain(domain_name));
   }
 
   /// @overload
--- a/src/serac/physics/functional_residual.hpp
+++ b/src/serac/physics/functional_residual.hpp
@@ -5,9 +5,10 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 
 /**
- * @file solid_residual.hpp
+ * @file functional_residual.hpp
  *
- * @brief Implement the residual interface for solid mechanics physics
+ * @brief Implements the residual interface using serac::ShapeAwareFunctional.
+ * Allows for generic specification of body and boundary integrals
  */
 
 #pragma once
@@ -36,7 +37,7 @@ class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>
   static constexpr int dim = OutputSpace::components;
 
   /**
-   * @brief Construct a new SolidResidual object
+   * @brief Construct a new FunctionalResidual object
    *
    * @param physics_name A name for the physics module instance
    * @param mesh The serac mesh
@@ -70,7 +71,7 @@ class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>
    *
    * @tparam BodyIntegralType The type of the body integral
    * // DependsOn<active_parameters...> can be indices into fields which the body integral may depend on
-   * @param domain_name The name of the registered domain over which the body force is applied. If nothing is supplied
+   * @param body_name The name of the registered domain over which the body force is applied. If nothing is supplied
    * the entire domain is
    * @param body_integral A function describing the body force applied
    * used.
@@ -86,17 +87,17 @@ class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>
    *
    */
   template <int... active_parameters, typename BodyIntegralType>
-  void addBodyIntegral(DependsOn<active_parameters...>, std::string domain_name, BodyIntegralType body_integral)
+  void addBodyIntegral(DependsOn<active_parameters...>, std::string body_name, BodyIntegralType body_integral)
   {
     residual_->AddDomainIntegral(Dimension<dim>{}, DependsOn<active_parameters...>{}, body_integral,
-                                 mesh_->domain(domain_name));
+                                 mesh_->domain(body_name));
   }
 
   /// @overload
   template <typename BodyForceType>
-  void addBodyIntegral(std::string domain_name, BodyForceType body_integral)
+  void addBodyIntegral(std::string body_name, BodyForceType body_integral)
   {
-    addBodyIntegral(DependsOn<>{}, domain_name, body_integral);
+    addBodyIntegral(DependsOn<>{}, body_name, body_integral);
   }
 
   /**
@@ -104,7 +105,7 @@ class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>
    *
    * @tparam NeumannType The type of the traction load
    * * // DependsOn<active_parameters...> can be indices into fields which the body integral may depend on
-   * @param domain_name The name of the registered domain over which the traction is applied. If nothing is supplied the
+   * @param boundary_name The name of the registered domain over which the traction is applied. If nothing is supplied the
    * entire boundary is
    * @param surface_function A function describing the traction applied to a boundary
    * used.
@@ -122,7 +123,7 @@ class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>
    *
    */
   template <int... active_parameters, typename NeumannType>
-  void addBoundaryIntegral(DependsOn<active_parameters...>, std::string domain_name, NeumannType surface_function)
+  void addBoundaryIntegral(DependsOn<active_parameters...>, std::string boundary_name, NeumannType surface_function)
   {
     residual_->AddBoundaryIntegral(
         Dimension<dim - 1>{}, DependsOn<active_parameters...>{},
@@ -130,14 +131,14 @@ class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>
           auto n = cross(get<DERIVATIVE>(X));
           return surface_function(t, get<VALUE>(X), normalize(n), params...);
         },
-        mesh_->domain(domain_name));
+        mesh_->domain(boundary_name));
   }
 
   /// @overload
   template <typename NeumannType>
-  void addBoundaryIntegral(std::string domain_name, NeumannType surface_function)
+  void addBoundaryIntegral(std::string boundary_name, NeumannType surface_function)
   {
-    addBoundaryIntegral(DependsOn<>{}, domain_name, surface_function);
+    addBoundaryIntegral(DependsOn<>{}, boundary_name, surface_function);
   }
 
   /// @overload
--- a/src/serac/physics/functional_residual.hpp
+++ b/src/serac/physics/functional_residual.hpp
@@ -1,0 +1,261 @@
+// Copyright (c) 2019-2024, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file solid_residual.hpp
+ *
+ * @brief Implement the residual interface for solid mechanics physics
+ */
+
+#pragma once
+
+#include "serac/physics/residual.hpp"
+#include "serac/physics/mesh.hpp"
+#include "serac/physics/state/state_manager.hpp"
+
+namespace serac {
+
+template <typename ShapeSpace, typename OutputSpace, typename parameters = Parameters<>,
+          typename parameter_indices = std::make_integer_sequence<int, parameters::n>>
+class FunctionalResidual;
+
+/**
+ * @brief The nonlinear residual class
+ *
+ * This uses Functional to compute the solid mechanics residuals and tangent
+ * stiffness matrices.
+ *
+ * @tparam dim The spatial dimension of the mesh
+ */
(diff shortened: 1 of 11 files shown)


src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,278 @@
...
+      : Residual(physics_name), mesh_(mesh)
+  {
+    std::array<const mfem::ParFiniteElementSpace*, sizeof...(parameter_space)> trial_spaces;
+
+    SLIC_ERROR_ROOT_IF(
+        sizeof...(parameter_space) != input_mfem_spaces.size(),
+        axom::fmt::format("{} parameter spaces given in the template argument but {} parameter names were supplied.",
+                          sizeof...(parameter_space), input_mfem_spaces.size()));
+
+    if constexpr (sizeof...(parameter_space) > 0) {
```
tupek2: no need to copy

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,253 @@
...
+ */
+template <typename ShapeSpace, typename OutputSpace, typename... parameter_space, int... parameter_indices>
+class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>,
+                         std::integer_sequence<int, parameter_indices...>> : public Residual {
+ public:
+  /// @brief extract residual dimension from the output space
+  static constexpr int dim = OutputSpace::components;
+
+  /**
+   * @brief Construct a new SolidResidual object
```
white238:
```suggestion
   * @brief Construct a new FunctionalResidual object
```

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,253 @@
...
+
+/**
+ * @brief The nonlinear residual class
+ *
+ * This uses Functional to compute the solid mechanics residuals and tangent
+ * stiffness matrices.
+ *
+ */
+template <typename ShapeSpace, typename OutputSpace, typename... parameter_space, int... parameter_indices>
+class FunctionalResidual<ShapeSpace, OutputSpace, Parameters<parameter_space...>,
```
white238: Is `FunctionalResidual` descriptive enough?

src/serac/physics/materials/solid_material.hpp
```diff
@@ -521,4 +479,52 @@ struct ConstantTraction {
   }
 };
 
+/**
+ * @brief Neo-Hookean material model
+ * This struct differs in style relative to the older materials as it needs to evaluate both stress
+ * and density.  As a result, we want to clearly name these functions.
+ * This is likely going to be a new design going forward, at the moment it works with the
+ * solid_residual_class.
```
white238:
```suggestion
 * SolidResidual class.
```

src/serac/physics/mesh.hpp
```diff
@@ -76,13 +79,13 @@ class Mesh {
   /// names/blocks/attributes from the mesh and create default domains
   void createDomains();
 
-  /// @brief
+  /// @brief String identifying mesh in the state manager
   std::string mesh_tag;
```
white238: This is missing the trailing underscore as well. Would you mind fixing this?

src/serac/physics/tests/test_functional_residual.cpp
```diff
@@ -0,0 +1,281 @@
...
+    residual->jvp(time, all_states, all_v_rhs_states, jvps);
+    EXPECT_NEAR(jvp_slow.Norml2(), jvp.Norml2(), 1e-12);
+  }
+}
+
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+
+  serac::initialize(argc, argv);
```
white238: Please use the ApplicationManager here

src/serac/physics/tests/test_solid_residual.cpp
```diff
@@ -30,12 +30,13 @@ auto getPointers(std::vector<T>& values)
...
 {
-  std::vector<T*> pointers;
+  assert(params.size() > 0);
+  std::vector<T*> pointers{&params[0]};
   for (auto& t : states) {
     pointers.push_back(&t);
   }
-  for (auto& t : params) {
-    pointers.push_back(&t);
+  for (size_t n = 1; n < params.size(); ++n) {
```
white238: Why is the 0 index being skipped here?

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,253 @@
+// Copyright (c) 2019-2024, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file solid_residual.hpp
```
white238:
```suggestion
 * @file functional_residual.hpp
```

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,253 @@
...
+ *
+ * @brief Implement the residual interface for solid mechanics physics
```
white238: Is this comment correct?

src/serac/physics/tests/test_functional_residual.cpp
tupek2: fixed

src/serac/physics/tests/test_solid_residual.cpp
tupek2: Added a comment to explain a bit more what it happening here.

src/serac/physics/mesh.hpp
tupek2: Fixed

src/serac/physics/materials/solid_material.hpp
tupek2: Fixed

src/serac/physics/functional_residual.hpp
tupek2: Fixed

src/serac/physics/functional_residual.hpp
tupek2: Fixed

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,258 @@
...
+    return J;
+  }
+
+  /// @overload
+  void jvp(double time, double dt, const std::vector<FieldPtr>& fields, const std::vector<FieldPtr>& vFields,
+           std::vector<DualFieldPtr>& jvpReactions) const override
+  {
+    SLIC_ERROR_IF(vFields.size() != fields.size(),
+                  "Invalid number of field sensitivities relative to the number of fields");
+    SLIC_ERROR_IF(jvpReactions.size() != 1, "Solid mechanics nonlinear system only supports 1 output residual");
```
btalamini: Solid mechanics -> Functional

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,258 @@
...
+
+    dt_ = dt;
+    auto jacs = jacobianFunctions(std::make_integer_sequence<int, sizeof...(input_indices) + 1>{}, time, fields);
+
+    *jvpReactions[0] = 0.0;
+
+    for (size_t input_col = 0; input_col < fields.size(); ++input_col) {
+      if (vFields[input_col] != nullptr) {
```
btalamini, line 201: I presume this is a performance optimization for the case where the user wants some of the tangent vectors to be logically zero. That's a nice feature. We'll need to communicate the users.

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }
+
+  /// @overload
+  void vjp(double time, double dt, const std::vector<FieldPtr>& fields, const std::vector<DualFieldPtr>& vReactions,
+           std::vector<FieldPtr>& vjpFields) const override
+  {
+    SLIC_ERROR_IF(vjpFields.size() != fields.size(),
+                  "Invalid number of field sensitivities relative to the number of fields");
+    SLIC_ERROR_IF(vReactions.size() != 1, "Solid mechanics nonlinear system only supports 1 output residual");
```
btalamini: Says "Solid mechanics" again. If the size is 0, this error message is technically correct, but probably misleading. Consider checking size == 0 and size > 1 separately.

src/serac/physics/functional_residual.hpp
tupek2: I think its in the comments, but I'll double check.  Also, could do that for the vjp as well to communicate they don't want a particular sensitivity computed.

btalamini: True. Probably worth doing.

src/serac/physics/tests/test_solid_residual.cpp
```diff
@@ -178,14 +183,14 @@ TEST_F(ResidualFixture, VjpConsistency)
   // test vjp
   serac::FiniteElementDual v(res_vector.space(), "v");
   pseudoRand(v);
-  auto all_jvps = getPointers(adjoint_states, adjoint_params);
+  auto all_jvps = getPointers(dual_states, dual_params);
```
btalamini: I'm confused by the naming. Did you mean `all_vjps`?

src/serac/physics/tests/test_solid_residual.cpp
```diff
@@ -178,14 +186,14 @@ TEST_F(ResidualFixture, VjpConsistency)
...
-  residual->vjp(time, all_states, getPointers(v), all_jvps);
+  residual->vjp(time, dt, all_states, getPointers(v), all_jvps);
 
   for (size_t i = 0; i < all_states.size(); ++i) {
     serac::FiniteElementState vjp = *all_states[i];
     vjp = 0.0;
-    auto J = residual->jacobian(time, all_states, jacobian_weights(i));
+    auto J = residual->jacobian(time, dt, all_states, jacobian_weights(i));
     J->MultTranspose(v, vjp);
     if (i == 0) vjp += 1.0;  // make sure jvp uses +=
```
btalamini:
```suggestion
    if (i == 0) vjp += 1.0;  // make sure vjp uses +=
```

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,266 @@
...
+   * @param input_mfem_spaces Vector of finite element spaces which are arguments to the residual
+   */
+  FunctionalResidual(std::string physics_name, std::shared_ptr<Mesh> mesh,
+                     const mfem::ParFiniteElementSpace& shape_disp_space,
+                     const mfem::ParFiniteElementSpace& output_mfem_space, SpacesT input_mfem_spaces)
+      : Residual(physics_name), mesh_(mesh)
+  {
+    std::array<const mfem::ParFiniteElementSpace*, sizeof...(InputSpaces)> trial_spaces;
+
+    SLIC_ERROR_ROOT_IF(
```
btalamini, line 54: I think we should unit test this.

btalamini (approved):
I really like how streamlined the solid residual class is now!

src/serac/physics/tests/test_functional_residual.cpp
```diff
@@ -0,0 +1,277 @@
+// Copyright (c) 2019-2024, Lawrence Livermore National Security, LLC and
```
ebchin, line 1: Is the naming redundant here?  Maybe we can remove `test_` from the name, but then it's easily confused with `functional_residual.hpp`.  Hmm, maybe just leave it.

src/serac/physics/tests/test_functional_residual.cpp
```diff
@@ -0,0 +1,277 @@
...
+}
+
+struct ResidualFixture : public testing::Test {
+  static constexpr int dim = 2;
+  static constexpr int disp_order = 1;
+
+  using VectorSpace = serac::H1<disp_order, dim>;
+  using DensitySpace = serac::L2<disp_order - 1>;
+
+  enum STATE
```
ebchin, line 61: `enum class STATE`?

src/serac/physics/tests/test_functional_residual.cpp
```diff
@@ -0,0 +1,277 @@
...
+  {
+    DISP,
+    VELO
+  };
+
+  enum PAR
```
ebchin, line 67: `enum class PAR`?

src/serac/physics/tests/test_functional_residual.cpp
```diff
@@ -0,0 +1,277 @@
...
+    mesh = std::make_shared<serac::Mesh>(mfem::Mesh::MakeCartesian2D(6, 20, element_shape, true, length, width),
+                                         "this_mesh_name", 0, 0);
+
+    serac::FiniteElementState disp = serac::StateManager::newState(VectorSpace{}, "displacement", mesh->tag());
+    serac::FiniteElementState velo = serac::StateManager::newState(VectorSpace{}, "velocity", mesh->tag());
+    serac::FiniteElementState shape_disp =
+        serac::StateManager::newState(VectorSpace{}, "shape_displacement", mesh->tag());
+    serac::FiniteElementState density = serac::StateManager::newState(DensitySpace{}, "density", mesh->tag());
+
+    states = {disp, velo};
```
ebchin, line 89: Can these be `move`d here, or does `serac::StateManager` need to know they exist?  Though I think there's still a copy required here, so it probably doesn't matter.

src/serac/physics/functional_residual.hpp
```diff
@@ -0,0 +1,266 @@
...
+      auto K = serac::get<DERIVATIVE>(jacs[input_col](time, fields));
+      std::unique_ptr<mfem::HypreParMatrix> J = assemble(K);
+      J->AddMultTranspose(*vReactions[0], *vjpFields[input_col]);
+    }
+  }
+
+ protected:
+  /// @brief Utility to evaluate residual using all fields in vector
+  template <int... i>
+  auto evaluateResidual(std::integer_sequence<int, i...>, double time, const std::vector<FieldPtr>& fs) const
```
ebchin: Doesn't look like this is used anywhere.  Should it be removed?

src/serac/physics/solid_residual.hpp
```diff
@@ -31,83 +29,38 @@ class SolidResidual;
  * @tparam order The order of the discretization of the displacement and velocity fields
  * @tparam dim The spatial dimension of the mesh
  */
-template <int order, int dim, typename... parameter_space, int... parameter_indices>
-class SolidResidual<order, dim, Parameters<parameter_space...>, std::integer_sequence<int, parameter_indices...>>
-    : public Residual {
+template <int order, int dim, typename... InputSpaces>
+class SolidResidual<order, dim, Parameters<InputSpaces...>>
```
ebchin, line 33: How does this work if we create a `SolidResidual` built on, say, `dFEM`?

src/serac/physics/functional_residual.hpp
tupek2: Fixed, I think.

src/serac/physics/tests/test_solid_residual.cpp
tupek2: Actually, I changed it to remove this confusion.

src/serac/physics/functional_residual.hpp
tupek2: I think this is fixed now

src/serac/physics/tests/test_solid_residual.cpp
tupek2: changed

src/serac/physics/tests/test_solid_residual.cpp
tupek2: changed

src/serac/physics/functional_residual.hpp
tupek2: next time!

ebchin (approved):
Thanks @tupek2, these changes are a definite improvement over the current state of the code.

Two things to consider going forward: 1) how this will work with potential replacements for `Functional` and 2) how it will work with constraints/contact.

src/serac/physics/functional_residual.hpp
tupek2: Fixed

src/serac/physics/functional_residual.hpp
tupek2: I hope the comments help, and I wouldn't really know what other words to use.  The idea: its the 'residual' implemented using the serac::Functional utilities.  Eventually we'll have a DfemResidual, or some such.

src/serac/physics/tests/test_functional_residual.cpp
tupek2: because its hard to tell if a file is a test, example, or library code, I've suggestion we come up with a naming convention.  LiDO does something similar, we could adopt their convention, though I'm ok with this one too.

src/serac/physics/tests/test_functional_residual.cpp
tupek2: enum class entries are not size_t, so it doesn't achieve what I'm trying to do here, which is simply give names to the indices to improve readability without having to add lies (aka, comments).

src/serac/physics/tests/test_functional_residual.cpp
tupek2: same, these are indented to be classical enums to index into things with names.

src/serac/physics/tests/test_functional_residual.cpp
tupek2: there are all shared_ptr, I'm not worried about the copy and reference counting cost at this high of a level.

src/serac/physics/functional_residual.hpp
tupek2: Good catch, removing.  It was needed in an earlier iteration of the design, no longer.

src/serac/physics/solid_residual.hpp
tupek2: I suppose we could call this SolidFunctionalResidual, when we are there, there will be a DfemResidual, then maybe a SolidDfemResidual based on that.  We can do the name changes at that point if we feel like it.

src/serac/physics/tests/test_functional_residual.cpp
white238: This should be a meeting topic. I have thoughts but wont die on this hill

PR #1357, opened 2025-04-08
Refactor serac init/ termination; make usage consistent

### LINKED ISSUE #1326: RAII `serac::finalize`
There is a possible issue in Serac where certain objects have destructors that make MPI calls. If we let unique pointers to those objects clean themselves when they get out of scope, their MPI calls will be called after `serac::exitGracefully` function (which finalizes MPI) is called. In LiDO, we resolved this by forcing our initalizer function return a "finalizer struct" whose destructor finalizes MPI.

Do we want something like this in Serac?

```cpp
struct SeracFinalizer {
  ~SeracFinalizer() { serac::finalize(...); }
};

...

{
   // since the finalizer is initialized first, it's destructor is guaranteed to be called last
  std::shared_ptr<serac::SeracFinalizer> seracFinalizer;
  seracFinalizer = serac::initialize(...); 

  // do stuff

  return 0;
}
```

This would also mean removing the `exit` call here https://github.com/LLNL/serac/blob/efadbe3cdc10cb4e5f325ec81a8780150f760183/src/serac/infrastructure/terminator.cpp#L73

More info here

* https://lc.llnl.gov/gitlab/topopt/lido-2.0/-/issues/491
* https://lc.llnl.gov/gitlab/topopt/lido-2.0/-/merge_requests/827

chapman39:
Replaces `serac::initialize(...);` and `serac::exitGracefully();` with a new `serac::ApplicationManager` class, which inits all of our important libraries, profiling, logging, mpi, etc. Its destructor automatically cleans up these libraries when it goes out of scope.

Also, makes init/ termination usage consistent across all tests and examples.

```cpp
// old way
serac::initialize(argc, argv);
...
serac::exitGracefully();

// new way
serac::ApplicationManager applicationManager(argc, argv);
...
return 0;
```

Fixes #1326


src/serac/physics/base_physics.cpp
```diff
@@ -10,9 +10,9 @@
 
 #include "axom/fmt.hpp"
 
-#include "serac/infrastructure/initialize.hpp"
+#include "serac/infrastructure/about.hpp"
 #include "serac/infrastructure/logger.hpp"
-#include "serac/infrastructure/terminator.hpp"
+#include "serac/infrastructure/application_manager.hpp"
```
tupek2: Is it necessary for base_physics to depend on application_manager?  Seems odd to me, as base_physics isn't setting up or terminating anything?

src/serac/numerics/equation_solver.cpp
```diff
@@ -12,7 +12,7 @@
 #include <iostream>
 
 #include "serac/infrastructure/logger.hpp"
-#include "serac/infrastructure/terminator.hpp"
+#include "serac/infrastructure/application_manager.hpp"
```
tupek2: Can we remove this dependency.  I was sort of expecting only tests and apps, examples, etc. to depend on this, not internal utilities.

src/serac/mesh/mesh_utils.cpp
```diff
@@ -12,7 +12,7 @@
 #include "axom/fmt.hpp"
 
 #include "serac/infrastructure/logger.hpp"
-#include "serac/infrastructure/terminator.hpp"
+#include "serac/infrastructure/application_manager.hpp"
```
tupek2: Same here, why is application_manager needed?

src/serac/infrastructure/cli.cpp
```diff
@@ -10,7 +10,7 @@
 
 #include "serac/infrastructure/input.hpp"
 #include "serac/infrastructure/logger.hpp"
-#include "serac/infrastructure/terminator.hpp"
+#include "serac/infrastructure/application_manager.hpp"
```
tupek2: Is this needed?

src/serac/infrastructure/input.cpp
```diff
@@ -12,7 +12,7 @@
 #include "axom/core.hpp"
 
 #include "serac/infrastructure/logger.hpp"
-#include "serac/infrastructure/terminator.hpp"
+#include "serac/infrastructure/application_manager.hpp"
```
tupek2: Here as well, include needed?

src/serac/infrastructure/output.cpp
```diff
@@ -6,9 +6,9 @@
 
 #include "serac/infrastructure/output.hpp"
 
-#include "serac/infrastructure/initialize.hpp"
+#include "serac/infrastructure/about.hpp"
 #include "serac/infrastructure/logger.hpp"
-#include "serac/infrastructure/terminator.hpp"
+#include "serac/infrastructure/application_manager.hpp"
```
tupek2: check about this include too!

src/serac/infrastructure/logger.cpp
```diff
@@ -5,9 +5,8 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 
 #include "serac/infrastructure/logger.hpp"
-
-#include "serac/infrastructure/initialize.hpp"
-#include "serac/infrastructure/terminator.hpp"
+#include "serac/infrastructure/about.hpp"
+#include "serac/infrastructure/application_manager.hpp"
```
tupek2: This include?

tupek2 (approved):
This looks great to me.  Just double check that all the include of the new app manager are actually needed.  I think I marked all those I feel wouldn't be needed (unless I'm missing something).

src/serac/physics/base_physics.cpp
chapman39: good point

src/serac/infrastructure/logger.cpp
chapman39: ill be pushing a commit removing these soon

chapman39: I removed includes by searching for "application_manager.hpp" in `src` but ignoring subdirs titled `tests`

examples/buckling/cylinder.cpp
```diff
@@ -87,8 +87,9 @@ int main(int argc, char* argv[])
   bool use_contact = true;
   auto contact_type = serac::ContactEnforcement::Penalty;
 
-  // Initialize Serac and all of the external libraried
-  serac::initialize(argc, argv);
+  // Initialize Serac and all of the external libraries.
+  // Automatically terminates libraries and finalizes MPI.
```
white238: This is a good comment. Can you add it to all examples?

src/serac/infrastructure/application_manager.cpp
```diff
@@ -27,38 +27,22 @@
...
   }
 
-  // Initialize the signal handler
-  terminator::registerSignals();
-
   // Initialize SLIC logger
-  if (!logger::initialize(comm)) {
-    serac::exitGracefully(true);
+  if (!logger::initialize(comm_)) {
+    std::cerr << "Failed to initialize SLIC logger" << std::endl;
```
white238, line 88: 👍

src/serac/infrastructure/application_manager.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <utility>
+#include <iostream>
+
+#include "mpi.h"
+
+namespace serac {
+
+/**
+ * @brief RAII Application Manager class. Initializes MPI and other important libraries as
+ * well as automatically terminates them upon going out of scope.
```
white238:
```suggestion
 * well as automatically finalizes them upon going out of scope.
```
terminate sounds so drastic

src/serac/infrastructure/terminator.cpp
```diff
@@ -1,76 +0,0 @@
...
-
-}  // namespace
-
-namespace serac {
-
-namespace terminator {
-
-void registerSignals()
-{
-  std::signal(SIGINT, signalHandler);
```
white238, line 38: This seems to have been lost. Unless I'm missing it.

examples/buckling/cylinder.cpp
tupek2: I would love to keep this short if its duplicated a lot.  I'd suggest:

// Initialize and automatically terminate Serac libraries

src/serac/physics/tests/thermomech_statics_patch.cpp
```diff
@@ -857,12 +858,6 @@ TEST(Thermomechanics, Patch3dQ3EssentialAndNaturalCoupled)
...
-
-  axom::slic::SimpleLogger logger;
-
-  int result = RUN_ALL_TESTS();
-  MPI_Finalize();
-
-  return result;
+  serac::ApplicationManager applicationManager(argc, argv);
+  return RUN_ALL_TESTS();
 }
```
white238:
```suggestion
}

```

src/serac/infrastructure/terminator.cpp
chapman39: i don't think its needed anymore since MPI is finalized automatically now, unless the printing of the signal is useful. let me know if you want me to re-add it

chapman39: turns out we needed `std::signal` if we wanted the application manager's destructor to be called. I re-added it and will be pushing it up soon. I didn't need to explicitly call the destructor, i found. Simply registering the signal handler does it for us.

Here's some example output.

```
[meemee@ruby962:build-ruby-toss_4_x86_64_ib-clang@14.0.6-debug]$ ./tests/parameterized_thermomechanics_example 
Logger activated: 'serac_serial_logger'
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Thermomechanics
[ RUN      ] Thermomechanics.ParameterizedMaterial
Saving data collection at time: '0' and cycle: '0' to path: 'parameterized_thermomechanics/mesh_datacoll'
^C[SIGNAL]: Received signal 2 (Interrupt), exiting
  0 SNES Function norm 2.143256126293e-03

...

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (9939 ms total)
[  PASSED  ] 1 test.
HELLO DESTROYING! APPLICATION
```

chapman39: ill be pushing a fix soon

```
[meemee@ruby962:build-ruby-toss_4_x86_64_ib-clang@14.0.6-debug]$ ./tests/parameterized_thermomechanics_example 
Logger activated: 'serac_serial_logger'
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Thermomechanics
[ RUN      ] Thermomechanics.ParameterizedMaterial
^C[SIGNAL]: Received signal 2 (Interrupt), exiting
HELLO DESTROYING! APPLICATION
```

src/serac/infrastructure/application_manager.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <iostream>
+
+#include "mpi.h"
+
+namespace serac {
+
+/**
+ * @brief Destroy MPI, signal handling, logging, profiling, hypre, sundials, petsc, and slepc.
+ */
+void finalizer();
```
white238: I don't think we want to expose this to users. We want it to be called either by the RAII of ApplicationManager or via the catching of the signals.

white238 (approved):

PR #1358, opened 2025-04-09
Enable curved high-order elements

### LINKED ISSUE #1356: High-order meshes read in from files drop to first order
When Serac reads in meshes from formats that support higher-order elements (such as Genesis), the mesh coordinate field is always linear. Curved adges/faces are lost, and the elements become straight-sided.

btalamini:
The position field in Serac meshes are currently always converted to first order. This PR enables high order meshes with curved faces/edges.

In my view, the state manager was doing too much. On the `setMesh()` call, it was checking that several conditions on the mesh were true, and if not, it *changed* the mesh to make it so. Now, I just abort if the mesh being set doesn't meet requirements, and post an error message telling the user why the mesh is invalid. 

High order ParaView output automatically works. We already had code that set the ParaView element order to the highest order field registered for output. I made it also consider the order of the mesh coordinates, in case that's the highest order field.

As an aside, I think this bug was telling us to have our own mesh class - it has invariants that we need to enforce. There are other good reasons to do it, as well.

Fixes #1356.


src/serac/physics/state/state_manager.cpp
```diff
@@ -256,15 +212,12 @@ mfem::ParMesh& StateManager::setMesh(std::unique_ptr<mfem::ParMesh> pmesh, const
...
 
-  // Generate the face neighbor information in the mesh. This is needed by the face restriction
-  // operators used by Functional
-  new_pmesh.ExchangeFaceNbrData();
-
   // We must construct the shape fields here as the mesh did not exist during the newDataCollection call
   // for the non-restart case
+  // BT: Consider storing shape fields on the mesh class and making them on mesh construction.
+  // The setMesh() call wouldn't mutate the mesh at all, just store it as name implies.
   constructShapeFields(mesh_tag);
```
tupek2, line 221: I think in the future we won't be making shape fields here at all?  But rather create then like a normal field during construction of the serac::Mesh.  Is that basically what that comment is indicating?

src/serac/physics/state/state_manager.cpp
```diff
@@ -322,4 +275,35 @@ double StateManager::time(std::string mesh_tag)
...
+                     "your mesh before setting it with the state manager.");
+
+  SLIC_ERROR_ROOT_IF(!pmesh.OwnsNodes(),
+                     "The mesh must own its node grid function, as ownership will be passed ot the state manager.");
+
+  SLIC_WARNING_ROOT_IF(nodes->FESpace()->FEColl()->GetContType() == mfem::FiniteElementCollection::DISCONTINUOUS,
+                       "Periodic mesh detected! This will only work on translational periodic surfaces for vector H1 "
+                       "fields and has not been thoroughly tested. Proceed at your own risk.");
+
+  std::string ordering_string{"byVDIM"};
```
tupek2: I would probably ternary this, but no need to change.

src/serac/physics/state/state_manager.cpp
btalamini: Yes.

src/serac/physics/tests/lce_Brighenti_tensile.cpp
```diff
@@ -36,6 +36,8 @@ TEST(LiquidCrystalElastomer, Brighenti)
   mfem::Mesh cuboid =
       mfem::Mesh(mfem::Mesh::MakeCartesian3D(nElem, 2 * nElem, 2 * nElem, mfem::Element::HEXAHEDRON, lx, ly, lz));
   auto mesh = std::make_unique<mfem::ParMesh>(MPI_COMM_WORLD, cuboid);
+  mesh->EnsureNodes();
```
tupek2, line 38: Is this needed just because its a test that doesn't register the mesh on the state_manager?

tupek2 (approved):

src/serac/physics/state/state_manager.cpp
btalamini: I almost did it that way. I went this way so that it's compile-time, but there's no real advantage to that here. I can change it.

tupek2: I think the ternary would be compile time too as the conditional is compile time known.  In the case of "by_nodes", it might even save a copy. None of this performance matters of course.  I don't care strongly, I'm just pro ternary in a lot of cases, as it avoids mutable state and the result can be const.

src/serac/physics/state/state_manager.cpp
```diff
@@ -322,4 +275,35 @@ double StateManager::time(std::string mesh_tag)
...
+void checkMesh(const mfem::ParMesh& pmesh)
+{
+  const mfem::GridFunction* nodes = pmesh.GetNodes();
+
+  SLIC_ERROR_ROOT_IF(!nodes,
+                     "The mesh must have a grid function for the nodes defined. Call the EnsureNodes() method on "
+                     "your mesh before setting it with the state manager.");
+
+  SLIC_ERROR_ROOT_IF(!pmesh.OwnsNodes(),
+                     "The mesh must own its node grid function, as ownership will be passed ot the state manager.");
```
white238:
```suggestion
                     "The mesh must own its node grid function, as ownership will be passed to the state manager.");
```

src/serac/physics/state/tests/test_state_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+}  // namespace serac
+
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+  serac::initialize(argc, argv);
+  int result = RUN_ALL_TESTS();
+  serac::exitGracefully(result);
+}
```
white238:
```suggestion
}

```

tupek2: also, should this use Alex's new RAII thing?

src/serac/physics/tests/lce_Brighenti_tensile.cpp
btalamini: No. The mesh is registered, on line 44. It's because Functional *requires* that the mesh have a coordinate grid function and that the face neighbor data exist. Before this PR, those properties were enforced by `setMesh`, but I removed that. I don't think the state manager should try to enforce mesh invariants - it should just set the mesh. If we need our meshes to have certain properties, our own mesh class is the place to do that.

btalamini: Maybe the short answer is that the other examples and tests call `refineAndDistribute()` to make the mesh parallel, and that function calls `ExchangeFaceNbrData()` and `EnsureNodes()`, so the driver doesn't have to.

src/serac/physics/state/tests/test_state_manager.cpp
btalamini: That may not work in this case. How do you use the RAII approach and return a value from `main()`? The gtest standard requires you to return the value of `RUN_ALL_TESTS()`.

src/serac/physics/state/state_manager.cpp
btalamini: I put in the ternary and made the variable `const`.

src/serac/physics/state/tests/test_state_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+}  // namespace serac
+
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+  serac::initialize(argc, argv);
+  int result = RUN_ALL_TESTS();
+  serac::exitGracefully(result);
```
white238:
```suggestion
  ::testing::InitGoogleTest(&argc, argv);
  serac::ApplicationManager applicationManager(argc, argv);
  return RUN_ALL_TESTS();
```

src/serac/physics/state/tests/test_state_manager.cpp
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+#include "serac/physics/state/state_manager.hpp"
+
+#include "axom/slic/core/SimpleLogger.hpp"
+#include <gtest/gtest.h>
+#include "mfem.hpp"
+
+#include "serac/infrastructure/initialize.hpp"
+#include "serac/infrastructure/terminator.hpp"
```
white238:
```suggestion
#include "serac/infrastructure/application_manager.hpp"
```

white238 (approved):

PR #1360, opened 2025-04-16
Tupek/residual redesign

tupek2:
Adding in a few additional missing features from the mesh class, and the functional residual class that ended up being needed in downstream applications.

### DIFF
--- a/src/serac/numerics/functional/shape_aware_functional.hpp
+++ b/src/serac/numerics/functional/shape_aware_functional.hpp
@@ -464,6 +464,61 @@ class ShapeAwareFunctional<shape, test(trials...), exec> {
    * lambda for compatibility with NVCC.
    */
   template <typename Integrand, int dim, int... args>
+  struct ShapeAwareInteriorBoundaryIntegrandWrapper {
+    /// @brief Constructor for functor
+    ShapeAwareInteriorBoundaryIntegrandWrapper(Integrand integrand) : integrand_(integrand) {}
+    /// @brief integrand
+    Integrand integrand_;
+    /**
+     * @brief integrand call
+     *
+     * @tparam PositionType position type
+     * @tparam ShapeValueType shape value type
+     * @tparam QFuncArgs type of variadic pack to forward to qfunc
+     * @param[in] time time
+     * @param[in] x position
+     * @param[in] shape_val shape
+     * @param[in] qfunc_args qfunc parameter pack
+     * @return qf function corrected for boundary area
+     */
+    template <typename PositionType, typename ShapeValueType, typename... QFuncArgs>
+    SERAC_HOST_DEVICE auto operator()(double time, PositionType x, ShapeValueType shape_val,
+                                      QFuncArgs... qfunc_args) const
+    {
+      auto unmodified_qf_returns = integrand_(time, x + shape_val, qfunc_args...);
+      auto correction = detail::compute_boundary_area_correction(x, shape_val);
+      return serac::tuple{correction * serac::get<0>(unmodified_qf_returns),
+                          correction * serac::get<1>(unmodified_qf_returns)};
+    }
+  };
+
+  /**
+   * @brief Adds an interior face integral term to the weak formulationx of the PDE
+   *
+   * @tparam dim The dimension of the element (2 for quad, 3 for hex, etc)
+   * @tparam args The type of the trial function input arguments
+   * @tparam lambda The type of the integrand functor: must implement operator() with an appropriate function signature
+   * @tparam domain_type The type of the integration domain (either serac::Domain or mfem::Mesh)
+   *
+   * @param[in] integrand The user-provided quadrature function, see @p Integral
+   * @param[in] domain The domain on which to evaluate the integral
+   *
+   * @note The @p Dimension parameters are used to assist in the deduction of the @a geometry_dim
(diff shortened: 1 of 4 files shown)


src/serac/physics/functional_residual.hpp
```diff
@@ -223,6 +223,10 @@ class FunctionalResidual<spatial_dim, ShapeSpace, OutputSpace, Parameters<InputS
     }
   }
 
+  /// @brief Accessor to get a reference to the underlying ShapeAwareFunctional in case more direct access is needed.
+  /// @return Reference to ShapeAwareFunctional instance.
+  ShapeAwareFunctional<ShapeSpace, OutputSpace(InputSpaces...)>& getFunctional() { return *residual_; }
```
btalamini, line 228: I trust that there is a good reason for the non-const reference. This undermines the abstraction, but would I be right to assume this is a temporary step to make incremental progress?  Is there a path forward to eventually dispensing with this?

src/serac/physics/mesh.cpp
```diff
@@ -38,12 +38,15 @@ void Mesh::createDomains()
...
   domains_.insert({entireBodyName(), serac::EntireDomain(*mfem_mesh_)});
   domains_.insert({entireBoundaryName(), serac::EntireBoundary(*mfem_mesh_)});
+  domains_.insert({internalBoundaryName(), serac::InteriorFaces(*mfem_mesh_)});
 }
 
 serac::Domain& Mesh::entireBody() const { return domain(entireBodyName()); }
 
 serac::Domain& Mesh::entireBoundary() const { return domain(entireBoundaryName()); }
 
+serac::Domain& Mesh::internalBoundary() const { return domain(internalBoundaryName()); }
```
btalamini, line 48: Is this the set of all inter-element boundaries? If so, I don't love this name. It's the same name that I would expect for surfaces embedded in the bulk, say for computing J-integrals. You don't need to address this now, but, if this is indeed the purpose, let's start thinking of a clearer name.

btalamini (approved):

src/serac/physics/functional_residual.hpp
tupek2: I suppose this is because I needed interior face integrals, but didn't want to add that to the FunctionalResidual interface proper.  This is kind of like a back door work around in case the interface is missing something.  I know the classic C++ idea is to hide as much as possible, but I'm not entirely all about that.  Non-const because sometimes one needs to add more integrals to the functional.

src/serac/physics/mesh.cpp
tupek2: functional calls it: interiorFaces.  but that is not quite right as for my use cases they are edges... internalElementBoundaries or something.  suggestions are welcome.  This whole PR is simply tracking things I ended up needing in smith for various things.

src/serac/physics/functional_residual.hpp
btalamini: Why *not* add it to the interface? If the interface is already too restrictive, before there are any real users, maybe we should extend it. I'd like to talk offline (not here on GitHub) about whatever the underlying need is  - I'd like to help out on this task if I can.

src/serac/numerics/functional/shape_aware_functional.hpp
```diff
@@ -459,6 +459,61 @@ class ShapeAwareFunctional<shape, test(trials...), exec> {
...
+    {
+      auto unmodified_qf_returns = integrand_(time, x + shape_val, qfunc_args...);
+      auto correction = detail::compute_boundary_area_correction(x, shape_val);
+      return serac::tuple{correction * serac::get<0>(unmodified_qf_returns),
+                          correction * serac::get<1>(unmodified_qf_returns)};
+    }
+  };
+
+  /**
+   * @brief Adds an interior face integral term to the weak formulationx of the PDE
```
white238:
```suggestion
   * @brief Adds an interior face integral term to the weak formulation of the PDE
```

src/serac/physics/mesh.cpp
```diff
@@ -69,4 +72,22 @@ serac::Domain& Mesh::addDomainOfBoundaryElements(const std::string& domain_name,
...
+  SLIC_ERROR_IF(domains_.find(domain_name) != domains_.end(),
+                axom::fmt::format("A domain named {0} already exists in mesh with tag {1}", domain_name, mesh_tag_));
+  domains_.emplace(domain_name, Domain::ofElements(*mfem_mesh_, func));
+  return domain(domain_name);
+}
+
+serac::Domain& Mesh::addDomainOfBodyElements(const std::string& domain_name,
+                                             std::function<bool(std::vector<vec3>, int)> func)
+{
+  SLIC_ERROR_IF(domains_.find(domain_name) != domains_.end(),
```
white238, line 87: Should there be a function for finding a domain?

Edit: I see there is one but how about checking if it exists.

white238 (approved):

src/serac/physics/mesh.cpp
tupek2: I think there could/should be something like "getValidDomainNames()"... At the moment, the user sort of has to keep track of their domain_names in use (which I suppose they were already sort of doing).  I'd suggest we add this when we (hopefully soon) start pushing the use of this mesh out throughout our examples and tests.

PR #1362, opened 2025-04-21
Tupek/redesign/objective2

tupek2:
Here is a version of the objective interface which is more consistent with the residual interface, accepts dt, is labeled as scalar objective, which will be in contrast with, perhaps, a vector constraint, as might be required for contact.


src/serac/physics/functional_objective.hpp
```diff
@@ -0,0 +1,124 @@
...
+#include "serac/physics/state/finite_element_dual.hpp"
+
+namespace serac {
+
+template <int spatial_dim, typename ShapeDispSpace, typename parameters = Parameters<>,
+          typename parameter_indices = std::make_integer_sequence<int, parameters::n>>
+class FunctionalObjective;
+
+/**
+ * @brief Construct a new FunctionalObjective object
```
white238: This comment doesn't seem to be in the right place.

src/serac/physics/functional_objective.hpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  /// @brief using
+  using FieldPtr = FiniteElementState*;
+
+  /**
+   * @brief register a custom domain integral calculation as part of the residual
+   *
+   * @tparam active_parameters a list of indices, describing which parameters to pass to the q-function
+   * @param body_name string specifing the domain to integrate over
```
white238:
```suggestion
   * @param body_name string specifying the domain to integrate over
```

src/serac/physics/functional_objective.hpp
```diff
@@ -0,0 +1,124 @@
...
+  mutable double dt_ = std::numeric_limits<double>::max();
+
+  /// @brief primary mesh
+  std::shared_ptr<Mesh> mesh_;
+
+  /// @brief scalar output shape aware functional
+  std::unique_ptr<ShapeAwareFunctional<ShapeDispSpace, double(InputSpaces...)>> objective_;
+};
+
+}  // namespace serac
```
white238:
```suggestion
}  // namespace serac

```

src/serac/physics/scalar_objective.hpp
```diff
@@ -0,0 +1,65 @@
...
+  using FieldPtr = FiniteElementState*;
+
+  /** @brief Virtual interface for computing objective gradient from a vector of serac::FiniteElementState*
+   *
+   * @param time time
+   * @param dt  time step
+   * @param fields vector of serac::FiniteElementState* as arguments to the residual
+   * @return double
+   */
+  virtual double evaluate(double time, double dt, const std::vector<FieldPtr>& fields) const = 0;
```
white238, line 46: What is the difference between `evaluate` and `gradient` below? Both are computing gradients but return different types...

src/serac/physics/scalar_objective.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+  /// @brief name
+  std::string name() const { return name_; }
+
+ private:
+  /// @brief name provided to objective
+  std::string name_;
+};
+
+}  // namespace serac
```
white238:
```suggestion
}  // namespace serac

```

src/serac/physics/tests/CMakeLists.txt
```diff
@@ -4,7 +4,16 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause) 
 
-set(physics_test_depends serac_physics serac_mesh gtest)
+set(physics_test_headers
+    physics_test_utils.hpp
+    )
+    
+blt_add_library(
```
white238, line 11: You don't need to create a header-only library unless you want to export it to other projects or if it is helpful in other places. Just include it in the cpp file.

src/serac/physics/tests/physics_test_utils.hpp
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2019-2025, Lawrence Livermore National Security, LLC and
+// other Serac Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file utils_physics_tests.hpp
```
white238:
```suggestion
 * @file physics_test_utils.hpp
```

src/serac/physics/tests/physics_test_utils.hpp
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+void pseudoRand(serac::FiniteElementDual& dual)
+{
+  int sz = dual.Size();
+  for (int i = 0; i < sz; ++i) {
+    dual(i) = -1.2 + 2.02 * (double(i) / sz);
+  }
+}
```
white238:
```suggestion
}

```

src/serac/physics/tests/test_kinematic_objective.cpp
```diff
@@ -0,0 +1,176 @@
...
+    }
+  }
+}
+
+int main(int argc, char* argv[])
+{
+  serac::ApplicationManager manager(argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
```
white238:
```suggestion
}

```

src/serac/physics/functional_objective.hpp
```diff
@@ -0,0 +1,124 @@
...
+ * @param shape_disp_space Shape displacement space
+ * @param input_mfem_spaces Vector of finite element spaces which are arguments to the residual
+ */
+template <int spatial_dim, typename ShapeDispSpace, typename... InputSpaces, int... parameter_indices>
+class FunctionalObjective<spatial_dim, ShapeDispSpace, Parameters<InputSpaces...>,
+                          std::integer_sequence<int, parameter_indices...>> : public ScalarObjective {
+ public:
+  using SpacesT = std::vector<const mfem::ParFiniteElementSpace*>;  ///< typedef
+
+  /// @brief construct a center of mass objective
```
btalamini: This is where the constructor docstrings should be

src/serac/physics/functional_objective.hpp
```diff
@@ -0,0 +1,124 @@
...
+  template <int... active_parameters, typename FuncOfTimeSpaceAndParams>
+  void addBodyIntegral(DependsOn<active_parameters...>, std::string body_name,
+                       const FuncOfTimeSpaceAndParams& qfunction)
+  {
+    objective_->AddDomainIntegral(serac::Dimension<spatial_dim>{}, serac::DependsOn<active_parameters...>{}, qfunction,
+                                  mesh_->domain(body_name));
+  }
+
+  /// @overload
+  virtual double evaluate(double time, double dt, const std::vector<FieldPtr>& fields) const
```
btalamini: These concrete implementations should probably be marked `override`, right?

src/serac/physics/scalar_objective.hpp
btalamini: I presume this is an error in the documentation comments. I think `evaluate` returns the value of the objective function, and only `gradient` returns the derivative. @tupek2 , can you confirm (and fix the comments if so)?

tupek2: My comments were wrong. fixing.

src/serac/physics/scalar_objective.hpp
```diff
@@ -0,0 +1,66 @@
...
+ * @brief Specifies interface for evaluating scalar objective from fields and their field gradients
+ */
+
+#pragma once
+
+#include <vector>
+#include "serac/physics/common.hpp"
+
+namespace mfem {
+class Vector;
```
btalamini, line 19: I'm a little uneasy moving to forward declarations of external library objects, mostly because I don't fully appreciate the potential downsides. I'm not requesting anything now, but we should figure out what our policy is.

btalamini (approved):

src/serac/physics/functional_objective.hpp
tupek2: There is no base objective addBodyIntegral.  So nothing to override.  This is specific to 'functional' objectives.  maybe we can eventually have a dfem version matching this interface too, but I'm not counting on them being the same.  The idea is you setup your objective/residual and on construction specify the math how ever that objective needs you to.  Then all downstream things like solver, time integrators, etc. work on the base  objective and residual interfaces.

src/serac/physics/scalar_objective.hpp
tupek2: I'd argue we should forward declare especially with external library objects... but read up on it.

src/serac/physics/tests/CMakeLists.txt
tupek2: I'm leaning toward leaving it as is.  I recall having compilation issues before I did this... maybe I was doing something else wrong, but I don't really want to go down a rabbit hole at this moment.

src/serac/physics/tests/test_kinematic_objective.cpp
```diff
@@ -0,0 +1,176 @@
...
+    using ObjectiveT = serac::FunctionalObjective<dim, VectorSpace, serac::Parameters<VectorSpace, DensitySpace>>;
+
+    double time = 0.0;
+    double dt = 0.0;
+    auto all_states = getPointers(states, params);
+    auto objective_states = {all_states[SHAPE_DISP], all_states[DISP], all_states[DENSITY]};
+
+    ObjectiveT::SpacesT param_space_ptrs{&all_states[DISP]->space(), &all_states[DENSITY]->space()};
+
+    ObjectiveT mass_objective("mass constraing", mesh, all_states[SHAPE_DISP]->space(), param_space_ptrs);
```
thartland: Typo: "mass constraing"

src/serac/physics/functional_objective.hpp
btalamini: My comment is about `evaluate()`, not `addBodyIntegral()`.

src/serac/physics/tests/test_kinematic_objective.cpp
```diff
@@ -0,0 +1,176 @@
...
+    params = {density};
+
+    std::string physics_name = "solid";
+    residual = constructResidual(physics_name);
+
+    params[0] = 1.2;  // set density before computing mass properties
+    constraints = constructConstraints();
+
+    // initialize displacement
+    states[FIELD::DISP].setFromFieldFunction([](serac::tensor<double, dim> x) {
```
thartland, line 137: Does this initialize the displacement field? Based on the enum, `FIELD::DISP` is `1`, since `states = {disp, velo, accel, shape_disp};` then isn't `states[FIELD::DISP]` actually `velo`?

src/serac/physics/tests/test_kinematic_objective.cpp
```diff
@@ -0,0 +1,176 @@
...
+  std::vector<std::shared_ptr<serac::ScalarObjective>> constraints;
+};
+
+TEST_F(ConstrainedResidualFixture, CanComputeResidualObjectivesAndTheirGradients)
+{
+  double time = 0.0;
+  double dt = 1.0;
+  auto all_states = getPointers(states, params);
+
+  serac::FiniteElementDual res_vector(states[0].space(), "residual");
```
thartland: For consistency perhaps access `states[0]` via the enum as elsewhere in this test.

src/serac/physics/tests/test_kinematic_objective.cpp
```diff
@@ -0,0 +1,176 @@
...
+    mesh = std::make_shared<serac::Mesh>(mfem::Mesh::MakeCartesian3D(6, 4, 4, element_shape, xlength, ylength, zlength),
+                                         "this_mesh_name", 0, 0);
+
+    serac::FiniteElementState disp = serac::StateManager::newState(VectorSpace{}, "displacement", mesh->tag());
+    serac::FiniteElementState velo = serac::StateManager::newState(VectorSpace{}, "velocity", mesh->tag());
+    serac::FiniteElementState accel = serac::StateManager::newState(VectorSpace{}, "acceleration", mesh->tag());
+    serac::FiniteElementState shape_disp = serac::StateManager::newState(VectorSpace{}, "shape_disp", mesh->tag());
+    serac::FiniteElementState density = serac::StateManager::newState(DensitySpace{}, "density", mesh->tag());
+
+    states = {disp, velo, accel, shape_disp};
```
thartland: Should this be consistent with ordering of the `FIELD` enum and be `states = {shape_disp, disp, velo, accel};`?

src/serac/physics/functional_objective.hpp
tupek2: Fixed

src/serac/physics/functional_objective.hpp
tupek2: Fixed.

src/serac/physics/scalar_objective.hpp
tupek2: Keeping as is.  Keep pushing the compilation into the .cpp files, out of headers.

src/serac/physics/tests/test_kinematic_objective.cpp
tupek2: Fixed, thanks.

src/serac/physics/tests/test_kinematic_objective.cpp
tupek2: chaned to states[DISP], which works I think.  Thanks.  The residual is 'dual' to the displacements in this physics.

src/serac/physics/tests/test_kinematic_objective.cpp
tupek2: I've made a bit of a conceptual change from the beginning.  Because ShapeAwareFunctional requires shape displacement to always go first.  I've just decided to always essentially treat it like a 'state'.  Otherwise there is a huge amount of copying and reordering of fields that has to happening, which is really confusing.  Eventually, I think we want 'shape_displacement' to be associated with the mesh object itself, and so it will removed from all these vectors of fields at that point in time.

tupek2: So, the 0th 'state field' is shape, then disp, then velo, etc.

src/serac/physics/tests/test_kinematic_objective.cpp
tupek2: Fantastic catch.  One could question how the tests were passing at all with this error.  The answer: in this tests, all those fields are the same space/size, and it was really just allocating their space.  More testing may be nice here eventually for more complex physics... Fixed for now though.

thartland (approved):

PR #1363, opened 2025-04-21
Tupek/print helper

tupek2:


src/serac/infrastructure/debug_print.hpp
```diff
@@ -16,6 +16,25 @@
 #include "serac/infrastructure/memory.hpp"
 #include "serac/numerics/functional/element_restriction.hpp"
 
+namespace serac {
+
+/**
+ * @brief get a readable sting_view of the underlying templated type
```
white238:
```suggestion
 * @brief get a readable string_view of the underlying templated type
```

src/serac/infrastructure/debug_print.hpp
```diff
@@ -140,4 +162,7 @@ void printCUDAMemUsage()
   std::cout << " Free Memory (MB): " << (freeBytes / 1024.0 / 1024.0) << std::endl;
   std::cout << " Used Memory (MB): " << (usedBytes / 1024.0 / 1024.0) << std::endl;
 }
+
 #endif
+
+}  // namespace serac
```
white238:
```suggestion
}  // namespace serac

```

white238 (approved):

src/serac/infrastructure/debug_print.hpp
```diff
@@ -16,6 +16,25 @@
...
+ * @brief get a readable string_view of the underlying templated type
+ * @tparam T the type to get a string name for
+ */
+template <class T>
+constexpr std::string_view type_name(const T& /*t*/)
+{
+  using namespace std;
+#ifdef __clang__
+  string_view p = __PRETTY_FUNCTION__;
+  return string_view(p.data() + 34, p.size() - 34 - 1);
```
white238: Explaining the magic numbers here would be great.

white238 (changes requested):
I'm playing with this and getting extra characters. The magic numbers look to be off.

btalamini:
> I'm playing with this and getting extra characters. The magic numbers look to be off.

I did the same this morning with Compiler Explorer and found the same. This is a great idea, we need to get it in. I think it can be fixed by replacing the magic numbers with offsets determined by the `find()` method.

src/serac/infrastructure/debug_print.hpp
```diff
@@ -16,6 +16,25 @@
...
 #include "serac/numerics/functional/element_restriction.hpp"
 
+namespace serac {
+
+/**
+ * @brief get a readable string_view of the underlying templated type
+ * @tparam T the type to get a string name for
+ */
+template <class T>
+constexpr std::string_view type_name(const T& /*t*/)
```
btalamini: For my money, I'd name this function simply `type` like in Python.

btalamini (changes requested):
Fantastic idea, given the degree of templating in our code. I have the same experience as Chris: the magic numbers are off. As a suggestion, I posted a Compiler Explorer link in our Teams PR channel that fixes it on at least that platform (both clang and gcc).

src/serac/infrastructure/debug_print.hpp
white238: @tupek2  and i went back and forth on naming... decided on `typeString` as a middle ground. But we could be convinced. My concern with `type` is it being slightly generic (though its guarded by the namespace) and maybe implies it is returning the actual type not a string of the type. I dunno....

btalamini: It’s not the hill I’m going to die on, either :) My gut response is that the return type is publicly conveyed by the function prototype, so no need to put it in the name. I’m comfortable with the name “type” from Python.

tupek2: I like typeString, then I dont have to dive to see what type it returns.  Not having to look at the return type of type() is nice.

btalamini: I see the return type on mouseover in vscode, which is easy to get when I want but without cluttering names in the code. Like I said though, not a strong preference in this case. Go with typeString.

white238 (approved):

src/serac/infrastructure/debug_print.hpp
```diff
@@ -16,14 +16,42 @@
 #include "serac/infrastructure/memory.hpp"
 #include "serac/numerics/functional/element_restriction.hpp"
```
btalamini, line 25:
```suggestion
#include <cxxabi.h>
```

btalamini: `abi::__cxa_demangle()` is provided by `cxxabi.h`. We're getting it here from some transitive include. I'd put it in explicitly in case someone upstream refactors.

btalamini (approved):
LGTM

src/serac/infrastructure/debug_print.hpp
white238: ah good catch.. forgot to copy/paste all my includes from my smaller example.

src/serac/infrastructure/tests/debug_print.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <gtest/gtest.h>
+
+#include "serac/infrastructure/debug_print.hpp"
+#include "serac/infrastructure/application_manager.hpp"
+#include "serac/serac_config.hpp"
+
+namespace serac {
+
+TEST(DebugPrint, typeString)
+{
```
chapman39, line 27: would it be worth testing a more complex templated object? i guess the downside of that is then this test would need to depend on serac physics cmake targets.

src/serac/infrastructure/debug_print.hpp
```diff
@@ -8,22 +8,57 @@
...
+
 /**
  * @brief write an array of values out to file, in a space-separated format
  * @tparam T the type of each value in the array
  * @param v the values to write to file
  * @param filename the name of the output file
  */
 template <typename T>
-void write_to_file(std::vector<T> v, std::string filename)
+void writeToFile(std::vector<T> v, std::string filename)
```
chapman39, line 61: are there any references to these functions in the repo?

white238: nope I grepped the code before i did this.

src/serac/infrastructure/tests/debug_print.cpp
white238: I was concerned about how differing compilers would end up with function signatures. You can see how even `std::string` ended up different on one compiler. I wanted to at least have a smoketest to have basic types to check if they were demangling at all. I am planning on trying to use this on basic embedded struct types for Quadrature Data so wanted to check that. Though now I remembered I didn't add that and will do now.

src/serac/infrastructure/debug_print.hpp
chapman39: unrelated to this pr then, but should we remove these?

white238: They were originally intended to be used when you needed to debug issues. I am fine with them staying.

PR #1365, opened 2025-04-23
Update tests/ examples to use new `serac::Mesh`

### LINKED ISSUE #1342: Add all examples as tests
is this example tested in ci?

_Originally posted by @chapman39 in https://github.com/LLNL/serac/pull/1312#discussion_r2003941438_

### LINKED ISSUE #1364: Bucking Cylinder example failing `std::out_of_range`
```
[meemee@rzgenie5:build-rzwhippet-toss_4_x86_64_ib-clang@14.0.6-debug]$ ./examples/buckling_cylinder
Logger activated: 'serac_serial_logger'
terminate called after throwing an instance of 'std::out_of_range'
  what():  _Map_base::at
[SIGNAL]: Received signal 6 (Aborted), exiting
```

chapman39:
- [x] Fixes #1364 
- [x] Fixes #1342
- [x] Rename `serac_mesh` target to `serac_mesh_utils`, to clarify it does not include the `serac::Mesh` class.
- [x] Use `serac::Mesh`across all tests and examples
- [x] Update docs


src/serac/numerics/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ set(numerics_sources
     petsc_solvers.cpp
     )
 
-set(numerics_depends serac_infrastructure serac_functional)
+set(numerics_depends serac_physics)
```
chapman39, line 26: Since `serac::Mesh` depends on the state manager, it means it depends on `serac_physics`. So, any test that has a mesh will depend on `serac_physics`.

It might make sense to move these tests from `numerics/tests` to `physics/tests`. That way the dependency tree matches the file structure. But, I have a feeling most tests aren't following this rule anyway and might just be better for a future PR.

src/serac/physics/mesh.hpp
```diff
@@ -14,6 +14,7 @@
 
 #include "mfem.hpp"
 #include "serac/numerics/functional/tensor.hpp"
+#include "serac/numerics/functional/domain.hpp"
```
chapman39, line 17: The `struct Domain;` forward declaration wasn't enough, it seems, since when I included `mesh.hpp` in some of the tests, I got a bunch of errors complaining about `Domain`.

src/serac/numerics/CMakeLists.txt
chapman39: nvm https://github.com/LLNL/serac/issues/1374

src/serac/physics/tests/quasistatic_solid_adjoint.cpp
```diff
@@ -221,22 +227,22 @@ TEST(quasistatic, finiteDifference)
 
   Estate = E0 + h;
   seracSolid->setParameter(0, Estate);
-  double fpE = forwardPass(seracSolid.get(), qoi.get(), meshPtr, nTimeSteps, timeStep, "fpE");
+  double fpE = forwardPass(seracSolid.get(), qoi.get(), &pmesh->mfemParMesh(), nTimeSteps, timeStep, "fpE");
```
tupek2: minor thing.  suggestion, you can go either way.  There would be less lines of changes if you just make a temporary meshPtr = &pmesh->mfemParMesh().

examples/contact/CMakeLists.txt
```diff
@@ -18,7 +18,12 @@ if(TRIBOL_FOUND)
         blt_add_executable(NAME        contact_${example_name}
                            SOURCES     ${filename}
                            OUTPUT_DIR  ${EXAMPLE_OUTPUT_DIRECTORY}
-                           DEPENDS_ON  serac_physics serac_mesh)
+                           DEPENDS_ON  serac_physics serac_mesh_utils)
+
+        # TODO Shorten runtime down to less than one minute
+        # blt_add_test(NAME            contact_${example_name}
+        #              COMMAND         contact_${example_name}
+        #              NUM_MPI_TASKS   4)
```
chapman39, line 26: ATM, these tests take too long to be in a regular CI job. Maybe we can have an option to tests examples nightly. Or we should somehow reduce the runtime of these examples similar to `--fast` in `cylinder.cpp`.

```
      Start 86: contact_beam_bending
86/93 Test #86: contact_beam_bending ....................   Passed   20.40 sec
      Start 87: contact_ironing
87/93 Test #87: contact_ironing .........................   Passed  277.27 sec
      Start 88: contact_sphere
88/93 Test #88: contact_sphere ..........................   Passed  294.65 sec
      Start 89: contact_twist
89/93 Test #89: contact_twist ...........................   Passed  143.91 sec
```

src/serac/mesh_utils/CMakeLists.txt
```diff
@@ -4,32 +4,34 @@
...
 
-set(mesh_depends serac_infrastructure)
+set(mesh_utils_depends serac_infrastructure)
 
 blt_add_library(
-    NAME        serac_mesh
-    HEADERS     ${mesh_headers}
-    SOURCES     ${mesh_sources}
-    DEPENDS_ON  ${mesh_depends}
+    NAME        serac_mesh_utils
```
chapman39, line 21: The motivation for renaming `serac_mesh` -> `serac_mesh_utils` is that with the introduction of `serac::Mesh` makes it is hard to differentiate what is "serac mesh". Adding "utils" makes it clearer that the CMake target just includes helper mesh functions and doesn't include `serac::Mesh`.

src/serac/physics/benchmarks/physics_benchmark_functional.cpp
```diff
@@ -29,14 +30,14 @@ void functional_test(int parallel_refinement)
   std::string filename =
       (dim == 2) ? SERAC_REPO_DIR "/data/meshes/star.mesh" : SERAC_REPO_DIR "/data/meshes/beam-hex.mesh";
 
-  auto mesh =
-      serac::mesh::refineAndDistribute(serac::buildMeshFromFile(filename), serial_refinement, parallel_refinement);
+  ::axom::sidre::DataStore datastore;
+  ::serac::StateManager::initialize(datastore, "sidreDataStore");
```
chapman39, line 34: Generally, I avoided adding `StateManager::initialize` unnecessarily. However, this benchmark already had `serac_physics`, so I figured it made sense to add it.

examples/buckling/CMakeLists.txt
```diff
@@ -8,9 +8,27 @@ if(TRIBOL_FOUND AND PETSC_FOUND)
...
+    #              COMMAND         buckling_cylinder --fast --contact --contact-type 1 --preconditioner 6 -options_file examples/cylinder_petsc_options.yml
+    #              NUM_MPI_TASKS   4)
+
+    blt_add_test(NAME            buckling_cylinder_no_contact
+                 COMMAND         buckling_cylinder --fast --no-contact
+                 NUM_MPI_TASKS   4)
+
+    configure_file(cylinder_petsc_options.yml
+                   ${PROJECT_BINARY_DIR}/examples/cylinder_petsc_options.yml
+                   COPYONLY)
```
white238:
```suggestion
    # TODO Fix this test
    # `InvIso: Newtons method did not converge.`
    # configure_file(cylinder_petsc_options.yml
    #                ${PROJECT_BINARY_DIR}/examples/cylinder_petsc_options.yml
    #                COPYONLY)

    # blt_add_test(NAME            buckling_cylinder_mortar_contact
    #              COMMAND         buckling_cylinder --fast --contact --contact-type 1 --preconditioner 6 -options_file examples/cylinder_petsc_options.yml
    #              NUM_MPI_TASKS   4)

    blt_add_test(NAME            buckling_cylinder_no_contact
                 COMMAND         buckling_cylinder --fast --no-contact
                 NUM_MPI_TASKS   4)
```

white238: If these options are only used in one example they should go with that example's directory... Am i missing another place?

examples/contact/CMakeLists.txt
white238: @ebchin

ebchin: We can coarsen the meshes and/or just run them with 1 timestep for CI, but it'd be good to run them all the way once in a while.  In my experience, convergence issues don't appear until towards the end of the simulation and with more refined meshes.  I guess I'd vote for a nightly test for examples?

white238 (approved):

examples/buckling/cylinder.cpp
```diff
@@ -80,61 +76,55 @@ int main(int argc, char* argv[])
...
   // Need to allow extra arguments for PETSc support
   app.set_help_flag("--help");
   app.allow_extras()->parse(argc, argv);
 
   nonlinear_options.force_monolithic = linear_options.preconditioner != Preconditioner::Petsc;
 
+  if (use_fast_options) {
+    dt = 1;
+    nonlinear_options.max_iterations = 5;
+    linear_options.max_iterations = 5;
```
tupek2, line 125: I have a silly question.  For these examples, do we just run them and make sure they exit cleanly, or is there any regression testing on the actual result that comes back?  If we are really just running them without checking their result at all, they are really just examples that need to keep compiling and not aborting.

examples/contact/CMakeLists.txt
tupek2: Do these test have a 'gold' value that we are comparing to?  If not, nightly or weekly, longer running tests don't add much anyways in my view.  Perhaps we need to start developing a regression testing system, where we establish some expected result to compare to.  Then, weekly running tests could catch changes in behavior.  Though, realistically, they should be run through CI.  Maybe we need to audit our CI for bottlenecks?  For example, do we currently launch our 93 test in parallel, or do they wait for eachother to complete?

src/serac/physics/mesh.hpp
tupek2: This include should not be needed.  The Domain object is not used in the header directly and its type comes from the forward-declaration.  This is another one of those dependency management habits, don't include files you don't need.  Forward-declare things as much as possible in .hpp and implementation details are in .cpp files.

tupek2 (approved):
Fantastic! Looks amazing to me.  Maybe not as much duplication removal as hoped, but there getting rid of the Domain is big, as users no longer have to keep them in scope.

examples/buckling/cylinder.cpp
chapman39: I added them as a test to simply verify they run without error, since beforehand they failed at runtime. I agree there is a chance these examples could start producing invalid answers without us knowing - as it is now.

If we were to test the result, we could use `gtest`, but that might be messy considering these are meant to be a reference for users. We could add some kind of `assert` or `SLIC_ERROR_IF` alternatively, but given these examples take a long time to run, it may not make sense to run the entire thing on-push. So maybe in a future PR I can work on moving these to nightly CI and add some `SLIC_ERROR_IF`'s?

src/serac/physics/mesh.hpp
chapman39: I think I tried that and it didn't work, but it's possible I did something wrong. I'll try again.

chapman39: fixed

examples/buckling/CMakeLists.txt
chapman39: i moved the yaml file to `examples/buckling`, though its now commented out as well

examples/buckling/cylinder.cpp
chapman39: https://github.com/LLNL/serac/issues/1393

PR #1367, opened 2025-04-23
Remove year from copyright

white238:


chapman39 (approved):
nice, i saw spack do this recently as well

chapman39 (changes requested):
i almost missed this, we should update the year here too https://github.com/LLNL/serac/blob/develop/LICENSE

comparing with spack again, they seem to add the date of creation only (no present year) https://github.com/spack/spack/blob/develop/COPYRIGHT

tupek2 (approved):
Great

white238:
> i almost missed this, we should update the year here too https://github.com/LLNL/serac/blob/develop/LICENSE
> 
> comparing with spack again, they seem to add the date of creation only (no present year) https://github.com/spack/spack/blob/develop/COPYRIGHT

I removed the year entirely in this branch and then read around and put it back in just that file.

chapman39 (approved):

PR #1370, opened 2025-04-25
Tupek/residual fast vjp

tupek2:

PR #1379, opened 2025-05-07
Li/heat new interface

lihanyu97:


src/serac/physics/heat_transfer_residual.hpp
```diff
@@ -0,0 +1,159 @@
...
+     * isotropic material
+     */
+    template <typename X, typename Temperature, typename dT_dt, typename... Params>
+    auto operator()(double /*time*/, X x, Temperature temperature, dT_dt dtemp_dt, Params... params) const
+    {
+      // Get the value and the gradient from the input tuple
+      auto [u, du_dX] = temperature;
+      auto du_dt = get<VALUE>(dtemp_dt);
+
+      auto [heat_capacity, heat_flux] = material_(x, u, du_dX, params...);
```
tupek2, line 164: @btalamini This is the pattern for thermal materials, returning a tuple.  In the new solid, I have 2 methods, 1 to get pkStress, the other density.  This could be material_.heat_flux(), and capacity(), or we could change the solid materials to return a pair/tuple like this instead?  But, I think being consistent is best, whichever way we go.

src/serac/physics/tests/test_heat_residual.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include "serac/mesh/mesh_utils.hpp"
+#include "serac/physics/materials/parameterized_thermal_material.hpp"
+#include "serac/physics/mesh.hpp"
+#include "serac/physics/state/state_manager.hpp"
+#include "serac/physics/heat_transfer_residual.hpp"
+
+auto element_shape = mfem::Element::QUADRILATERAL;
+
+template <typename T>
+auto getPointers(std::vector<T>& values)
```
tupek2, line 19: If we keep copying this pattern (essentially getting pointers into vectors to pass into the new interface), we should probably put it in some common location so the different tests and examples can use the same code.

src/serac/physics/tests/test_heat_residual.cpp
```diff
@@ -0,0 +1,231 @@
...
+  return pointers;
+}
+
+template <typename T>
+auto getPointers(T& v)
+{
+  return std::vector<T*>{&v};
+}
+
+void pseudoRand(serac::FiniteElementState& dual)
```
tupek2, line 48: same here, there are a few places for the rand-like things.  Mostly my fault, but we should consolidate.

src/serac/physics/heat_transfer_residual.hpp
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * @pre MaterialType must return a serac::tuple of volumetric heat capacity and thermal flux when operator() is called
+   * with the arguments listed above.
+   *
+   */
+  template <int... active_parameters, typename MaterialType>
+  void setMaterial(DependsOn<active_parameters...>, std::string body_name, const MaterialType& material)
+  {
+    ThermalMaterialFunctor<MaterialType> material_functor(material);
+    BaseResidualT::residual_->AddDomainIntegral(Dimension<dim>{},
```
tupek2, line 92: The implementation for these is changing very shortly in the fast_vjp PR.  I think you might want to wait, and update to that new pattern.

tupek2 (commented):
Looks good.  Some minor suggestions, but mainly I'm going to suggest it wait until the fast_vjp pr is through.  That way, someone else goes through that new pattern, so I know others can work with it, not just me.

src/serac/physics/heat_transfer_residual.hpp
btalamini: In the general case, you need to solve nonlinear ODE equations to get the outputs, so I wouldn't want the interface to force us to re-solve them for every output. So in general, I prefer the tuple. Density is a special case, because inertia is not really a constitutive response, it's a fundamental law, so it doesn't force a solve. So maybe that one can bear having the separate method.

src/serac/physics/tests/test_heat_residual.cpp
```diff
@@ -0,0 +1,231 @@
...
+    std::string physics_name = "heat";
+
+    using HeatResidualT = serac::HeatTransferResidual<order, dim, serac::Parameters<ParamSpace>>;
+    auto heat_transfer_residual = std::make_shared<HeatResidualT>(physics_name, mesh, states[SHAPE].space(),
+                                                                  states[TEMP].space(), getSpaces(params));
+
+    ThermalMaterial mat(1.0, 1.0, 1.0);
+    heat_transfer_residual->setMaterial(serac::DependsOn<0>{}, mesh->entireBodyName(), mat);
+
+    std::string source_name = "souce";
```
btalamini: not really important, but "source" is misspelled.

src/serac/physics/tests/test_heat_residual.cpp
```diff
@@ -0,0 +1,231 @@
...
+    EXPECT_NEAR(jvp_slow.Norml2(), jvp.Norml2(), 1e-12);
+  }
+}
+
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+  serac::ApplicationManager applicationManager(argc, argv);
+  return RUN_ALL_TESTS();
+}
```
btalamini:
```suggestion
}

```

btalamini (approved):
Wait for #1370 to go through and then add that capability. Nice work.

tupek2 (approved):

PR #1384, opened 2025-05-19
Prevent maybe-uninitialized error on GCC Release builds

chapman39:
I noticed this benchmark was failing: https://lc.llnl.gov/gitlab/smith/serac/-/jobs/2618056#L272

Seems like on GCC Release builds, there is a warning that others do not catch.

### DIFF
--- a/src/serac/numerics/equation_solver.cpp
+++ b/src/serac/numerics/equation_solver.cpp
@@ -628,7 +628,7 @@ class TrustRegion : public mfem::NewtonSolver {
     num_subspace_solves = 0;
     num_jacobian_assembles = 0;
 
-    real_t norm, norm_goal, prev_norm;
+    real_t norm, norm_goal, prev_norm = 0.0;
     norm = initial_norm = computeResidual(X, r);
     norm_goal = std::max(rel_tol * initial_norm, abs_tol);
     if (print_options.first_and_last && !print_options.iterations) {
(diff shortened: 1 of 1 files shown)


white238:
Are benchmarks allowed to fail or did this just fail on develop due to a system fluke?

white238 (approved):

chapman39:
> Are benchmarks allowed to fail or did this just fail on develop due to a system fluke?

Since benchmarks are run weekly, they are technically allowed to fail before being merged into develop. I don't think Gcc release builds are tested in CI on-push, which is how the warning escaped.

btalamini:
This was introduced in a very recent change. Good catch. I'm disturbed that our CI didn't catch a use of an uninitialized variable. @chapman39 or @white238 , can you look into this and see why the warning in clang was suppressed?

btalamini (approved):

btalamini:
On our clang version (14.0.0), compilation should fail for this if `-Wuninitialized` is set (or any flag that includes this) [(example)](https://www.godbolt.org/z/WGK38scze) Please change the CI builds to include this if we don't already have it.

btalamini (approved):
Good catch.

chapman39:
@btalamini since prev_norm is only printed if one iteration has past, prev_norm is guaranteed to be set before it is printed. so it is kind of weird the error is being produced. i have not been able to produce a warning in this case:

```cpp
    double norm = 1.0, prev_norm;
    for (int i = 0; i < 2; i++) {
        if (i > 0) {
            std::cout << "quotient: " << prev_norm/norm << std::endl;
        }
        prev_norm = norm;
    }
    return 0;
```

https://www.godbolt.org/z/vv3vWY5ba

tupek2:
I think this pattern was added to 2 solvers.  Maybe double check that the issue isn't in that file twice?

PR #1387, opened 2025-05-23
Merging scalar wave solver support files

chvillanuevap:
This branch includes supporting changes required for the scalar wave solver:
- Made checkpointed solution accessors public to allow test-time comparison with analytical solution.
- Modified boundary condition interface to support DOF lists and scalar fields (it only worked for vector fields before).

### DIFF
--- a/src/serac/physics/base_physics.hpp
+++ b/src/serac/physics/base_physics.hpp
@@ -481,6 +481,15 @@ class BasePhysics {
    */
   std::string name() const { return name_; }
 
+  /**
+   * @brief Accessor for getting all of the primal solutions from the physics modules at a given
+   * checkpointed cycle index
+   *
+   * @param cycle The cycle to retrieve state from
+   * @return A map containing the primal field names and their associated FiniteElementStates at the requested cycle
+   */
+  std::unordered_map<std::string, FiniteElementState> getCheckpointedStates(int cycle);
+
  protected:
   /**
    * @brief Create a paraview data collection for the physics package if requested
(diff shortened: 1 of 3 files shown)


src/serac/physics/base_physics.hpp
```diff
@@ -503,15 +512,6 @@ class BasePhysics {
...
   void initializeBasePhysicsStates(int cycle, double time);
 
-  /**
-   * @brief Accessor for getting all of the primal solutions from the physics modules at a given
-   * checkpointed cycle index
-   *
-   * @param cycle The cycle to retrieve state from
-   * @return A map containing the primal field names and their associated FiniteElementStates at the requested cycle
-   */
-  std::unordered_map<std::string, FiniteElementState> getCheckpointedStates(int cycle);
```
tupek2, line 513: I'm confused by this change?  We have other tests that use loadCheckpointedState(const std::string& state_name, int cycle), which is public.  Why not use that?

src/serac/physics/boundary_conditions/boundary_condition.cpp
```diff
@@ -99,7 +99,7 @@ void BoundaryCondition::setDofs(mfem::Vector& vector, const double time) const
     // a single component of a vector-valued function
     auto scalar_coef = get<std::shared_ptr<mfem::Coefficient>>(coef_);
     scalar_coef->SetTime(time);
-    if (component_) {
+    if (component_ || attr_markers_.Size() == 0) { // TODO: If attribute markers is empty, that means this is applied to the entire domain (component could be zero).
```
tupek2, line 102: why is this needed?  was there a bug?

PR #1391, opened 2025-05-30
Fix vjp dual types

tupek2:
The types where backward for the vjp.  The 'v', should be a ParGridFunction as its interpolated.  All derivatives are linear forms (FiniteElementDual).  This cleans up a few things in smith implementations, avoids copies, etc.

### DIFF
--- a/src/serac/physics/functional_residual.hpp
+++ b/src/serac/physics/functional_residual.hpp
@@ -241,8 +241,8 @@ class FunctionalResidual<spatial_dim, ShapeDispSpace, OutputSpace, Parameters<In
     SLIC_ERROR_IF(vs.size() != 1, "FunctionalResidual nonlinear systems only supports 1 output residual");
 
     dt_ = dt;
-    auto vecJacs = vectorJacobianFunctions(std::make_integer_sequence<int, sizeof...(input_indices) + 1>{}, time,
-                                           vs[0], fields);
+    auto vecJacs =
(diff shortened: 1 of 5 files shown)


btalamini (approved):

src/serac/physics/functional_residual.hpp
```diff
@@ -233,22 +233,22 @@ class FunctionalResidual<spatial_dim, ShapeDispSpace, OutputSpace, Parameters<In
   }
 
   /// @overload
-  void vjp(double time, double dt, const std::vector<FieldPtr>& fields, const std::vector<DualFieldPtr>& v_reactions,
-           const std::vector<FieldPtr>& vjp_fields) const override
+  void vjp(double time, double dt, const std::vector<FieldPtr>& fields, const std::vector<FieldPtr>& vs,
```
white238: I am a bit concerned how the doxygen docs are gonna map these renamed variables, ie. `vs` vs `vReactions` and `vjp_sensitivies` vs `vjpFields`.

https://github.com/LLNL/serac/blob/15bade029d4ba88ad4121957ef3925d636a8cbbc/src/serac/physics/residual.hpp#L84-L93

white238 (approved):

src/serac/physics/functional_residual.hpp
tupek2: Good catch, I will unify the naming between base class and derived.

PR #1392, opened 2025-05-31
Add BDF Integrator, Field Extrapolator, and Tensor Utilities

chvillanuevap:
Adds infrastructure to support higher-order time integration and multiphysics coupling:
- Backward Differentiation Formula (BDF) time integrator.
- Field extrapolation utility for multi-step schemes.
- Tensor utility functions to support physics modules.


src/serac/physics/base_physics.cpp
```diff
@@ -385,6 +385,88 @@ std::unordered_map<std::string, FiniteElementState> BasePhysics::getCheckpointed
   return previous_states_map;
 }
 
+::std::vector<::serac::FiniteElementState>
+BasePhysics::getPreviousStates(
```
tupek2, line 390: what is this for?  we are really trying to get away from this type of thing with the new design.

src/serac/numerics/extrapolation_operator.hpp
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @brief Implements a cycle-aware extrapolator for time-dependent fields.
+ *
+ * This operator handles extrapolation of any time-dependent field (e.g., pressure, velocity)
+ * using a compile-time specified maximum order and a runtime cycle count to determine
+ * how many states are available.
+ *
+ * @tparam r_order The desired extrapolation order (0 = zero, 1 = constant, 2 = linear).
+ */
+template <int r_order>
```
tupek2, line 29: what is the point of templating this?  Is there some performance optimization.  If not, I'd suggest just having it be a constructor argument or and input to the () operator.  In term, the details should be implemented in a .cpp file.

src/serac/physics/base_physics.hpp
```diff
@@ -446,6 +446,47 @@ class BasePhysics {
    */
   virtual double getCheckpointedTimestep(int cycle) const;
 
+  /**
```
tupek2, line 453: As a general rule:  The base_physics interface exists as a public API for external use.  At the moment, that external use is in LiDO.  Do these newly added functions need to be called by LiDO for your new physics to work?  If so, I'd like to understand that.  If LiDO does not need to call these functions directly, they should be implementation details of your particular physics so that we do not bloat the base_physics interface even more.

src/serac/physics/base_physics.hpp
```diff
@@ -446,6 +446,47 @@ class BasePhysics {
...
+   * @brief Computes the absolute change (L2 norm of the difference) between two successive states.
+   *
+   * This function compares checkpointed states at cycles `k` and `k-1` for a given field.
+   *
+   * @param cycle      The solution cycle `k` (must be >= 1).
+   * @param state_name The name of the checkpointed state field to compare.
+   *
+   * @return The absolute L2 norm of the difference: || u^k - u^{k-1} ||.
+   */
+  double
```
tupek2, line 470: Operations like this, that may be reused by other physics, should be provided in, e.g., numerics, as a free function.  Once again, we don't want to expand the public base_physics interface when its not strictly required.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -1890,6 +1954,38 @@ SERAC_HOST_DEVICE constexpr int size(const double&) { return 1; }
...
+ *
+ * @tparam T The datatype stored in the tensor.
+ * @tparam n The extents of each dimension.
+ *
+ * @return The number of dimensions (rank) of the tensor.
+ */
+template <typename T, int... n>
+SERAC_HOST_DEVICE int constexpr
+rank(
+  ::serac::tensor<T, n...> const &
```
tupek2, line 1968: stylistically, we don't really do the pre :: for ::serac or ::std in serac.  I'd argue its best to keep the style consistent, even though its not enforced by our style guide.

src/serac/numerics/functional/tensor.hpp
```diff
@@ -696,6 +704,62 @@ SERAC_HOST_DEVICE constexpr auto inner(const tensor<S, m>& A, const tensor<T, m>
  */
 SERAC_HOST_DEVICE constexpr auto inner(double A, double B) { return A * B; }
 
+/**
```
tupek2, line 707: These are confusing me (inner products with zeros).  I thought they were already in the repo?  I cannot figure out why its showing up here as a change.

PR #1398, opened 2025-06-03
dFEM build

ebchin:
Adds mfem w/dFEM to serac codevelop builds


cmake/thirdparty/SetupSeracThirdParty.cmake
```diff
@@ -195,6 +195,29 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
...
 
+    #------------------------------------------------------------------------------
+    # Enzyme
+    #------------------------------------------------------------------------------
+    if (ENZYME_DIR)
+        serac_assert_is_directory(DIR_VARIABLE ENZYME_DIR)
+        set(Enzyme_ROOT ${ENZYME_DIR} CACHE PATH "")
+        find_dependency(Enzyme REQUIRED)
+
+        # serac_assert_find_succeeded(PROJECT_NAME Enzyme
```
white238: Did this not work?

ebchin: No, it does.  dFEM doesn't use it, so I temporarily disabled it while I had Tribol manually disabled.  Added it back in.

PR #1401, opened 2025-06-05
Tupek/residual const and sign correctness

### LINKED ISSUE #1399: Make sure mesh is guaranteed to be in a valid state
Even if given a parallel mfem mesh.

tupek2:
The goal of this is

1. Modify the new residual and scalar_objective interfaces to that they are more const correct (the pointers are pointing to const things when fields are used as inputs).
2. Flip the sign of the solid mechanics and thermal residuals.  This means tractions and body forces will now have the sign convention that most people know and expect.
3. We were not calling the   ->EnsureNodes() and ->ExchangeFaceNbrData() when the new serac::mesh structure is passed a ParMesh.  Now we will do it, to ensure the mesh is in a valid state.
4. There was a lot of duplicated functions to create vectors of pointers for the new interface.  These are now unified and available in a new header file, along with useful and consolidated type information.

closes #1399

### DIFF
--- a/src/serac/physics/field_types.hpp
+++ b/src/serac/physics/field_types.hpp
@@ -32,7 +32,7 @@ using ConstDualFieldPtr = FiniteElementDual const*;
 /// @brief Get a vector of FieldPtr or DualFieldPtr from a vector of shared_pointers to FiniteElementState or
 /// FiniteElementDual
 template <typename T>
-auto getFieldPointers(std::vector<std::shared_ptr<T>>& states, std::vector<std::shared_ptr<T>>& params)
+auto getFieldPointers(const std::vector<std::shared_ptr<T>>& states, const std::vector<std::shared_ptr<T>>& params)
 {
   static_assert(std::is_same_v<T, FiniteElementState> || std::is_same_v<T, FiniteElementDual>,
                 "Type must be either FiniteElementState or FiniteElementDual");
@@ -50,6 +50,8 @@ auto getFieldPointers(std::vector<std::shared_ptr<T>>& states, std::vector<std::
 template <typename T>
 auto getFieldPointers(std::vector<T>& states, std::vector<T>& params)
 {
+  static_assert(std::is_same_v<T, FiniteElementState> || std::is_same_v<T, FiniteElementDual>,
+                "Type must be either FiniteElementState or FiniteElementDual");
   std::vector<T*> pointers;
   for (auto& t : states) {
     pointers.push_back(&t);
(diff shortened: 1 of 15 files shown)


src/serac/physics/field_types.hpp
```diff
@@ -0,0 +1,106 @@
...
+/// @brief using
+using ConstFieldPtr = FiniteElementState const*;
+
+/// @brief using
+using ConstDualFieldPtr = FiniteElementDual const*;
+
+/// @brief Get a vector of FieldPtr or DualFieldPtr from a vector of shared_pointers to FiniteElementState or
+/// FiniteElementDual
+template <typename T>
+auto getFieldPointers(std::vector<std::shared_ptr<T>>& states, std::vector<std::shared_ptr<T>>& params)
```
btalamini: Should `states` and `params` be marked as `const` in these functions?

src/serac/physics/field_types.hpp
```diff
@@ -0,0 +1,106 @@
...
+  for (auto& t : params) {
+    pointers.push_back(t.get());
+  }
+  return pointers;
+}
+
+/// @brief Get a vector of FieldPtr or DualFieldPtr from a vector of FiniteElementState or FiniteElementDual
+template <typename T>
+auto getFieldPointers(std::vector<T>& states, std::vector<T>& params)
+{
```
btalamini, line 52: In the previous function, you statically checked the template type. Why not here?

src/serac/physics/tests/test_functional_residual.cpp
```diff
@@ -60,13 +25,13 @@ struct ResidualFixture : public testing::Test {
 
   enum STATE
   {
+    SHAPE_DISP,
     DISP,
     VELO
   };
```
btalamini, line 31: Why make this enum? Is the one in the SolidResidual class not sufficient?

btalamini (approved):

src/serac/physics/tests/test_functional_residual.cpp
tupek2: Its not using (or even including) solid_residual.  those are sort of made up generic names for some residual.

src/serac/physics/field_types.hpp
tupek2: Yes, I meant to put those all over.

src/serac/physics/field_types.hpp
tupek2: I was missing a few consts.  Some could not be const.

src/serac/physics/field_types.hpp
tupek2: Fixed now

src/serac/physics/heat_transfer_residual.hpp
```diff
@@ -163,7 +172,7 @@ class HeatTransferResidual<order, dim, Parameters<InputSpaces...>>
 
       auto [heat_capacity, heat_flux] = material_(x, u, du_dX, params...);
 
-      return serac::tuple{heat_capacity * du_dt, -1.0 * heat_flux};
+      return serac::tuple{-heat_capacity * du_dt, heat_flux};
```
lihanyu97, line 175: the reason there is a "-1.0" in front of heat_flux is because physically, the flux is computed by "q = -\kappa \nabla u" and this is what the material functor returns. In the mean time, the variational form for diffusion is "(\kappa \nabla u, \nabla v) = (-1.0 * q, \nabla v)"

lihanyu97: I guess you are moving every term to the RHS to calculate the residual. In that sense, this is correct

lihanyu97 (approved):

src/serac/physics/heat_transfer_residual.hpp
tupek2: yeah, I'm switching the convention of the residual from what it has been.  This will make Neumann conditions more intuitive.  Thanks.

PR #1402, opened 2025-06-05
Restore tdof ebc method

### LINKED ISSUE #1366: Tdof BC interface
Consider adding back to base_physics

btalamini:
Restore the ability to set boundary conditions by a list of true dofs in the `SolidMechanics` class.

Removed the method to set boundary conditions by a spatial range query. This overlapped in capability with the domain-based method, and was somewhat dangerous  in that you can set BCs on part of an element edge.

The unit test checks that setting the boundary conditions with the new method gets you the exact same answer as setting them through the standard way (using domains). The particular problem is just made up.

closes #1366


src/serac/physics/solid_mechanics.hpp
```diff
@@ -490,50 +490,42 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+   * @note The displacement function is required to be vector-valued. However, only the dofs specified in the @a
+   * true_dofs array will be set. This means that if the @a true_dofs array only contains dofs for a specific vector
+   * component in a vector-valued finite element space, only that component will be set.
    *
    * @note This method must be called prior to completeSetup()
    */
-  void setDisplacementBCs(std::function<bool(const mfem::Vector&)> is_node_constrained,
-                          std::function<double(const mfem::Vector&, double)> disp, int component)
+  template <typename AppliedDisplacementFunction>
+  void setDisplacementBCsByDofList(AppliedDisplacementFunction applied_displacement, const mfem::Array<int> true_dofs)
```
tupek2, line 516: could we implement this in base_physics, so that it works for thermal and other physics as well by default?

btalamini: Certainly. I put it here mechanically because that is where it was originally, but base physics is a better idea.

btalamini: actually ... this wouldn't be convenient to do now. The problem is that the coefficient function needs to know the vector dimension of the field that is being set, and base_physics doesn't know that. We *could* do it, if we were to take a container of values for the bcs. I think we should do that soon, but not in this PR.

kswartz92 (approved):

tupek2 (approved):

PR #1403, opened 2025-06-06
Tupek/shape on mesh

tupek2:
This change moves the shape_displacement (and its dual, gradient wrt shape_displacement) onto the new mesh class, as shape_disp is naturally associated with the mesh coordinates.  This gets shape_displacement out of a huge number of interface functions.  In addition, base_physics now holds a shared_ptr to the mesh to get these shape displacements, etc.  This completely removes all direct usage of domains, which should simply our transition to dfem, and also makes the ownership clearer.

Finally, by now requiring the new mesh interface, we can ensure that users have setup the mesh correctly.


src/serac/physics/base_physics.hpp
```diff
@@ -226,7 +228,7 @@ class BasePhysics {
    *
    * @return The shape displacement finite element state
    */
-  virtual const FiniteElementState& shapeDisplacement() const { return shape_displacement_; }
+  virtual const FiniteElementState& shapeDisplacement() const;
```
btalamini, line 231: Should this be `virtual`? Is there any good reason for a base class to override it?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -1617,7 +1527,7 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     // Get the nodal positions for the displacement vector in grid function form
     mfem::ParGridFunction nodal_positions(
         const_cast<mfem::ParFiniteElementSpace*>(&displacement_.space()));  // MRT mfem const correctness issue
-    mesh_.GetNodes(nodal_positions);
+    mesh().GetNodes(nodal_positions);
```
btalamini: I think this whole method is dead code. It was used only in the "apply bcs by range query" method that I removed. I should have removed this, too.

src/serac/physics/solid_residual.hpp
```diff
@@ -31,11 +31,11 @@ class SolidResidual;
  */
 template <int order, int dim, typename... InputSpaces>
 class SolidResidual<order, dim, Parameters<InputSpaces...>>
-    : public FunctionalResidual<dim, H1<order, dim>, H1<order, dim>,
+    : public FunctionalResidual<dim, H1<order, dim>,
```
btalamini, line 34: Well this is nice.

examples/buckling/cylinder.cpp
```diff
@@ -133,23 +133,21 @@ int main(int argc, char* argv[])
 
   // Create and refine mesh
   std::string filename = SERAC_REPO_DIR "/data/meshes/hollow-cylinder.mesh";
-  auto mesh = serac::buildMeshFromFile(filename);
-  auto refined_mesh = mesh::refineAndDistribute(std::move(mesh), serial_refinement, parallel_refinement);
-  auto& pmesh = serac::StateManager::setMesh(std::move(refined_mesh), mesh_tag);
+  auto mesh = std::make_shared<serac::Mesh>(filename, mesh_tag, serial_refinement, parallel_refinement);
```
btalamini, line 136: This is great! The boilerplate is gone now.

btalamini (approved):
The changes had to be propagated far, but they're straightforward. This looks clearly better to me.

src/serac/physics/base_physics.hpp
tupek2: This is required now for certain things in the new smith interfaces.  Its possible that eventually this can be non-virtual... but that day is not today.

src/serac/physics/benchmarks/physics_benchmark_thermal.cpp
```diff
@@ -48,8 +48,7 @@ void functional_test_static()
 
   // Construct a functional-based heat transfer solver
   serac::HeatTransfer<p, dim> thermal_solver(serac::heat_transfer::default_nonlinear_options, linear_options,
-                                             serac::heat_transfer::default_static_options, "thermal_functional",
-                                             "default_mesh");
+                                             serac::heat_transfer::default_static_options, "thermal_functional", pmesh);
```
chapman39: would it make sense to rename all `pmesh`'s to `mesh`, if they're the new `serac::Mesh`?

src/serac/physics/base_physics.cpp
```diff
@@ -14,33 +14,21 @@
...
-BasePhysics::BasePhysics(std::string physics_name, std::string mesh_tag, int cycle, double time,
+BasePhysics::BasePhysics(std::string physics_name, std::shared_ptr<serac::Mesh> mesh, int cycle, double time,
                          bool checkpoint_to_disk)
     : name_(physics_name),
-      mesh_tag_(mesh_tag),
-      mesh_(StateManager::mesh(mesh_tag_)),
-      comm_(mesh_.GetComm()),
-      shape_displacement_(StateManager::shapeDisplacement(mesh_tag_)),
-      bcs_(mesh_),
+      mesh_tag_(mesh->tag()),
```
white238: Is this member variable needed anymore? For that matter, any of the member variables that are directly from the mesh class...

src/serac/physics/heat_transfer.hpp
```diff
@@ -159,7 +161,7 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
              *nonlin_solver_, bcs_)
   {
     SLIC_ERROR_ROOT_IF(
-        mesh_.Dimension() != dim,
+        mesh().Dimension() != dim,
```
white238: I am concerned about the subtle difference between when you need to call `mesh()` vs using `mesh_` directly.

Should we have different functions for our mesh vs what `mesh()` returns, which is `mfem::ParMesh&`?

src/serac/physics/solid_mechanics.hpp
```diff
@@ -150,7 +151,7 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
    * @param solver The nonlinear equation solver for the implicit solid mechanics equations
    * @param timestepping_opts The timestepping options for the solid mechanics time evolution operator
    * @param physics_name A name for the physics module instance
-   * @param mesh_tag The tag for the mesh in the StateManager to construct the physics module on
+   * @param serac_mesh Serac mesh used for physics
```
white238, line 154: I think this should just be `mesh`

src/serac/physics/tests/lce_Brighenti_tensile.cpp
```diff
@@ -29,25 +30,28 @@ TEST(LiquidCrystalElastomer, Brighenti)
...
+      mesh_tag);
+
+  mesh->addDomainOfBoundaryElements("xmin_face", by_attr<dim>(5));
+  mesh->addDomainOfBoundaryElements("ymin_face", by_attr<dim>(2));
+  mesh->addDomainOfBoundaryElements("zmin_face", by_attr<dim>(1));
+  mesh->addDomainOfBoundaryElements("xmax_face", by_attr<dim>(3));
+  mesh->addDomainOfBoundaryElements("ymax_face", by_attr<dim>(4));
+  mesh->addDomainOfBoundaryElements("zmax_face", by_attr<dim>(6));
+
+  // auto xmin_face = Domain::ofBoundaryElements(mesh, by_attr<dim>(5));
```
white238: Delete?

white238:
This is great! I just have concerns that I have noted above that we may be adding (maybe not fixing pre-existing) confusion based on our mesh class vs MFEM's. I am up for a discussion at any point if it's needed.

src/serac/physics/base_physics.cpp
tupek2: not really. removing

src/serac/physics/heat_transfer.hpp
tupek2: I renamed this to mfemParMesh.  I added a new method mesh() which returns serac::Mesh.  This simplifies some tests.

src/serac/physics/solid_mechanics.hpp
tupek2: The name has to be distinct.  mesh_ is a member variable.  mesh() is a base class function.

src/serac/physics/tests/lce_Brighenti_tensile.cpp
tupek2: yep

src/serac/physics/solid_mechanics.hpp
tupek2: I will let you remove this in a follow on, as I dont understand it as well and there is enough change here.

src/serac/physics/benchmarks/physics_benchmark_thermal.cpp
tupek2: yes, I've tried.  I may have missed a few more.  But I did just update this one.

src/serac/physics/base_physics.cpp
```diff
@@ -46,9 +45,11 @@ int BasePhysics::minCycle() const { return min_cycle_; }
 
 const std::vector<double>& BasePhysics::timesteps() const { return timesteps_; }
 
-const mfem::ParMesh& BasePhysics::mesh() const { return mesh_->mfemParMesh(); }
+const serac::Mesh& BasePhysics::mesh() const { return *mesh_; }
```
white238, line 48: Should this return the shared pointer?

tupek2: its an option... I don't have a strong view.  I suppose returning a shared_ptr in a way makes it like a non-const access (as someone could muck with the Mesh on the outside).  I was 50/50 on that.

white238: If we don't want the user to be able to get it back out of the class, would be better to have it be a unique_ptr then? I feel like we either need to return the thing we are sending in, or use something else. Anyone else have thoughts? @btalamini @tepperly ?

tupek2: unique_ptr indicates that the base_physics is giving up ownership, which is not the intent.
shared_ptr indicates that some class can take join ownership.  This more or less invalidates the const on the method.
const & indicates you can get a handle to it, but will not own.  Also, you can only use const methods of that class.  This is basically the intent.

btalamini: I agree, we want const references from getters like these.

white238: I'm fine with that. I just wanted to have a conversation about this. The users can always hold onto the original shared pointer.

white238 (approved):

chapman39 (approved):

PR #1404, opened 2025-06-09
Fix ARPACK include path

ebchin:
ARPACK 3.9.1 changed the include path.  See https://github.com/opencollab/arpack-ng/commit/459f46a6dd6b9ea4b00fbdaee10d7d146edec87a.  This adds the new path so the `FindARPACK.cmake` script works for old and new versions.


chapman39 (approved):
thanks!

chapman39 (approved):

white238 (approved):

tupek2 (approved):

PR #1405, opened 2025-06-10
Constraint interface

thartland:
This PR adds a constraint interface. 

For context -- we currently have the ability to specify a fixed (mesh-independent) number of [ScalarObjective](https://github.com/LLNL/serac/blob/develop/src/serac/physics/scalar_objective.hpp) constraints in which the constraints are not distributed among MPI processes as in this [test](https://github.com/LLNL/serac/blob/develop/src/serac/physics/tests/test_kinematic_objective.cpp#L59).

One use case for the new constraint interface is for a mesh-dependent number of constraints wherein we want the constraints to be distributed among MPI processes and so that the rows of the constraint Jacobian are owned by different MPI processes. 

constraints.hpp includes abstract class structure for a constraint
1. Evaluation of the constraint (required)
2. Evaluation of the jacobian of the constraint (required)
3. Evaluation of an approximation constraint Jacobian (option, defaults to true constraint Jacobian in 2.)
4. Constraint contribution to the residual, which in the context of optimization is (constraint Jacobian^T * Lagrange multipliers)
5. Jacobian of the residual contribution. I did not make this a pure virtual function since a "solver" may not need Jacobian of the residual contribution (e.g., if the "solver" utilizes a Gauss-Newton type approach). This Jacobian may also not be available since it would typically involve 2nd derivatives of the constraints.


thartland:
Design question: should a constraint enforcement method, such as `LagrangeMultiplier` or `Penalty`, be associated with a `serac::Constraint`?

src/serac/physics/constraint.hpp
```diff
@@ -0,0 +1,124 @@
...
+   * @param dt time step
+   * @param fields vector of serac::FiniteElementState*
+   * @param direction index for which field to take the gradient with respect to
+   * @return std::unique_ptr<mfem::HypreParMatrix>
+   */
+  virtual std::unique_ptr<mfem::HypreParMatrix> jacobian_tilde(double time, double dt,
+                                                               const std::vector<ConstFieldPtr>& fields,
+                                                               int direction) const
+  {
+    return jacobian(time, dt, fields, direction);
```
tupek2, line 73: am I remembering correctly?  typically jac_tilde will be transposed as its the 0,1 block?  but here we return the untransposed one.  I think that is fine, but just want to make sure I understand.

src/serac/physics/constraint.hpp
```diff
@@ -0,0 +1,124 @@
...
+   * @param multipliers mfem::Vector of Lagrange multipliers
+   * @param direction index for which field to take the gradient with respect to
+   * @return std::unique_ptr<mfem::HypreParMatrix>
+   */
+  virtual std::unique_ptr<mfem::HypreParMatrix> residual_contribution_jacobian(
+      [[maybe_unused]] double time, [[maybe_unused]] double dt,
+      [[maybe_unused]] const std::vector<ConstFieldPtr>& fields, [[maybe_unused]] const mfem::Vector& multipliers,
+      [[maybe_unused]] int direction) const
+  {
+    SLIC_ERROR_ROOT(axom::fmt::format("Base class must override residual_contribution_jacobian before usage"));
```
tupek2, line 112: this is the one that needs second derivative info for the constraints?

tupek2 (approved):
Good progress.  I think we should discuss moving toward running a full application at some point!  Maybe integrating this version of the interface with the complementarity solvers as well.

src/serac/physics/constraint.hpp
thartland: Yes, Jacobian tilde will show up in the 0, 1 block via a transpose and will be an approximation of the transpose of the true Jacobian that sits in the 1, 0 block.

src/serac/physics/constraint.hpp
thartland: Yes, the idea is that residual contribution will be the Jtilde^T (Lagrange multipliers) contribution to the gradient of the Lagrangian. The Jacobian of this object (with respect to the independent variables of the constraint function) will involve second derivative terms. I chose to not make it a pure virtual function because perhaps not all optimizers will need such second derivative information.

thartland:
> Good progress. I think we should discuss moving toward running a full application at some point! Maybe integrating this version of the interface with the complementarity solvers as well.

I agree. I'll set up a meeting.

thartland:
The [blueos-clang_14_0_5-src](https://lc.llnl.gov/gitlab/smith/serac/-/jobs/2891131) failure shouldn't be concerning given that we intend to disable this test, see [here](https://github.com/LLNL/serac/pull/1453).

examples/contact/constraint_twist.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <string>
+
+#include "axom/slic.hpp"
+
+#include "mfem.hpp"
+
+#include "serac/serac.hpp"
+
+int main(int argc, char* argv[])
+{
```
ebchin, line 20: We should have this eventually replace twist.cpp (once we have solvers that work with the new constraint interface).  Add a note so we know that is the eventual goal.

examples/contact/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ if(TRIBOL_FOUND AND STRUMPACK_DIR)
         ironing.cpp
         sphere.cpp
         twist.cpp
+	constraint_twist.cpp
```
ebchin: Fix indent

src/serac/physics/tests/test_contact_constraint.cpp
```diff
@@ -0,0 +1,107 @@
...
+  auto gaps_tribol = constraint->evaluate(time, dt, input_fields);
+
+  // gap should be -(tributary area * interpen).  for the unit cube mesh, tributary area should be 0.25.
+  // only check nonmortar dofs
+  constexpr int num_mortar_dofs = 4;
+  for (int i{num_mortar_dofs}; i < gaps_tribol.Size(); ++i) {
+    EXPECT_NEAR(gaps_tribol[i], -0.25 * interpen, 1.0e-12);
+  }
+}
+
```
ebchin, line 101: Add a test for Jacobian contributions (I can do this)

src/serac/physics/constraint.hpp
```diff
@@ -0,0 +1,124 @@
...
+   * @param dt time step
+   * @param fields vector of serac::FiniteElementState*
+   * @param direction index for which field to take the gradient with respect to
+   * @return std::unique_ptr<mfem::HypreParMatrix>
+   */
+  virtual std::unique_ptr<mfem::HypreParMatrix> jacobian(double time, double dt,
+                                                         const std::vector<ConstFieldPtr>& fields,
+                                                         int direction) const = 0;
+
+  /** @brief Virtual interface for computing constraint Jacobian_tilde from a vector of serac::FiniteElementState*
```
ebchin, line 60: Add a definition for Jacobian tilde

src/serac/physics/tests/test_contact_constraint.cpp
ebchin: Let's defer this for the next PR

src/serac/physics/constraint.hpp
thartland: Added

examples/contact/constraint_twist.cpp
thartland: Note has been added.

ebchin (approved):

PR #1415, opened 2025-06-24
DG parallelization

lihanyu97:
This is work in progress to parallelize DG scheme relying MFEM's parallel communication infrastructure (ExchangeFaceNbrData). The first phase of this work to configure residual computation in parallel is complete. The main effort is to properly prepare input data that includes ghost dofs for Functional and record the correct index in element restriction to ensure the evaluation and AD at each quadrature point accesses the right ghost dof value. The comment associated with each major implementation is included.

Please point out anything that is obviously wrong or clumsy, and I will avoid them in the next phase.


src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -545,6 +519,11 @@ axom::Array<DoF, 2, serac::detail::host_memory_space> GetFaceDofs(const serac::f
         // 3. get the dofs for the entire element
         mfem::Array<int> elem_dof_ids;
         fes->GetElementDofs(elem, elem_dof_ids);
+        // // Debug hanyu
```
tupek2: can we remove?

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -545,6 +519,11 @@ axom::Array<DoF, 2, serac::detail::host_memory_space> GetFaceDofs(const serac::f
...
+        // mpi::out << " Shared face = " << info.IsShared() << " Face id = " << f << " Elem id = " << elem << std::endl
```
tupek2: good to remove, or will be helpful in the next phase?

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -695,6 +711,9 @@ void ElementRestriction::ScatterAdd(const mfem::Vector& E_vector, mfem::Vector&
       for (uint64_t j = 0; j < nodes_per_elem; j++) {
         uint64_t E_id = (i * components + c) * nodes_per_elem + j;
         uint64_t L_id = GetVDof(dof_info(i, j), c).index();
+        // // debug
```
tupek2: remove?

src/serac/numerics/functional/functional.hpp
```diff
@@ -247,23 +292,36 @@ class Functional<test(trials...), exec> {
...
     std::array<Family, num_trial_spaces> trial_families = {trials::family...};
     std::array<int, num_trial_spaces> trial_orders = {trials::order...};
     std::array<int, num_trial_spaces> trial_components = {trials::components...};
+
     for (uint32_t i = 0; i < num_trial_spaces; i++) {
       trial_function_spaces_[i] = {trial_families[i], trial_orders[i], trial_components[i]};
+
+      P_trial_[i] = trial_space_[i]->GetProlongationMatrix();
+
+      // For L2 spaces, configure a ParGridFunction with external data to exchange face neighbor data.
```
tupek2, line 325: will any of this introduce an extra cost when using L2 elements that do not require interior surface integrals?  For example, in top opt, the density is often L2, but no dg jump terms are needed.  Are things like this likely to add significant cost or memory?  Or only when DG terms are added?

src/serac/numerics/functional/functional.hpp
```diff
@@ -357,10 +424,10 @@ class Functional<test(trials...), exec> {
     std::vector<uint32_t> arg_vec = {args...};
     for (uint32_t i : arg_vec) {
       domain.insert_restriction(trial_space_[i], trial_function_spaces_[i]);
-      check_interior_face_compatibility(domain.mesh_, trial_function_spaces_[i]);
+      // check_interior_face_compatibility(domain.mesh_, trial_function_spaces_[i]);
```
tupek2: why turn this off?  performance?

src/serac/numerics/functional/functional.hpp
```diff
@@ -466,36 +565,61 @@ class Functional<test(trials...), exec> {
...
+        }
       } else {
-
         P_trial_[i]->Mult(*input_T[i], input_L_[i]);
-
       }
-#else
-      P_trial_[i]->Mult(*input_T[i], input_L_[i]);
-#endif
+      // // Debug
```
tupek2: remove?

src/serac/numerics/functional/functional.hpp
```diff
@@ -515,13 +639,30 @@ class Functional<test(trials...), exec> {
       output_E_buffer_.SetSize(int(G_test.ESize()));
       output_E_.Update(output_E_buffer_, G_test.bOffsets());
       integral.Mult(t, input_E_, output_E_, wrt, update_qdata_);
+      // // Debug
```
tupek2: remove?

src/serac/numerics/functional/tests/dg_restriction_operators.cpp
```diff
@@ -351,12 +353,12 @@ void parametrized_test(int permutation)
 
   using space = L2<polynomial_order>;
 
-  auto pmesh = mesh::refineAndDistribute(std::move(mesh), 0);
-  Domain pinterior_faces = InteriorFaces(*pmesh);
+  // auto pmesh = mesh::refineAndDistribute(std::move(mesh), 0);
```
tupek2: not needed because you call refineAndDistibute... so might as well delete

src/serac/numerics/functional/tests/dg_restriction_operators.cpp
```diff
@@ -351,12 +353,12 @@ void parametrized_test(int permutation)
...
+  // Domain pinterior_faces = InteriorFaces(*pmesh);
 
-  auto L2_pfes = mfem::ParFiniteElementSpace(pmesh.get(), L2_fec.get());
+  // auto L2_pfes = mfem::ParFiniteElementSpace(pmesh.get(), L2_fec.get());
```
tupek2: delete

src/serac/numerics/functional/tests/functional_basic_dg.cpp
```diff
@@ -69,8 +69,8 @@ void L2_test(std::string meshfile)
 
   double t = 0.0;
 
-  auto value = residual(t, U);
-  // check_gradient(residual, t, U);
+  // auto value = residual(t, U);
```
tupek2, line 72: why is value commented out now?  And why was check_gradient commented out before!?

tupek2 (commented):
So, maybe I'm going dense... but where is the test that you get the same field solution on both sides of the dg interface for pre-specified linear/quadratic fields?

Looks like amazing progress.  Are you intending to push this in incrementally, or wait until the jacobian is done?

src/serac/numerics/functional/element_restriction.cpp
lihanyu97: Yeah I might still need these outputs in the next phase so I marked a "debug" here. I will remove all the "debug" after fixing the jacobian

src/serac/numerics/functional/functional.hpp
lihanyu97: It shouldn't add significant memory or computational cost. 1. we are using prolongation matrix to go from True dof to local dof in the first place. 2. we are not constructing a fresh `pargridfunction`. This variable serves as a cache. We link these pargridfunction to external data, aka the local vector we got from prolongation matrix. So this process does not create additional data. The additional data is only created during ExchangeFaceNbrData to get the ghost dof values. This memory footprint is probably negligible as compared to the local dof data.

src/serac/numerics/functional/functional.hpp
lihanyu97: This is Sam's safeguard to make sure DG is not running in parallel. Basically if it detects shared interior faces, it will raise an error. We will ultimately remove this function when the parallelization is complete.

src/serac/numerics/functional/tests/functional_basic_dg.cpp
lihanyu97: I honestly don't know why check_gradient was commented out before. But to test if residual calculation is working fine, I used "value". Now that I'm checking jacobian, I switched back to "check_gradient".

lihanyu97:
> So, maybe I'm going dense... but where is the test that you get the same field solution on both sides of the dg interface for pre-specified linear/quadratic fields?
> 
> Looks like amazing progress. Are you intending to push this in incrementally, or wait until the jacobian is done?

The new test is here: https://github.com/LLNL/serac/blob/42804174f653ba821530ff9d09210a401aa283e6/src/serac/numerics/functional/tests/dg_ghost_face_index.cpp

I would say let's push everything in after jacobian is fixed bacause this PR is making some fundamental changes on the typedef.
mfem::Mesh --> mfem::ParMesh
mfem::FiniteElementSpace --> mfem::ParFiniteElementSpace

lihanyu97:
@tupek2 @btalamini This DG PR is ready for review. All debug footprints (std::out / mpi::out) are removed.

src/serac/numerics/functional/tests/dg_ghost_face_index.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <fstream>
+#include <iostream>
+
+#include "mfem.hpp"
+
+#include <gtest/gtest.h>
+
+#include "axom/slic/core/SimpleLogger.hpp"
+#include "serac/infrastructure/application_manager.hpp"
+#include "serac/serac_config.hpp"
```
white238:
```suggestion
#include "serac/serac_config.hpp"
#include "serac/infrastructure/application_manager.hpp"
```

src/serac/numerics/functional/tests/dg_ghost_face_index.cpp
```diff
@@ -0,0 +1,101 @@
...
+      [=](double /*t*/, auto X, auto velocity) {
+        // compute the surface normal
+        auto dX_dxi = get<DERIVATIVE>(X);
+        auto n = normalize(cross(dX_dxi));
+
+        // extract the velocity values from each side of the interface
+        // note: the orientation convention is such that the normal
+        //       computed as above will point from from side 1->2
+        auto [u_1, u_2] = velocity;
+        std::cout << "One side = " << u_1 << ", The other side = " << u_2 << ", Jump = " << u_1 - u_2 << std::endl;
```
white238: We should use `SLIC_INFO_ROOT`

src/serac/numerics/functional/tests/domain_tests.cpp
```diff
@@ -406,17 +419,18 @@ TEST(domain, of2dBoundaryElementsFindsDofs)
 {
   constexpr int dim = 2;
   constexpr int p = 2;
-  auto mesh = import_mesh("patch2D_tris_and_quads.mesh");
+  auto bmesh = import_mesh("patch2D_tris_and_quads.mesh");
+  auto mesh = serac::mesh::refineAndDistribute(std::move(bmesh));
```
white238, line 423: My understanding is noone should be calling this function anymore and move to the new Mesh class. Examples in PR #1365

I may be missing some subtlety

src/serac/numerics/functional/typedefs.hpp
```diff
@@ -12,9 +12,10 @@ namespace serac {
 //      mfem::ParMesh / mfeme::ParFiniteElementSpace, but I'm not ready to pull the trigger on a big
 //      interface change like that, so these typedefs mark the parts that would need to eventually change
 
+// hanyu: I'm pulling the trigger for sam and the bullet might hit myself
```
white238:
```suggestion
```

src/serac/numerics/functional/tests/domain_tests.cpp
tupek2: In this case, we can use this function still.  The dependency is that serac::Mesh depends on functional.  So we actually cannot use the mesh class here.

tupek2: or, another way of thinking about it... this is a lower level test which is using the components, but not the final mesh abstraction.

white238: understood! I figurd i was missing something..

src/serac/numerics/functional/tests/dg_ghost_face_index.cpp
lihanyu97: SLIC_INFO_ROOT will only output those quadrature point evaluations on the root rank? I switched to using SLIC_INFO, which will output all quadrature point evaluations.

white238: Yes, unless you want it to output on all nodes, you should use the ROOT macros, if you are unsure use the non-ROOT macros.

lihanyu97: Yes I want output on all nodes. If the ghost dofs are aligned properly, the jump output for every quadrature point evaluation should give 0.

src/serac/numerics/functional/functional.hpp
lihanyu97: This function is removed now.

src/serac/numerics/functional/domain.cpp
```diff
@@ -535,6 +535,49 @@ void Domain::insert_restriction(const serac::fes_t* fes, FunctionSpace space)
...
+      }
+
+      for (int f : mfem_quad_ids_) {
+        mfem::Mesh::FaceInformation info = mesh_.GetFaceInformation(f);
+
+        if (info.IsShared()) {
+          shared_interior_face_ids_.push_back(i);
+        }
+
+        ++i;
```
tupek2, line 575: are these i even used?

tupek2 (approved):

src/serac/numerics/functional/domain.cpp
lihanyu97: Yes, in functional_qoi.inl, in `operator()` and `ActionOfGradient`, this list is used to scale integrals of shared interior faces by 0.5 because they are integrated twice by two processors, so the global sum is correct.

tupek2: I mean the index 'i' itself?

lihanyu97: Yeah, i is the i'th element index in the domain and corresponds to the i'th entry in output_E_.

Then in `ActionalOfGradient` line 315, there's a block
```
        // make sure shared interior faces are integrated only once
        if (dom.type_ == Domain::Type::InteriorFaces) {
          for (int shared_id : dom.shared_interior_face_ids_) {
            output_E_[shared_id] *= 0.5;
          }
        }
```

`shared_id` is the `i` pushed into `shared_interior_face_ids_`

tupek2: oh, sorry.  i missed it before.  nevermind.

lihanyu97: No worries. It's also worth to note that this list is only used for qoi functionals. Plus, this list is only constructed if the qoi contains interior face integrals, through `AddInteriorFaceIntegral` in `Functional<double(trials...)>`. Otherwise, it's an empty container.

src/serac/numerics/functional/domain.hpp
```diff
@@ -96,6 +96,9 @@ struct Domain {
    */
   std::map<FunctionSpace, BlockElementRestriction> restriction_operators;
 
+  /// @brief Shared interior face list
```
btalamini: This comment just repeats the variable name, and so isn't useful. Think of a user who has never seen this class before and is reading these. What does "shared" mean in this context? Tell them what this is for. Here's a suggestion:

```c++
/// @brief Ids of interior faces that lie on the boundary shared by two processors
```

But feel free to change this if you can make it clearer.

src/serac/numerics/functional/domain.hpp
```diff
@@ -244,6 +247,12 @@ struct Domain {
   /// thereby populate the element lists.
   void addElements(const std::vector<int>& geom_id, const std::vector<int>& elem_id,
                    mfem::Geometry::Type element_geometry);
+
+  /**
+   * @brief Find the list of interior faces shared by two processors to make sure
+   *        these faces are only integrated once.
+   */
+  void insert_shared_interior_face_list();
```
btalamini, line 255: I think this function should be considered logically private, and therefore be called *inside* `InteriorFaces(...)` (and any other factory function - maybe `ofEdges`?) . In my opinion, if one calls `InteriorFaces(...)` and the resulting Domain has an empty `shared_interior_face_ids_`, it's in an invalid state. The user shouldn't need to know to call an additional method for the class to be usable in parallel. @lihanyu97 and @tupek2, do you agree?

src/serac/numerics/functional/domain.cpp
```diff
@@ -535,6 +535,49 @@ void Domain::insert_restriction(const serac::fes_t* fes, FunctionSpace space)
 
 const BlockElementRestriction& Domain::get_restriction(FunctionSpace space) { return restriction_operators.at(space); };
 
+void Domain::insert_shared_interior_face_list()
+{
+  // Weights only need to be computed for Domain of InteriorFaces type
+  SLIC_ERROR_IF(type_ != Domain::Type::InteriorFaces, "This method is only for interior face domains");
+
+  // make a list if we don't already have one
+  if (shared_interior_face_ids_.size() == 0) {
```
btalamini: I prefer this, as the intent is clearer:

```suggestion
  if (shared_interior_face_ids_.empty()) {
```

src/serac/numerics/functional/element_restriction.cpp
```diff
@@ -561,34 +532,65 @@ axom::Array<DoF, 2, serac::detail::host_memory_space> GetFaceDofs(const serac::f
...
-          int stride = (by_vdim) ? components_per_node : 1;
-
-          // sam: is this right?
           // byVDIM == x y z x y z x y z x y z
           // byNODES == x x x x y y y y z z z z
-          for (int k = 0; k < dofs_per_face; k++) {
-            face_dofs.push_back(uint64_t(fes->VDofToDof(shared_elem_vdof_ids[k * stride]) + LSize));
-            mpi::out << face_dofs.back().index() << " ";
+          int dofs_per_element = shared_elem_vdof_ids.Size() / components_per_node;
+          int stride = (by_vdim) ? components_per_node : 1;
```
btalamini, line 563: Is this necessary? Aren't we aborting on line 547 if the ordering is not by vdim?

src/serac/numerics/functional/functional_qoi.inl
```diff
@@ -202,9 +215,10 @@ class Functional<double(trials...), exec> {
     std::vector<uint32_t> arg_vec = {args...};
     for (uint32_t i : arg_vec) {
       domain.insert_restriction(trial_space_[i], trial_function_spaces_[i]);
-      check_interior_face_compatibility(domain.mesh_, trial_function_spaces_[i]);
     }
 
+    domain.insert_shared_interior_face_list();
```
btalamini: As noted above, I think the shared face list should be filled on the domain construction and not be the responsibility of the user here.

src/serac/numerics/functional/functional.hpp
```diff
@@ -166,6 +147,88 @@ generateParFiniteElementSpace(mfem::ParMesh* mesh)
   return std::pair(std::move(fes), std::move(fec));
 }
 
+/**
+ * @brief helper function to locally cast away const on FE space to update face neighbor data.
+ *  this is ok because the original FE space is declared without const and we constrained
+ *  the non-constness locally
```
btalamini: I don't understand this comment. What I hope it means is that this function exists because the `MakeRef` method in MFEM is not const correct, but we know and trust that it does not actually change the trial space. If my statement is wrong, we need to revisit this.

src/serac/numerics/functional/domain.cpp
```diff
@@ -535,6 +535,49 @@ void Domain::insert_restriction(const serac::fes_t* fes, FunctionSpace space)
 
 const BlockElementRestriction& Domain::get_restriction(FunctionSpace space) { return restriction_operators.at(space); };
 
+void Domain::insert_shared_interior_face_list()
+{
+  // Weights only need to be computed for Domain of InteriorFaces type
+  SLIC_ERROR_IF(type_ != Domain::Type::InteriorFaces, "This method is only for interior face domains");
```
btalamini: I think in this case you should use `SLIC_ERROR_ROOT_IF` to avoid echoing the output on all ranks.

src/serac/numerics/functional/tests/dg_ghost_face_index.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include "serac/mesh_utils/mesh_utils_base.hpp"
+#include "serac/numerics/stdfunction_operator.hpp"
+#include "serac/numerics/functional/functional.hpp"
+#include "serac/numerics/functional/tensor.hpp"
+
+using namespace serac;
+using namespace serac::profiling;
+
+template <int dim, int p>
+void L2_index_test(std::string meshfile)
```
btalamini, line 37: I would appreciate if you write a short comment block that explains the intent of the test. If it ever fails, it will help the person who broke it understand what the test's job is and hopefully give a clue as to how they broke it. As a group, we do not usually do this, so this is a hypocritical complaint.

btalamini (changes requested):

src/serac/numerics/functional/element_restriction.cpp
lihanyu97: `int stride = (by_vdim) ? components_per_node : 1;` is part of Sam's initial implementation that is commented out. I didn't remove it just in case that we happened to want to support `byNODE` ordering.

src/serac/numerics/functional/functional_qoi.inl
lihanyu97: The reason I implemented this way is because 1) shared interior face list is only used in qoi functionals, and 2) if the qoi involves integrating over a set of interior faces. This means if we always fill this `interior_face_ids_` container for all `InteriorFace` domains, we are requesting additional memory that we may never use. Therefore, I only require this container to be filled automatically from `AddInteriorFaceIntegral` in `functional_qoi.inl`. The user doesn't have to worry about filling this container. You can see that `AddInteriorFaceIntegral` in `functional.hpp` doesn't have this line.

Also, this method is similar to `insert_restriction`, which doesn't do anything if there's already an element restriction for a given trial space. So for example, if we our functional qoi is computed by adding two different interior face integrals, we need to call `AddInteriorFaceIntegral` twice. But `insert_shared_interior_face_list` only execute once.

src/serac/numerics/functional/functional.hpp
lihanyu97: `MakeRef` is not the problem. `ExchangeFaceNbrData` is the method that doesn't have const keyword because it creates/recreates the `FaceNbrData` vector and other infos related to the face neighbor data. The catch here is `MakeRef` links the `ParGridFunction` with external data, so it doesn't create any additional memory footprint. Now if the external data is const, then we cannot invoke `ExchangeFaceNbrData` on the `ParGridFunction`.

So what I meant in the comment is that, we can cast away the const locally in this method because we want to invoke `ExchangeFaceNbrData`. This is ok because we are not changing any of the locally owned data on the trial space. All we are changing/updating is the face neighbor data.

src/serac/numerics/functional/tests/dg_ghost_face_index.cpp
lihanyu97: Sure I will add comments to this test.

src/serac/numerics/functional/functional_qoi.inl
btalamini: I understand. But I think the decision of whether to allocate that memory is made when the user decides to instantiate a `Domain` of interior faces. The point of the domains is to integrate over them, and the shared face information is needed to compute integrals on this type of domain. If it doesn't have it, it can't do its job in all cases. I would view that as having an invalid state.

"Users" in this case are not just people making `Functional` objects, they're also developers using the `Domain` class. If someone writes another integrator - like the dFEM one, or example - it won't be obvious to them that they need to call an additional method on interior face domains to be able to use the shared face info. 

I can see why the restriction operators lead you to this. But it turns out that this is not a good model. They were tacked onto `Domain` as an afterthought. We might pull them back out. I don't want to replicate that pattern.

Really, the underlying problem is that `Domain` should be a class with constructors, data hiding, and a clear contract. That's not your fault. We'll address it later.

btalamini (approved):

PR #1416, opened 2025-07-03
Run some code checks

chapman39:
Running basic code checks to ensure quality

- [x] cppcheck
- [x] run-clang-tidy

# cppcheck

didn't catch too much

Screenshot 2025-07-07 at 3 05 08 PM

# clang-tidy

### how i ran clang-tidy

```
module load clang/14.0.6
module load python
run-clang-tidy -p <build-dir>
```

### [ignored] warning related to logger

* i think this can be safely ignored. from [axom documentation](https://axom.readthedocs.io/en/develop/axom/slic/docs/sphinx/sections/getting_started.html#step-5-register-a-log-stream):

> Slic maintains ownership of all registered [Log Stream](https://axom.readthedocs.io/en/develop/axom/slic/docs/sphinx/sections/architecture.html#logstream) instances and will deallocate them when slic::finalize() is called.

```
/usr/WS2/meemee/serac/repo/src/serac/infrastructure/logger.cpp:67:3: warning: Potential leak of memory pointed to by 'i_logstream' [clang-analyzer-cplusplus.NewDeleteLeaks]
   67 |   addStreamToMsgLevel(d_logstream, slic::message::Debug);
      |   ^
```

### [solved] warning related to `stateNames()`

* Clang-tidy didn't like when virtual functions are called in the constructor. I moved the block that calls `stateNames()` into `resetStates()` instead (see PR diff, also simplified godbolt example showcasing warning https://godbolt.org/z/P391j9YP3)

### [ignored] memory leak warning related to `generateParFiniteElementSpace()`

* This warning is about [this function](https://github.com/LLNL/serac/blob/develop/src/serac/numerics/functional/functional.hpp#L138), `generateParFiniteElementSpace`, which creates a `mfem::FiniteElementCollection` (fec) and a `mfem::ParFiniteElementSpace` (fes). Fes takes a raw pointer of fec in its constructor.
* I think the warning is about the the case if fec gets deleted before fes, fes will have a dangling pointer. Or, clang-tidy is simply unsure whether the raw pointer fec will be properly deleted in MFEM.
* Similar to the logger warning, this can also probably be safely ignored assuming MFEM manages their raw pointers properly.
* We could use shared_ptr instead of unique_ptr, which would likely resolve the issue but could have greater implications.
* Adding `// NOLINT` is kind of unreasonable, since it would have to be added to every time `generateParFiniteElementSpace` is called (it cannot be added to functions - it can only be added to lines of code that triggers the warning).

```
/usr/WS2/meemee/serac/repo/src/serac/physics/boundary_conditions/tests/boundary_cond.cpp:109:61: warning: Potential leak of memory pointed to by '.first._M_t._M_t._M_head_impl' [clang-analyzer-cplusplus.NewDeleteLeaks]
  serac::mfem_ext::GetEssentialTrueDofsFromElementAttribute(*l2_fes, elem_attr_is_ess, ess_tdof_list);
                                                            ^
1m/usr/WS2/meemee/serac/repo/src/serac/physics/boundary_conditions/tests/boundary_cond.cpp:108:27: note: Calling 'generateParFiniteElementSpace<serac::L2<0, 1>>'
  auto [l2_fes, l2_fec] = serac::generateParFiniteElementSpace<L2<0>>(&pmesh);
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


src/serac/numerics/tests/odes.cpp
```diff
@@ -234,7 +235,8 @@ int which_kind_of_ode(serac::TimestepMethod m)
   if (m == serac::TimestepMethod::CentralDifference) return 2;
   if (m == serac::TimestepMethod::FoxGoodwin) return 2;
 
-  return -1;
+  SLIC_ERROR(axom::fmt::format("Unsupported serac::TimestepMethod {0}", static_cast<int>(m)));
+  return 0;
```
chapman39, line 251: Clang-tidy complained this was returning -1, because it figured out the return value was being used as an index. I can see we'd still want to fail in this case, so I decided to explicitly print an error instead.

src/serac/physics/heat_transfer.hpp
```diff
@@ -307,6 +299,14 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
     BasePhysics::initializeBasePhysicsStates(cycle, time);
     initializeThermalStates();
+
+    if (!checkpoint_to_disk_) {
+      checkpoint_states_.clear();
+      auto state_names = stateNames();
+      for (const auto& state_name : state_names) {
+        checkpoint_states_[state_name].push_back(state(state_name));
+      }
+    }
```
chapman39, line 309: Removes virtual function call in constructor. Warning was the following:

> Call to virtual method 'Thermal::stateNames' during construction bypasses virtual dispatch [clang-analyzer-optin.cplusplus.VirtualCall]

src/serac/physics/thermomechanics_monolithic.hpp
```diff
@@ -734,6 +734,16 @@ class ThermomechanicsMonolithic<order, dim, Parameters<parameter_space...>,
...
+
+    if (checkpoint_to_disk_) {
+      outputStateToDisk();
+    } else {
+      checkpoint_states_.clear();
+      auto state_names = stateNames();
+      for (const auto& state_name : state_names) {
+        checkpoint_states_[state_name].push_back(state(state_name));
+      }
+    }
```
chapman39, line 746: Added resetting of checkpoint states to `completeSetup()` to fix `thermomech_finite_diff` segfault. I think this was supposed to be here anyway.

src/serac/numerics/equation_solver.cpp
```diff
@@ -730,9 +730,8 @@ class TrustRegion : public mfem::NewtonSolver {
 
       bool have_computed_Hvs = false;
 
-      bool happyAboutTrSize = false;
       int lineSearchIter = 0;
-      while (!happyAboutTrSize && lineSearchIter <= nonlinear_options.max_line_search_iterations) {
+      while (lineSearchIter <= nonlinear_options.max_line_search_iterations) {
```
chapman39, line 739: @tupek2 let me know if removing this var was the right call

tupek2: I think this needs to stay how it was.  Was there some tidy issue that is was causing?  Seems like fairly mundane bool logic.  Maybe it just wants parentheses?

while ( (!happyAboutTrSize) && (lineSearchIter <= nonlinear_options.max_line_search_iterations) ) {

tupek2 (changes requested):
Good, other than the trust-region logic change.

src/serac/numerics/equation_solver.cpp
chapman39: i think the issue is after `happyAboutTrSize` is set to true, the while loop breaks, so there is no need to it to be a while loop condition, and therefore no need for the variable at all

tupek2: I see.  OK, it looks to be good to me now.

tupek2 (approved):

white238 (approved):

btalamini (approved):

PR #1422, opened 2025-07-14
Inertia Relief

thartland:
Adding an inertia relief example with a NeoHookean material model.

This PR adds an inertia relief example problem that utilizes solid_weak_form to define a residual and a vector of ScalarObjective's to define a small set of constraints. This is then wrapped in a class that inherits from GeneralNLMCProblem (NLMC = nonlinear mixed-complementarity) so that the problem can be passed to the HomotopySolver of the ContinuationSolvers submodule.


tupek2 (approved):
Did you figure out the sign issue with the 'stiffness matrix'?  Looks like it, give you are back to Neohookean.

thartland:
> Did you figure out the sign issue with the 'stiffness matrix'? Looks like it, give you are back to Neohookean.

@tupek2 yes, I just needed to change back the sign.

examples/inertia_relief/inertia_relief_example.cpp
```diff
@@ -0,0 +1,585 @@
...
+ */
+
+#include "serac/serac.hpp"
+
+#include "mfem.hpp"
+
+// ContinuationSolver headers
+#include "problems/Problems.hpp"
+#include "solvers/HomotopySolver.hpp"
+#include "utilities.hpp"
```
white238, line 23: You should consider putting your headers in a named directory to avoid conflicts. This is outside of the scope of this PR clearly.

white238 (approved):
Thanks @thartland ! My only comment is this could use more comments.

examples/inertia_relief/inertia_relief_example.cpp
thartland: Thanks for the reminder. I have some work to do on this external solver submodule and will get this done.

thartland: I agree that utilities is too generic of a header title for serac and it would be good to keep them all solver relevant headers in one location.

PR #1426, opened 2025-07-21
Add extra argument to allow specification of the communicator

tepperly:
`serac::Mesh::Mesh(mfem::Mesh &&mesh...)` should have an argument for the communicator. Otherwise, it only works with meshes that are defined over `MPI_COMM_WORLD`. For optimization under uncertainty problems in LiDO, it's common to define the meshes over a strict subset of `MPI_COMM_WORLD`.

@chapman39 for awareness.


white238 (approved):

chapman39 (approved):
thank you tom

PR #1427, opened 2025-07-25
Include What You Use

chapman39:
Updating serac and system includes based on whether we're using them in the file directly using [include-what-you-use](https://github.com/include-what-you-use/include-what-you-use/).

This PR:

* adds (or removes) standard library includes that's directly used (or not used) in each file
* adds (or removes) serac includes that's directly used (or not used) in each file'
* tries to group include headers by class header (if cpp), system, tpl, and project includes

also,

* removes usage of `using namespace serac::profiling;` to better track profiling function usage (e.g. concat, SERAC_MARK_SCOPE, etc)

# my setup for iwyu

```bash
# build iwyu outside serac proj
git clone git@github.com:include-what-you-use/include-what-you-use.git
git checkout clang_14
cd include-what-you-use
mkdir build && cd build
module load clang/14.0.6
cmake -G "Unix Makefiles" -DCMAKE_PREFIX_PATH=/usr/tce/packages/clang/clang-14.0.6 ..
srun -N1 make -j

# add build to path
# PATH=/usr/workspace/meemee/iwyu/include-what-you-use/build/bin/:$PATH
# PATH=/usr/workspace/meemee/iwyu/include-what-you-use/:$PATH

# build serac with clang 14
module load python
./config-build.py -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE=include-what-you-use
srun -N1 make -j 2> /tmp/iwyu.out
fix_includes.py -b < ../iwyu.out
```

after this, i one-by-one remove the comments and check if the header they added is worth adding or not. there is a `--nocomments` option, but it's useful to see why iwyu added the header.

### workaround petsc circular includes

* Petsc internally has a circular include error, so I [commented out this error in the iwyu source](https://github.com/include-what-you-use/include-what-you-use/blob/clang_14/iwyu_include_picker.cc#L1081). I emailed the petsc team so they could be aware of this.
* Cycles are actually ignored in iwyu develop, but since we needed the `clang_14` branch this workaround was needed.

```
Error running 'include-what-you-use': Cycle in include-mapping:
  <petscmacros.h> ->
  <petsc/private/petscadvancedmacros.h> ->
  <petsc/private/petscadvancedmacros.h>
```


src/serac/infrastructure/about.cpp
```diff
@@ -5,9 +5,12 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 
 #include "serac/infrastructure/about.hpp"
-
 #include "serac/serac_config.hpp"
 
+#include <mpi.h>
```
white238:
```suggestion
#include "mpi.h"
```

chapman39: should i do the same for googletest headers?

src/serac/infrastructure/accelerator.cpp
```diff
@@ -6,11 +6,9 @@
 
 #include "serac/infrastructure/accelerator.hpp"
 
-#include <memory>
-
 #include "mfem.hpp"
 
-#include "serac/infrastructure/logger.hpp"
```
white238, line 13: This is used in the file.

src/serac/infrastructure/profiling.cpp
```diff
@@ -4,14 +4,15 @@
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
 
-#include "serac/infrastructure/about.hpp"
 #include "serac/infrastructure/profiling.hpp"
-#include "serac/infrastructure/logger.hpp"
```
white238, line 9: Same here

src/serac/numerics/petsc_solvers.cpp
```diff
@@ -5,6 +5,10 @@
 // SPDX-License-Identifier: (BSD-3-Clause)
 
 #include "serac/numerics/petsc_solvers.hpp"
+
+#include <cstddef>  // for NULL
```
white238: Two points:

* I personally dislike comments that tell you what each header is for
* Why are we using `NULL` and not `nullptr`?

src/serac/physics/boundary_conditions/boundary_condition_manager.cpp
```diff
@@ -4,12 +4,10 @@
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
 
-#include "serac/physics/boundary_conditions/boundary_condition_manager.hpp"
-
 #include <algorithm>
 #include <iterator>
 
-#include "serac/infrastructure/logger.hpp"
```
white238, line 12: Readd this

white238 (approved):

white238:
This PR shows that IWYU has issues with our way of using SLIC through our transitive `logger.hpp` header and it won't pick up the need for that header for the SLIC macros. We either need to be fine with it not recognizing the need for that header and/or manually adding it ourselves.

Our logger and config headers should be added to all files. The logger header for ease of use and most files log anyways and the config header because the consequences of not having it listed could be unexpected behavior.

src/serac/numerics/petsc_solvers.cpp
chapman39: my guess as to why there is NULL is the creator of this file was directly interacting with petsc, which is mostly written in C.

but i changed it to nullptr and removed the include

tupek2 (approved):

PR #1429, opened 2025-07-30
Ryans 2 d mortar smooth

rjl33:
2D smoothed integrated Mortar Method


tupek2 (commented):
Why is there a patch test in contact_finite_diff.cpp.  And, there is a contact_patch_2D.cpp, and an energy_mortar_patch.cpp.  But I thought the only 2D contact right now is energy_mortar?  Or is that not correct?

ebchin:
> Why is there a patch test in contact_finite_diff.cpp. And, there is a contact_patch_2D.cpp, and an energy_mortar_patch.cpp. But I thought the only 2D contact right now is energy_mortar? Or is that not correct?

Added a comment to the contact_finite_diff.cpp test. It was originally set up to finite diff a whole Smith solve with contact to find issues with solid_mechanics_contact.hpp integration with functional.

Good catch on the multiple patch tests. I removed contact_patch_2D.cpp.

chapman39 (approved):
thanks eric

tupek2 (approved):

PR #1412, opened 2025-06-19
Spack Enzyme build

white238:
Adds Enzyme as a TPL that is not fully enabled but exists in the TPL build. Set `-DSERAC_USE_ENZYME=ON` on the command line for `config-build.py` or `cmake` to enable.

Notable Changes:
* Updates to Spack 1.0
* Updates to TOSS4 to clang@19 and gcc@13
* Updates to new Tribol that includes tet support

Things currently out of scope of this PR but up for debate and will hopefully done shortly after:
* Updating mac build
* Updating docker containers (working on it but radiuss-docker repo is quite broken and needs to be updated)
* Updating blueos and toss4_cray
* Updating to clang-format 19

Updated docs:
https://serac.readthedocs.io/en/task-white238-spack_enzyme_compiler/sphinx/quickstart.html#building-serac-s-dependencies-via-spack-uberenv
https://serac.readthedocs.io/en/task-white238-spack_enzyme_compiler/sphinx/dev_guide/tpl_build.html

PRs this is dependent on:
https://github.com/LLNL/uberenv/pull/147
https://github.com/LLNL/radiuss-spack-configs/pull/134

Worked around Spack issue:
https://github.com/spack/spack/issues/51244

PR #1431, opened 2025-08-01
Tet support with contact

ebchin:
Tribol was updated to include tet support in #1412.

This PR just modifies the contact patch test to include tets, so it is included in our CI.


src/smith/physics/tests/contact_patch.cpp
```diff
@@ -11,22 +11,23 @@
 #include "gtest/gtest.h"
 #include "mpi.h"
 #include "mfem.hpp"
+#include "shared/mesh/MeshBuilder.hpp"
```
white238, line 14: took a bit to find out this came from tribol.. why is it not just "tribol/meshbuilder.hpp"?

white238 (approved):

src/smith/physics/tests/contact_patch.cpp
ebchin: I'll tell the Tribol guy to fix it over there

tupek2 (approved):

PR #1436, opened 2025-08-12
Tupek/residual interface adjustments

tupek2:
This semi-reverts a previous change putting the shape_displacement on the serac::Mesh.  Now the serac::Mesh knows the space of the shape displacements, and how create them, but does not own.   Also, flipped the sign convention in the serac::Residual.  AddBoundaryIntegral(traction) will give a positive traction force for a positive value.  However, the resulting residual has a flipped sign, i.e., the residual the derivative of the energy (minus the force).  This ensures the resulting Jacobian is positive definite (ignoring buckling), which is an assumption our solvers make. 

Also, shape_disp and its dual must be passed as arguments to a lot more things in the new interface.  Also, stubbed out a spot for where quadrature fields will go, but at the moment they are placeholders until we have settled on a type for our quadrature fields.

Change name of residual to WeakForm... This is because residual is highly overloaded.  Now residual is just the vector.  WeakForm is the base class which you can add terms to, and evaluate residuals and jacobians on.


src/serac/physics/field_types.hpp
```diff
@@ -32,7 +41,7 @@ using ConstDualFieldPtr = FiniteElementDual const*;
 /// @brief Get a vector of FieldPtr or DualFieldPtr from a vector of shared_pointers to FiniteElementState or
 /// FiniteElementDual
 template <typename T>
-auto getFieldPointers(const std::vector<std::shared_ptr<T>>& states, const std::vector<std::shared_ptr<T>>& params)
+auto getFieldPointers(std::vector<std::shared_ptr<T>>& states, std::vector<std::shared_ptr<T>>& params)
```
chapman39, line 44: why was const removed?

src/serac/physics/field_types.hpp
```diff
@@ -46,6 +55,20 @@ auto getFieldPointers(const std::vector<std::shared_ptr<T>>& states, const std::
   return pointers;
 }
 
+/// @brief Get a vector of FieldPtr or DualFieldPtr from a vector of shared_pointers to FiniteElementState or
+/// FiniteElementDual
+template <typename T>
+auto getFieldPointers(std::vector<std::shared_ptr<T>>& states)
```
chapman39, line 61: could this be const?

chapman39: nvm, i see there is a const counterpart

src/serac/physics/functional_residual.hpp
```diff
@@ -99,13 +100,16 @@ class FunctionalResidual<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
...
+        Dimension<spatial_dim>{}, DependsOn<active_parameters...>{},
+        [body_integral](double t, auto X, auto... inputs) { return -body_integral(t, X, inputs...); },
+        mesh_->domain(body_name));
+
     v_residual_->AddDomainIntegral(
         Dimension<spatial_dim>{}, DependsOn<0, 1 + active_parameters...>{},
         [body_integral](double t, auto X, auto V, auto... inputs) {
           auto orig_tuple = body_integral(t, X, inputs...);
-          return serac::inner(get<VALUE>(V), get<VALUE>(orig_tuple)) +
+          return -serac::inner(get<VALUE>(V), get<VALUE>(orig_tuple)) -
```
chapman39: would this be more readable if you did something like `(serac::inner(...) + serac::inner(...)) * -1`?

if not, i think its fine the way it is

src/serac/physics/field_types.hpp
tupek2: This is to get a non-const vector of FiniteElementState* from a vector of shared_ptr<FiniteElementState>.... technically, if you understand C++ constness well, these vector arguments could be const... but I think its more confusing for most people.  For example, to get a vector<FiniteElementState*> from a vector<FiniteElementState>, the vector<FiniteElementState> CANNOT be const.  So, I removed the const to be consistent with that case.

src/serac/physics/functional_residual.hpp
tupek2: I think its basically the same.  I tend to minimize parenthesis, but its a style choice.

src/serac/physics/tests/solid_finite_diff.cpp
```diff
@@ -195,10 +195,10 @@ void finite_difference_shape_test(LoadingType load)
 
   // Construct the appropriate dimension mesh and give it to the data store
   std::string filename = SERAC_REPO_DIR "/data/meshes/patch2D_tris.mesh";
-
+  printf("a1\n");
```
chapman39: debug prints!

src/serac/physics/thermomechanics_monolithic.hpp
```diff
@@ -190,7 +190,7 @@ class ThermomechanicsMonolithic<order, dim, Parameters<parameter_space...>,
     block_nonlinear_oper_ = std::make_unique<mfem::BlockOperator>(block_thermomech_offsets_);
     block_nonlinear_oper_transpose_ = std::make_unique<mfem::BlockOperator>(block_thermomech_offsets_);
 
-    mesh_->shapeDisplacement() = 0.0;
+    shape_displacement_ = 0.0;
```
chapman39: does `shape_displacement_` need to be set here? should base physics do this instead?

tupek2: I think doing it the base class makes sense.  In fact, it already does and this wasn't really needed.  Good catch.

src/serac/physics/tests/solid_finite_diff.cpp
tupek2: fixed

chapman39 (approved):
thank you @tupek2

src/serac/physics/functional_residual.hpp
```diff
@@ -83,8 +83,9 @@ class FunctionalResidual<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
    * // DependsOn<active_parameters...> can be indices into fields which the body integral may depend on
    * @param body_name The name of the registered domain over which the body force is applied. If nothing is supplied
    * the entire domain is
-   * @param body_integral A function describing the body force applied
-   * used.
+   * @param body_integral A function describing the body force applied.  Our convention for the sign of the residual
+   * vector is to negate the body force internally so the resulting residual is a 'negative force'.  This is to ensure
+   * that the Jacobian of the residual is positive definite for most physics.
```
btalamini: I find this documentation confusing, but I admit it's hard to articulate. I think of this as documentation for people trying to use a physics module. They need to look up what the sign convention is, but the reason for the convention is an implementation detail they don't need to know. 

```suggestion
   * @param body_integral A function returning a volumetric source. Expected sign convention is that
   * the source would do positive work through a positive change in the conjugate generalized
   * displacement.
```

btalamini: As an aside, the variable name `body_integral` is misleading. It's more accurately the `body_integrand`. We can deal with that later.

src/serac/physics/functional_residual.hpp
```diff
@@ -99,13 +100,16 @@ class FunctionalResidual<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
   template <int... active_parameters, typename BodyIntegralType>
   void addBodyIntegral(DependsOn<active_parameters...>, std::string body_name, BodyIntegralType body_integral)
   {
-    residual_->AddDomainIntegral(Dimension<spatial_dim>{}, DependsOn<active_parameters...>{}, body_integral,
-                                 mesh_->domain(body_name));
+    residual_->AddDomainIntegral(
```
btalamini: I'd put another comment in here to future maintainers & modifiers:

```suggestion
    // The body force is negated here so that the residual (internal forces - external forces), and thus its
    // derivative is the stiffness matrix.
    // Applying the negative sign here makes the interface easier, so that users don't have to return
    // negative body forces in body_integral.
    residual_->AddDomainIntegral(
```

src/serac/physics/residual.hpp
```diff
@@ -43,53 +47,72 @@ class Residual {
...
    * @param dt time step
-   * @param fields vector of serac::FiniteElementState* as arguments to the residual
+   * @param shape_disp serac::FiniteElementState*, change in model coordinates relative to the initially read in mesh
+   * @param fields vector of serac::FiniteElementState*
+   * @param quad_fields vector of ConstQuadratureFieldPtr
    * @param v_fields left hand side 'v' fields
+   * @param vjp_shape_disp_sensitivity jvp for shape_displacement: v_fields[i] * d{r}_i / d{shape_disp}
    * @param vjp_sensitivities output jvps, 1 per input field: v_fields[i] * d{r}_i / d{fields}_j
+   * @param vjp_quadrature_sensivities output jvps, 1 per input quadrature field: v_fields[i] * d{r}_i /
+   * d{quadrature_field}_j
```
btalamini: There are a few typos with "jvp" appearing where "vjp" is intended. The `@brief` comment up top references `vjpFields`, which is an out-of-date name.

src/serac/physics/scalar_objective.hpp
```diff
@@ -38,30 +38,34 @@ class ScalarObjective {
...
+  virtual double evaluate(double time, double dt, ConstFieldPtr shape_disp,
+                          const std::vector<ConstFieldPtr>& fields) const = 0;
 
   /** @brief Virtual interface for computing objective gradient from a vector of serac::FiniteElementState*
    *
    * @param time time
    * @param dt time step
-   * @param fields vector of serac::FiniteElementState* as arguments to the residual
+   * @param shape_disp shape displacement const serac::FiniteElementState*
+   * @param fields vector of serac::FiniteElementState*
```
btalamini:
```suggestion
   * @param shape_disp shape displacement
   * @param fields Field inputs to residual operator
```

src/serac/physics/scalar_objective.hpp
```diff
@@ -38,30 +38,34 @@ class ScalarObjective {
...
+  virtual mfem::Vector gradient(double time, double dt, ConstFieldPtr shape_disp,
+                                const std::vector<ConstFieldPtr>& fields, int field_ordinal) const = 0;
 
   /** @brief Virtual interface for computing objective gradient with respect to the mesh coordinates
    *
    * @param time time
    * @param dt time step
-   * @param fields vector of serac::FiniteElementState* as arguments to the residual
+   * @param shape_disp shape displacement const serac::FiniteElementState*
+   * @param fields vector of const serac::FiniteElementState*
```
btalamini:
```suggestion
   * @param shape_disp shape displacement
   * @param fields Field inputs to residual operator
```

btalamini (approved):
Making the shape displacement a required argument is a good solution to me.

src/serac/physics/functional_residual.hpp
```diff
@@ -121,67 +126,71 @@ class FunctionalResidual<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
...
-  void addBoundaryIntegral(DependsOn<active_parameters...>, std::string boundary_name, NeumannType surface_function)
+  void addBoundaryIntegral(DependsOn<active_parameters...>, std::string boundary_name, NeumannType integrand)
   {
     residual_->AddBoundaryIntegral(
         Dimension<spatial_dim - 1>{}, DependsOn<active_parameters...>{},
-        [surface_function](double t, auto X, auto... params) {
+        [integrand](double t, auto X, auto... params) {
           auto n = cross(get<DERIVATIVE>(X));
-          return surface_function(t, get<VALUE>(X), normalize(n), params...);
+          return integrand(t, get<VALUE>(X), normalize(n), params...);
```
btalamini: Should we sent in `X` with its derivatives to the integrand (not just the value) so that users can compute normals in the current configuration?

btalamini (approved):

src/serac/physics/functional_weak_form.hpp
```diff
@@ -0,0 +1,417 @@
...
+   * @tparam active_parameters Type for indices into fields which the body integral may depend on
+   * @tparam BoundaryFluxType The type of the traction load
+   * @param depends_on Andices into fields which the body integral may depend on
+   * @param boundary_name The name of the registered domain over which the boundary integral is applied.
+   * @param flux_function A function describing the boundary integral term to include in the weak form.
+   * Our convention for the sign of the residual
+   * vector is that it is expected to be a 'negative force', so the mass terms show up with a positive sign in the
+   * residual.  This also ensures that the Jacobian of the residual is positive definite for most physics.  A body
+   * integrand involving 'right hand side' contributions like a body load, should be supplied by the user with a
+   * negative sign.
```
btalamini: This comment makes it sound like the user should put the negative sign in manually in this function. I think it was accidentally copied and pasted from`addBoundaryIntegral`.

btalamini (approved):

src/serac/physics/functional_weak_form.hpp
tupek2: I pushed a fix.  maybe double check if you have a moment.

src/serac/physics/functional_weak_form.hpp
```diff
@@ -0,0 +1,413 @@
...
+    addBodyIntegral(DependsOn<>{}, body_name, body_integral);
+  }
+
+  /**
+   * @brief Add a body source (body load) to the weak form
+   *
+   * @tparam active_parameters Type for indices into fields which the body integral may depend on
+   * @tparam BodyLoadType The type of the body load function
+   * @param body_name The name of the registered domain over which the body loads are applied.
+   * @param depends_on Andices into fields which the body integral may depend on
```
btalamini:
```suggestion
   * @param depends_on Indices into fields which the body integral may depend on
```

src/serac/physics/functional_weak_form.hpp
```diff
@@ -0,0 +1,413 @@
...
+   * @param load_function A function describing the body force applied.
+   * @pre load_function must be a object that can be called with the following arguments:
+   *    1. `double t` the time
+   *    2. `tensor<T,dim> X` the spatial coordinates for the quadrature point.
+   *    3. `value`, a variadic list of field values, one tuple for each of the trial spaces specified in the
+   * `DependsOn<...>` argument.
+   *    The expected return is the value (either double or tensor<double,field_dim>) of the body source/load flux
```
btalamini:
```suggestion
   *    The expected return is the value of the source at X.
```

I removed the output shape requirement because the either/or description was vague about the circumstances. I started to clarify it, saying that it had to have the same tensor shape as the test function. However, I think this is worse. For people who know what a source is, it's redundant, and for those who don't, it's not enough to help.

src/serac/physics/functional_weak_form.hpp
```diff
@@ -0,0 +1,413 @@
...
+   * @param boundary_name The name of the registered domain over which the boundary integral is applied.
+   * @param flux_function A function describing the outward normal flux applied.
+   * @pre flux_function must be a object that can be called with the following arguments:
+   *    1. `double t` the time
+   *    1. `tensor<T,dim> X` the spatial coordinates for the quadrature point
+   *    3. `tensor<T,dim> n` the outward-facing unit normal for the quadrature point
+   *    4. `value`, a variadic list of tuples of field values at quadrature points,
+   *            one for each of the trial spaces specified in the `DependsOn<...>` argument.
+   *   The expected return is the value of the boundary flux (either double or tensor<double,field_dim>), oriented
+   * relative to the outward normal.
```
btalamini:
```suggestion
   *   The expected return is the value of the boundary flux oriented in the sense of the outward normal.
```

btalamini (approved):

PR #1438, opened 2025-08-13
Configure script update

chapman39:
Switching to Release can be a discussion. It was suggested by @tupek2 , and I think it makes sense. One pro for Release is new users may accidentally run their production work on Debug and get confused. One pro for Debug is when you run into an issue, you can quickly debug without rebuilding with debug symbols.

To capture asserts in CI, most CI jobs have been switched explicitly to Debug mode (they were Debug before this PR too).

To further reduce confusion, Application Manager now prints build information when constructed. Example print:

```
**************************** Serac Run Information *****************************
Version: v0.0.1-a45e10072
Build Type: Release
User Name: chapman39
Host Name: firion
MPI Rank Count: 1
********************************************************************************
```


config-build.py
```diff
@@ -294,10 +294,10 @@ def main():
        else:
           return False
 
-    # CMake build type is Debug by default, but if CMAKE_BUILD_TYPE is an unknown argument (i.e. a CMake argument), then
+    # CMake build type is Release by default, but if CMAKE_BUILD_TYPE is an unknown argument (i.e. a CMake argument), then
     # use that option instead.
     if args.buildtype == None:
-        args.buildtype = "Debug"
+        args.buildtype = "Release"
```
white238, line 299: Not sure I agree with this. At least we should do RelWithDebInfo

tupek2: We could revert for now, and get the other part through.  But, I like default release in the long term.  For developer who known enough to use a debugging tool on serac, changing the build to debug is not too hard.  For new devs and students, they often try to build with minimal options (as the options will be confusing to them to start) and then tell us a simple simulation takes hours and times out on LC.  This has happened at least 3 times in the past 2 years, and cost a few of them some days.

chapman39: Maybe we can print a more noticeable warning somewhere if you're on a Debug build. In the configure script, or maybe in ApplicationManager constructor?

white238: I am also concerned that this may disable our asserts. Though this could be mitigated in a couple ways (CI debug jobs or forcing them on).

tupek2: yeah, if we change this, our CI and testing shouldn't change, as we want those asserts in CI.  But, the default build for new users could be changed.  That is the thought at least.

btalamini: I agree with Mike here - the expectation is that if no custom build flags are supplied, the build should be release.  Also agree with Chris and Mike that we want the assertions in CI builds.

chapman39: What are peoples thoughts on this @white238 @btalamini @tupek2 :

* switch default to Release, but test CI in Debug (with some exceptions, like benchmarks)
* leave asserts as they are currently for now
* rather than clearing build/ install dir, print error message, supply rm command, and exit

btalamini: I like these. If it turns out we find the third bullet annoying, we could later add an option (like `--force` or something like that) to make it behave as it does now (deleting the old directory first).

chapman39: actually, i like that idea, im adding it now

chapman39: For the sake of limiting changes this PR does, I reverted the "error on existing build/ install dir". @btalamini

white238 (approved):
Thanks @chapman39

tupek2 (approved):

PR #1439, opened 2025-08-14
Reduce CI allocation times

chapman39:
i can't help but create another tiny PR..

this change was motivated by me having to increase lassen CI alloc time to 60. but, it was probably a fluke, since now its running back at ~20 minutes. https://lc.llnl.gov/gitlab/smith/serac/-/jobs/2830694

i noticed i kept increasing CI alloc times, when they were likely unnecessary. also in this PR, i compared the recent run-times of each CI pipeline and reduced them reasonably such that they won't time out, but won't wait in the queue longer than they need to be.


btalamini (approved):
🙌 Thanks for revisiting the data and looking for gains. Faster CI helps everyone!

tupek2 (approved):

PR #1441, opened 2025-08-18
Nonlin Benchmark use enums from `solver_config.hpp`

### LINKED ISSUE #1231: Consolidate enums of nonlinear solvers and preconditioners
These maps and string things can potentially move into the library.  In fact, I made something similar  a few months back. [see inline std::string nonlinearName(const NonlinearSolver& s), line ~159 in solver_config.hpp].  Maybe think about consolidating those designs.  I'm not particular to the way I did it, so just choose one and put it all in solver_config.hpp?

_Originally posted by @tupek2 in https://github.com/LLNL/serac/pull/1222#discussion_r1752289573_

chapman39:
* Removes duplicating enums that were created in the Solid Nonlinear Solve Benchmark
* Removes unused function `functional_solid_test_euler()`

Fixes #1231


src/serac/physics/benchmarks/physics_benchmark_solid_nonlinear_solve.cpp
```diff
@@ -14,261 +14,89 @@
...
-  switch (nonlinSolve) {
-    case NonlinSolve::NEWTON: {
-      SLIC_INFO_ROOT("using newton solver");
+  SLIC_INFO_ROOT("================================================================================");
+  SLIC_INFO_ROOT(axom::fmt::format("Nonlinear Solver = {}", serac::nonlinearName(nonlinearSolver)));
+  SLIC_INFO_ROOT(axom::fmt::format("Linear Solver    = {}", serac::linearName(linearSolver)));
+  SLIC_INFO_ROOT(axom::fmt::format("Preconditioner   = {}", serac::preconditionerName(preconditioner)));
+  SLIC_INFO_ROOT("================================================================================");
+
+  // TODO do we always want to set min iters and max line serach?
```
tupek2: I think so... for pure newton's method, max_line_search should be 0.  The default is probably ok for the rest.  Generally we recommend running with 0 min_iters, but Kenny has a case where he puts it to 1.

tupek2 (approved):
I feel like this is a nice reduction in duplication.  Thanks!

src/serac/physics/benchmarks/physics_benchmark_solid_nonlinear_solve.cpp
chapman39: ill remove the comment and leave it as is

white238 (approved):

PR #1446, opened 2025-08-28
Add method to insert serac::Domain on serac::Mesh

kswartz92:
This PR adds a simple method to add existing domains to serac meshes. I did a minor refactoring of error checking and used the new method in one of the existing patch tests.


src/serac/physics/mesh.cpp
```diff
@@ -59,6 +59,19 @@ serac::Domain& Mesh::entireBoundary() const { return domain(entireBoundaryName()
...
 
+void Mesh::errorIfDomainExists(const std::string& domain_name) const
+{
+  SLIC_ERROR_IF(domains_.find(domain_name) != domains_.end(),
+                axom::fmt::format("A domain named {0} already exists in mesh with tag {1}", domain_name, mesh_tag_));
+}
+
+void Mesh::insertDomain(const std::string& domain_name, const Domain& domain)
+{
+  SLIC_ERROR_IF(&this->mfemParMesh() != &domain.mesh_, "A domain inserted onto a mesh must be defined on that mesh");
```
kswartz92, line 70: Is there a better way than pointer comparison to make sure the mesh is consistent?

tupek2: honestly, its not a bad way

tupek2 (approved):
Did you want to test the boolean logic part specifically, or presume that is tested independently?

white238 (approved):

src/serac/physics/mesh.hpp
```diff
@@ -89,6 +89,12 @@ class Mesh {
   /// @brief Returns domain boundary corresponding to the internal boundary elements
   serac::Domain& internalBoundary() const;
 
+  /// @brief Helper function used to check if a domain name already exists
+  void errorIfDomainExists(const std::string& domain_name) const;
```
btalamini: This function should probably be private.

btalamini: Thanks for useful the addition, @kswartz92!

kswartz92: agreed, just switched it to private

PR #1457, opened 2025-09-04
Fix beam bending example

ebchin:
The kinsol solver is not converging.  Disabling kinsol for this test and adding some google test checks to verify the solution is correct.


btalamini:
I removed this solver from patch tests for the same reason. Maybe we deprecate it.

ebchin:
> I removed this solver from patch tests for the same reason. Maybe we deprecate it.

Agreed, I think deprecation makes sense.

tupek2 (approved):

btalamini (approved):

PR #1458, opened 2025-09-04
Add enzyme to mfem cleared vars

ebchin:
This doesn't seem to affect toss 4 codevelop builds, but it appears toss 4 cray codevelop builds need this.


white238 (approved):
I really need to track down why this happens in MFEM.

tupek2 (approved):

PR #1463, opened 2025-09-08
Contact-Homotopy

thartland:
The purpose of this PR is to add the ability to solve (tied) contact problems with the homotopy solver.

Tribol calls are wrapped in a constraint object which will define a NLMCProblem which is the object that the homotopy solver acts on.

Another addition of this PR is the complete avoidance of `delete` and replacing usage of raw pointers with `std::unique_ptr` in inertia relief and tied contact problem `EqualityConstrainedHomotopyProblem`'s 

Removing finite difference check task as such a check will only pass if we compile with enzyme and enable exact derivatives.

- [X]  create a contact NLMCProblem and pass it to the HomotopySolver.
- [X] Have complementarity problems return Operator* instead of HypreParMatrix* to support a wider range of applications wherein the Operators are not assembled. _Caveat: more work is needed for the HomotopySolver to fully support other operator types_
- [X] More verbose function defs for serac wrapper of problems so things are more easily understood by others.
- [X] ~~Utilize exceptions from Tribol compuations to signal bad contact computations to the solver and to reduce solver step length parameters. (Currently via a branch of Tribol)~~. While exceptions are now in place they are not being explicitly used in this PR as with "milder" homotopy solver parameters the nonlinear iterates do not vary so wildly as to require the exceptions that are only available via a specific Tribol branch.
- [X] Given updates to algorithmic parameters to the HomotopySolver is it still necessary to catch contact mechanics computation errors for the contact mechanics application problem? This will potentially impact the PR making it being able to merge into master dependent on if we update Tribol. **Answer:** it is not strictly necessary to catch Tribol contact mechanics errors for the two_block example that runs with the Homotopy solver.
- [X] Hide the infastructure needed to utilize "fixed" and "displacement" `ess_tdof_list`'s. That is the "user" need just pass the `ess_tdof_list`'s and the ContinuationSolver problem class should just take care of things. As is the "user" shouldn't need to construct `prolongation_` and `restriction_` `mfem::HypreParMatrix`'s as in the current implementation of `examples/contact/homotopy/two_blocks.cpp`.

For a future PR: (1) finish work begun here to cache contact mechanics function calls and (2) update the manner in which we do Dirichlet BC's. Currently we eliminate essential/displacement dofs and only expose the reduced linear system to the Homotopy solver. This has implications for the linear solvers used, such as AMG with SystemsOptions.


examples/contact/constraint_twist.cpp
```diff
@@ -13,20 +13,95 @@
...
+  std::shared_ptr<SolidWeakFormType> weak_form;
+  std::unique_ptr<serac::FiniteElementState> shape_disp;
+  std::shared_ptr<serac::Mesh> mesh;
+  std::shared_ptr<serac::ContactConstraint> constraints;
+  double time = 0.0;
+  double dt = 0.0;
+  std::vector<double> jacobian_weights = {1.0, 0.0, 0.0, 0.0};
+
+ public:
+  TiedContactProblem(std::vector<serac::FieldPtr> contact_states_, std::vector<serac::FieldPtr> all_states_,
```
white238: Our project naming conventions are class members variables end in `_`. Please swap the parameters to not have the trailing underscore but have the member variables above to have them.

thartland: Thank you @white238 for the reminder. I now recall you telling me this before with respect to another serac effort. I have addressed this issue.

examples/contact/homotopy/two_blocks.cpp
```diff
@@ -0,0 +1,394 @@
...
+  // Set the options for the paraview output files
+  paraview_dc->SetLevelsOfDetail(max_order_in_fields);
+  paraview_dc->SetHighOrderOutput(true);
+  paraview_dc->SetDataFormat(mfem::VTKFormat::BINARY);
+  paraview_dc->SetCompression(true);
+
+  return ParaviewWriter(std::move(paraview_dc), output_states, {});
+}
+
+// this example is intended to eventually replace twist.cpp
```
tupek2: this example replaces twist.cpp

examples/contact/homotopy/two_blocks.cpp
```diff
@@ -0,0 +1,394 @@
...
+  constant_force[dim - 1] = 0.0;
+
+  solid_mechanics_weak_form->addBodyIntegral(mesh->entireBodyName(), [constant_force](double /* t */, auto x) {
+    return smith::tuple{constant_force, 0.0 * smith::get<smith::DERIVATIVE>(x)};
+  });
+
+  auto residual_state_ptrs = smith::getFieldPointers(states, params);
+  auto contact_state_ptrs = smith::getFieldPointers(contact_states);
+
+  // homogeneous Dirichlet boundary conditions
```
tupek2: Are these homogeneous?

src/smith/physics/contact_constraint.hpp
```diff
@@ -83,17 +122,21 @@ class ContactConstraint : public Constraint {
...
+  mfem::Vector evaluate(double time, double dt, const std::vector<ConstFieldPtr>& fields,
+                        bool new_point = true) const override
   {
     contact_.setDisplacements(*fields[ContactFields::SHAPE], *fields[ContactFields::DISP]);
     tribol::setLagrangeMultiplierOptions(interaction_id_, tribol::ImplicitEvalMode::MORTAR_GAP);
 
-    // note: Tribol does not use cycle.
-    int cycle = 0;
-    contact_.update(cycle, time, dt);
+    if (new_point) {
```
tupek2: what does point mean in this context?  Is this checking if we need to updating things in Tribol again because contact interactions may have changed?

tupek2 (approved):

src/smith/physics/contact_constraint.hpp
thartland: Precisely. Like is the point that we are doing such and such computations at a "new_point" or was it "seen" in the previous call. This flag is intended to allow caching contact mechanics computations.

examples/contact/homotopy/two_blocks.cpp
thartland: No. The comment was outdated but has now been updated.

examples/contact/homotopy/two_blocks.cpp
thartland: It no longer does. It is a different problem and I'm not sure we even want to replace `twist.cpp`. The comment has been removed.

src/smith/physics/contact_constraint.hpp
btalamini: Thanks for the explanation. Can you make the name more clear and specific? `new_point` is too vague. "point" makes me think of points in space. It might help to make the name an imperative statement, like "refresh_contact_search" or something like that.

btalamini (changes requested):

examples/contact/homotopy/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+                        DEPENDS_ON smith_physics smith_mesh_utils continuation_solver
+		      )
+    smith_add_example_test(NAME          two_blocks
+                           COMMAND       two_blocks_example
+                           NUM_MPI_TASKS 4)
+    install(
+	FILES
+	    two_blocks.cpp
+	DESTINATION
+	    examples/smith/two_blocks
```
white238:
```suggestion
        FILES
            two_blocks.cpp
        DESTINATION
            examples/smith/two_blocks
```

src/smith/physics/contact_constraint.hpp
```diff
@@ -107,39 +150,148 @@ class ContactConstraint : public Constraint {
...
-          continue;
-        }
-        if (!J_contact->IsZeroBlock(i, j)) {
-          delete &J_contact->GetBlock(i, j);
-        }
+    auto dgdu = safelyObtainBlock(J_contact_.get(), iblock, jblock, own_blocks_);
+    return dgdu;
+  };
+
+  /** @brief Interface for computing residual contribution Jacobian_tilde^T multiplier from a vector of
```
white238: `Jacobian_tilde^T`

Is this a typo or a math thing? =)

white238:
Approved after minor nitpicks but there is a conflict.

src/smith/physics/contact_constraint.hpp
thartland: A math thing :-) but I updated this to be more descriptive.

src/smith/physics/contact_constraint.hpp
thartland: @btalamini I used "fresh_evaluation" to not be so vague as "point"/"new_point". I didn't want to use "refresh_contact_search" as this language is specific for contact but caching constraint evaluations/constraint Jacobian evaluations can be done in applications other than contact.

btalamini (approved):

PR #1465, opened 2025-09-10
dFEM: explicit dynamics gpu example

ebchin:
This is an attempt to bring the dFEM functionality online in smaller, more easily reviewable chunks.

This first PR enables required dFEM functionality to solve a forward explicit dynamics problem.  Parameters, jvp, and vjp functionality will likely be the next chunk.

Note: waiting for #1469 to merge before merging this one in


host-configs/rzvernal-toss_4_x86_64_ib_cray-rocmcc@6.2.1_hip.cmake
```diff
@@ -141,4 +143,9 @@ set(ENABLE_CLANGTIDY OFF CACHE BOOL "")
 
 set(ENABLE_DOCS OFF CACHE BOOL "")
 
+set(SERAC_ENABLE_CODEVELOP ON CACHE BOOL "")
+
+set(SERAC_DISABLE_TRIBOL ON CACHE BOOL "")
```
chapman39: imo, it would be better to unset `TRIBOL_DIR` here, or at least rename this var to `SERAC_ENABLE_TRIBOL` to avoid double-negative

cmake/thirdparty/SetupSmithThirdParty.cmake
```diff
@@ -294,6 +328,10 @@ if (NOT SERAC_THIRD_PARTY_LIBRARIES_FOUND)
         endif()
         set(MFEM_USE_UMPIRE OFF CACHE BOOL "")
         set(MFEM_USE_ZLIB ON CACHE BOOL "")
+        if(ENZYME_DIR)
+            serac_assert_is_directory(DIR_VARIABLE ENZYME_DIR)
+            set(MFEM_USE_ENZYME ON CACHE BOOL "")
```
chapman39, line 333: in case `MFEM_USE_ENZYME`'s default changes

```suggestion
            set(MFEM_USE_ENZYME ON CACHE BOOL "")
        else()
            set(MFEM_USE_ENZYME OFF CACHE BOOL "")
```

host-configs/rzvernal-toss_4_x86_64_ib_cray-rocmcc@6.2.1_hip.cmake
ebchin: I'll have a separate tribol-related PR up soon that will eliminate the need for this variable.

cmake/thirdparty/SetupSmithThirdParty.cmake
ebchin: fixed

tupek2 (approved):

src/smith/physics/dfem_mass_weak_form.hpp
```diff
@@ -0,0 +1,129 @@
...
+    if (bc_manager_) {
+      u_pred.SetSubVector(bc_manager_->allEssentialTrueDofs(), 0.0);
+    }
+
+    auto m_inv = mass_weak_form_->residual(time, dt, &u, {states[COORDINATES], params[DENSITY]});
+    m_inv.Reciprocal();
+
+    std::vector<ConstFieldPtr> pred_states = {&u_pred, &v_pred, &a, states[COORDINATES], params[DENSITY]};
+
+    auto zero_mass_resid = weak_form_->residual(time, dt, &u_pred, pred_states);
```
btalamini: I see how you arrived at this name, but it's making the mechanics opaque IMO. How about `force_residual`?

src/smith/physics/dfem_solid_weak_form.hpp
```diff
@@ -0,0 +1,157 @@
...
+   * @param mesh Problem mesh
+   * @param parameter_fe_spaces Vector of finite element spaces which are parameter arguments to the residual
+   * @return Vector of input spaces that are passed along to the dFEM differentiable operator constructor
+   */
+  std::vector<const mfem::ParFiniteElementSpace*> makeInputSpaces(
+      const mfem::ParFiniteElementSpace& test_space, const std::shared_ptr<Mesh>& mesh,
+      const std::vector<const mfem::ParFiniteElementSpace*>& parameter_fe_spaces)
+  {
+    std::vector<const mfem::ParFiniteElementSpace*> input_spaces;
+    input_spaces.reserve(4 + parameter_fe_spaces.size());
```
btalamini: I suggest storing the magic numbers in constants with a comment, like in our old physics modules.

btalamini (approved):
Fantastic work!

src/smith/physics/dfem_solid_weak_form.hpp
ebchin: fixed

src/smith/physics/dfem_mass_weak_form.hpp
ebchin: Agreed.  Done.

cmake/thirdparty/SetupSmithThirdParty.cmake
```diff
@@ -593,7 +599,7 @@ if (NOT SMITH_THIRD_PARTY_LIBRARIES_FOUND)
             $<INSTALL_INTERFACE:include>
         )
         
-        set(TRIBOL_FOUND TRUE CACHE BOOL "" FORCE)
+        set(TRIBOL_FOUND ON CACHE BOOL "" FORCE)
```
chapman39:
```suggestion
        set(TRIBOL_FOUND TRUE CACHE BOOL "" FORCE)
```

examples/explicit/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) Lawrence Livermore National Security, LLC and
+# other Smith Project Developers. See the top-level LICENSE file for
+# details.
+#
+# SPDX-License-Identifier: (BSD-3-Clause)
+
+if(SMITH_USE_DFEM)
```
chapman39, line 7: it seems like `SMITH_USE_DFEM` isn't associated with any package, but rather its to enable or disable certain tests/ examples. maybe `SMITH_ENABLE_DFEM` is more appropriate?

i still don't fully understand the difference between dfem and enzyme, so maybe this question doesn't make any sense, but could we use SMITH_USE_ENZYME here?

also last question, could this option be removed once https://github.com/LLNL/smith/pull/1477 merges?

chapman39 (changes requested):
just had a couple questions about `SMITH_USE_DFEM` option (see above) besides that lgtm

examples/explicit/CMakeLists.txt
ebchin: Yes, this bugged me too.  I like `SMITH_ENABLE_DFEM` more, but it looks so out of place here: https://github.com/LLNL/smith/blob/develop/src/smith/smith_config.hpp.in#L32-L55.  Also, we can make it go away once #1477 is merged.  Maybe we just keep it as `SMITH_USE_DFEM` for now and get rid the variable when #1477 merges?

ebchin: A few more thoughts:

Re: `SMITH_USE_ENZYME`, you could use dFEM without Enzyme, so it doesn't quite make sense here.

Also, IMO, using the `SMITH_ENABLE_DFEM` name would make sense at the CMake level (i.e. do conditional header includes instead of doing a `#cmakedefine SMITH_USE_DFEM`), but it breaks the header check in CI.

chapman39: its fine to leave it as is then, and we can remove it later. may be worth creating an issue to track it

chapman39 (approved):
ty @ebchin !

PR #1466, opened 2025-09-11
contact constraint bug fix

thartland:
I noticed the contact_constraint_twist example was not working and causing CI failures in develop. This bugfix resolves said issue. The deletes are being done in a more careful way. I assume that the source of this issue is due to not carefully destroying the objects as well as exact Jacobian being broken with latest TPLs.

### DIFF
--- a/src/serac/physics/contact_constraint.hpp
+++ b/src/serac/physics/contact_constraint.hpp
@@ -116,11 +116,22 @@ class ContactConstraint : public Constraint {
     contact_.update(cycle, time, dt);
     auto J_contact = contact_.mergedJacobian();
     J_contact->owns_blocks = false;
-    delete &J_contact->GetBlock(0, 0);
-    delete &J_contact->GetBlock(0, 1);
-    delete &J_contact->GetBlock(1, 1);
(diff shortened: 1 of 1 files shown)


tupek2 (approved):
I assume that this logic will have to be done in a few places once the other contact constraint terms are implemented.  At that point, makes sense to make a function for this safe deleting of blocks...  For now, looks good.  Regarding the exact Jacobians, I think they are only exact for enzyme builds, and sometimes we will be testing with Enzyme off?

thartland:
> I assume that this logic will have to be done in a few places once the other contact constraint terms are implemented. At that point, makes sense to make a function for this safe deleting of blocks... For now, looks good. Regarding the exact Jacobians, I think they are only exact for enzyme builds, and sometimes we will be testing with Enzyme off?

Yes, I have this safe delete propagated in various places in the other PR. A safe delete function seems like a good idea and I'll revisit that in the other PR. For now, I want ti fix the CI falures in develop that are due to the constraint interface and its usage.

ebchin (approved):

PR #1468, opened 2025-09-22
Make Application Manager print optional

chapman39:
motivational smith mr: https://lc.llnl.gov/gitlab/smith/smith/-/merge_requests/181

### DIFF
--- a/src/serac/infrastructure/about.cpp
+++ b/src/serac/infrastructure/about.cpp
@@ -62,8 +62,7 @@ std::string about()
   std::string about = "\n";
 
   // Version info
-  about += format("Serac Version:   {0}\n", version(false));
-  about += format("Git Commit SHA:  {0}\n", gitSHA());
+  about += format("Serac Version:   {0}\n", version());
   about += "\n";
 
   // General configuration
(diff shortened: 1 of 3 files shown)


tupek2:
Do we want this to call down to other things?  I think the main place is in equation_solver.cpp, where there is an mfem verbosity.  At the moment, an mfem verbosity of 0 still prints a limited amount of info.

tupek2:
Also, a bool may not exactly be what we want.  Presumable we want: 0 (no output), 1 (mfems default minimal output), 2 (lots of output).  But even that oversimplifies it.  You want to be able to turn on verbosity specifically per solver still.

chapman39:
@tupek2 I might need to rethink how I would approach usage of this `verbosity` variable if we want it to affect other aspects of the code. Since in the associated Smith MR (https://lc.llnl.gov/gitlab/smith/smith/-/merge_requests/181), I set the verbosity to `false` in order to prevent Serac from double printing run info. If setting this affects other things, that may cause unintentional changes to verbosity in Smith.

For now, maybe I should change the variable name to `doesPrintRunInfo` or something to not mislead people into thinking it does more than that. And in a future PR we could revisit Application Manager setting overall verbosity?

white238 (approved):

tupek2 (approved):

PR #1471, opened 2025-09-25
Tupek/print level adjustments

tupek2:
It looks like print_level=0 for the nonlinear solvers doesn't print much, but our default was 1.  This PR makes it so print_level=0 will print nothing (even solver error warnings), but keeps the default to 1, which prints a bit of nonlinear solver convergence information.


src/serac/numerics/equation_solver.cpp
```diff
@@ -197,10 +204,10 @@ class NewtonSolver : public mfem::NewtonSolver {
     final_iter = it;
     final_norm = norm;
 
-    if (print_options.summary || (!converged && print_options.warnings) || print_options.first_and_last) {
-      mfem::out << "Newton: Number of iterations: " << final_iter << '\n' << "   ||r|| = " << final_norm << '\n';
+    if (print_level == 1) {
```
chapman39, line 207: if the print level is 2, wouldn't you want to print this section too? there are a couple other sections like this

`if (print_level >= 1)` ?

tupek2: Its a bit subtle/confusing, but for print level 2, this print happens in a loop below.  If I do ">=1" here, it will print that line twice.

src/serac/numerics/equation_solver.cpp
```diff
@@ -88,13 +92,16 @@ class NewtonSolver : public mfem::NewtonSolver {
     MFEM_ASSERT(oper != NULL, "the Operator is not set (use SetOperator).");
     MFEM_ASSERT(prec != NULL, "the Solver is not set (use SetSolver).");
 
+    print_level = print_options.iterations ? 1 : print_level;
+    print_level = print_options.summary ? 2 : print_level;
```
chapman39, line 96: where is print_options set/ defined?

src/serac/numerics/equation_solver.cpp
chapman39: oh that makes sense, thanks for clarifying

chapman39 (approved):
thanks @tupek2 !

src/serac/numerics/equation_solver.cpp
tupek2: This is a member of the base mfem solver class.  I basically have to reverse engineer how they map print level into these options.

btalamini (approved):

PR #1476, opened 2025-10-02
Add check for no elements on by local rank

kswartz92:
On multiple occasions I've lost hours debugging when a mesh has no elements on a particular rank (this usually comes up when I have both a 2D and a 3D mesh on the same communicator). I'm not 100% sure where the error message comes from but it's a cryptic out-of-range error that isn't easy to debug, so I thought this check would be useful to alert users of what is happening.

### DIFF
--- a/src/serac/physics/mesh.cpp
+++ b/src/serac/physics/mesh.cpp
@@ -49,7 +49,7 @@ Mesh::Mesh(mfem::ParMesh&& mesh, const std::string& meshtag) : mesh_tag_(meshtag
 
 Mesh::errorIfRankHasNoElements() const
 {
-  SLIC_ERROR_IF(mfem_mesh_.GetNE() == 0, "Local size of mesh is 0 which will cause out-of-range error");
+  SLIC_ERROR_IF(mfem_mesh_->GetNE() == 0, "Local size of mesh is 0 which will cause out-of-range error");
 }
 
 MPI_Comm Mesh::getComm() const { return mfem_mesh_->GetComm(); }
--- a/src/serac/physics/mesh.cpp
+++ b/src/serac/physics/mesh.cpp
@@ -47,7 +47,7 @@ Mesh::Mesh(mfem::ParMesh&& mesh, const std::string& meshtag) : mesh_tag_(meshtag
   createDomains();
 }
 
-Mesh::errorIfRankHasNoElements() const
+void Mesh::errorIfRankHasNoElements() const
 {
   SLIC_ERROR_IF(mfem_mesh_->GetNE() == 0, "Local size of mesh is 0 which will cause out-of-range error");
 }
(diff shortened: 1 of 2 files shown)


tupek2 (approved):
Great idea.

src/serac/physics/mesh.cpp
```diff
@@ -41,9 +43,15 @@ Mesh::Mesh(mfem::ParMesh&& mesh, const std::string& meshtag) : mesh_tag_(meshtag
   meshtmp->EnsureNodes();
   meshtmp->ExchangeFaceNbrData();
   mfem_mesh_ = &serac::StateManager::setMesh(std::move(meshtmp), mesh_tag_);
+  errorIfRankHasNoElements();
   createDomains();
 }
 
+Mesh::errorIfRankHasNoElements() const
+{
+  SLIC_ERROR_IF(mfem_mesh_.GetNE() == 0, "Local size of mesh is 0 which will cause out-of-range error");
```
chapman39:
```suggestion
  SLIC_ERROR_IF(mfem_mesh_->GetNE() == 0, "Local size of mesh is 0 which will cause out-of-range error");
```

src/serac/physics/mesh.cpp
```diff
@@ -41,9 +43,15 @@ Mesh::Mesh(mfem::ParMesh&& mesh, const std::string& meshtag) : mesh_tag_(meshtag
   meshtmp->EnsureNodes();
   meshtmp->ExchangeFaceNbrData();
   mfem_mesh_ = &serac::StateManager::setMesh(std::move(meshtmp), mesh_tag_);
+  errorIfRankHasNoElements();
   createDomains();
 }
 
+Mesh::errorIfRankHasNoElements() const
+{
+  SLIC_ERROR_IF(mfem_mesh_->GetNE() == 0, "Local size of mesh is 0 which will cause out-of-range error");
```
white238:
```suggestion
  SLIC_ERROR_IF(mfem_mesh_->GetNE() == 0, "After refining and distributing mesh, local size of mesh is 0");
```

not sure the last part is helpful and may cause confusion.

src/serac/physics/mesh.cpp
```diff
@@ -41,9 +43,15 @@ Mesh::Mesh(mfem::ParMesh&& mesh, const std::string& meshtag) : mesh_tag_(meshtag
   meshtmp->EnsureNodes();
   meshtmp->ExchangeFaceNbrData();
   mfem_mesh_ = &serac::StateManager::setMesh(std::move(meshtmp), mesh_tag_);
+  errorIfRankHasNoElements();
   createDomains();
 }
 
+Mesh::errorIfRankHasNoElements() const
```
white238:
```suggestion
void Mesh::errorIfRankHasNoElements() const
```

white238 (approved):
Thanks @kswartz92 ! We need more of these helpful checks before things go horribly wrong.

PR #1474, opened 2025-09-26
Update Docker Containers w/ Enzyme+Spack 1.0 + clang-format 19

chapman39:
- [x] updates docker containers to ubuntu 22 -> 24
  - [x] clang 14 -> 19
  - [x] gcc  13 -> 14
  - [x] build hdf5 w/o shared libs fixes #1480
- [x] cleanup dockerfile setup
- [x] adds `serac_add_library` and `serac_add_executable` cmake macros to apply device target everywhere (only really relevant to cuda/ hip builds)
- [x] temporarily removes cuda docker container (will revisit in future pr see issue https://github.com/LLNL/serac/issues/1479)
- [x] new docker host configs
- [x] setup and apply clang-format 19 style fixes #1455

Note: the only changes in cpp/ hpp `src` are style-related! https://github.com/LLNL/serac/pull/1474/commits/e546e6b810b4cd37f7aa31678beb18aac538fe5d

### before merge

- [x] modify github actions ci in serac github settings

### related links

- https://github.com/LLNL/axom/pull/1673
- https://github.com/LLNL/axom/pull/1690

### followup issues

- https://github.com/LLNL/serac/issues/1481
- https://github.com/LLNL/serac/issues/1479

PR #1478, opened 2025-10-06
Fix `ContactConstraint::evaluate()`

ebchin:
Explicitly copies gap values from an `mfem::HypreParVector` to an `mfem::Vector` (instead of being moved) to prevent the data in the `mfem::HypreParVector` from going out of scope.  See [this MFEM issue](https://github.com/mfem/mfem/issues/5029) for more details.

More generally, we might want to consider returning a `serac::FiniteElementDual` in the constraint interface here.  Thoughts @thartland @tupek2?


tupek2 (approved):
strictly speaking, this is not a TODO.  Nice find!

src/serac/physics/contact_constraint.hpp
```diff
@@ -94,7 +94,11 @@ class ContactConstraint : public Constraint {
     // note: Tribol does not use cycle.
     int cycle = 0;
     contact_.update(cycle, time, dt);
-    return contact_.mergedGaps(false);
+    auto gaps_hpv = contact_.mergedGaps(false);
+    // TODO: this copy is needed to prevent the HypreParVector pointer from going out of scope.  see
```
white238:
```suggestion
    // Note: this copy is needed to prevent the HypreParVector pointer from going out of scope.  see
```
If this is a TODO please add what the TODO is

chapman39 (approved):
thanks @ebchin !

thartland (approved):

thartland:
> Explicitly copies gap values from an `mfem::HypreParVector` to an `mfem::Vector` (instead of being moved) to prevent the data in the `mfem::HypreParVector` from going out of scope. See [this MFEM issue](https://github.com/mfem/mfem/issues/5029) for more details.
> 
> More generally, we might want to consider returning a `serac::FiniteElementDual` in the constraint interface here. Thoughts @thartland @tupek2?

I am ok returning a FiniteElementDual in the constraint interface and can work on that if that's what we'd like. There is a bit assumed in that the constraints will then be assumed to come from a discretization. But perhaps that will not at all limit us in the context of applications we are targetting.

white238 (approved):

tupek2:
Oh, we probably should not have merged this.  That was our only exodus reading text.  Like 90% of our usage is reading exodus files, which seems to be broken.

chapman39:
@tupek2 it was only broken on the docker containers due to an improperly installed hdf5. i fixed the issue and re-enabled that exodus file on this pr https://github.com/LLNL/serac/pull/1474

PR #1494, opened 2025-11-18
Guard and clarify team scripts and builds

white238:
This tripped up a user. This should make it very clear.... hopefully


tupek2 (commented):
Just curious, if the user tries to build with -DSMITH_ENABLE_CODEVELOP=ON, will that work even if they don't have that access?

scripts/llnl/common_build_functions.py
```diff
@@ -593,6 +596,39 @@ def build_devtools(builds_dir, timestamp, short_path, report_to_stdout = False):
     return res
 
 
+def ensure_on_lc_and_group_permissions():
```
chapman39: i think we should still allow the use of these scripts on non-LC machines. for example, this is breaking the github actions CI. as long as SYS_TYPE is set, these scripts work completely fine.

chapman39:
@tupek2 codevelop builds still rely on a host config to find other TPLs, so if a user doesn't have access to them, it will not work.

scripts/llnl/common_build_functions.py
white238: Yea, I knew I missed something. I think the `build_src.py` script is the only one that makes sense to run not on LC.

chapman39 (approved):

tupek2 (approved):

PR #1504, opened 2025-12-11
Enable coverage using GCC docker image

chapman39:
eventually we should use the llvm image but enzyme doesn't seem to pair well with llvm

related issue https://github.com/LLNL/smith/issues/1481


chapman39:
devtools aren't built on GCC images. next TPL build i will attempt to change that, in the meantime, i'll just close this pr

PR #1511, opened 2025-12-16
Tupek/gretl cmake

tupek2:
this enables smith to be built by other projects containing smith as a submodule


chapman39 (approved):
this looks good to me

cmake/thirdparty/SetupSmithThirdParty.cmake
```diff
@@ -568,19 +568,20 @@ if (NOT SMITH_THIRD_PARTY_LIBRARIES_FOUND)
...
         else()
             set(GRETL_SOURCE_DIR "${PROJECT_SOURCE_DIR}/gretl" CACHE PATH "")
-
             if (NOT EXISTS "${GRETL_SOURCE_DIR}/CMakeLists.txt")
-                message(FATAL_ERROR
-                    "The gretl repo is not present. "
-                    "Either run the following command in your git repository: \n"
-                    "    git submodule update --init --recursive\n"
-                    "Or add -DGRETL_SOURCE_DIR=/path/to/gretl to your CMake command." )
+                set(GRETL_SOURCE_DIR "${PROJECT_SOURCE_DIR}/smith/gretl" CACHE PATH "" FORCE)
```
white238: I think something like this would be cleaner:

```
if(NOT DEFINED GRETL_SOURCE_DIR)
    set(GRETL_SOURCE_DIR "${PROJECT_SOURCE_DIR}/gretl" CACHE PATH "")
endif()

// then error if ${GRETL_SOURCE_DIR}/CMakeLists.txt doesnt exist
```

tupek2: I think I cleaned it up a bit in the latest commit, but I still needed to try looking in /smith/gretl in some cases

btalamini (approved):

PR #1526, opened 2026-01-21
Fix contact memory leaks + a bug

ebchin:


src/smith/physics/contact/contact_data.cpp
```diff
@@ -161,7 +161,7 @@ std::unique_ptr<mfem::BlockOperator> ContactData::mergedJacobian() const
       // zero out rows and cols not in the active set
       auto inactive_dofs = interactions_[i].inactiveDofs();
       dgdu->EliminateRows(inactive_dofs);
-      dfdp->EliminateCols(inactive_dofs);
+      auto dfdp_elim = std::unique_ptr<mfem::HypreParMatrix>(dfdp->EliminateCols(inactive_dofs));
```
tupek2, line 164: Was this the main leak?

tupek2 (approved):
Good finds.  It looks like most of this is just cleanup to make the logic easier?  But maybe also they were memory leaking.  That first one above certainly looked like a real issue, am I missing others?

src/smith/physics/contact/contact_data.cpp
```diff
@@ -279,11 +279,11 @@ void ContactData::residualFunction(const mfem::Vector& u_shape, const mfem::Vect
...
-std::unique_ptr<mfem::BlockOperator> ContactData::jacobianFunction(mfem::HypreParMatrix* orig_J) const
+std::unique_ptr<mfem::BlockOperator> ContactData::jacobianFunction(std::unique_ptr<mfem::HypreParMatrix> orig_J) const
 {
   auto J_contact = mergedJacobian();
   if (J_contact->IsZeroBlock(0, 0)) {
-    J_contact->SetBlock(0, 0, orig_J);
+    J_contact->SetBlock(0, 0, orig_J.release());
   } else {
     J_contact->SetBlock(0, 0,
                         mfem::Add(1.0, *orig_J, 1.0, static_cast<mfem::HypreParMatrix&>(J_contact->GetBlock(0, 0))));
```
ebchin, line 289: `orig_J` was leaking here too

src/smith/physics/contact/contact_data.cpp
ebchin: One more leak in jacobianFunction().  Added a comment to show where.

white238 (approved):
I am fine with this being merged due to a codevelop customer needing this fixed and the assumption we will update tribol very soon including the RDC fix.

PR #1538, opened 2026-02-11
Less ambiguous ContinuationSolver includes

thartland:
In order to include files such as "utilities.hpp" from ContinuationSolvers in a smith code the line previously was fairly ambiguous

`#include "utilities.hpp"`

this has now been updated so that smith example/test codes have the include:

`#include "continuationsolvers/utilities.hpp"`.

similarly for including other ContinuationSolver specific headers.


white238 (approved):
Thanks @thartland ! I restarted your pipeline because it failed for a reason outside of this PR.

tupek2 (approved):

PR #1540, opened 2026-02-12
Fix ASAN bugs and add nightly ASAN pipeline schedule

chapman39:
Adds CI job with address sanitizer enabled as a part of our nightly CI pipelines.

https://lc.llnl.gov/gitlab/smith/smith/-/pipelines/689568

fixes 13 tests:

- [x] domain_tests
- [x] equationsolver (PetscKSPSolver needed destructor)
- [x] equationsolver_petsc
- [x] eigensolver (internal mvapich2 leak, added to suppressions)
- [x] contact_finite_diff
- [x] tribol_finite_diff
- [x] parameterized_thermomechanics_example
- [x] solid_shape
- [x] contact_patch (fixed with tribol tpl update)
- [x] contact_patch_tied
- [x] contact_beam (fixed with tribol tpl update)
- [x] contact_beam_amgf (fixed with tribol tpl update)
- [x] mesh_generation (leak when calling mfem::Extrude2D)
- [x] test_solid_mechanics_state_advancer (heap use after free) https://github.com/llnl/gretl/pull/7

TODO before merge

- [x] set pipeline back to develop branch


.gitlab-ci.yml
```diff
@@ -90,6 +90,7 @@ stages:
       - _smith_build_and_test_*/build-*/output.log*.txt
       - _smith_build_and_test_*/build-*/*.*.logs/*.log
       - _smith_build_and_test_*/build-*/*.*.logs/*.err
+      - asan_logs/
```
chapman39:
```suggestion
```

src/smith/mesh_utils/mesh_utils.cpp
```diff
@@ -240,7 +240,8 @@ mfem::Mesh buildCylinderMesh(int radial_refinement, int elements_lengthwise, dou
     mesh.SetVertices(new_vertices);
   }
 
-  return mfem::Mesh(*mfem::Extrude2D(&mesh, elements_lengthwise, height));
+  auto extruded = std::unique_ptr<mfem::Mesh>(mfem::Extrude2D(&mesh, elements_lengthwise, height));
```
white238: This change concerns me. It relies on mfem::Mesh move construction being safe and correct, since we are moving *extruded into the return value and then immediately destroying the moved-from heap object via unique_ptr.

It is also hard to read, because we allocate an object, move from it, and destroy it right away.

What was the specific ASan error being fixed here?

Would it make more sense to have these functions return a unique_ptr?

chapman39: > Direct leak of 34752 byte(s) in 6 object(s) allocated from: #0 0x555555d43cad in operator new(unsigned long) /root/.spack/stage/sly1/spack-stage-llvm-19.1.3-gy2lu5xbi4csr2k47emlajzfs5mlsd4g/spack-src/compiler-rt/lib/asan/asan_new_delete.cpp:86:3 #1 0x555556251bea in mfem::Extrude2D(mfem::Mesh*, int, double) (/usr/WS2/meemee/smith/repo/[build-rzwhippet-toss_4_x86_64_ib-llvm@19.1.3-debug](mailto:build-rzwhippet-toss_4_x86_64_ib-llvm@19.1.3-debug)/tests/mesh_generation+0xcfdbea)

my understanding is previously we were creating a copy of the extrude2D mesh, but not removing the original.

instead of using unique pointer, i could alternatively delete the original mesh object after the copy.

or like you said, we could change the return type to unique ptr mfem mesh.

chapman39: to prevent an inconsistency between the other mesh functions, and limiting api changes, i opted to delete the raw pointer rather than change the return type of those cylinder mesh functions.

```cpp
  mfem::Mesh* extruded_ptr = mfem::Extrude2D(&mesh, elements_lengthwise, height);
  mfem::Mesh extruded_obj(*extruded_ptr);
  delete extruded_ptr;
  return extruded_obj;
```

now its a little more clear whats going on.

if you want, we could switch the return type to unique ptr, but in that case, we'd probably want to change all of those mesh functions to return a unique ptr as well. lmk what you think.

white238 (approved):
Thanks @chapman39 !

tupek2 (approved):

PR #1542, opened 2026-02-16
Algebraic multigrid with filtering contact (AMGFContact) preconditioner for contact linear systems

thartland:
This PR adds the capability of using AMGF (algebraic multigrid with filtering [arXiv preprint](https://arxiv.org/abs/2505.18576))-preconditioned Krylov-subspace linear solvers for the Newton linear solves that arise in outer loop solves of contact problems.

- [x] Obtain list of contact dofs via `ContactData::addContactInteraction` calls.
- [x] Create a boundary_dof --> subspace transfer operator (`mfem::HypreParMatrix`) call. Much of this infastructure is in the utilities of continuationsolvers but a version most suitable for usage in `smith` would be best for less complex code.
- [x] Obtain subspace transfer (displacement dofs --> contact dofs) operator via the list of contact dofs in appropriate place for passing the subspace transfer operator to the preconditioner provided that said linear solver preconditioner is AMGF. 
- [x] Resolve usage errors conflating `print_level` and `preconditioner_print_level` from linear solver options in the `buildPreconditioner` method.
- [x] Test.

For future investigation/potential PR(s)
(1) Make `AMGFContact` robust to indefiniteness issues stemming from using exact Hessians in Newton solves.
(2) More careful consideration for setting outer/inner tolerances (as well as norms used) for Newton-Krylov.

Note -- `AMGFContact` preconditioned Krylov-subspace methods will fails unless the problem is a contact problem. The failure is indicated by mfem generated error messages indicating that the subspace transfer operator has not been set.


src/smith/numerics/solver_config.hpp
```diff
@@ -327,7 +341,8 @@ enum class Preconditioner
   HypreILU,         /**< Hypre's Incomplete LU */
   AMGX,             /**< NVIDIA's AMGX GPU-enabled algebraic multi-grid, GPU builds only */
   Petsc,            /**< PETSc preconditioner,  */
-  None              /**< No preconditioner used */
+  None,             /**< No preconditioner used */
```
ebchin: Should we swap these? AMGF before None

src/smith/numerics/solver_config.hpp
```diff
@@ -351,6 +366,8 @@ inline std::string preconditionerName(Preconditioner p)
       return "Petsc";
     case Preconditioner::None:
       return "None";
+    case Preconditioner::AMGFContact:
```
ebchin, line 368: Swap here too

src/smith/physics/contact/contact_data.cpp
```diff
@@ -43,6 +43,26 @@ void ContactData::addContactInteraction(int interaction_id, const std::set<int>&
     num_pressure_dofs_ += interactions_.back().numPressureDofs();
     offsets_up_to_date_ = false;
   }
+  // specify all contact boundaries
+  mfem::Array<int> ess_bdry;
```
ebchin: I think we could pick a name for this that better reflects what it's doing. Maybe "contact_bdry_attribs" or something?

ebchin (approved):
Thanks @thartland, looking forward to using it!

src/smith/physics/contact/contact_data.cpp
```diff
@@ -363,6 +383,62 @@ void ContactData::updateDofOffsets() const
...
+std::unique_ptr<mfem::HypreParMatrix> ContactData::contactSubspaceTransferOperator()
+{
+  const MPI_Comm comm = reference_nodes_->ParFESpace()->GetComm();
+  HYPRE_BigInt* col_offsets = reference_nodes_->ParFESpace()->GetTrueDofOffsets();
+  HYPRE_BigInt ncols_glb = reference_nodes_->ParFESpace()->GlobalTrueVSize();
+
+  // number of rows of the restriction
+  // operator owned by the local MPI process
+  int nrows_loc = contact_dofs_.Size();
+  // should nrows_glb be of type HYPRE_BigInt?
```
tupek2, line 395: Ideally, yes.  this would limit us to 2 billions dofs, max.  There are likely other places we'd need to fix this though.

src/smith/physics/tests/contact_beam_amgf.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  auto mesh = std::make_shared<smith::Mesh>(buildMeshFromFile(filename), "beam_mesh", 1, 0);
+
+  LinearSolverOptions linear_options{.linear_solver = LinearSolver::CG,
+                                     .preconditioner = Preconditioner::AMGFContact,
+                                     .relative_tol = 0.0,
+                                     .absolute_tol = 1.0e-13,
+                                     .print_level = 2};
+#ifndef MFEM_USE_STRUMPACK
+  SLIC_INFO_ROOT("Contact requires MFEM built with strumpack.");
```
tupek2, line 58: technically, contact with AMGF solver requires strumpack

src/smith/physics/contact/contact_data.cpp
```diff
@@ -428,6 +504,13 @@ void ContactData::setDisplacements([[maybe_unused]] const mfem::Vector& u_shape,
 {
 }
 
+std::unique_ptr<mfem::HypreParMatrix> ContactData::contactSubspaceTransferOperator()
+{
+  // TODO: if there is no tribol contact what should be done here?
+  std::unique_ptr<mfem::HypreParMatrix> transfer_operator = nullptr;
```
tupek2, line 509: presumably we error out.  maybe that is already happening elsewhere?

src/smith/numerics/solver_config.hpp
```diff
@@ -250,6 +250,20 @@ struct AMGXOptions {
...
 };
 
+/**
+ * @brief Stores the configuration information for an AMGFContact preconditioner
+ */
+struct AMGFContactOptions {
+  /**
+   * @brief The amg relaxation type
+   */
+  int relax_type = 88;
```
white238: What do these numbers mean?

src/smith/numerics/solver_config.hpp
```diff
@@ -369,6 +386,7 @@ inline std::map<std::string, Preconditioner> preconditionerMap = {
     {"AMGX", Preconditioner::AMGX},
     {"Petsc", Preconditioner::Petsc},
     {"None", Preconditioner::None},
+    {"AMGFContact", Preconditioner::AMGFContact},
```
white238, line 389: And here.

src/smith/physics/tests/contact_beam_amgf.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "mpi.h"
+#include "gtest/gtest.h"
+#include "mfem.hpp"
+
+#include "smith/physics/solid_mechanics_contact.hpp"
+#include "smith/numerics/functional/domain.hpp"
+#include "smith/physics/state/state_manager.hpp"
+#include "smith/physics/mesh.hpp"
+#include "smith/physics/materials/solid_material.hpp"
+#include "smith/smith_config.hpp"
```
white238:
```suggestion
#include "smith/smith_config.hpp"
#include "smith/physics/solid_mechanics_contact.hpp"
#include "smith/numerics/functional/domain.hpp"
#include "smith/physics/state/state_manager.hpp"
#include "smith/physics/mesh.hpp"
#include "smith/physics/materials/solid_material.hpp"
```

src/smith/physics/tests/contact_beam_amgf.cpp
white238, line 58: Should this be an error and you can also make it a compile time failure like this:

```
#ifndef MFEM_USE_STRUMPACK
#error "Strupack is required for contact amgf tests"
#endif
```

src/smith/physics/solid_mechanics_contact.hpp
```diff
@@ -377,8 +385,27 @@ class SolidMechanicsContact<order, dim, Parameters<parameter_space...>,
...
+      if (amgf_prec) {
+        // compute contact_dof_prolongation
+        computeContactSubspaceTransferOperator();
+        // set AMGF subspace transfer operator
+        amgf_prec->SetFilteredSubspaceTransferOperator(*(contact_dof_prolongation_.get()));
+        // set the filteredsubspace solver component of AMGF
+        auto iterative_solver = dynamic_cast<mfem::IterativeSolver*>(&lin_solver);
+        SLIC_ERROR_ROOT_IF(!iterative_solver,
+                           "AMGFContact should only be used as a preconditioner for an iterative solver");
+#ifdef SMITH_USE_MPI
```
white238: We do not allow MPI to be disabled. This isn't needed.

tupek2 (commented):
Are these subspace transfers being computed once up front, or would they occur every time a new Jacobian is assembled (say in a Newton iteration loop).  Maybe the performance if not so bad, but something to consider down the line perhaps.

src/smith/physics/tests/contact_beam_amgf.cpp
thartland: Done.

src/smith/physics/tests/contact_beam_amgf.cpp
thartland: This is addressed now.

thartland:
> Are these subspace transfers being computed once up front, or would they occur every time a new Jacobian is assembled (say in a Newton iteration loop). Maybe the performance if not so bad, but something to consider down the line perhaps.

Good point. The subspace transfer operator is computed once every time `SolidMechanicsContact::warmStartDisplacementContact` is called which seems to be once per time-step. This does seem like too much and don't see a case for which the set of potential contact dofs would evolve in time.

thartland:
> Are these subspace transfers being computed once up front, or would they occur every time a new Jacobian is assembled (say in a Newton iteration loop). Maybe the performance if not so bad, but something to consider down the line perhaps.

This has now been updated and the subspace transfer operator is computed once per contact simulation.

src/smith/numerics/solver_config.hpp
thartland: Updated comments now explain what these numbers corresponds to.

src/smith/physics/tests/contact_beam_amgf.cpp
thartland: This suggestion has been incorporated in recent updates to this PR.

thartland:
@white238 -- regarding the guarding against usage of the `contact_beam_amgf.cpp` test in non-Tribol/non-Strumpack builds. Currently it is guarded against usage in Tribol and similar Tribol contact test cases have similar error messages when using non-Strumpack builds. See [here](https://github.com/llnl/smith/blob/develop/src/smith/physics/tests/contact_finite_diff.cpp#L74-L80). Do any of the mfem TPLs that we test against have non STRUMPACK builds?

Nonetheless I put STRUMPACK specific guards for including this test in non strumpack builds.

white238:
> @white238 -- regarding the guarding against usage of the `contact_beam_amgf.cpp` test in non-Tribol/non-Strumpack builds. Currently it is guarded against usage in Tribol and similar Tribol contact test cases have similar error messages when using non-Strumpack builds. See [here](https://github.com/llnl/smith/blob/develop/src/smith/physics/tests/contact_finite_diff.cpp#L74-L80). Do any of the mfem TPLs that we test against have non STRUMPACK builds?
> 
> Nonetheless I put STRUMPACK specific guards for including this test in non strumpack builds.

Thanks for putting the check in.. we should verify turning off our optional TPLs to see if they are properly guarded but that is outside the scope of this PR.

tupek2 (approved):

white238 (approved):

PR #1547, opened 2026-02-23
Setup Codex sandbox

white238:
This is aimed at a good base for using the Codex Sandbox on LC. I put in effort to allow it to work on non-LC machines but haven't tried it personally. For example, the compute/login node script will default to compute if not on LC and put some guardrails recommended by ChatGPT that might sandbox you outside of a sandbox.

To be very clear, these are here to help but cannot be fully trusted. Only use Agentic AI on repos that you know dont have any work you care to lose since it can still revert/commit/delete.


AGENTS.md
```diff
@@ -0,0 +1,24 @@
...
+- Keep changes minimal and localized; avoid drive-by refactors.
+
+## Testing Guidelines
+- Tests use GoogleTest (`TEST(...)`) and are registered via CMake/BLT.
+- Add or update tests near the code they cover (e.g., `src/smith/physics/state/tests/`).
+- If a test requires MPI/GPU/TPLs, gate it behind the existing CMake options and keep a CPU-only unit test when possible.
+
+## Commit & Pull Request Guidelines
+- Commit messages in this repo are typically short and imperative (e.g., “fix …”, “update …”, “add …”).
+- PRs should include: a clear description, how to reproduce/verify, and the relevant `ctest` command(s) you ran.
```
tupek2: git commits should probably be done via /commit custom slash commands so they can be done with less token usage.

AGENTS.md
```diff
@@ -0,0 +1,24 @@
...
+- `cmake/` + `host-configs/`: BLT-based CMake build system and machine configs.
+
+## Coding Style & Naming Conventions
+- C++ formatting is enforced via `.clang-format`; prefer running `cmake --build build --target style` before pushing.
+- Use existing naming patterns: `PascalCase` types, `camelCase` methods, `snake_case_` members, `SCREAMING_SNAKE_CASE` constants.
+- Keep changes minimal and localized; avoid drive-by refactors.
+
+## Testing Guidelines
+- Tests use GoogleTest (`TEST(...)`) and are registered via CMake/BLT.
+- Add or update tests near the code they cover (e.g., `src/smith/physics/state/tests/`).
```
tupek2: I would only add this if it makes this mistake semi-consistently, which I doubt it will.

AGENTS.md
```diff
@@ -0,0 +1,24 @@
...
+- `src/drivers/`: driver entrypoints (currently mostly disabled in CMake).
+- `examples/`: runnable example executables (e.g., `examples/conduction/`).
+- `src/tests/` and `src/smith/**/tests/`: GoogleTest-based unit/smoke tests.
+- `tests/`: external test-data submodule (not compiled; used by CI/integration).
+- `cmake/` + `host-configs/`: BLT-based CMake build system and machine configs.
+
+## Coding Style & Naming Conventions
+- C++ formatting is enforced via `.clang-format`; prefer running `cmake --build build --target style` before pushing.
+- Use existing naming patterns: `PascalCase` types, `camelCase` methods, `snake_case_` members, `SCREAMING_SNAKE_CASE` constants.
+- Keep changes minimal and localized; avoid drive-by refactors.
```
tupek2: I would only add this if it makes this mistake semi-consistently, which I doubt it will.

AGENTS.md
```diff
@@ -0,0 +1,24 @@
...
+## Project Structure & Module Organization
+- `src/smith/`: core C++ libraries (physics, numerics, infrastructure).
+- `src/drivers/`: driver entrypoints (currently mostly disabled in CMake).
+- `examples/`: runnable example executables (e.g., `examples/conduction/`).
+- `src/tests/` and `src/smith/**/tests/`: GoogleTest-based unit/smoke tests.
+- `tests/`: external test-data submodule (not compiled; used by CI/integration).
+- `cmake/` + `host-configs/`: BLT-based CMake build system and machine configs.
+
+## Coding Style & Naming Conventions
+- C++ formatting is enforced via `.clang-format`; prefer running `cmake --build build --target style` before pushing.
```
tupek2: there should be a style hook after refactors or slash command to format.  I don't want it wasting tokens with this, and it can just be enforced at the end.

AGENTS.md
```diff
@@ -0,0 +1,24 @@
...
+- Use existing naming patterns: `PascalCase` types, `camelCase` methods, `snake_case_` members, `SCREAMING_SNAKE_CASE` constants.
+- Keep changes minimal and localized; avoid drive-by refactors.
+
+## Testing Guidelines
+- Tests use GoogleTest (`TEST(...)`) and are registered via CMake/BLT.
+- Add or update tests near the code they cover (e.g., `src/smith/physics/state/tests/`).
+- If a test requires MPI/GPU/TPLs, gate it behind the existing CMake options and keep a CPU-only unit test when possible.
+
+## Commit & Pull Request Guidelines
+- Commit messages in this repo are typically short and imperative (e.g., “fix …”, “update …”, “add …”).
```
tupek2: This should be handled maybe in a /commit custom slash command

AGENTS.md
```diff
@@ -0,0 +1,9 @@
+# Repository Guidelines
```
tupek2: Having done this stuff now for a bit, all the reading and experience I have says not to have a common AGENTS.md file for the team.  The skills are fine, also the sandboxing is important of course, but I will setup my AGENTS.md to be very specific to the computer I'm on, the task I'm doing, the things I see the LLM do wrong.  its an iterative process.  You could look at my current AGENTS.md and see its not something that makes sense to be common or shared by the team.  Every LLM session injects this into context, and its very situation specific what I want to be injecting into my sessions like this.  So, my recommendation is we rename this file (claude as well, though I guess claude should probably be deleted for now) to STARTER_AGENTS.md or SUGGESTED_AGENTS.md or INITIAL_AGENTS.md.  People can choose to copy that to use as their own AGENTS.md (or not) as they see fit.  Otherwise I'll be dealing with merge conflicts on the file forever, and I'd argue its best for people learning this stuff to figure out how to manage their own AGENTS.md to get experience.

skills/building/SKILL.md
```diff
@@ -0,0 +1,76 @@
...
+1) (Recommended) Initialize required submodules (at least `cmake/blt`):
+
+```bash
+git submodule update --init --recursive
+```
+
+2) Configure (recommended wrapper around CMake):
+
+```bash
+./config-build.py -bp build -ip install -hc host-configs/<file>.cmake -bt Debug --exportcompilercommands
```
tupek2: Is this likely to default to everyone running debug by default?   Its pretty common for people to tell me: 'this or that smith problem is slow' only for me to (after a day of work) realize they were default building debug.  Now my default response to these this is: just build in release (which they often still ignore).  I personally rarely build, test, or debug in debug.  Maybe we want a release_build skill, and a debug_build skill.

tupek2 (changes requested):
If you are interested in getting into the next level of this agentic stuff (which I haven't done yet), I'd recommend you research custom /slash commands (maybe for build or testing), and also the use of LLM hooks (where, for example, you can make a hook that after the agent makes code changes, it automatically builds the code, runs the tests, makes style, etc).  Just something to be aware of that I'm not sure the initial tutorial covered.  I'm personally more interested in these things than skills, which are mostly ignored.

AGENTS.md
tupek2: we would then put agents.md, AGENTS.md, CLAUDE.md and claude.md in .gitignore

white238: Removed the agents file. Do you think its worth it to add an ignore if we are recommending against them?

skills/building/SKILL.md
white238: Swapped to RelWithDebInfo by default unless specifically asked for a debug build.

AGENTS.md
tupek2: We are not recommending against them.  We are highly encouraging them.  But its an individual thing, so I've added CLAUDE.md, claude.md, AGENT.md, agent.md to my local git ignore.

skills/building/SKILL.md
tupek2: Release please.  I people want to debug, the should probably run in debug.

white238: Maybe we should discuss this in the technical meeting. RelWithDebInfo is nearly as fast (-O2 vs -O3) but does get you debug symbols. I was envisioning these files are mainly for out development. Are you expecting them to be used by people outside of our team? I feel this is a healthy middle ground and if someone really cares they can put it in their agents file or just tell it to use release.

AGENTS.md
white238: done.

btalamini (approved):

tupek2 (approved):

white238:
dont merge this.. i just found out its not using my skills and how to fix it.

skills/building/SKILL.md
white238: I removed the build type from all command except the asan build where i told it to default to debug. I also told it to use the `-bt <CMake Build Type>` option if asked. I verified it is now finding the skills w/ the new symlink and tested several different variations of prompts and it seems to work like I would expect at this point.

Side thought: I like that I asked codex to make these skills and iterated on them with it.. but it couldn't find them after. It does now.

white238:
> dont merge this.. i just found out its not using my skills and how to fix it.

Fixed now.

white238:
Example of `AGENTS.md` that defaults to building debug if that is better for you:

```
# AGENTS.md

## Build Configuration

Unless explicitly instructed otherwise, always configure and build the project in **Debug** mode.

Do not default to Release or RelWithDebInfo builds.
```

You can also say "build in debug and with GCC". That worked for me.

skills/cpp-style/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+cmake --build build --target style
+```
+
+## File structure
+
+- **License header**: Keep the existing copyright + SPDX block at the top of C++ files.
+- **Header guards**: Use `#pragma once` for headers.
+- **Doxygen file header**: Use a `/** ... */` block with `@file` and `@brief`.
+- **Namespaces**: Prefer nested namespaces like `namespace smith::input { ... }`.
+- **Namespace closing comment**: Close with `}  // namespace smith::input` (match the opened namespace).
```
ebchin, line 29: Maybe mention adding a new line at the end of the file?

ebchin (approved):

skills/cpp-style/SKILL.md
white238: good idea. done.

PR #1551, opened 2026-03-10
Adding setTemperatureBC method to HeatTransfer that uses Domain

kswartz92:
I was recently using smith::HeatTransfer and was frustrated by the lack of method to set essential bc's (temperature in this case) on a smith::Domain. I added a method to do so, but since the plan is to move away from the monolithic HeatTransfer class in heat_transfer.hpp in the near future, I did not want to break the interface by removing the old method in case others are still using it. I did switch the thermal statics patch test to the new essential bc method to make sure it works as expected.


src/smith/physics/heat_transfer.hpp
```diff
@@ -308,6 +308,26 @@ class HeatTransfer<order, dim, Parameters<parameter_space...>, std::integer_sequ
...
+  template <typename AppliedTemperatureFunction>
+  void setTemperatureBCs(AppliedTemperatureFunction applied_temperature, Domain& domain)
+  {
+    auto mfem_coefficient_function = ([applied_temperature](const mfem::Vector& X_mfem, double t) {
+      auto X = make_tensor<dim>([&X_mfem](int k) { return X_mfem[k]; });
+      return applied_temperature(X, t);
+    });
+    temp_bdr_coef_ = std::make_shared<mfem::FunctionCoefficient>(mfem_coefficient_function);
+    auto dof_list = domain.dof_list(&temperature_.space());
+    bcs_.addEssential(dof_list, temp_bdr_coef_, temperature_.space(), 0);
```
kswartz92, line 328: I originally tried to write this line without providing the `std::optional<int> component`, but that caused the test to seg fault. If I provide the `0` the test passes, but I don't understand why I need to provide a component for a scalar field. Any ideas @btalamini @tupek2 ?

tupek2: I remember seeing this before.  I think the path where you dont specify the int is just implemented wrong somewhere down in the depths.  We should probably get rid of that option and force the component to always be passed.

src/smith/physics/tests/thermal_statics_patch.cpp
```diff
@@ -264,4 +273,4 @@ int main(int argc, char* argv[])
   ::testing::InitGoogleTest(&argc, argv);
   smith::ApplicationManager applicationManager(argc, argv);
   return RUN_ALL_TESTS();
-}
+}
```
tupek2:
```suggestion
}

```

tupek2 (approved):

white238 (approved):

PR #1552, opened 2026-03-10
Notify instead of error when no elements on rank

sethwatts:
In #1476, we added a `SLIC_ERROR` if the partitioning of the `mfem::ParMesh/smith::Mesh` left zero elements on any rank. However, per recent discussions with the MFEM team, the `mfem::ParMesh` and everything built on top of it (`ParFiniteElementSpaces`, `ParGridFunctions`, ...) should Just Work even when the partitioning results in no elements/dofs being owned by one or more ranks. We have confirmed that this is the case, and also that (at least in common usage), so does Smith.

In some cases, the user may not intend for such a partitioning, and may therefore want to kill the job and adjust the run size or mesh creation. However, in other use cases, particularly when there are meshes of different dimension/refinement on the same MPI communicator, this may be expected and intended. This PR therefore replaces the `SLIC_ERROR` with a `SLIC_INFO` message so that the user is notified about the partitioning, but otherwise code execution keeps moving.

Update: #1549 Removed the `SLIC_ERROR` calls entirely. That resolves the intent of this PR. If the `SLIC_INFO` is more than necessary, I'm happy to close without merging.


white238:
> /style

This functionality was disabled for reasons... you have engaged Chris Style Bot! Commit coming soon!

sethwatts:
I broke my build doing codevelop things and couldn't easily run styling myself. Thank you for jumping on it!

sethwatts:
> > /style
> 
> This functionality was disabled for reasons... you have engaged Chris Style Bot! Commit coming soon!

Shame... was going to try `/deliver latte and perhaps a cranberry biscotti` next 😄

btalamini (approved):
Thanks. I think it's a good idea if users are notified that there are empty ranks.

src/smith/physics/mesh.cpp
```diff
@@ -41,9 +43,39 @@ Mesh::Mesh(mfem::ParMesh&& mesh, const std::string& meshtag) : mesh_tag_(meshtag
...
+    }
+    if (0 < ranksWithNoElements.size()) {
+      ::std::ostringstream msg;
+      msg << "\nAfter refining and distributing mesh, the following " << ranksWithNoElements.size()
+          << " ranks own zero elements:\n[";
+      for (const ::std::size_t& rank : ranksWithNoElements) {
+        msg << "  " << rank;
+      }
+      msg << "]\n";
+      SLIC_INFO(msg.str());
```
white238, line 73: How noisy is this? Once a simulation/timestep/something else?

btalamini: It's in the mesh constructor, so should be once per simulation for the normal use case.

white238 (approved):

PR #1556, opened 2026-03-16
Tupek/system solver

tupek2:
- [x]     Consolidated Time Integrators: Removed specific integrators (e.g., solid_mechanics_time_integrator.hpp) and replaced them with a generic MultiphysicsTimeIntegrator capable of handling coupled systems.
- [x]     Generic System Templates: Replaced specific dynamic/static struct definitions (solid_dynamics_system.hpp, solid_statics_with_internal_vars_system.hpp) with configurable templates (solid_mechanics_system.hpp, solid_mechanics_with_internal_vars_system.hpp) that accept custom time integration rules.
- [x]     Solver Relaxation: Introduced a relaxation_factor into the CoupledSystemSolver to support under-relaxation for staggered multi-physics solves (e.g., x n e w = ω x s o l v e d + ( 1 − ω ) x o l d ).
- [x]     Cycle-Zero Acceleration Solves: Built explicit support into the systems and time integrators for "cycle zero" solves to establish initial accelerations from initial displacements and velocities.
- [x]     System Solver Robustness: Improved SystemSolver convergence logic (using global L2 norm of the un-staggered residual) and replaced brittle string-matching with O ( 1 ) block index mappings.
- [x]     Deprecation of Old Physics Forms: Deleted legacy solid_weak_form.hpp and heat_transfer_weak_form.hpp from the smith/physics directory, routing users entirely through the new differentiable numerics block solvers.
- [x]     Extended Field Store: Added dependent fields and state indexing to the FieldStore to handle states and time derivatives ( VAL , DOT , DDOT ) seamlessly for multiple variables.
- [x]     Test Suite Overhaul: Re-wrote tests (test_thermo_mechanics.cpp, test_solid_dynamics.cpp, test_thermal_static.cpp) to utilize the updated block solvers and newly abstracted simulation builders.


tlroy:
Does `LinearSolverOptions.sub_block_linear_solver_options` allow for nested block solvers? We should test that.

tlroy:
`LinearSolverOptions.sub_block_linear_solver_options` requires a linear solver even if one wants to apply the preconditioner by itself. We should set up a `LinearSolver:None`

examples/inertia_relief/inertia_relief_example.cpp
```diff
@@ -1,3 +1,4 @@
+
```
ebchin, line 1: Remove extra line

src/smith/differentiable_numerics/field_store.hpp
```diff
@@ -0,0 +1,440 @@
...
+   * @param mesh The mesh associated with the fields.
+   * @param storage_size Initial storage size for fields (default: 50).
+   */
+  FieldStore(std::shared_ptr<Mesh> mesh, size_t storage_size = 50);
+
+  /**
+   * @brief Enum for different types of time derivatives.
+   */
+  enum class TimeDerivative
+  {
```
ebchin, line 56: Does it make sense to add an integral time derivative type? I'm thinking of places where the independent field might already be a time derivative.

src/smith/differentiable_numerics/multiphysics_time_integrator.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    if (!mapping.history_name.empty()) {
+      size_t hist_idx = field_store_->getFieldIndex(mapping.history_name);
+      new_states[hist_idx] = u_new;
+    }
+  }
+
+  return {new_states, reactions};
+}
```
ebchin, line 171: Maybe I haven't gotten there yet, but is there something that works with multi-stage time integration schemes like RK4?

src/smith/differentiable_numerics/coupled_system_solver.hpp
```diff
@@ -0,0 +1,79 @@
...
+  CoupledSystemSolver(std::shared_ptr<NonlinearBlockSolverBase> single_solver);
+
+  /// @brief Construct a CoupledSystemSolver for staggered iteration.
+  /// @param max_staggered_iterations Maximum number of staggered sweeps across all stages.  When
+  ///        @p exact_staggered_steps is false, the solver may exit early once all stage solvers
+  ///        report convergence.
+  /// @param exact_staggered_steps If true, always perform exactly @p max_staggered_iterations
+  ///        sweeps with no early-exit convergence check.  Useful when a fixed number of
+  ///        partitioned-stagger steps is required regardless of residual level.
+  CoupledSystemSolver(int max_staggered_iterations, bool exact_staggered_steps = false);
```
ebchin, line 45: It seems like you can set this up to do Uzawa iterations to solve for constraints?

src/smith/differentiable_numerics/nonlinear_solve.cpp
```diff
@@ -491,27 +271,38 @@ std::vector<FieldState> block_solve(const std::vector<WeakForm*>& residual_evals
...
       for (size_t col_j = 0; col_j < num_rows; ++col_j) {
         size_t field_index_to_diff = block_indices[row_i][col_j];
-        tangent_weights[field_index_to_diff] = 1.0;
-        auto jac_ij = residual_evals[row_i]->jacobian(time_info, shape_disp_ptr.get(),
-                                                      getConstFieldPointers(row_field_inputs), tangent_weights);
-        jacobians[row_i].emplace_back(std::move(jac_ij));
-        tangent_weights[field_index_to_diff] = 0.0;
+        if (field_index_to_diff != invalid_block_index) {
+          tangent_weights[field_index_to_diff] = 1.0;
+          auto jac_ij = residual_evals[row_i]->jacobian(time_info, shape_disp_ptr.get(),
```
ebchin, line 276: Should we refactor `WeakForm::jacobian()` to only take the derivative w.r.t. a single field?  If you had a thermomechanical system or a constrained system, the size of the blocks would be different anyway.

ebchin (approved):
Very elegant, very flexible; a lot of good stuff here.

I know the focus here is coupled PDEs, but it seems like a lot of this can be used for solving constraints too. Excited to try it out for contact!

tlroy (approved):

src/smith/differentiable_numerics/multiphysics_time_integrator.cpp
tupek2: not yet

src/smith/differentiable_numerics/field_store.hpp
tupek2: I think we'll have a way to mark that dot(u) is the independent.  but, haven't dont it until we have a real use-case

src/smith/differentiable_numerics/coupled_system_solver.hpp
tupek2: I think so.  Nearly, we need a fixed 1 iteration Newton step type of thing to mimic Uzawa.  I have a demo of this coming up in a future PR.

src/smith/differentiable_numerics/nonlinear_solve.cpp
tupek2: I think we can consider this down the line.  for now, I think its not really a performance overhead, its just a more complex interface.

src/docs/doxygen/CMakeLists.txt
```diff
@@ -4,6 +4,22 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause)
 
+find_package(Doxygen COMPONENTS dot)
```
chapman39, line 7: this shouldnt have been required, doxygen is found via BLT

PR #1558, opened 2026-03-17
ContactAMGF bugfix for no warm start contact

thartland:
This small PR addresses a bug when using `AMGFContact` when `use_warm_start` is set to false. As important calls for setting up AMGFContact was previously in a `if (use_warm_start)` block.

The contact_beam_amgf test now includes tests with and without warm starting. The test code has been reformatted for (in my opinion) easier readability.


ebchin (approved):
Looks good, @thartland

tupek2 (approved):

PR #1559, opened 2026-03-18
Per-contact interaction adjoint support

ebchin:


tupek2 (approved):
Lets get it out there and start trying it!

src/smith/physics/tests/contact_solid_adjoint.cpp
```diff
@@ -298,6 +298,126 @@ TEST_F(ContactSensitivityFixture, ReactionShapeSensitivities)
...
+  FiniteElementDual displacement_adjoint_load(solid_solver->state("displacement").space(),
+                                              "contact_force_qoi_adjoint_load");
+  displacement_adjoint_load = 0.0;
+  J00->MultTranspose(f_base, displacement_adjoint_load);
+
+  // Our shape sensitivity has an explicit term (dJ/d(shape) with u=const) which is only dependent on the contact
+  // interaction force ( = f_contact,i^T * d(f_contact,i)/d(shape) )
+  FiniteElementDual explicit_shape_term(solid_solver->shapeDisplacement().space(), "contact_force_qoi_explicit_shape");
+  explicit_shape_term = 0.0;
+  J00->MultTranspose(f_base, explicit_shape_term);
```
btalamini: I'm sure this part is correct, but it's not clear to me how line 334 follows from the comment. Is the comment trying to say the `J00` factor is used because `d(f_contact,i)/d(shape)` is the equal to `d(f_contact,i)/du`? If so, can you write that in?

btalamini (approved):

src/smith/physics/tests/contact_solid_adjoint.cpp
ebchin: Comment expanded

PR #1560, opened 2026-03-19
Set Tribol displacements after nonlinear solve

ebchin:
The displacement Tribol sees is usually changed after calling `nonlin_solver_->solve()` and this displacement doesn't necessarily match the final displacement at the converged solution (e.g. if the residual is already at 0, TrustRegion will put NaNs in for displacement).  This fix just resets the displacements to the value returned by the nonlinear solver, ensuring the `contact_.update()` call sees the right values.

### DIFF
--- a/src/smith/physics/solid_mechanics_contact.hpp
+++ b/src/smith/physics/solid_mechanics_contact.hpp
@@ -12,6 +12,7 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


tupek2 (approved):

ebchin:
Fixed by #1561 and #1567

PR #1563, opened 2026-03-19
Fix mac build warnings

tupek2:
so there is not a wall of un-fixable, uninteresting war…nings.


chapman39 (approved):
the comments are a bit verbose but besides that lgtm. this is an issue that bothers me too so thanks.

ebchin (approved):
Thanks for figuring this out!

white238:
We should look into figuring out if we should fix this upstream in Spack. @becker33

PR #1566, opened 2026-03-19
Macnider/interior faces subdomain

bmacnider:
Adding methods and test to create a domain out of a subset of interior faces.


src/smith/physics/mesh.cpp
lihanyu97, line 145: We can implement 2D and 3D `addDomainOfInternalBoundaryElements` for the mesh class here

src/smith/numerics/functional/domain.cpp
```diff
@@ -660,6 +660,80 @@ Domain InteriorFaces(const mesh_t& mesh)
...
+
+    auto geom = mesh.GetFaceGeometry(f);
+
+    mfem::Array<int> vertex_ids;
+    mesh.GetFaceVertices(f, vertex_ids);
+
+    auto x = gather<d>(vertices, vertex_ids);
+
+    int bdr_id = face_id_to_bdr_id[f];
+    int attr = (bdr_id >= 0) ? mesh.GetBdrAttribute(bdr_id) : -1;
```
lihanyu97, line 692: If face `f` is an interior face, then `face_id_to_bdr_id[f]` will return -1, which will lead to attr = -1. I am not sure if this implementation will always return the correct attribute. In 3D, you can easily point to a face and get its attribute like this
```suggestion
    int attr = mesh.GetFace(f)->GetAttribute();
```
I've trying to find a way to point to an edge element for 2D scenarios but I could not find any. Maybe @btalamini and @tupek2 knows a way.

lihanyu97: I guess your original implementation will work. side sets in cubit are considered boundaries, even if the side set is defined on a set of interior faces. When loaded in MFEM, the mesh still consider them to be boundaries and assign them a boundary attribute. But the faces on the side set has `IsInterior() == true`.

lihanyu97 (approved):

tupek2 (approved):

PR #1567, opened 2026-03-20
Refactor `ContactData::update()`

ebchin:
- Refactors the `ContactData::update()` call to send in updated fields and only call the Tribol API functions needed to update Tribol for the specific updated fields
- Adds a `ContactData::updateGaps()` method that just updates the gaps and, optionally, the Jacobian
- Remove unnecessary calls to update()


tupek2 (approved):

thartland (approved):
Looks great!

PR #1569, opened 2026-03-20
C++20

white238:
Changes the default and now requires C++20

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,6 +82,8 @@ endif()
 # Tune BLT to our needs
 if (NOT BLT_CXX_STD)
     set(BLT_CXX_STD "c++20" CACHE STRING "")
+elseif (BLT_CXX_STD MATCHES "^c\\+\\+([0-9]+)$" AND CMAKE_MATCH_1 VERSION_LESS 20)
+    message(FATAL_ERROR "Smith requires BLT_CXX_STD to be c++20 or higher.")
 endif()
 
 # These BLT tools are not used in Smith, turn them off
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,7 +81,7 @@ endif()
 
 # Tune BLT to our needs
 if (NOT BLT_CXX_STD)
-    set(BLT_CXX_STD "c++17" CACHE STRING "")
+    set(BLT_CXX_STD "c++20" CACHE STRING "")
 endif()
 
 # These BLT tools are not used in Smith, turn them off
--- a/cmake/SmithCompilerFlags.cmake
+++ b/cmake/SmithCompilerFlags.cmake
@@ -27,8 +27,9 @@ if(DEFINED ENV{SYS_TYPE} AND "$ENV{SYS_TYPE}" STREQUAL "blueos_3_ppc64le_ib_p9")
 endif()
 
 # Enable warnings for things not covered by -Wall -Wextra
-set(_extra_flags "-Wshadow -Wdouble-promotion -Wconversion -Wundef -Wnull-dereference -Wold-style-cast -Wpedantic")
+set(_extra_flags "-Wshadow -Wdouble-promotion -Wconversion -Wundef -Wnull-dereference -Wold-style-cast")
 blt_append_custom_compiler_flag(FLAGS_VAR CMAKE_CXX_FLAGS DEFAULT ${_extra_flags})
 
 # Clang specific warnings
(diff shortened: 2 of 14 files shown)


cmake/SmithCompilerFlags.cmake
```diff
@@ -14,6 +14,29 @@ endif()
 # Need to add symbols to dynamic symtab in order to be visible from stacktraces
 string(APPEND CMAKE_EXE_LINKER_FLAGS " -rdynamic")
 
+# ROCm's amdclang++ can drop its implicit C++ standard library on HIP links
+# once Fortran runtime libraries are introduced. On this platform the shared
+# libstdc++ linker script still points at the old system runtime, so use the
+# compiler's GCC 13 static archives instead.
+if(ENABLE_CODEVELOP AND
```
white238:
```suggestion
if(SMITH_ENABLE_CODEVELOP AND 
```

cmake/SmithCompilerFlags.cmake
```diff
@@ -14,6 +14,29 @@ endif()
 # Need to add symbols to dynamic symtab in order to be visible from stacktraces
 string(APPEND CMAKE_EXE_LINKER_FLAGS " -rdynamic")
 
+# ROCm's amdclang++ can drop its implicit C++ standard library on HIP links
+# once Fortran runtime libraries are introduced. On this platform the shared
+# libstdc++ linker script still points at the old system runtime, so use the
+# compiler's GCC 13 static archives instead.
+if(SMITH_ENABLE_CODEVELOP AND
```
white238, line 21: I can't figure out why this is happening. I guarded it to the only case that it showed up in. I even asked around and noone else has hit this.

tupek2: Just so I understand... this doesn't effect our C++20 transition, its just that on this platform, some libraries have to be gcc compiled?  Or is this something else?

tupek2 (approved):

cmake/SmithCompilerFlags.cmake
white238: This only affects the codevelop/amdclang build that has amdflang. It does not stop us from going forward to C++20. Without this it is bringing in the system libstdc++ and ending in the standard c++ error that happens in when you mixing an old c++ std library, usually something about basicstring being messed up.

src/smith/physics/tests/thermomech_statics_patch.cpp
```diff
@@ -218,39 +218,39 @@ class ManufacturedSolution {
     tm.setDisplacementBCs(disp_ebc_func, disp_ess_bdr);
 
     // natural BCs
-    auto flux = [=](auto X, auto n0, auto /* time */, auto /* T */) {
+    auto flux = [this, &material](auto X, auto n0, auto /* time */, auto /* T */) {
```
chapman39, line 221: is `this` keyword more commonly used in c++20?

CMakeLists.txt
```diff
@@ -81,7 +81,9 @@ endif()
 
 # Tune BLT to our needs
 if (NOT BLT_CXX_STD)
-    set(BLT_CXX_STD "c++17" CACHE STRING "")
+    set(BLT_CXX_STD "c++20" CACHE STRING "")
+elseif (BLT_CXX_STD MATCHES "^c\\+\\+([0-9]+)$" AND CMAKE_MATCH_1 VERSION_LESS 20)
+    message(FATAL_ERROR "Smith requires BLT_CXX_STD to be c++20 or higher.")
```
chapman39, line 86: is there a mechanism in cmake/ blt that sets minimum c++ version? maybe we could add this to blt. i can imagine other codes wanting this.

chapman39 (approved):
just had minor clarification questions but overall lgtm. maybe we should create an issue to track the weird codevelop/amdclang issue. maybe it would be resolved once we move to rocm 7+

https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/packages/llvm_amdgpu/package.py#L71

src/smith/physics/tests/thermomech_statics_patch.cpp
white238: `this` refers to the current instance of the class you are in so you can gain access to its members. It has always been there but C++20 deprecated the implicit capturing of `this` in lambdas.

CMakeLists.txt
white238: https://github.com/llnl/blt/issues/758

cmake/SmithCompilerFlags.cmake
white238: #1575

PR #1570, opened 2026-03-23
Li/periodic geometric factor

lihanyu97:
This PR is to fix issues when integrating on interior faces of periodic meshes.

mfem uses DG space to store nodal coordinates of periodic meshes because faces on periodic boundaries have two sets of nodal coordinates. While this does not cause any problems for H1-based physics solvers, it produces inconsistent facet normal for DG-based physics solvers during interior face integral. The main issue comes from geometric factor computation of these faces. Since nodal coordinate is in DG space, the element restriction returns dofs on both side of the face while interpolation of value and shape gradient assumes H1 nodal field. The fix is to collapse two sets of nodal coordinates on both sides of the face into a single set.


tupek2 (commented):
When I run the tests, I see a lot of output, e.g.:
[1] One side = {0.333333, 0.291667, 0}, The other side = {0.333333, 0.291667, 1}, Jump = {0, 0, -1}
[2] One side = {0.0416667, 0, 0.0833333}, The other side = {0.0416667, 1, 0.0833333}, Jump = {0, -1, 0}
[3] One side = {0.541667, 0, 0.0833333}, The other side = {0.541667, 1, 0.0833333}, Jump = {0, -1, 0}
[1] One side = {0.458333, 0.291667, 0}, The other side = {0.458333, 0.291667, 1}, Jump = {0, 0, -1}
[0] One side = {0.0833333, 0.0416667, 0}, The other side = {0.0833333, 0.0416667, 1}, Jump = {0, 0, -1}

From codex review:
 1. Missing 2D periodic coverage. The new API adds ofInteriorFaces for both 2D and 3D, and the geometric-factor special case also applies to 1D face geometry, but the new test only covers 3D tets/hexes. See domain.hpp:169, geometric_factors.cpp:83, dg_periodic_bc_index.cpp:94.
  2. Missing higher-order periodic DG coverage. The new coordinate path rewrites interior-face node data by halving nodes_per_elem, which is exactly the kind of indexing logic that can break at p>1, but the test only exercises p=1. See geometric_factors.cpp:86, dg_periodic_bc_index.cpp:25, dg_periodic_bc_index.cpp:94.

tupek2 (approved):

PR #1577, opened 2026-03-26
Roy/block precon custom operators

tlroy:
Can pass custom operators to be used in subblock preconditioners. For e.g., to set custom Schur complement approximations.


src/smith/differentiable_numerics/tests/test_porous_heat_sink.cpp
```diff
@@ -39,7 +39,9 @@ enum class BlockPrecondType
   SchurLower,
   SchurUpper,
   SchurDiag,
-  SchurFull
+  SchurFull,
+  SchurFullA22,
```
tupek2, line 43: Is it possible to give a short summary of the mathematical differences between these different Schur versions.  Maybe just some comments here (or whereever seems most appropriate.

tlroy: You want more description in that test or in block_preconditioner.hpp? The mathematical description is there in the hpp file, but I can add something about when appropriate.

tupek2: I see that now.  reading in order, not finished.  Are these enums useful outside of this test too?  I think its not entirely clear to me yet how users are going to interact with these options.  But we should meet to discuss.

tlroy: These enums were simply to loop over different solver options in the test. You did add `Preconditioner` enums in your own PR for the block preconditioning that existed at that point. We can discuss how to make that sustainable, but I guess that's also a `solver_config` problem.

tupek2 (approved):

lihanyu97 (approved):

PR #1580, opened 2026-04-08
Swap to std::format

white238:
The update to C++20 now allows us to use the format that comes in the standard library

### DIFF
--- a/src/smith/infrastructure/format.hpp
+++ b/src/smith/infrastructure/format.hpp
@@ -1,0 +1,44 @@
+// Copyright (c) Lawrence Livermore National Security, LLC and
+// other Smith Project Developers. See the top-level LICENSE file for
+// details.
+//
+// SPDX-License-Identifier: (BSD-3-Clause)
+
+/**
+ * @file format.hpp
+ *
+ * @brief Shared helpers for C++ standard formatting support in Smith
+ */
+
+#pragma once
+
+#include <format>
+#include <sstream>
+#include <string>
+#include <string_view>
+
+namespace smith::format {
+
+/// @brief Converts a value to a string using its stream insertion operator.
+template <typename T>
+std::string streamed(const T& value)
+{
+  std::ostringstream stream;
+  stream << value;
+  return stream.str();
+}
+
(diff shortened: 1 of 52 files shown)


src/smith/infrastructure/format.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <format>
+#include <sstream>
+#include <string>
+#include <string_view>
+
+namespace smith::format {
+
+/// @brief Converts a value to a string using its stream insertion operator.
+template <typename T>
+std::string streamed(const T& value)
```
chapman39, line 24: why not [to_string()](https://en.cppreference.com/w/cpp/string/basic_string/to_string)?

chapman39: nvm that might not work for the templated function below

src/smith/infrastructure/format.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+  std::ostringstream stream;
+  stream << value;
+  return stream.str();
+}
+
+/// @brief Formatter that renders values through `operator<<`.
+struct OstreamFormatter {
```
chapman39, line 32: so was this originally in Axom and we're moving it to smith?

chapman39:
should we remove `SMITH_USE_FMT`? or do we use fmt in other ways

src/smith/infrastructure/format.hpp
white238: This specifically is just a helper function that is used to create the formatter structs. It's a pattern used in `std::format` and `fmt::format`.

src/smith/infrastructure/format.hpp
white238: This is just reducing the boiler plate needed for the formatter classes. To my understanding, `std::format` is not a full replacement for the full `fmt` library. `fmt` had fallback cases if you had `operator<<` so you didn't need to create formatter classes for every struct and class.

chapman39 (approved):

tupek2 (approved):

PR #1585, opened 2026-04-16
Make some of the tensor metadata easier to access for clients

tepperly:
Make the outer size and type of stored data available in the class.

The `static constexpr int outer_size()` method returns the number of elements in `data`.

The `static constexpr int order()` method returns the order of the tensor.

The `inner_type` type alias is the type of `data`.


tepperly:
According to wikipedia, a `tensor<double,1>` has order zero, and a `tensor<double,2>` has order one. A `tensor<double, 3, 3>` has an order of 2. Should a `tensor<double, 2, 1>` have an order of 2 or 1? Right now, the code would say 2.

tepperly:
The main thing I am looking for in this PR is the ability to add some assertions in LiDO to ensure that we don't accidentally reference data outside the scope of the tensor (i.e., use operator[] to access data that is beyond the end of the tensor).

src/smith/numerics/functional/tensor.hpp
```diff
@@ -51,10 +51,26 @@ struct tensor<T, m, n...> {
     return data[i](jklm...);
   }
 
+  /// return the number of elements in data
+  SMITH_HOST_DEVICE constexpr static int outer_size() { return m; }
```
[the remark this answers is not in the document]
tepperly: I kind of waffled back and forth about what style guidelines I should follow. Should I follow the tensor style or the Smith style. Initially, I tried following Smith guidelines and then I switched to tensor's style which seems to be snake case.

src/smith/numerics/functional/tensor.hpp
```diff
@@ -51,10 +51,26 @@ struct tensor<T, m, n...> {
...
+
   SMITH_HOST_DEVICE constexpr auto& operator[](int i) { return data[i]; }
   SMITH_HOST_DEVICE constexpr const auto& operator[](int i) const { return data[i]; }
 
-  tensor<T, n...> data[m];
+  /// the type of data
+  using inner_type = tensor<T, n...>;
+
+  /// return the order of the tensor
+  SMITH_HOST_DEVICE constexpr static int order()
```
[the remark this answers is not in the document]
tepperly: I waffled between Smith style and tensor style and ended up following tensor-style. The PR comments lagged.

src/smith/numerics/functional/tensor.hpp
```diff
@@ -84,7 +100,23 @@ struct tensor<T, m> {
...
+  {
+    if constexpr (m == 1) {
+      return 0;
+    } else {
+      return 1;
+    }
+  }
```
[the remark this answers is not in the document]
tepperly: I added a link to the Wikipedia section where order of a tensor is discussed. There are other parts of the tensor class that are specialized for tensor<T,1>, and I believe my addition is consistent with those other specializations.

src/smith/numerics/functional/tests/tensor_unit_tests.cpp
```diff
@@ -35,7 +35,16 @@ TEST(Tensor, BasicOperations)
   tensor<double, 3> u = {1, 2, 3};
   tensor<double, 4> v = {4, 5, 6, 7};
 
+  static_assert(tensor<double, 3>::outer_size() == 3, "Outer size is wrong");
```
white238, line 38: Why not check against the correct value?

tepperly: 3 is the correct value.

tepperly: The message is printed if the test fails.

white238: im dumb i get it.

white238 (approved):

btalamini:
Sam took pains to keep this class as lean as possible. There's both a compile-time burden and maintenance burden to expanding a class template, especially one so heavily used throughout our code. For that reason, I prefer the size function to be a free function. I pushed a draft revision that does that. (I also renamed it, we can discuss what name would be clearest).

I think `order` and `inner_type` are less compelling. For `order`, the degenerate behavior of the scalar case is likely to trip people up. In my mind, there isn't a complling use case for this, so I'd rather avoid expanding the class, at least for now.

I think `inner_type` is available already by `decltype(A[0])` for a `tensor` object `A`.

src/smith/numerics/functional/tensor.hpp
```diff
@@ -84,7 +87,10 @@ struct tensor<T, m> {
     return data[0];
   }
 
-  T data[m];
+  /// the type of data
+  using inner_type = T;
```
tupek2: this fine. but remember we are likely to switch to mfem::tensor at some point soon?  So, anything we add to our interface here that they don't support will cause more churn in the future.

tepperly: `mfem::tensor` has `first_dim` which is like the `outer_size` and `ndim` that's order without the special case.

Perhaps, this PR should be canceled, and I'll get what I want when `mfem::tensor` comes.

src/smith/numerics/functional/tensor.hpp
```diff
@@ -259,6 +265,13 @@ SMITH_HOST_DEVICE constexpr zero dot(zero, const T&)
   return zero{};
 }
 
+/** @brief return the size of the leftmost tensor dimension */
+template <typename T, int m, int... n>
+SMITH_HOST_DEVICE constexpr int outer_size(tensor<T, m, n...>)
+{
+  return m;
+}
```
tepperly, line 267: When I used a free function like this in LiDO, I got error messages shared with @btalamini by email. For reasons, I don't fully understand, I had to write it like this:
```c++
template<typename ValueType, int N, int... remainder>
constexpr int tensorOuterSize(const ::smith::tensor<ValueType, N, remainder...> &)
{
  return N;
}
```
The change from pass by value to pass by reference makes the compiler issues go away.

btalamini: I copied this pattern, and test it with a static assertion.

btalamini:
@white238 and @tupek2 , I modified this. Would you review again?
Regarding the MFEM compatibility: I named the function `first_dim` which matches the MFEM name. They put `first_dim` on the struct itself, while I prefer to keep the struct as simple as possible and put this in a free function. We'll have to settle these eventually, but at least the name is the same.

tupek2 (approved):

PR #1588, opened 2026-04-21
fix `ld: library 'mpifort' not found` on macmini codevelop

chapman39:
i manually add -lmpifort to the mfem target to fix a toss4_cray issue i was having. turns out, this is an issue when building with macmini, because openmpi does not have a libmpifort.so

the following pr checks the systype before adding this mpifort lib manually, so it only happens on toss4 cray systems.


tupek2 (approved):
I was having this issue too.  but did not figure it out!  thanks.

white238 (approved):

btalamini:
This was hitting me, too. Thank you!

PR #1589, opened 2026-04-21
Tupek/system solver extended

tupek2:


white238:
https://llnlsmith.readthedocs.io/en/tupek-system_solver_extended/

white238:
There are quite a few `std::cout`` calls.. should we use SLIC and std::format for consistency and utilizing any changes we do to the logger globally? does it make it more readable?

white238:
On the examples since they are user facing they should have good comments explaining what is going on.

PR #1590, opened 2026-04-22
Fix MacOS "blas not found" when codevelop + asan

chapman39:
* enforce blas provider to be openblas in spack, when building on macos. this will add openblas to the cmake prefix path, so it can be found and used
* on mac, when codeveloping + asan enabled, temporarily disable c, cxx flags to avoid issues in the find blas cmake module.

fixes the following issue

```
-- Looking for Fortran sgemm
-- Looking for Fortran sgemm - not found
CMake Error at /Applications/CMake.app/Contents/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:233 
(message): Could NOT find BLAS (missing: BLAS_LIBRARIES)
Applications/CMake.app/Contents/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE) 
Applications/CMake.app/Contents/share/cmake-3.30/Modules/FindBLAS.cmake:1417 (find_package_handle_standard_args) 
cmake/thirdparty/SetupSmithThirdParty.cmake:347 (find_package) CMakeLists.txt:174 (include)
```

if you experience this issue, you shouldn't need to rebuild tpls. the apple acceleration library will used instead of openblas, which is probably fine.


ebchin (approved):

btalamini:
This was biting me. I am giving this a shot on my system.

white238 (approved):

btalamini:
I tried this on my system. I turned on codevelop and asan in a debug build. The build configure hangs after hypre is found.

cmake/thirdparty/SetupSmithThirdParty.cmake
```diff
@@ -338,6 +338,17 @@ if (NOT SMITH_THIRD_PARTY_LIBRARIES_FOUND)
             set(MFEM_USE_UMPIRE OFF CACHE BOOL "")
         endif()
 
+        # Temporarily unset asan flags when finding blas, in order to avoid conflicts with fortran and asan.
```
chapman39:
```suggestion
        # Temporarily unset compiler and linker flags when configuring MFEM, in order to avoid conflicts with fortran and asan.
```

PR #1595, opened 2026-04-30
type alias mfem::future::tuple to smith::tuple

white238:
This removes a large portion of the implementation in Smith. Originally I was going to swap completely over to the `mfem::tuple` but it exists in a `future` namespace and it would have bled over all codebase. We might want to do that when it gets out of the `future` namespace and when we don't have large PRs up.

### DIFF
--- a/src/smith/numerics/functional/tuple.hpp
+++ b/src/smith/numerics/functional/tuple.hpp
@@ -7,202 +7,21 @@
 /**
  * @file tuple.hpp
  *
- * @brief Implements a std::tuple-like object that works in CUDA kernels
+ * @brief Smith tuple compatibility layer over mfem::future::tuple
  */
 #pragma once
 
-#include <utility>
+#include "mfem/fem/dfem/tuple.hpp"
 
 #include "smith/infrastructure/accelerator.hpp"
 
-namespace smith {
+namespace mfem::future {
 
 /**
- * @tparam T the types stored in the tuple
- * @brief This is a class that mimics most of std::tuple's interface,
- * except that it is usable in CUDA kernels and admits some arithmetic operator overloads.
+ * @brief Smith compatibility extension for the MFEM tuple implementation.
  *
- * see https://en.cppreference.com/w/cpp/utility/tuple for more information about std::tuple
- */
-template <typename... T>
-struct tuple {};
-
-/**
- * @brief Type that mimics std::tuple
- *
- * @tparam T0 The first type stored in the tuple
- */
-template <typename T0>
-struct tuple<T0> {
-  T0 v0;  ///< The first member of the tuple
-};
-
-/**
- * @brief Type that mimics std::tuple
- *
- * @tparam T0 The first type stored in the tuple
- * @tparam T1 The second type stored in the tuple
- */
-template <typename T0, typename T1>
-struct tuple<T0, T1> {
-  T0 v0;  ///< The first member of the tuple
-  T1 v1;  ///< The second member of the tuple
-};
-
-/**
- * @brief Type that mimics std::tuple
- *
- * @tparam T0 The first type stored in the tuple
- * @tparam T1 The second type stored in the tuple
- * @tparam T2 The third type stored in the tuple
- */
-template <typename T0, typename T1, typename T2>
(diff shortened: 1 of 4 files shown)


tupek2 (approved):
Strategy looks good to me!

src/smith/numerics/functional/tuple.hpp
```diff
@@ -7,202 +7,21 @@
...
- * @tparam T2 The third type stored in the tuple
- * @tparam T3 The fourth type stored in the tuple
- * @tparam T4 The fifth type stored in the tuple
- * @tparam T5 The sixth type stored in the tuple
- * @tparam T6 The seventh type stored in the tuple
- * @tparam T7 The eighth type stored in the tuple
- * @tparam T8 The ninth type stored in the tuple
- * @tparam T9 The tenth type stored in the tuple
+ * MFEM's tuple copy currently stores up to nine elements, while Smith's
+ * historical tuple API supports ten and eleven elements.
```
chapman39, line 24: should we try to add those extra two elements to the mfem tuple implementation at some point?

chapman39 (approved):
seems like this simplifies a lot of things and reduces duplication between mfem's implementation

src/smith/numerics/functional/tuple.hpp
white238: yep

src/smith/numerics/functional/tuple_tensor_dual_functions.hpp
```diff
@@ -14,6 +14,12 @@
 
 namespace smith {
 
+template <typename... T>
+SMITH_HOST_DEVICE auto get_value(const smith::tuple<T...>& tuple_of_values);
+
+template <typename... T>
+SMITH_HOST_DEVICE auto get_gradient(smith::tuple<T...> tuple_of_values);
```
btalamini: I'm curious, why are these declarations needed?

src/smith/numerics/functional/boundary_integral_kernels.hpp
```diff
@@ -105,7 +105,7 @@ template <int i, int dim, typename... trials, typename lambda>
 auto get_derivative_type(lambda qf)
 {
   using qf_arguments = smith::tuple<typename QFunctionArgument<trials, smith::Dimension<dim>>::type...>;
-  return tuple{get_gradient(apply_qf(qf, double{}, tensor<double, dim + 1>{}, make_dual_wrt<i>(qf_arguments{}))),
+  return tuple{smith::get_gradient(apply_qf(qf, double{}, tensor<double, dim + 1>{}, make_dual_wrt<i>(qf_arguments{}))),
```
btalamini: I see this fully qualified name as a pattern in this PR: for my knowledge, was this necessary? It can't hurt, but I'm curious if the mfem version was being found first without the explicit `smith` namespace scope?

btalamini (approved):

src/smith/numerics/functional/tuple_tensor_dual_functions.hpp
white238: That is a good point. It is putting them into the smith namespace but they may not be used anywhere else.  Let me try it w/o it when I get some time. This week is nuts.

src/smith/numerics/functional/boundary_integral_kernels.hpp
white238: Without it there were ambigious compiler errors but possibly if it is removed the get_gradient and get_value it wouldn't be. I'll at the same time to revert these.

src/smith/numerics/functional/tuple_tensor_dual_functions.hpp
white238: I should have spent more time looking into this. This was to fix an ambigious compiler error due to ordering of functions and now conflicting with generic mfem's functions (i think). I found a better and more minimal change.

white238:
Some of these changes can be removed after we do a bulk change to mfem's tuple, after either we move the 10,11 elements to them or to the more generic tuple I'm told might come. But this is the most minimal set of changes I could come up with w/o bleeding everywhere through the code.

PR #1602, opened 2026-05-11
Shared mesh in StateManager

### LINKED ISSUE #1443: Make serac state manager take a shared_ptr to the mesh instead of a unique pointer
```c++
   ::mfem::ParMesh &setMesh(::std::shared_ptr<mfem::ParMesh> mesh, const ::std::string& mesh_tag);
```

white238:
Fixes #1443 

* Adds a `shared_ptr` overload to `setMesh()`
* Converts the given `unique_ptr` mesh to a `shared_ptr`
* Adds two new lists to handle the memory lifecycle of duals and states created by the `StateManager`
* Clarified a lot of comments about ownership since `StateManager` handles most of our unique memory ownership issues like restart/not, shared/owned mesh


white238:
@SchmidtMat

src/docs/sphinx/dev_guide/state_manager.rst
```diff
@@ -102,9 +102,12 @@ the tag identifying the mesh.  As in the nominal case, this tag is not necessary
   in the previous simulation run.
 
 ``StateManager::load()`` will reconstruct the ``mfem::ParMesh``, ``mfem::GridFunction``, and ``mfem::QuadratureFunction``
-objects.  The ``StateManager`` factory methods can be used in the exact same way as they would in a nominal run, though
-the internal logic is of course different.  In particular, it will search through the restored data for a field with the
-requested name and use that instead of constructing a new field via the process described above.
+objects. In a restart workflow, the reconstructed mesh is owned by the underlying ``MFEMSidreDataCollection``.
```
tupek2, line 105: I think the only real 'debate' we could consider.  Does it make sense to handle this complexity of having the Sidre own the mesh only when its a restart.  Or is it somehow easier to just require that on restart the user is requires to point to the same mesh again.  If the extra code is minimal to handle this complexity of changing what owns the mesh depending on restart vs. an initial simulation, then I think this is OK.  If the code is a lot simpler to just say that the mesh is ALWAYs a shared_ptr, then I'd be happy to force users to just point out the mesh file again.  No real strong opinion, its just what came to mind for me.  @btalamini  thoughts?

white238: This has always been a complicated section, memory lifecycle wise. We could also add functionality to `MFEMSidreDataCollection` allow transfer of ownership back to us.

At some point we should revisit restarts as a whole, I am not entirely sure they work at this point, but the concept is what the other customer uses.

white238: Either way i feel this is out of the scope of this PR and could be a follow up.

src/smith/physics/state/state_manager.cpp
```diff
@@ -20,13 +20,16 @@ namespace smith {
...
+std::unordered_map<std::string, std::shared_ptr<mfem::ParMesh>> StateManager::meshes_;
 std::unordered_map<std::string, std::unique_ptr<FiniteElementState>> StateManager::shape_displacements_;
 std::unordered_map<std::string, std::unique_ptr<FiniteElementDual>> StateManager::shape_displacement_duals_;
 bool StateManager::is_restart_ = false;
 axom::sidre::DataStore* StateManager::ds_ = nullptr;
 std::string StateManager::output_dir_ = "";
 std::unordered_map<std::string, mfem::ParGridFunction*> StateManager::named_states_;
 std::unordered_map<std::string, mfem::ParGridFunction*> StateManager::named_duals_;
+std::vector<std::unique_ptr<mfem::ParGridFunction>> StateManager::owned_states_;
+std::vector<std::unique_ptr<mfem::ParGridFunction>> StateManager::owned_duals_;
```
chapman39, line 32: did we figure out why these needed to be added? i remember there was some question over this

chapman39 (approved):

src/smith/physics/state/state_manager.cpp
white238: I figured it out... basically now that the data collection is no longer owning data (needed for the shared mesh), we need to keep the unique ptrs to hold onto the states and duals that are created by us (always except during a restart).

src/docs/sphinx/dev_guide/state_manager.rst
tupek2: I good with how it is, and revisiting once we revive restart more.

tupek2 (approved):

PR #1609, opened 2026-05-20
Tupek/system solver extended

tupek2:
This introduces a new interface for constructing multiphysics systems of equations, their boundary conditions, time integration rules, and solving them using potentially customized staggered or coupled algorithms.


src/smith/differentiable_numerics/time_integration_rule.hpp
```diff
@@ -114,13 +114,15 @@ class QuasiStaticRule : public TimeIntegrationRule {
 
   /// @brief evaluate value of the ode state as used by the integration rule
   template <typename T1>
+  /// @brief Return the static field value.
```
btalamini, line 116: Looks like the agent added a second set of doxygen comments, (one before the template, the other after. Delete one set. The new set seems more specific then the old, maybe delete the old ones.

src/smith/physics/functional_weak_form.hpp
```diff
@@ -99,50 +98,60 @@ class FunctionalWeakForm<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
...
   {
-    weak_form_->AddDomainIntegral(Dimension<spatial_dim>{}, DependsOn<active_parameters...>{}, integrand,
-                                  mesh_->domain(body_name));
+    const double* dt = &dt_;
+    const size_t* cycle = &cycle_;
+    const TimeInfo::EvaluationMode* mode = &mode_;
+    weak_form_->AddDomainIntegral(
+        Dimension<spatial_dim>{}, DependsOn<all_params...>{},
+        [dt, cycle, mode, integrand](double time, auto X, auto... inputs) {
+          return integrand(TimeInfo(time, *dt, *cycle, *mode), X, inputs...);
```
lihanyu97: I think reconstructing time_info here will also result in double adding dt, as we observed in `time_discretized_weak_form`. Since we are already storing `dt_`, `cycle_`, and `mode_`, can we store `time_info_ = time_info` when we invoke `residual()` and `jacobian()`?

Then the lambda injected for these add integral functions would be
```
[time_info, integrand] (double /* time */, auto X, auto... inputs) {
  return integrand(time_info, X, inputs...);
}
```

src/smith/differentiable_numerics/coupling_params.hpp
```diff
@@ -0,0 +1,387 @@
...
+          return applyCouplingTimeRules(
+              coupling, t_info,
+              [&](auto... interpolated_coupling) {
+                return std::forward<Callback>(callback)(std::get<Si>(all)..., interpolated_coupling...);
+              },
+              std::get<n_self + Ti>(all)...);
+        }(std::make_index_sequence<n_self>{}, std::make_index_sequence<tail_count>{});
+      },
+      raw_args...);
+}
```
lihanyu97: The nested return here is not very human-readable. Can we simplify this method?

src/smith/differentiable_numerics/dirichlet_boundary_conditions.hpp
```diff
@@ -158,12 +160,24 @@ class DirichletBoundaryConditions {
...
-  /// @brief Return the smith BoundaryConditionManager
+  /// @brief Return the value-level smith BoundaryConditionManager.
   const smith::BoundaryConditionManager& getBoundaryConditionManager() const { return bcs_; }
 
+  /// @brief Return the BC manager whose prescribed values are the second time derivative of the
+  ///        value-level BC, computed by one-sided three-point forward FD with step
+  ///        @c h = 1e-4 * max(1,|t|).
+  ///
+  /// Rebuilt on each call from current value-level essentials, so late additions are reflected.
+  const smith::BoundaryConditionManager& getSecondDerivativeManager() const;
```
btalamini, line 173: What is the purpose of this? It doesn't seem to be used. Is this to enable acceleration BCs?

src/smith/numerics/solver_config.hpp
```diff
@@ -103,12 +103,13 @@ struct TimesteppingOptions {
...
-  Strumpack, /**< Strumpack MPI-enabled direct frontal solver*/
-  PetscCG,   /**< PETSc MPI-enabled conjugate gradient solver */
-  PetscGMRES /**< PETSc MPI-enabled generalize minimal residual solver */
+  CG,         /**< Conjugate gradient */
+  GMRES,      /**< Generalized minimal residual method */
+  SuperLU,    /**< SuperLU MPI-enabled direct nodal solver */
+  Strumpack,  /**< Strumpack MPI-enabled direct frontal solver*/
+  PetscCG,    /**< PETSc MPI-enabled conjugate gradient solver */
+  PetscGMRES, /**< PETSc MPI-enabled generalize minimal residual solver */
+  None        /**< Preconditioner application only, No linear solver Krylov iterations */
```
btalamini: Maybe the intent would be clearer to call this one `PrecondOnly` or something like that?

examples/solid_mechanics/composable_solid_mechanics.cpp
```diff
@@ -0,0 +1,205 @@
...
+      0.0 * smith::evaluateObjective(qoi, physics->getShapeDispFieldState(), qoi_fields,
+                                     smith::TimeInfo(physics->time(), dt, static_cast<size_t>(physics->cycle())));
+  for (int step = 0; step < num_steps; ++step) {
+    physics->advanceTimestep(dt);
+    qoi_state = qoi_state +
+                smith::evaluateObjective(qoi, physics->getShapeDispFieldState(), current_qoi_fields(),
+                                         smith::TimeInfo(physics->time(), dt, static_cast<size_t>(physics->cycle())));
+  }
+
+  std::cout << "reaction norm: " << physics->getReactionStates().front().get()->Norml2() << '\n';
```
white238: I'd prefer SLIC_INFO + std::format, easier to read imo.

examples/thermo_mechanics/composable_thermo_mechanics_advanced.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include "smith/differentiable_numerics/evaluate_objective.hpp"
+#include "smith/differentiable_numerics/differentiable_test_utils.hpp"
+#include "smith/differentiable_numerics/make_time_info_material.hpp"
+#include "smith/physics/materials/green_saint_venant_thermoelastic.hpp"
+// _includes_end
+
+int main(int argc, char* argv[])
+{
+  // _init_start
+  smith::ApplicationManager application_manager(argc, argv);
```
white238, line 41: These examples are user facing so should have some inline comments

src/smith/differentiable_numerics/tests/test_combined_thermo_mechanics.cpp
```diff
@@ -0,0 +1,524 @@
...
+#include "smith/differentiable_numerics/thermal_system.hpp"
+#include "smith/differentiable_numerics/thermo_mechanics_system.hpp"
+#include "smith/differentiable_numerics/combined_system.hpp"
+#include "smith/differentiable_numerics/multiphysics_time_integrator.hpp"
+#include "smith/differentiable_numerics/differentiable_test_utils.hpp"
+#include "smith/differentiable_numerics/evaluate_objective.hpp"
+#include "smith/differentiable_numerics/nonlinear_solve.hpp"
+#include "smith/differentiable_numerics/make_time_info_material.hpp"
+#include "smith/physics/functional_objective.hpp"
+#include "gretl/wang_checkpoint_strategy.hpp"
```
white238, line 31: I think this is a larger problem than this PR but we have gretl as an optional TPL guarded by the CMake option `SMITH_ENABLE_GRETL`. This will fail if someone turns that off, which I don't think anyone has tried.

btalamini (approved):
Creating multi-physics is very nice with this!

src/smith/differentiable_numerics/tests/test_combined_thermo_mechanics.cpp
btalamini: That's a good point. Gretl really isn't optional, core functionality (differentiability) depends on it. Unless I'm mistaken? If not, the CMake option should be removed IMO.

white238: Issue created. we can handle it outside of this PR https://github.com/llnl/smith/issues/1630

chapman39 (commented):
It would be good to track this PRs new feature(s) in the changelog's unreleased section under "added"

https://github.com/llnl/smith/blob/develop/RELEASE-NOTES.md?plain=1#L22

src/smith/physics/functional_weak_form.hpp
```diff
@@ -99,50 +98,55 @@ class FunctionalWeakForm<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
...
-  void addBodyIntegral(std::string body_name, BodyForceType body_integral)
+  /// @brief Add a body integral depending only on selected input fields.
+  template <int... active_parameters, typename BodyIntegralType>
+  void addBodyIntegral(DependsOn<active_parameters...>, std::string body_name, BodyIntegralType integrand)
   {
-    addBodyIntegral(DependsOn<>{}, body_name, body_integral);
+    addBodyIntegralImpl(body_name, integrand, std::integer_sequence<int, active_parameters...>{});
+  }
+
+  /// @brief Add a body integral to the weak form.
```
lihanyu97:
```suggestion
  /// @brief Add a body integral depending on all input fields to the weak form.
```

src/smith/physics/functional_weak_form.hpp
```diff
@@ -180,93 +186,108 @@ class FunctionalWeakForm<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
...
 
-  /// @overload
+  /// @brief Add a boundary integral depending only on selected input fields.
+  template <int... active_parameters, typename BoundaryIntegrandType>
+  void addBoundaryIntegral(DependsOn<active_parameters...>, std::string boundary_name, BoundaryIntegrandType integrand)
+  {
+    addBoundaryIntegralImpl(boundary_name, integrand, std::integer_sequence<int, active_parameters...>{});
+  }
+
+  /// @brief Add a boundary integral to the weak form.
```
lihanyu97:
```suggestion
  /// @brief Add a boundary integral depending on all input fields to the weak form.
```

src/smith/physics/functional_weak_form.hpp
```diff
@@ -180,93 +186,108 @@ class FunctionalWeakForm<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
...
-  /// @overload
+  /// @brief Add an interior boundary integral depending only on selected input fields.
+  template <int... active_parameters, typename InteriorIntegrandType>
+  void addInteriorBoundaryIntegral(DependsOn<active_parameters...>, std::string interior_name,
+                                   InteriorIntegrandType integrand)
+  {
+    addInteriorBoundaryIntegralImpl(interior_name, integrand, std::integer_sequence<int, active_parameters...>{});
+  }
+
+  /// @brief Add an interior boundary integral to the weak form.
```
lihanyu97:
```suggestion
  /// @brief Add an interior boundary integral depending on all input fields to the weak form.
```

src/smith/physics/functional_weak_form.hpp
```diff
@@ -398,19 +419,49 @@ class FunctionalWeakForm<spatial_dim, OutputSpace, Parameters<InputSpaces...>,
     }
   }
 
-  /// @brief Accessor to get a reference to the underlying ShapeAwareFunctional in case more direct access is needed.
-  /// @return Reference to ShapeAwareFunctional instance.
-  ShapeAwareFunctional<ShapeDispSpace, OutputSpace(InputSpaces...)>& getShapeAwareResidual() { return *weak_form_; }
+ protected:
```
lihanyu97, line 422: Are these methods still necessary? Overloaded functions to add integrals with all inputs are defined previously already.

src/smith/physics/functional_objective.hpp
```diff
@@ -67,12 +67,29 @@ class FunctionalObjective<spatial_dim, Parameters<InputSpaces...>, std::integer_
...
   template <int... active_parameters, typename FuncOfTimeSpaceAndParams>
   void addBodyIntegral(DependsOn<active_parameters...>, std::string body_name,
                        const FuncOfTimeSpaceAndParams& qfunction)
   {
-    objective_->AddDomainIntegral(smith::Dimension<spatial_dim>{}, smith::DependsOn<active_parameters...>{}, qfunction,
-                                  mesh_->domain(body_name));
+    addBodyIntegralImpl(body_name, qfunction, std::integer_sequence<int, active_parameters...>{});
+  }
+
+  /// @brief Add a body integral to the objective function.
```
lihanyu97:
```suggestion
  /// @brief Add a body integral depending on all input fields.
```

src/smith/physics/functional_objective.hpp
```diff
@@ -82,48 +99,72 @@ class FunctionalObjective<spatial_dim, Parameters<InputSpaces...>, std::integer_
...
   template <int... active_parameters, typename FuncOfTimeSpaceAndParams>
   void addBoundaryIntegral(DependsOn<active_parameters...>, std::string boundary_name,
                            const FuncOfTimeSpaceAndParams& qfunction)
   {
-    objective_->AddBoundaryIntegral(smith::Dimension<spatial_dim>{}, smith::DependsOn<active_parameters...>{},
-                                    qfunction, mesh_->domain(boundary_name));
+    addBoundaryIntegralImpl(boundary_name, qfunction, std::integer_sequence<int, active_parameters...>{});
+  }
+
+  /// @brief Add a boundary integral to the objective function.
```
lihanyu97:
```suggestion
  /// @brief Add a boundary integral depending on all input fields.
```

src/smith/physics/functional_objective.hpp
```diff
@@ -82,48 +99,72 @@ class FunctionalObjective<spatial_dim, Parameters<InputSpaces...>, std::integer_
...
 
   /// @overload
-  virtual mfem::Vector gradient(TimeInfo time_info, ConstFieldPtr shape_disp, const std::vector<ConstFieldPtr>& fields,
-                                size_t field_ordinal) const override
+  virtual mfem::Vector gradient(const TimeInfo& time_info, ConstFieldPtr shape_disp,
+                                const std::vector<ConstFieldPtr>& fields, size_t field_ordinal) const override
   {
-    dt_ = time_info.dt();
-    cycle_ = time_info.cycle();
+    current_time_info_ = &time_info;
```
lihanyu97, line 143: Why do we have a guard to prevent `current_time_info_` from being nullified prematurally in `functional_weak_form` but not for `functional_objective`

lihanyu97 (commented):
The PR looks great and allows flexibility of constructing multiphysics modules from previously defined single physics modules, which can significantly speedup future developments.

Considering the substaintial changes, can you summarize a list of interface breaking changes (eg. removal of time_discretized_weak_form)? It will be easier for other PRs created before this one to adjust to the new API.

chapman39 (changes requested):

src/smith/differentiable_numerics/coupling_params.hpp
tupek2: I tried, see if you think its better. the complexity of time discretizing possible fields in somewhat arbitrary order is hard.

tupek2:
> It would be good to track this PRs new feature(s) in the changelog's unreleased section under "added"
> 
> https://github.com/llnl/smith/blob/develop/RELEASE-NOTES.md?plain=1#L22

I added some notes

lihanyu97 (approved):

chapman39 (approved):

PR #1610, opened 2026-05-29
update blt with new sphinx feature

chapman39:
* updates blt with the following functionality: https://github.com/llnl/blt/pull/764/
* removes explicit find doxygen command (blt does this for us)
* unrelated: change hubcast setting to display each ci job separately on github

### DIFF
--- a/src/docs/doxygen/CMakeLists.txt
+++ b/src/docs/doxygen/CMakeLists.txt
@@ -4,23 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause)
 
-find_package(Doxygen COMPONENTS dot)
-
-if(DOXYGEN_DOT_FOUND)
-    set(DOXYGEN_HAVE_DOT "YES")
-else()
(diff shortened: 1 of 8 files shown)


src/docs/doxygen/CMakeLists.txt
```diff
@@ -4,23 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause)
 
-find_package(Doxygen COMPONENTS dot)
-
-if(DOXYGEN_DOT_FOUND)
```
chapman39, line 9: @tupek2 let me know why this was needed and we can look into re-adding this. blt handles setting up doxygen. and im not sure about the pdflatex thing.

tupek2: I didn't have those packages on my mac, but still wanted to check if make docs worked so I didn't have to go to LC.  We can put it back and I can just figure out how to get them installed.

tupek2 (approved):

white238 (approved):

PR #1612, opened 2026-06-02
Fix dual adjoint BC setter

ebchin:
- Improves parsing of dual adjoint BC field names (first tries to match in derived class, then calls base class version)
- Removes `contact_forces` as a dual field in favor of a dual field for each contact force (e.g. `contact_force_0`, `contact_force_1`, etc.)

### DIFF
--- a/src/smith/physics/solid_mechanics_contact.hpp
+++ b/src/smith/physics/solid_mechanics_contact.hpp
@@ -382,6 +382,7 @@ class SolidMechanicsContact<order, dim, Parameters<parameter_space...>,
       return true;
     }
 
+    // No match found for dual_name in SolidMechanicsContact. Try the base class.
     return SolidMechanicsBase::trySetDualAdjointBc(dual_name, bc);
   }
 
--- a/src/smith/physics/solid_mechanics_contact.hpp
+++ b/src/smith/physics/solid_mechanics_contact.hpp
@@ -93,10 +93,13 @@ class SolidMechanicsContact<order, dim, Parameters<parameter_space...>,
       : SolidMechanicsBase(std::move(solver), timestepping_opts, physics_name, smith_mesh, parameter_names, cycle, time,
                            checkpoint_to_disk, use_warm_start),
         contact_(BasePhysics::mfemParMesh()),
-        forces_(StateManager::newDual(displacement_.space(), detail::addPrefix(physics_name, "contact_forces")))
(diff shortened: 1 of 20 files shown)


src/smith/physics/solid_mechanics_contact.hpp
```diff
@@ -291,6 +299,10 @@ class SolidMechanicsContact<order, dim, Parameters<parameter_space...>,
   /// @overload
   const FiniteElementState& dualAdjoint(const std::string& dual_name) const override
   {
+    if (isAggregateContactForceName(dual_name)) {
```
tupek2: I would vote for us not supporting aggregated contact interaction forces at all.  it makes no sense to me personally as an output in the first place and LiDO users can just compute it themselves by summing contact interactions if they ever come up with some justification.  I think it adds too much unneeded complications.

tupek2 (approved):

src/smith/physics/solid_mechanics_contact.hpp
```diff
@@ -354,47 +369,39 @@ class SolidMechanicsContact<order, dim, Parameters<parameter_space...>,
...
+    if (interaction_id.has_value()) {
       auto it = contact_interaction_force_adjoint_bcs_.find(*interaction_id);
       SLIC_ERROR_ROOT_IF(it == contact_interaction_force_adjoint_bcs_.end(),
                          std::format("No contact force adjoint BC registered for interaction_id={}", *interaction_id));
-
       *it->second = bc;
+      return true;
     }
+
+    return SolidMechanicsBase::trySetDualAdjointBc(dual_name, bc);
```
white238, line 386: Can you add a comment here to make it clear that this isn't an infinite loop? Something like "Not set on the derived class, try base class"

white238 (approved):

src/smith/physics/solid_mechanics_contact.hpp
ebchin: added

src/smith/physics/solid_mechanics_contact.hpp
ebchin: fixed

chapman39:
@lc-hubcast approve

chapman39 (commented):
@lc-hubcast approve

chapman39:
i was testing to see if the hubcast `approve` command would force the ci to run as me and turns out it'll still run as eric i guess.

ebchin:
@lc-hubcast approve

ebchin (commented):
@lc-hubcast approve

PR #1613, opened 2026-06-04
Create script to help developers avoid exceeding disk quota on LC

chapman39:
Now that we are using Hubcast, CI will run as the developer who pushes the commit, merges the branch, etc. Because of this, each Smith developer is at risk of exceeding disk quota. This script helps prevent that by doing the following two things

1. setup symlink pointing the jacamar ci directory out of the HOME directory and to the workspace
2. create a cron job that will periodically (daily at midnight) remove the contents of the jacamar ci

I'm sort of worried that people will create conflicting cron jobs across multiple machine nodes so I created some checks to avoid it.

### DIFF
--- a/scripts/setup_jacamar_ci.sh
+++ b/scripts/setup_jacamar_ci.sh
@@ -54,7 +54,7 @@ cron_script_contents() {
 0,20,40 * * * * crontab ${CRON_SCRIPT_PATH}
 
 # Remove contents of Jacamar CI that are older than 48 hours
-0 0 * * * srun -N1 find "${WORKSPACE_CI_DIR}" -mindepth 1 -maxdepth 1 -mtime +1 -exec rm -rf -- {} +
+0 0 * * * srun -N1 find "${WORKSPACE_CI_DIR}" -mindepth 1 -mtime +1 -exec rm -rf -- {} +
 EOF
 }
 
--- a/scripts/setup_jacamar_ci.sh
+++ b/scripts/setup_jacamar_ci.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+# Redirect Jacamar CI from your HOME directory to your LC workspace and periodically cleanup CI jobs to avoid exceeding
+# disk quota
+#
+# Some things to note:
+# - Running this script on oslic is recommended, because its best suited for file IO
+# - It's important to understand that Cron jobs are installed per node (not per machine!), so do not run this script
+#   more than once unless you know what you're doing.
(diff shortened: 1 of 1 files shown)


chapman39:
also i should mention i am aware of mpifileutils allowing you to remove directories in parallel. however, i could not for the life of me get it to work for this use-case, because of the sheer amount of files needed to be removed, `drm` would often hang and/ or fail. i think its some sort of memory limitation. https://myconfluence.llnl.gov/spaces/RAM/pages/211290486/File+info+-+aka+where+are+all+my+files+and+how+do+i+clean+up

i've been using regular `rm` for weeks and have had no issues.

scripts/setup_jacamar_ci.sh
```diff
@@ -0,0 +1,64 @@
...
+# This cron job has been installed on $HOST_NAME.
+
+# For more information about using Cron on LC:
+# https://myconfluence.llnl.gov/spaces/RAM/pages/417729120/Cron+-+how+it+can+be+useful+for+triggering+work+at+a+given+time
+
+# Allows you to update your Cron job without having to go to the specific node
+0,20,40 * * * * crontab ${CRON_SCRIPT_PATH}
+
+# Remove contents of Jacamar CI
+0 0 * * * srun -N1 rm -rf ${WORKSPACE_CI_DIR}/*
```
white238: This will cause builds to fail if they are running at night. I recommend only deleting things that are of a certain age.. like a day or two old.

scripts/setup_jacamar_ci.sh
```diff
@@ -0,0 +1,64 @@
...
+# This cron job has been installed on $HOST_NAME.
+
+# For more information about using Cron on LC:
+# https://myconfluence.llnl.gov/spaces/RAM/pages/417729120/Cron+-+how+it+can+be+useful+for+triggering+work+at+a+given+time
+
+# Allows you to update your Cron job without having to go to the specific node
+0,20,40 * * * * crontab ${CRON_SCRIPT_PATH}
+
+# Remove contents of Jacamar CI that are older than 48 hours
+0 0 * * * srun -N1 find "${WORKSPACE_CI_DIR}" -mindepth 1 -maxdepth 1 -mtime +1 -exec rm -rf -- {} +
```
chapman39:
```suggestion
0 0 * * * srun -N1 find "${WORKSPACE_CI_DIR}" -mindepth 1 -mtime +1 -exec rm -rf -- {} +
```

PR #1623, opened 2026-06-15
Call force at the end of finite differencing step to make sure contac…

tupek2:
…t state is as it should be going into the next time step.


ebchin (approved):

white238 (approved):
Thanks @tupek2 !

chapman39 (commented):
@lc-hubcast approve

PR #1626, opened 2026-06-24
Make mfem::FiniteElementSpace a std::shared_ptr in FiniteElementVector

ebchin:
- Allows StateManager's FESpace to remain valid even if a user-held FE Space or Dual's space is overwritten or goes out of scope
- Reinstates move assignment for FE Dual and FE Space, since they no longer break StateManager

### DIFF
--- a/src/smith/physics/state/finite_element_vector.hpp
+++ b/src/smith/physics/state/finite_element_vector.hpp
@@ -194,13 +194,15 @@ class FiniteElementVector : public mfem::HypreParVector {
   std::reference_wrapper<mfem::ParMesh> mesh_;
 
   /**
-   * @brief Handle to the FiniteElementCollection. Shared to ensure it outlives any Sidre grid functions.
+   * @brief Handle to the FiniteElementCollection. Shared to allow safe memory management when multiple vectors or grid
+   * functions use the same collection.
    * @note Must be const as FESpaces store a const reference to their FEColls
    */
   std::shared_ptr<mfem::FiniteElementCollection> coll_;
 
   /**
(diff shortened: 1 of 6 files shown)


src/smith/physics/state/tests/test_state_manager.cpp
```diff
@@ -253,6 +253,42 @@ TEST(StateManager, ResetDoesNotDeleteExternallySharedMesh)
...
 }
 
+TEST(StateManager, MoveAssignmentLifetime)
+{
+  std::string name = "move_assignment_test";
+  axom::sidre::DataStore datastore;
+  const std::string output_dir = name + "_data";
+
+  if (axom::utilities::filesystem::pathExists(output_dir)) {
+    GTEST_SKIP() << "Output directory already exists from a prior run: " << output_dir;
```
white238, line 263: Seems like these should be a failure not a skip

src/smith/physics/state/finite_element_vector.hpp
```diff
@@ -162,6 +162,16 @@ class FiniteElementVector : public mfem::HypreParVector {
...
 
+  /**
+   * @brief Get a shared pointer to the internal ParFiniteElementSpace
+   */
+  std::shared_ptr<mfem::ParFiniteElementSpace> shared_space() const { return space_; }
+
+  /**
+   * @brief Get a shared pointer to the internal FiniteElementCollection
+   */
+  std::shared_ptr<mfem::FiniteElementCollection> shared_coll() const { return coll_; }
```
white238, line 173: `coll` seems too generic since this is user-facing

PR #1629, opened 2026-07-13
Tupek/tr core mfem subspace

tupek2:
The intent here is mostly to get away from requiring PETSc to use the 'subspace' option for the trust-region solver, and to have additional header files for the trust-region implementation so the equations_solver.cpp does not keep getting larger.  There were some additional performance optimizations, such as reducing the total dot products and the number of parallel reductions, saving off subspace results when doing linesearching (as the subspace does not change), tuned some defaults based on a suite of benchmarks that I will add in a follow on PR, and some other small adjustments.

  - Remove PETSc/SLEPc dependency from trust-region subspace solver.
  - Replace PETSc reduced solve with MFEM/LAPACK implementation.
  - Cache reduced subspace projections across trust-radius retries.
  - Extract Steihaug-Toint CG into private implementation helper.
  - Keep public entry through NonlinearSolver::TrustRegion.
  - Use one batched global dot-product path for trust-region math.
  - Add previous accepted steps to subspace basis, default 2.
  - Add tuned trust-region defaults from benchmark runs.
  - Add MFEM subspace and Steihaug-Toint CG tests.
  - Guard LAPACK subspace tests on MFEM_USE_LAPACK.
  - Add shallow-arch snap-through physics coverage.


src/smith/numerics/solver_config.hpp
```diff
@@ -492,6 +492,33 @@ struct NonlinearSolverOptions {
 
   /// Number of extra leftmost eigenvector to be stored between solves
   int num_leftmost = 1;
+
+  /// Number of previous accepted steps to include in trust-region subspace solves
+  int num_previous_steps = 2;
+
+  /// Relative CG forcing term for TrustRegion model solves.
+  double cg_forcing_rel = 1.2981521889723316e-05;
```
btalamini: Could you give this a more meaningful name and explanation? I can suggest `cg_residual_reduction_for_early_exit`, but I'm open to better ideas.

src/smith/physics/weak_form.hpp
```diff
@@ -12,6 +12,7 @@
 
 #pragma once
 
+#include <cstddef>
```
chapman39: is this required?

src/smith/numerics/solver_config.hpp
```diff
@@ -499,6 +499,33 @@ struct NonlinearSolverOptions {
...
+  double tr_eta1 = 1.0e-9;
+
+  /// rho below which the radius shrinks.
+  double tr_eta2 = 0.12631114;
+
+  /// rho above which a boundary step grows the radius.
+  double tr_eta3 = 0.54944455;
+
+  /// rho ceiling: steps above this are distrusted.
+  double tr_eta4 = 1.8368325;
```
chapman39, line 529: should some of these be const?

tupek2: these are setting defaults.  they can be overwritten.

src/smith/physics/weak_form.hpp
tupek2: nope, fixed.

chapman39 (approved):

src/smith/numerics/mfem_trust_region_subspace.cpp
```diff
@@ -0,0 +1,448 @@
...
+struct SubspaceProjections {
+  mfem::DenseMatrix sAs;
+  mfem::DenseMatrix ss;
+  mfem::Vector sb;
+};
+
+void checkProjectionInputs(const std::vector<const mfem::Vector*>& states,
+                           const std::vector<const mfem::Vector*>& Astates, const mfem::Vector& b)
+{
+  MFEM_VERIFY(states.size() == Astates.size(),
```
btalamini: Should these be our own assertions, rather than MFEM ones?

src/smith/numerics/mfem_trust_region_subspace.cpp
```diff
@@ -0,0 +1,448 @@
...
+double norm(const mfem::Vector& x) { return x.Norml2(); }
+
+double sumAbs(const mfem::Vector& x)
+{
+  double total = 0.0;
+  for (int i = 0; i < x.Size(); ++i) {
+    total += std::abs(x[i]);
+  }
+  return total;
+}
```
btalamini: Your tests probably cover this, but just to be thorough: are these ok in parallel?

src/smith/numerics/equation_solver.cpp
```diff
@@ -26,6 +32,22 @@
 namespace smith {
 
 namespace {
+
+size_t rootOnlyPrintLevel(const mfem::NewtonSolver& solver, size_t level)
+{
+#ifdef MFEM_USE_MPI
```
btalamini: I'll need @white238 to confirm, but I think we always build with MPI and the guards are unnecessary. In fact, this function could possibly be reduced to a clear one-liner.

src/smith/numerics/steihaug_toint_cg.cpp
```diff
@@ -0,0 +1,124 @@
...
+  if (deltadelta_m_zz == 0.0) return;
+  const double tau = (std::sqrt(deltadelta_m_zz * dd + zd * zd) - zd) / dd;
+  z.Add(tau, d);
+}
+
+}  // namespace
+
+void steihaugTointCG(const mfem::Vector& r0, mfem::Vector& rCurrent, const mfem::Operator& H, const mfem::Solver* P,
+                     const TrustRegionSettings& settings, double& trSize, TrustRegionResults& results,
+                     double r0_norm_squared, const DotManyFunction& dot_many)
```
btalamini, line 25: For the dot product function, should there be a fallback? Naive, one-at-a-time dot products?

src/smith/numerics/mfem_trust_region_subspace.cpp
```diff
@@ -0,0 +1,448 @@
...
+  gram_copy.Eigensystem(evals, evecs);
+
+  trace_mag = 0.0;
+  for (int i = 0; i < evals.Size(); ++i) {
+    trace_mag += std::abs(evals[i]);
+  }
+
+  std::vector<mfem::Vector> kept_columns;
+  for (int i = 0; i < evals.Size(); ++i) {
+    if (evals[i] > 1e-9 * trace_mag) {
```
btalamini: Probably ok here, but hard-coded tolerances are a code smell. Want to make sure you're ok with this one.

btalamini (approved):
I have a few questions, but they're small.

src/smith/numerics/tests/test_steihaug_toint_cg.cpp
```diff
@@ -0,0 +1,102 @@
...
+  smith::TrustRegionResults results(size);
+
+  mfem::Vector rCurrent(size);
+
+  smith::steihaugTointCG(r0, rCurrent, H, nullptr, settings, trSize, results, r0.Norml2() * r0.Norml2(), localDotMany);
+
+  // For negative curvature, it should go to boundary
+  EXPECT_NEAR(results.z.Norml2(), 2.0, 1e-9);
+  EXPECT_EQ(results.interior_status, smith::TrustRegionResults::Status::NegativeCurvature);
+}
```
ebchin, line 102: Do we need to add some more tests? Maybe for `NonDescentDirection`? Or maybe some tests that aren't covered by `Status`: initial convergence, iteration limits, zero curvature? I think we had an initial convergence issue before, so a test might be helpful.

src/smith/numerics/tests/test_trust_region_solver_mfem.cpp
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) Lawrence Livermore National Security, LLC and
```
ebchin: Maybe add `subspace` to the test name so it's clear it only tests the reduced-subspace routine?

src/smith/numerics/tests/CMakeLists.txt
```diff
@@ -10,13 +10,19 @@ set(numerics_serial_test_sources
...
     test_block_preconditioner.cpp
     test_block_preconditioner_backend.cpp
     test_block_preconditioner_custom_operators.cpp
     test_linear_solver_precond_only.cpp
     )
 
-smith_add_tests(SOURCES       ${numerics_serial_test_sources}
+if(MFEM_USE_LAPACK)
+    list(APPEND numerics_serial_test_sources
+         test_trust_region_solver_mfem.cpp)
```
ebchin: I see some MPI reductions in the new code. Should we do some tests in parallel?

src/smith/numerics/mfem_trust_region_subspace.cpp
```diff
@@ -0,0 +1,448 @@
...
+  const double e2 = quadraticEnergy(A, b, neg_x);
+
+  x *= (e2 < e1 ? -delta : delta) / norm(x);
+
+  return std::make_pair(x, success);
+}
+
+}  // namespace
+
+void TrustRegionSubspaceCache::prepare(const std::vector<const mfem::Vector*>& directions,
```
ebchin: I had trouble finding this implementation since it's declaration is in `trust_region_subspace_cache.hpp`. Can you clarify the choices behind the file names?

src/smith/numerics/trust_region_subspace_cache.hpp
```diff
@@ -0,0 +1,46 @@
...
+namespace smith {
+
+/// Cached reduced trust-region subspace problem reusable across trust-region radius updates.
+struct TrustRegionSubspaceCache {
+  /// @brief prepares reduced trust-region subspace data reusable across trust-region radius updates
+  void prepare(const std::vector<const mfem::Vector*>& directions, const std::vector<const mfem::Vector*>& A_directions,
+               const mfem::Vector& b, int num_leftmost, MPI_Comm comm = MPI_COMM_WORLD);
+
+  /// @brief solves cached reduced trust-region problem for given trust-region radius
+  TrustRegionSubspaceResult solve(double delta) const;
```
ebchin, line 30: Is there any testing for re-use of the cached values?

src/smith/numerics/tests/test_trust_region_solver_mfem.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  EXPECT_THROW(smith::solveSubspaceProblem(states, astates, b, 1.0, 1), smith::TrustRegionException);
+}
+
+int main(int argc, char* argv[])
+{
+  ::testing::InitGoogleTest(&argc, argv);
+  smith::ApplicationManager applicationManager(argc, argv);
+  return RUN_ALL_TESTS();
+}
```
ebchin: Maybe add a regression test for that NaN thing Codex found?

src/smith/numerics/equation_solver.cpp
```diff
@@ -790,15 +710,13 @@ class TrustRegion : public mfem::NewtonSolver, public ConvergenceManagedNonlinea
   void hessVec(const mfem::Vector& x_, mfem::Vector& v_) const
   {
     SMITH_MARK_FUNCTION;
-    ++num_hess_vecs;
     grad->Mult(x_, v_);
   }
 
   /// apply trust region specific preconditioner
   void precond(const mfem::Vector& x_, mfem::Vector& v_) const
```
ebchin: Is this unused now?

ebchin (commented):
Overall, looks good. Mostly questions about testing.

src/smith/numerics/tests/test_trust_region_solver_mfem.cpp
tupek2: the nan thing was not entirely relevant.  It for a specific case where the reduced hessian is exactly indefinite, which is rare, and anyways, the case was handled on the outside as the fallback.  Will rewrite to make that decision a bit clearer.

src/smith/numerics/steihaug_toint_cg.cpp
tupek2: I tend to avoid optional args when possible.  Its an arbitrary style preference.

src/smith/numerics/tests/test_steihaug_toint_cg.cpp
tupek2: Adding a test for the initial convergence issue.  I'm going to defer more testing for later, as need dictates, to avoid going down rabbit holes.

src/smith/numerics/tests/CMakeLists.txt
tupek2: Yes

src/smith/numerics/tests/test_trust_region_solver_mfem.cpp
tupek2: yes

src/smith/numerics/mfem_trust_region_subspace.cpp
tupek2: I'm going to keep as is to avoid more churn.  can revisit later.

src/smith/numerics/solver_config.hpp
tupek2: renamed

src/smith/numerics/equation_solver.cpp
tupek2: I removed the guards.

src/smith/numerics/equation_solver.cpp
tupek2: yes, removed

src/smith/numerics/trust_region_subspace_cache.hpp
tupek2: not directly... I decided not to as it bloats the PR even more.  but, it works.  could in a follow up.

ebchin (approved):

Issue #1632, opened 2026-07-15
Time not tracked for adjoint with new differentiable physics


tlroy:
`DifferentiablePhysics` does not track time history. So for e.g. in LiDO if your objective explicitly depends on time, the adjoint solve will be stuck at the final time.

tlroy:
Fixed by PR #1634

PR #1634, opened 2026-07-27
Time history for time-dependent objective

tlroy:
Add time history to the new differentiable physics framework. Fixes Issue #1632


tupek2 (approved):
I think there will be, in the long run, a better way to handle this within smith by putting time actually on the gretl graph.  But, that is more involved of a change, so I think we can and should do this for now.  For transient problems, lido relies on us for storage, while for static, the handle their own storage.  Though a bit strange, having us save the time is consistent with us being on the hook to also save displacement fields (which we already do).

btalamini (approved):
+1 On Mike's assessment.

Thanks for fixing, Thomas!

PR #1638, opened 2026-08-05
Tupek/solid axisym

tupek2:


src/smith/physics/solid_mechanics.hpp
```diff
@@ -1027,6 +1248,37 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     setPressure(DependsOn<>{}, pressure_function, domain);
   }
 
+  /**
+   * @brief Apply an axisymmetric pressure-type follower load.
+   *
+   * @note Only valid for 2D meshes interpreted as (r, z).
```
white238, line 1335: Is this eventually going to be 3D or does that not make sense? Just curious if we should track it in an issue.... My guess is it doesnt make sense.

src/smith/physics/solid_mechanics.hpp
```diff
@@ -819,6 +944,30 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     setRateDependentMaterial(DependsOn<>{}, material, domain, qdata);
   }
 
+  /**
+   * @brief Set an axisymmetric material that gets dt as an argument.
+   */
+  template <int... active_parameters, typename RateDependentMaterialType, typename StateType = Empty>
+  void setAxisymmetricRateDependentMaterial(DependsOn<active_parameters...>, const RateDependentMaterialType& material,
```
white238, line 993: These parameters should be documented

src/smith/physics/solid_mechanics.hpp
```diff
@@ -743,6 +811,32 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
     setMaterial(DependsOn<>{}, material, domain, qdata);
   }
 
+  /**
+   * @brief Set an axisymmetric material stress response and mass properties.
+   *
+   * @note Only valid for 2D meshes interpreted as (r, z). The material receives a full 3D displacement gradient.
+   * @note This method must be called prior to completeSetup()
```
white238, line 840: Should we have a boolean that completeSetup has been called and check?

src/smith/physics/tests/test_solid_axisymmetric.cpp
```diff
@@ -0,0 +1,286 @@
...
+  });
+
+  constexpr double expected_radial = pressure * 2.0;
+  expectResultant(residual, tensor<double, 2>{{expected_radial, 0.0}});
+}
+
+TEST(SolidMechanics, AxisymmetricPressureInnerResultant)
+{
+  axom::sidre::DataStore datastore;
+  StateManager::initialize(datastore, "axisymmetric_pressure_inner_resultant");
```
white238, line 264: Would it be helpful to have a helper macro that gives you the test name?

Something like:

```
#define GTEST_FULL_NAME()                                                  \
    ([]() -> std::string {                                                 \
        const auto* info =                                                 \
            ::testing::UnitTest::GetInstance()->current_test_info();       \
        return info                                                        \
            ? std::string(info->test_suite_name()) + "_" + info->name()    \
            : std::string{};                                               \
    }())
```

It would return `SolidMechanics_AxisymmetricPressureInnerResultant` in this case.

I could do this outside of this PR if you think its useful.

src/smith/physics/solid_mechanics.hpp
tupek2: We have discussed this a few times.  These setup ordering issues have been here for years.  I would be happy to have that added, but perhaps a follow on PR.

white238: I'll make an issue and one of us can do it at some point. Maybe ill handle it after this goes in.

white238: #1639

src/smith/physics/tests/test_solid_axisymmetric.cpp
white238: #1640

btalamini (approved):
This looks good to me

src/smith/physics/tests/test_solid_axisymmetric.cpp
tupek2: I'd say no need.  I happens that a convenient/lazy name here is sometimes the test name, or similar.

btalamini:
I see you've changed the convention by putting the 2*pi factor in. I've seen both with and without in codes -- either works -- but the convention should be communicated to the user. For example, they need to know whether to supply tractions as force per unit length, or force per unit length and per radian.

src/smith/physics/solid_mechanics.hpp
tupek2: done

src/smith/physics/solid_mechanics.hpp
tupek2: it can only every be 2D.

src/smith/physics/tests/test_solid_axisymmetric.cpp
white238: we have been bit by this problem in the past

src/smith/physics/solid_mechanics.hpp
```diff
@@ -1027,6 +1313,50 @@ class SolidMechanics<order, dim, Parameters<parameter_space...>, std::integer_se
...
+  template <int... active_parameters, typename PressureType>
+  void setAxisymmetricPressure(DependsOn<active_parameters...>, PressureType pressure_function, Domain& domain)
+  {
+    static_assert(dim == 2, "axisymmetric pressure requires a 2D mesh");
+    residual_->AddBoundaryIntegral(
+        Dimension<dim - 1>{}, DependsOn<0, 1, active_parameters + NUM_STATE_VARS...>{},
+        [pressure_function](double t, auto X, auto displacement, auto /* acceleration */, auto... params) {
+          auto x = X + displacement;
+          auto n = cross(get<DERIVATIVE>(x));
+          auto X0 = get<VALUE>(X);
```
kswartz14, line 1350: might be more clear to call this `r`

white238 (approved):

chapman39 (commented):
itd be good to mention the new feature in the changelog

tupek2:
> itd be good to mention the new feature in the changelog

done

src/smith/physics/solid_mechanics.hpp
tupek2: added some rs where used (for clarify)

PR #1642, opened 2026-08-06
Energy mortar quadrature point gap enforcement

ebchin:
Also add FD testing to test_adjoint_workflow and finding related bugs.


cmake/thirdparty/SetupSmithThirdParty.cmake
```diff
@@ -341,22 +341,27 @@ if (NOT SMITH_THIRD_PARTY_LIBRARIES_FOUND)
...
         # NOTE: we check if an hdf5 target is namespaced or not, since some versions
         #       of hdf5 do not namespace their targets and others do
+        find_package(HDF5 CONFIG QUIET COMPONENTS C HL
+                     PATHS "${HDF5_DIR}/cmake"
+                     NO_DEFAULT_PATH)
         set(HDF5_TARGET_NAMES "" CACHE STRING "")
         if(TARGET hdf5::hdf5-static)
             list(APPEND HDF5_TARGET_NAMES hdf5::hdf5-static)
-        else()
+        elseif(TARGET hdf5-static)
```
tupek2: are these changes still needed after Chris' recent hdf5 related changes?

src/smith/physics/tests/test_adjoint_workflow.cpp
```diff
@@ -175,6 +193,87 @@ std::unique_ptr<BasePhysics> createContactSolver(std::shared_ptr<Mesh> mesh)
...
+        auto displacement = 0.0 * x;
+        displacement[1] = -1.0e-2;
+        return displacement;
+      },
+      mesh->domain("ymax_face"), Component::Y);
+
+  solid->addContactInteraction(config.interaction_id, {3}, {5}, config.options);
+#ifdef SMITH_USE_ENZYME
+  if (config.options.method == ContactMethod::EnergyMortar) {
+    tribol::setEnergyMortarEnforcementOption(config.interaction_id, config.gap_option);
```
tupek2: What is the default gap_option now?  should it be the quadrature one?

src/smith/physics/solid_mechanics_contact.hpp
```diff
@@ -681,6 +676,40 @@ class SolidMechanicsContact<order, dim, Parameters<parameter_space...>,
 #endif
   }
 
+#ifdef SMITH_USE_TRIBOL
```
tupek2, line 679: The old version had some bug in the adjoint, right? or was it just related to shape displacement adjoint?  either way, looks like this is mostly pulling out a function and some cleanup? or fundamental bug?

tupek2 (approved):

src/smith/physics/tests/test_adjoint_workflow.cpp
ebchin: Yep, the default (set in Tribol, see Parameters.hpp there) is QuadraturePoint.

cmake/thirdparty/SetupSmithThirdParty.cmake
ebchin: Don't think so. Removed.

src/smith/physics/solid_mechanics_contact.hpp
ebchin: The adjoint was OK, the issue is in computeTimestepShapeSensitivity() with a contact force adjoint BC: the effect on the QoI of a force change caused by a shape change (`(df_c/ds)^T (force_adjoint)`) wasn't being included.

tupek2 (approved):

src/smith/numerics/equation_solver.cpp
btalamini, line 1: Why is the trust region solver refactored in this particular PR?

src/smith/physics/contact/contact_data.cpp
```diff
@@ -25,10 +25,28 @@ namespace smith {
...
+    // vector but recomputes it from the current gap on update; quadrature-point gaps do not expose a nodal pressure
+    // vector. In both cases there is no independent pressure state to reset or set here.
+    return false;
+  }
+#endif
+  return true;
+}
+
+}  // namespace
+
```
btalamini, line 44: I'm ok with this going in the "classic contact" physics module, but it is an indication that the API between Smith and Tribol is blurry. This is a function that lives in Smith that responds to queries about how Tribol works. When we put contact in the new interface, we should revisit the design.

src/smith/physics/contact/contact_interaction.hpp
```diff
@@ -46,12 +46,13 @@ class ContactInteraction {
    * @param bdry_attr_surf1 MFEM boundary attributes for the first (mortar) surface
    * @param bdry_attr_surf2 MFEM boundary attributes for the second (nonmortar) surface
    * @param current_coords Reference to the grid function holding current mesh
+   * @param shaped_reference_coords Reference to the grid function holding the shape-displaced reference mesh
    * @param contact_opts Defines contact method, enforcement, type, and penalty
    * coordinates
```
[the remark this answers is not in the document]
btalamini: Actually, I think the stray "coordinates" was orphaned from the previous function parameter. That is, it belongs on the end of line 49. @ebchin, can you confirm?

src/smith/numerics/steihaug_toint_cg.cpp
```diff
@@ -18,8 +18,41 @@ void projectToBoundaryWithCoefs(mfem::Vector& z, const mfem::Vector& d, double d
...
+  } else if (cauchy_norm_squared >= trust_region_size_squared) {
+    step = cauchy_point;
+    step *= std::sqrt(trust_region_size_squared / cauchy_norm_squared);
+  } else {
+    step = cauchy_point;
+    const double cauchy_newton = dot_many({{&cauchy_point, &newton_point}})[0];
+    projectToBoundaryBetweenWithCoefs(step, newton_point, trust_region_size, cauchy_norm_squared, cauchy_newton,
+                                      newton_norm_squared);
+  }
+}
```
btalamini: The logic for the dogleg step is different than what was removed from `equation_solver.cpp`, meaning that this will change the behavior of the solver. @tupek2, have you looked over these changes carefully?

src/smith/numerics/tests/test_steihaug_toint_cg.cpp
```diff
@@ -133,3 +146,46 @@ TEST(SteihaugTointCG, LowInitialResidualHonorsMinimumIterations)
...
+  smith::steihaugTointCG(initial_residual, current_residual, hessian, &preconditioner, settings, trust_region_size,
+                         results, initial_residual * initial_residual, localDotMany);
+
+  EXPECT_NEAR(results.z[0], -0.5, 1.0e-9);
+  EXPECT_NEAR(results.z[1], -0.25, 1.0e-9);
+  EXPECT_EQ(results.interior_status, smith::TrustRegionResults::Status::Interior);
+}
+
+TEST(TrustRegionDogleg, SelectsFeasibleNewtonPointWhenCauchyPointIsLonger)
+{
```
btalamini: This is a different behavior than our original implementation - I believe the CP was taken when it was longer. Was the old behavior wrong?

btalamini (commented):
This changes the behavior of our core solver. Please don't merge this until we've talked about it.

src/smith/numerics/tests/test_steihaug_toint_cg.cpp
ebchin: Removed from this PR

src/smith/numerics/steihaug_toint_cg.cpp
ebchin: This needs to be reworked. I took it out of this PR.

examples/contact/ironing_2D.cpp
```diff
@@ -218,10 +245,14 @@ int main(int argc, char* argv[])
...
 
   std::string selected_case = "square";
+  std::string energy_mortar_gap_mode = "quadrature-point";
   int num_steps = -1;
   axom::CLI::App app{"2D contact ironing example"};
   app.add_option("--case", selected_case, "Ironing case: square, circle, or twisted")
       ->check(axom::CLI::IsMember({"square", "circle", "twisted"}));
+  app.add_option("--energy-mortar-gap-mode", energy_mortar_gap_mode,
+                 "EnergyMortar gap evaluation mode: nodal or quadrature-point")
+      ->check(axom::CLI::IsMember({"nodal", "quadrature-point"}));
```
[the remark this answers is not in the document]
ebchin: Yeah, I'm probably not going to do that.

src/smith/physics/contact/contact_data.cpp
ebchin: Yes, agreed. We will need to distinguish between 2 types of penalty enforcement: done internally in the method and done in Smith using constraint matrices.

btalamini (approved):
There are still a couple small things picked up by the bot. Please fix those before merging.

PR #1645, opened 2026-08-14
CMake fix on examples

thartland:
In a previous PR #1621 I erroneously changed `STRUMPACK_DIR` to `STRUMPACK_FOUND` in various CMake files associated to the examples. I am now reverting back to using `STRUMPACK_DIR`. Upon looking closer at `SetupSmithThirdParty.cmake` it appears that since strumpack will come via mfem, if it is made available at all, then smith doesn't set/use/define `STRUMPACK_FOUND` rather all the logic is done via `STRUMPACK_DIR`.

### DIFF
--- a/examples/contact/homotopy/CMakeLists.txt
+++ b/examples/contact/homotopy/CMakeLists.txt
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause)
 
-if (TRIBOL_FOUND AND STRUMPACK_FOUND AND SMITH_ENABLE_CONTINUATION)
+if (TRIBOL_FOUND AND STRUMPACK_DIR AND SMITH_ENABLE_CONTINUATION)
     smith_add_executable(NAME       two_blocks_example
                          SOURCES    two_blocks.cpp
                          OUTPUT_DIR ${EXAMPLE_OUTPUT_DIRECTORY}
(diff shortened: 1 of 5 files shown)


examples/contact/homotopy/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: (BSD-3-Clause)
 
-if (TRIBOL_FOUND AND STRUMPACK_FOUND AND SMITH_ENABLE_CONTINUATION)
+if (TRIBOL_FOUND AND STRUMPACK_DIR AND SMITH_ENABLE_CONTINUATION)
```
white238, line 7: This should be `STRUMPACK_FOUND` and if it is not being created that is a bug.

It is because it coming from MFEM and we aren't creating variables for it's dependencies. This is a larger issue you didn't create and I can fix it if you like.

It causes cascading issues as well. Like this config header check isn't being handled either:

https://github.com/llnl/smith/blob/d083f8b8277143c0b1d0c83ac37e8eb21eec6ce7/cmake/SmithConfigHeader.cmake#L10-L15

thartland: @white238  I can this PR go and let you put in the right plumbing to the smith setup thirdparty so that various mfem options are exposed in smith via variables like STRUMPACK_FOUND.

white238: Thats fine. Issue created #1646

white238 (approved):

PR #1649, opened 2026-08-19
State-dependent custom operators for block preconditioning

tlroy:
Adds state-dependent custom Schur complement approximations for block preconditioners.

Users can provide a custom Schur operator directly, or define one from a weak form whose coefficients depend on the current nonlinear solution. Before each Newton linearization, the preconditioner rebuilds that operator using the latest block state, then uses it for the Schur solve. The PR also adds the weak-form helper API and tests for fixed and state-dependent custom Schur approximations.


src/smith/differentiable_numerics/weak_form_block_operator.cpp
```diff
@@ -0,0 +1,206 @@
...
+      MFEM_VERIFY(block_begin >= 0 && block_size >= 0 && block_begin + block_size <= state.Size(),
+                  "Weak-form operator block offsets are inconsistent with the state size");
+
+      FieldState& field = fields_[static_cast<size_t>(binding.field_index)];
+      MFEM_VERIFY(field.get()->Size() == block_size,
+                  "Weak-form operator cannot update a field from a block with incompatible size");
+
+      mfem::Vector block_view;
+      block_view.MakeRef(const_cast<mfem::Vector&>(state), block_begin, block_size);
+      *field.get() = block_view;
```
tupek2: Directly updating the FiniteElementState within a FieldState is a bit scary because the
  FieldState is tracked on the graph, so this could potentially make the graph invalid; perhaps
  a scratch FiniteElementState could be used instead.

tlroy: Ok I made a change so that the weak-form block operator assembles from temporary copies of the input fields, rather than overwriting FieldStates.

src/smith/differentiable_numerics/tests/test_mixed_poisson.cpp
```diff
@@ -177,6 +192,8 @@ TEST_P(BlockPreconditionerTest, BlockSolve)
   auto time = graph->create_state<double, double>(0.0);
   auto dt = graph->create_state<double, double>(0.025);
   size_t cycle = 0;
+  const auto time_info = smith::TimeInfo(time.get(), dt.get(), cycle);
+  std::unique_ptr<mfem::Solver> diff_precond;
```
tupek2: rename: diffusion_ (diff means too many things).

src/smith/differentiable_numerics/tests/test_mixed_poisson.cpp
```diff
@@ -203,7 +220,19 @@ TEST_P(BlockPreconditionerTest, BlockSolve)
     case BlockPrecondType::SchurFullCustom:
       linear_options.preconditioner = smith::Preconditioner::BlockSchur;
       linear_options.block_schur_type = smith::BlockSchurType::Full;
-      /// linear_options.schur_approx_type = smith::BlockSchurType::Custom;
+      linear_options.schur_approx_type = smith::SchurApproxType::Custom;
+      {
```
tupek2: why this scoping?  is there a name conflict?  maybe try to do without this scoping

src/smith/differentiable_numerics/tests/test_porous_heat_sink.cpp
```diff
@@ -1,5 +1,11 @@
 #include <gtest/gtest.h>
```
tupek2, line 1: Does this test add something important that the nonlinear mixed diffsusion doesn't test (or couldn't be easily modified to test).  I worry about about us getting too much testing bloat.  So consolidating or unifying, or extracting out common things from these tests would actually be preferred for me.

tupek2: Or, between test_mixed_poisson, mixed_diffusion, porous_heat_sink... what do we really need to be testing to be sufficiently confident without over testing early on an evolving capability.

src/smith/differentiable_numerics/tests/test_porous_heat_sink.cpp
```diff
@@ -290,7 +299,21 @@ TEST_P(BlockPreconditionerTest, BlockSolve)
     case BlockPrecondType::SchurFullCustom:
       linear_options.preconditioner = smith::Preconditioner::BlockSchur;
       linear_options.block_schur_type = smith::BlockSchurType::Full;
-      /// linear_options.schur_approx_type = smith::BlockSchurType::Custom;
+      linear_options.schur_approx_type = smith::SchurApproxType::Custom;
+      {
```
tupek2: what the scoping here?

src/smith/differentiable_numerics/tests/test_state_dependent_preconditioner.cpp
```diff
@@ -0,0 +1,185 @@
...
+namespace smith {
+namespace {
+
+class StateScaledIdentityProvider : public BlockOperatorProvider {
+ public:
+  explicit StateScaledIdentityProvider(int size) : size_(size) { rebuild(1.0); }
+
+  void updateForState(const mfem::Vector& state, const mfem::Array<int>& block_offsets) override
+  {
+    MFEM_VERIFY(block_offsets.Size() == 3, "StateScaledIdentityProvider expects a two-block state");
```
tupek2: seems like a strange restriction for a identity preconditioner.  sort of doesnt matter as its in a test.

tupek2 (approved):

src/smith/differentiable_numerics/tests/test_mixed_poisson.cpp
tlroy: done

src/smith/differentiable_numerics/tests/test_mixed_poisson.cpp
tlroy: unnecessary scoping. removed

src/smith/differentiable_numerics/tests/test_porous_heat_sink.cpp
tlroy: unnecessary scoping. removed

src/smith/differentiable_numerics/tests/test_porous_heat_sink.cpp
tlroy: test_mixed_poisson is different from the other two since it is a saddle point problem and is a good starting point for other problems like that, where Schur complement approximations are important. nonlinear_diffusion is essentially a nonlinear version of the porous heat sink test. the linear test tests a bunch of the different factorizations, and the nonlinear version only looks at the custom operators. I guess those two could be combined to reduce the numbers of test files or lines of code, but it might increase test time since we'd do more nonlinear solves.

src/smith/differentiable_numerics/tests/test_state_dependent_preconditioner.cpp
tlroy: Check is unnecessary for this test. removed it for simplicity.

src/smith/numerics/equation_solver.cpp
```diff
@@ -1265,21 +1286,19 @@ std::unique_ptr<mfem::NewtonSolver> buildNonlinearSolver(NonlinearSolverOptions
   return nonlinear_solver;
 }
 
-std::pair<std::unique_ptr<mfem::Solver>, std::unique_ptr<mfem::Solver>> buildLinearSolverAndPreconditioner(
-    LinearSolverOptions linear_opts, MPI_Comm comm)
-{
-  auto preconditioner = buildPreconditioner(linear_opts, comm);
+namespace {
```
lihanyu97, line 1289: What is this namespace for?

tlroy: The anonymous namespace keeps `buildLinearSolver()` private to this .cpp file since it is only an implementation helper.

src/smith/numerics/state_dependent_solver.hpp
lihanyu97, line 1: I wonder if this class can inherit `MFEM::Solver`, so `SolverWithPreconditioner` and `BlockPreconditioner` can just inherit this class instead of both.

src/smith/numerics/block_preconditioner.hpp
```diff
@@ -1,29 +1,101 @@
...
+
 #include "mfem.hpp"
 
+#include "smith/numerics/state_dependent_solver.hpp"
+
 namespace smith {
+/**
+ * @brief Supplies the current concrete operator for a block solver override.
+ */
+class BlockOperatorProvider {
```
lihanyu97, line 16: Is this class essentially `StateDependentSolver`?

src/smith/numerics/state_dependent_solver.hpp
tlroy: Good idea. I'll change it

src/smith/numerics/block_preconditioner.hpp
tlroy: They share `updateForState()`, but have different roles. `StateDependentSolver` is a solver, while `BlockOperatorProvider` gives an operator to a solver and may be fixed or state-dependent. I think keeping them separate is clearer.

src/smith/differentiable_numerics/weak_form_block_operator.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+BlockProviderOverride makeStateDependentWeakFormBlockProviderOverride(
+    int block_index, const WeakForm& weak_form, FieldState shape_disp, std::vector<FieldState> fields,
+    std::vector<double> jacobian_weights, TimeInfo time_info, const BoundaryConditionManager* bc_manager,
+    std::vector<StateBlockBinding> state_block_bindings)
+{
+  return makeStateDependentWeakFormBlockProviderOverride(
+      block_index, weak_form, std::move(shape_disp), std::move(fields), std::move(jacobian_weights), time_info,
+      copyEssentialTrueDofs(bc_manager), std::move(state_block_bindings));
+}
```
lihanyu97, line 230: These overloaded methods are returning a lambda function to update the operator with latest field states. I can't find where these methods are used or tested. Would it be safer to just pass the builder to the nonlinear solver then invoke `build` or `buildAndUpdate`? The nonlinear solver should have all information needed such as weak_form, field_states, and etc. I have not seen any Smith implementations use `std::move` in the lambda function capture section.

tlroy: These are used through `makeStateDependentWeakFormBlockProviderOverride()`. There’s a focused test called `StateDependentOverrideUpdatesFromStateBlock`, and a full use example in the nonlinear mixed-diffusion test. `NonlinearBlockSolver` doesn’t own the weak form, fields, etc, so passing the builder there would add weak-form-specific details to a generic solver. The `move` capture just lets the returned callback own the builder after this function returns. I can add a comment to make that ownership clearer.

